PR #2, opened 2023-03-13
fix: reply encoding

dai-shi:
following https://github.com/facebook/react/pull/26360

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,7 +1,7 @@
 import type { ReactNode } from "react";
 import RSDWClient from "react-server-dom-webpack/client";
 
-const { createFromFetch } = RSDWClient;
+const { createFromFetch, encodeReply } = RSDWClient;
 
 export function serve<Props>(rscId: string, render: (ele: ReactNode) => void) {
   return async (props: Props) => {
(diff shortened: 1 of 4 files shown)


src/client.ts
```diff
@@ -19,7 +20,7 @@ export function serve<Props>(rscId: string, render: (ele: ReactNode) => void) {
         }
         const response = fetch(`/?${searchParams}`, {
           method: "POST",
-          body: JSON.stringify(args),
+          body: await encodeReply(args),
```
dai-shi, line 23: Not sure if doing this will set `content-type` header for form data.

dai-shi:
Did an experiment:

```diff
diff --git a/examples/04_callserver/src/Counter.tsx b/examples/04_callserver/src/Counter.tsx
index c55c50d..d0dac48 100644
--- a/examples/04_callserver/src/Counter.tsx
+++ b/examples/04_callserver/src/Counter.tsx
@@ -21,6 +21,18 @@ export const Counter = ({
         >
           greet("client:" + count) = {text}
         </button>
+        <button
+          onClick={async () => {
+            const res = await greet(
+              new Promise((resolve) =>
+                setTimeout(() => resolve("from promise"), 1000)
+              )
+            );
+            console.log(res);
+          }}
+        >
+          greet with delay
+        </button>
       </p>
       <h3>This is a client component.</h3>
     </div>
diff --git a/examples/04_callserver/src/funcs.ts b/examples/04_callserver/src/funcs.ts
index 119d09f..d7540fb 100644
--- a/examples/04_callserver/src/funcs.ts
+++ b/examples/04_callserver/src/funcs.ts
@@ -1,3 +1,3 @@
 "use server";
 
-export const greet = (name: string) => `Hello ${name} from server!`;
+export const greet = async (name: string) => `Hello ${await name} from server!`;
```

image
image
image

src/client.ts
dai-shi: It works.

PR #4, opened 2023-03-19
fix: send client bundle in the initial response

dai-shi:
image

Both `rsc` and `Counter.tsx` fetching start immediately.


dai-shi:
image

PR #8, opened 2023-03-26
alternative API for mutate

dai-shi:
https://twitter.com/dai_shi/status/1639937951596703745

image


dai-shi:
This is a breaking change, but release as a patch, considering an API design bug fix.

PR #11, opened 2023-03-30
experimental: add jotai example

dai-shi:
Highly experimental and not for merge:

- context on server is not isolated (no idea, don't like HoC either)
- we can't access values and methods from client reference (workaround for now, custom handler for jotai atoms is necessary for a proper solution)
- using custom hacky jotai store (this is easiest to solve, because it's on our end)
- async atom not considered yet


dai-shi:
I'm still very interested in integrating jotai and waku, but this branch is too old.

PR #12, opened 2023-04-01
Prefer using createElement before passing to renderToPipeableStream

dai-shi:
...because @gaearon suggested so in @bholmesdev 's stream https://www.twitch.tv/videos/1779925131.


dai-shi:
Maybe this changes the behavior of top-level async component.

PR #19, opened 2023-04-05
Chore: add postgres example

sayjeyhi:


examples/08_postgres/src/Users.tsx
```diff
@@ -0,0 +1,21 @@
+// @ts-ignore
+import db from "./db.js";
+
+export const UsersList = async ({ searchText }: { searchText: string }) => {
+  // WARNING: This is for demo purposes only.
+  // We don't encourage this in real apps. There are far safer ways to access
+  // data in a real application!
+  const users = (await db.query(
```
dai-shi, line 8: UserList.tsx imported from Couter.tsx, which is a client component.
In that case, you can't use `db` here under client component tree.
You want to import UserList.tsx from App.tsx, which is a server component.

sayjeyhi: Is it a part of the limitations of `wakuwork` implementation or it is a global concept?
cuz I saw documents like this, which allows importing the Server component into the Client component.
4cisrfhurwmv7ja9b4ei

dai-shi: It's not the limitation of wakuwork. It's how RSC is designed.
The shared component can't have server specific code.

dai-shi:
ref: https://github.com/dai-shi/wakuwork/issues/17#issuecomment-1535566006

Closing this for now, but I'm _very_ curious how people use it.

PR #20, opened 2023-04-12
Loading spinner

dai-shi:
Just to mitigate UX by showing a little spinner when we don't have SSR (which is default.)

https://user-images.githubusercontent.com/490574/231416726-929a3212-2b74-47dc-949a-ce0649288732.mov

### DIFF
--- a/examples/01_counter/index.html
+++ b/examples/01_counter/index.html
@@ -3,9 +3,31 @@
   <head>
     <meta charset="utf-8" />
     <title>Wakuwork example</title>
+    <style>
+      @keyframes spinner {
+        to {
+          transform: rotate(360deg);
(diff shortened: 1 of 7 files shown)


examples/01_counter/index.html
```diff
@@ -3,9 +3,31 @@
...
+        left: 50%;
+        width: 36px;
+        height: 36px;
+        margin-top: -18px;
+        margin-left: -18px;
+        border: 2px solid #ddd;
+        border-top-color: #222;
+        border-radius: 50%;
+        animation: spinner 1s linear infinite;
+      }
```
dai-shi, line 24: I wonder if there's smaller code version.

PR #31, opened 2023-05-08
fix: support single bracket

Himself65:


src/node-loader.ts
```diff
@@ -29,7 +29,7 @@ export async function load(url: string, context: any, nextLoad: any) {
     }
     // HACK pull directive to the root
     // Hope we can configure tsx to avoid this
-    const p = source.match(/(?:^|\n|;)("use (client|server)";)/);
+    const p = source.match(/(?:^|\n|;)(['"]use (client|server)["'];)/);
```
dai-shi, line 32: This is after transpilation, so it never becomes single quote.
I just confirmed that `'use client'` in 01_counter/src/Counter.tsx works without this change.

PR #32, opened 2023-05-08
feat: support e2e test using playwright

Himself65:
Wakuwork is a nice project, but the one concern is the lack of test cases. I tried to add vitest support, but since this is more like a server cli, I think `playwright` is more suitable for this repo

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,3 +12,11 @@ jobs:
         with:
           node-version: 18
       - run: npm test
+
+  e2e:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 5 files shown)


dai-shi:
Thanks for working on it. Yeah, e2e test seems a better fit.
Please feel free to add more tests with other examples. (Can we test "Pending..."?)

dai-shi (approved):
Looks good. Thanks for your contribution!
I will modify github workflows later.

dai-shi:
It would also be nice to test with production build and SSG.

Himself65:
I think playwright also supports no javascript so you can test pure server side result as well

PR #50, opened 2023-05-26
fix: rscPrefix config without trailing slash

dai-shi:
To support uncommon config just in case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- fix: rscPrefix config without trailing slash #50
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- fix: rscPrefix config without trailing slash #50
```
Himself65, line 5: do you have a CI to auto-generate the CHANGELOG?

dai-shi: No, for now it's done manually. (In jotai, it's generated by GitHub. We may follow it eventually.)

PR #53, opened 2023-05-29
vercel build output api

dai-shi:
there's an issue with the website: https://twitter.com/dai_shi/status/1661858992375894017
this should fix it.


src/lib/builder.ts
```diff
@@ -184,5 +233,21 @@ export async function build() {
     JSON.stringify(packageJson, null, 2)
   );
 
+  // https://vercel.com/docs/build-output-api/v3
+  // So far, only static sites are supported.
+  createVercelOutput(
```
dai-shi, line 238: We'll make it configurable with options once we have more sophisticated CLI, but for now it's enabled by default.

dai-shi (commented):
Should be good.

Anyway, the plan is to rewrite the builder for the next step.

PR #60, opened 2023-06-03
feat: enhance waku cli

Himself65:
``` shell
> waku --version  
0.11.1
```

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2,35 +2,30 @@
 
 import fs from "node:fs";
 import path from "node:path";
-import { parseArgs } from 'node:util'
-import { createRequire } from 'node:module'
+import { parseArgs } from "node:util";
+import { createRequire } from "node:module";
 
-const require = createRequire(
-  new URL('.', import.meta.url)
-)
+const require = createRequire(new URL(".", import.meta.url));
 
-const {
-  values,
-  positionals
-} = parseArgs({
+const { values, positionals } = parseArgs({
   args: process.argv.splice(2),
   allowPositionals: true,
   options: {
     config: {
-      type: 'string'
+      type: "string",
     },
     version: {
-      type: 'boolean',
-      short: 'v'
+      type: "boolean",
+      short: "v",
     },
     help: {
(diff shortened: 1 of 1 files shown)


Himself65:
I don't know why it failed. Could you please give me the log file?

dai-shi:
> I don't know why it failed. Could you please give me the log file?

The config is changed for #53. Once it's merged into main, you can merge it here.

src/cli.ts
```diff
@@ -27,6 +31,11 @@ switch (cmd) {
   case "start":
     runStart();
     break;
+  case "--version": {
```
dai-shi, line 34: I feel like we should consider adding some CLI lib. (Same applies to `create-waku`.)
I will look for something that fits well.

dai-shi (commented):
I could merge it, as long as it's short-lived until we refactor soon. (because I want to add more options in the future like `--disable-ssr` 😉 )

But, I'm curious if this is super important and what's your use case.

src/cli.ts
Himself65: Nodejs has official support for args. If you like it I could add the support

https://nodejs.org/dist/latest-v18.x/docs/api/util.html#utilparseargsconfig

dai-shi: Oh, that looks good. I think we can use it meanwhile.

dai-shi:
Would you try using https://nodejs.org/dist/latest-v18.x/docs/api/util.html#utilparseargsconfig ?

I'm still curious about your use case.

src/cli.ts
```diff
@@ -2,33 +2,67 @@
 
 import fs from "node:fs";
 import path from "node:path";
+import { parseArgs } from 'node:util'
+import { createRequire } from 'node:module'
+import * as process from 'process'
```
dai-shi, line 7: Can we make it this for consistency?
```suggestion
import process from 'node:process'
```

Himself65: My mistake, process is a global var

dai-shi (approved):
LGTM!

Issue #49, opened 2023-05-25
"use client" warnings


Tobbe:
Running `yarn build` I see 
`Module level directives cause errors when bundled, "use client" in "src/Counter.tsx" was ignored.`
![image](https://github.com/dai-shi/waku/assets/30793/5e4de403-04cd-4280-84fd-55ce456c9b18)

I thought this would have been solved as of `@vitejs/plugin-react` v4. And just to make sure I was actually using the correct version I ran `yarn why`
![image](https://github.com/dai-shi/waku/assets/30793/c3c1010a-e70a-46fb-9b59-290970f7f63e)

The fix is mentioned under "Other notable changes" in the release notes https://github.com/vitejs/vite-plugin-react/releases/tag/plugin-react%404.0.0

So not sure why I'm still seeing it

dai-shi:
> So not sure why I'm still seeing it

Me neither 😓

Himself65:
Oh I know why!

Himself65:
It use `warning.message.includes('use client')`, but we are using `"use client"` which is double quote😅

https://github.com/vitejs/vite-plugin-react/blob/bf8e967fd63b863b794db1a902c956a31d72d4ff/packages/plugin-react/src/index.ts#L287

Himself65:
But now we still have warning `Error when using sourcemap for reporting an error: Can't resolve original location of error.`

PR #63, opened 2023-06-03
fix: 'use client' warning

Himself65:
Fixes: https://github.com/dai-shi/waku/issues/49


src/lib/builder.ts
```diff
@@ -69,6 +70,27 @@ const resolveFileName = (fname: string) => {
   return "";
 };
 
+// Upstream issue: https://github.com/rollup/rollup/issues/4699
```
dai-shi, line 73: I hope it will be solved upstream soon.

src/lib/builder.ts
```diff
@@ -131,6 +153,7 @@ export async function build() {
     build: {
       ssr: true,
       rollupOptions: {
+        onwarn,
```
dai-shi, line 156: As I understand, this will overwrite previously defined option. I'm not sure if Vite provides a way to only merge the option if unavailable. Let me know if there's such a way.

dai-shi (approved):
For now, let's merge it.

PR #67, opened 2023-06-04
add prettier and eslint

dai-shi:
so that we can develop collaboratively! Waku Waku!


Himself65 (changes requested):
Add lint check to CI?

dai-shi:
Our test includes formatting, linting, and type checking altogether.

Issue #66, opened 2023-06-04
ssr middleware


dai-shi:
https://twitter.com/dai_shi/status/1664804156501098497

I will make `ssr` middleware totally isolated from `rsc` middlware (assuming they work on different machines). Not sure how it goes yet.
Refactoring builder would be required in advance, and we might need some breaking change for `getBuilder` API.

PR #68, opened 2023-06-08
refactor: split builder

dai-shi:
This is a refactor to prefer for #66.


dai-shi:
I'll continue refactoring in follow-up PRs.

PR #74, opened 2023-06-12
feat: ssr middlware

### LINKED ISSUE #66: ssr middleware

dai-shi:
Close #66 

- [x] first experiment
- [x] dev server with examples/01_counter
- [x] dev server with examples/07_router
- [x] dev server with examples/* (except for 06_nesting)
- [x] prd server
- [x] build and generate static sites

https://excalidraw.com/#json=pSsBQOqkYX4O-TIgxNrZj,i_CsymP1VZiHIAa0dlLbNw

image


dai-shi:
Found a relevant comment: https://github.com/facebook/react/issues/25627#issuecomment-1444073227

> You can suppress the error you've thrown from the console in onRecoverableError handler in your hydrateRoot options.

Let's do that.

src/lib/middleware/ssr.ts
```diff
@@ -0,0 +1,169 @@
...
+      }
+      if (!htmlRes.body || !rscRes.body) {
+        passthrough.destroy(new Error("No body"));
+        return;
+      }
+      const htmlStr = await htmlRes.text(); // Hope stream works, but it'd be too tricky
+      const [preamble, postamble] = splitHTML(htmlStr);
+      passthrough.write(preamble, "utf8");
+      const data = createFromNodeStream(
+        Readable.fromWeb(rscRes.body as any), // FIXME how to avoid any?
```
Himself65, line 91: Update to `"@types/node": "^20.3.1"` could fix this issue

This is upstream bug

Himself65 (changes requested):

src/lib/config.ts
```diff
@@ -2,18 +2,49 @@ import { resolveConfig as viteResolveConfig } from "vite";
...
+  }
+  return [splitted[0] + startStr, endStr + splitted[1]];
+};
+
+const getFallback = (id: string) => {
+  if (id.endsWith("#Waku_SSR_Capable_Link")) {
+    return "waku/server#ClientOnly";
+  }
+  return "waku/server#ClientFallback";
+};
```
dai-shi, line 26: I tried to create a full SSR (= also render client components into HTML on server) example, with configuring `getFallback`, but there are a few issues and I gave it up doing in this PR:
- We need a module graph in DEV, which we don't have in our current architecture.
- In production mode, it should be easier, but the configuration would be tricker.
- What's really blocking is probably as we load React in a separate worker, we get "Invalid hook call." error. I don't think we can do SSR in the same worker, so we'd need a new separate worker and duplicate the work. That doesn't feel very RSC-first framework.

We will see it again in the future, maybe after we migrate to `react-server-dom-vite`, which is not yet planned.

examples/02_async/entries.ts
```diff
@@ -18,4 +18,14 @@ export default defineEntries(
       },
     };
   }
+  // getSsrConfig
+  // FIXME SSR streaming and RSC streaming conflict. Let's revisit it later.
```
dai-shi, line 22: I think I understand what's happening. We need to send our `<script>` in the middle of SSR streaming. Let me see how we could do that in a simple way... 🤔

dai-shi: #78

PR #82, opened 2023-06-25
redesign the website completely

ahmeddots:
This PR redesigns the website, based on the design [here](https://www.figma.com/file/YiYbwKYWpPqjajvyyxeQXg/waku.gg?type=design&node-id=0%3A1&mode=design&t=jfJeeR71ufJcyWxM-1).

It will also create a prettierrc for unified formatting, and change tailwind colours to the ones used in the theme.

There are some issues currently:
- sidebar of docs does not highlight the page you're on (possibly solve using user requested url)
- all icons are svg components (possibly better to have an 'assets' folder for storing svg as files)
- original link component was restyled, but not sure whether to keep it or change into something else


.prettierrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "tabwidth": 2,
+  "useTabs": false
```
dai-shi, line 3: These are defaults and we don't plan to change the defaults anyway. Doesn't it work without this file?

website/index.html
```diff
@@ -29,9 +49,6 @@
     <div id="root">
       <div class="spinner"></div>
     </div>
-    <!--/placeholder1-->
-    <script src="/src/index.tsx" async type="module"></script>
-    <!--placeholder2-->
-    <!--/placeholder2-->
```
dai-shi, line 35: Those placeholder tags (both open and close) are required now.

dai-shi (changes requested):

dai-shi:
Thanks for your work!!

> * sidebar of docs does not highlight the page you're on (possibly solve using user requested url)

Ah, this is tricky. I need to think about it.

> * all icons are svg components (possibly better to have an 'assets' folder for storing svg as files)

It think 'assets' folder works. I haven't tried though.

> * original link component was restyled, but not sure whether to keep it or change into something else

I'll see what's changed.

.prettierrc.json
ahmeddots: Some people have different prettier configuration (like me), and this will force their editor to use these settings.

website/index.html
ahmeddots: I added them again in the latest commit. I believe it is correct now.

ahmeddots:
It now deploys correctly, so you can view it.

.prettierrc.json
dai-shi: Oh, interesting. Does having an empty config in `package.json` help?

dai-shi:
> It now deploys correctly, so you can view it.

Cool! Looks great!

Here are some feedbacks:
- The logo looks a bit different. Can we use a similar font like the site?
- I clicked the down arrow and nothing happened.
- The dot loading indicator is nice and pretty (I like it) but it shows unnecessarily (some logic changed?)
- "Introduction" is always highlighted 👉 #84

dai-shi:
https://github.com/dai-shi/waku/pull/82#discussion_r1242887384

Please remove `.prettierrc.json` file and add a "prettier" entry with empty config in package.json.

dai-shi (approved):
Added some commits to be mergeable.

dai-shi:
> * The logo looks a bit different. Can we use a similar font like the site?

It still remains.

> * I clicked the down arrow and nothing happened.

It still remains.

> * The dot loading indicator is nice and pretty (I like it) but it shows unnecessarily (some logic changed?)

Fixed, but in a hacky way.

> * "Introduction" is always highlighted 👉 [how to get current pathname in server component? #84](https://github.com/dai-shi/waku/issues/84)

It remains, but no easy solution yet.

I see a warning in console.
`Warning: validateDOMNesting(...): <pre> cannot appear as a descendant of <p>.`

Please feel free to work on them and file follow-up PRs.

ahmeddots:
> > * I clicked the down arrow and nothing happened.
> 
> It still remains.

As far as I know, the only way to create a collapse menu is to use js, which I tried not to do. For now it can act as a heading.

dai-shi:
> For now it can act as a heading.

I thought it jumps to introduction, the same as `Get Started >` button does.

Merging it anyway.

PR #86, opened 2023-06-27
feat: server context

### LINKED ISSUE #54: Feature: support reading request "context" in server actions
My goal here is to support session-based authentication for server actions in RSC.

Authentication for server actions seems like an open question at this point, but step one is making some kind "current context" from the caller available to the action. For a client holding access tokens, it's probably trivial to wire up helpers to invoke actions with. However, for battle tested session cookie-based authentication, there seems to be some work to do.

The same problem could surface for any other data sourced from the server request that needs to be supplied to actions.

As far as prior art, the next.js examples imply a backend like AsyncLocalStorage, given [this code](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions#actions):
```js
import { cookies } from 'next/headers';
 
export default function AddToCart({ productId }) {
  async function addItem(data) {
    'use server';
 
    const cartId = cookies().get('cartId')?.value;
    await saveToDb({ cartId, data });
  }
 
  return (
    <form action={addItem}>
      <button type="submit">Add to Cart</button>
    </form>
  );
}
```

To start, I briefly explored adding support for reading request headers in server actions. Two things got in the way:
1. The action handlers are run in a worker, separately from the express handler. This means even on the server, anything originating from the request cannot share module state with an action, and instead must cross a thread boundary via postMessage. This stopped a simple `AsyncLocalStorage` approach from working. I'm curious if RSC prescribes the worker in some way
2. The server code is bundled and loaded by Vite. This is what stopped me (for now) from simply importing a module from the server code and seeding `AsyncLocalStorage` with something around this spot: https://github.com/dai-shi/waku/blob/0848fa64eac2a578f602d7ab460889a8124bbbb0/src/lib/rsc-handler-worker.ts#L228C1-L229 The bundled action code gets its own module graph, and I couldn't easily figure out how to access a known module in that graph from the worker source to seed it with a value.

That fiddling can be found here, if interested: https://github.com/dai-shi/waku/compare/main...jwbay:request-headers?expand=1

The interesting output is:
```
✓ built in 1.15s
clientEntries { 'assets/rsc0.js': 'assets/rsc0-1b0f0351.js' }
Creating header store @     at eval (/home/admin/source/waku/examples/05_mutation/dist/assets/headers-storage-b5646013.js:5:40)
3:44:00 PM [vite] page reload dist/public/index.html
Listening on 8080
Creating header store @     at file:///home/admin/source/waku/dist/lib/headers-storage.js:2:40
Creating header store @     at eval (/home/admin/source/waku/examples/05_mutation/dist/assets/headers-storage-b5646013.js:5:40)
user-agent in worker src:  Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/114.0
user-agent in action:  undefined
```

Possibly the simplest alternative is to `.call/.apply` the server action with a `this` value containing request context. Making it a first/last parameter is also possible, but would require some TS trickery.

dai-shi:
(It's different from React Server Context.)

It's to communicate with `req` and `res`.

Close #54


src/lib/middleware/rsc.ts
```diff
@@ -14,9 +14,14 @@ type Middleware = (
 
 const { decodeReply, decodeReplyFromBusboy } = RSDWServer;
 
-export function rsc(options: {
+export function rsc<Context>(options: {
   mode: "development" | "production";
+  prehook?: (req: IncomingMessage) => Context;
+  posthook?: (res: ServerResponse, ctx: Context) => void;
```
dai-shi, line 20: Any suggestion for naming these?

dai-shi:
It's kind of WIP and we have some hacks, which should be fixed later.
But, merging it anyway to collect some feedbacks.

PR #101, opened 2023-07-07
fix: update bin to fix command not found when run examples

EasonYou:
# What's the issue
When clone the repo at the **first time**, can not run examples after running pnpm install.
It tells waku: command not found.
If waku link was built in the dir before, it would not warn the error.
image

# What's the platform
Run in mac OS 13.2.1 (22D68), Node v18.16.0

# How to fix
I found the bin property is corrent, but it can't work
I re-clone the repo and update the bin then run pnpm install, it fixed.
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,9 @@
       "default": "./dist/router/server.js"
     }
   },
-  "bin": "./dist/cli.js",
+  "bin": {
+    "waku": "./dist/cli.js"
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
I don't know why this fixes (a bug in pnpm?), but fine to merge. Thanks for your contribution!

Issue #90, opened 2023-06-28
Support `react-tweet`


dai-shi:
If we wanted to use https://github.com/vercel-labs/react-tweet, CSS Modules should be supported on server.

~Currently, it shows some errors with Vite ssrLoadModule.~ Fixed in #98 and WIP #105

We still have issues. Can anyone help?

himself65:
What is the error? I think the logic here is to collect `*.module.css` files and append them to response html?

dai-shi:
`ssrLoadModule` doesn't understand `.module.css` extension. It would be the easiest if we can simply ignore it.

dai-shi:
Here's the error message. My expectation is that Waku doesn't handle the logic directly, but instead let Vite do it.

```
[vite] Error when evaluating SSR module /src/components/App.tsx: failed to import "/src/components/App.module.css"
|- TypeError: Cannot read properties of undefined (reading 'set')
    at TransformContext.transform (file:///.../waku/node_modules/.pnpm/vite@4.3.9_@types+node@20.3.1/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:37919:29)
    at async Object.transform (file:///.../waku/node_modules/.pnpm/vite@4.3.9_@types+node@20.3.1/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:42919:30)
    at async loadAndTransform (file:///.../waku/node_modules/.pnpm/vite@4.3.9_@types+node@20.3.1/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:53385:29)
```

dai-shi:
with #98 the error is gone, but there are no styles.

dai-shi:
I will create a follow-up issue for #98 to support bare css modules.

dai-shi:
#105 is created and this issue is specific to support `react-tweet.`

dai-shi:
It seems to somewhat work for production build, but not for dev mode. We need someone who is familiar with Vite to help.

Aslemammad:
#171 This might fix this.

Aslemammad:
I'll work on this tomorrow! It's a matter of seeing if it works or not :)

Issue #96, opened 2023-07-01
build error in SWR 2.2.0


himself65:
```tsx
"use client";

import { useState } from "react";
import useSWR from 'swr'

export const Counter = () => {
  // @ts-ignore
  const { data } = useSWR('hello', () => 'world', {
    suspense: true
  })
  const [count, setCount] = useState(0);
  return (
    <div style={{ border: "3px blue dashed", margin: "1em", padding: "1em" }}>
      <p>Count: {count}</p>
      <button onClick={() => setCount((c) => c + 1)}>Increment</button>
      <h3>This is a client component.</h3>
      <div>
        {data}
      </div>
    </div>
  );
};
```

```shell
/Users/himself65/.nvm/versions/node/v18.16.1/bin/npm run build

> affine-waku@0.1.0 build
> waku build

vite v4.3.9 building SSR bundle for production...
✓ 6 modules transformed.
✓ built in 42ms
"default" is not exported by "node_modules/swr/core/dist/react-server.mjs", imported by "src/Counter.tsx".
file: /Users/himself65/Code/affine-waku/src/Counter.tsx:4:7
2: 
3: import { useState } from "react";
4: import useSWR from 'swr'
          ^
5: 
6: export const Counter = () => {
file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:2124
        base = Object.assign(new Error(base.message), base);
                             ^

Error [RollupError]: "default" is not exported by "node_modules/swr/core/dist/react-server.mjs", imported by "src/Counter.tsx".
    at error (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:2124:30)
    at Module.error (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:13463:16)
    at Module.traceVariable (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:13884:29)
    at ModuleScope.findVariable (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:12429:39)
    at ReturnValueScope.findVariable (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:6965:38)
    at ChildScope.findVariable (file:///Users/himself65/Code/affine-waku/node_modules/rollup/dist/es/shared/node-entry.js:6965:38)
[...]

Issue #79, opened 2023-06-23
not compatible with prisma


vincenteof:
Trying to add prisma to an waku project, but seems that they are not compatible. `@prisma/client` doesn't have esm export. When the server component render, error shows:
`Cannot render RSC ReferenceError: module is not defined`
[Here](https://github.com/vincenteof/waku-with-prisma) is the repo link which reproduces the error. 
I am not familiar with vite, but I have followed [Server-Side Rendering](https://vitejs.dev/guide/ssr.html) section on vite official doc and add prisma to it, and this kind of error doesn't show, not sure it's a bug from vite or the waku behavior. 
[Here](https://github.com/vincenteof/vite-ssr-with-prisma) is the repo link in which prisma works with vite.

dai-shi:
Hi, thanks for reporting!
I'll look into it next week.
Meanwhile, if anyone is interested, feel free to dig in.

vincenteof:
After debugging for sometime, it seems that `@prisma/client` is not externalized when running in vite ssr mode. 
```
  ssr: {
    noExternal: /^(?!node:)/,
    // FIXME this is very adhoc.
    external: ["react", "minimatch", "react-server-dom-webpack"],
  }
```
I can provide the config and override it, and everything is ok.
But shouldn't we externalize most of the dependencies when rendering RSC?

dai-shi:
> After debugging for sometime, it seems that `@prisma/client` is not externalized when running in vite ssr mode.

Nice finding!

> ```
>   ssr: {
>     noExternal: /^(?!node:)/,
>     // FIXME this is very adhoc.
>     external: ["react", "minimatch", "react-server-dom-webpack"],
>   }
> ```
> 
> I can provide the config and override it, and everything is ok.

Do you mean you edit your `vite.config.ts` to add the `external` field? That sounds like a valid solution for now.

> But shouldn't we externalize most of the dependencies when rendering RSC?

Yes, we should. I just haven't been able to find a solution yet. I would appreciate for any help.

vincenteof:
> Do you mean you edit your `vite.config.ts` to add the `external` field? That sounds like a valid solution for now.

Yes, I provide something like
```
  ssr: {
    noExternal: /^(?!node:)/,
    external: ["react", "minimatch", "react-server-dom-webpack", "@prisma/client", "prisma"],
  }
```
in the vite config file.
> Yes, we should. I just haven't been able to find a solution yet. I would appreciate for any help.

[...]

Issue #21, opened 2023-04-15
Support server bundling with vite


dai-shi:
Currently, wakuwork uses `react-server-dom-webpack/node-loader` for handing `"use client"` and `"use server"`.
We use vite for client bundling, but for server bundling we use a naive approach.
It's limited. For example, we now don't support `"use client"` in `node_modules`.

I'm hoping we could use vite for server bundling. Wish [vite-rsc](https://github.com/cyco130/vite-rsc/tree/main/packages/vite-rsc) does that.

@nksaraf What's the current status?

dai-shi:
I will see if I can do it in this repo.

hyf0:
Have you thought through it?

I guess we just need to adapt the [register logic](https://github.com/facebook/react/blob/main/packages/react-server-dom-webpack/src/ReactFlightWebpackNodeRegister.js#L200) in Rollup and apply it for `use client` `components` in `transform` hook.

dai-shi:
Well, I'm learning vite. Will open a PR if something is ready.

dai-shi:
Having a trouble...

PR #103, opened 2023-07-10
feat: use node-loader again

### LINKED ISSUE #44: pnpm doesn't work with waku/router
https://github.com/dai-shi/waku/blob/07ac23eb174e439ae73185495924a8a4bde90f75/src/middleware/lib/rsc-handler-worker.ts#L180-L186

dai-shi:
Relates with #79 and #96

It partly reverts the idea in #21.

Closes #44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Changed
 - fix: partially for css modules #98
+- feat: use node-loader again #103
 
 ## [0.13.0] - 2023-07-06
 ### Changed
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ yarn create waku
 ```
 
 ```bash
-pnpm create waku # pnpm not working for now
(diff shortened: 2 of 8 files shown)


dai-shi:
This doesn't fix #90. react-tweet uses CSS Modules, so externalizing it causes this error: `[ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".css" for ...`

I have no idea how to fix this.

dai-shi:
If I added `vite.config.ts` for `react-tweet`,
```ts
import { defineConfig } from "waku/config";

export default defineConfig({
  ssr: {
    noExternal: ["react-tweet"],
  },
});
```
Now, it shows a different error: `Warning: Invalid hook call. Hooks can only be called inside of the body of a function component.`

Maybe because it has two different React instances?

dai-shi:
Not sure how many issues this will solve, but it should be good for now. (We don't know the future. There might be another/better way.)

Issue #81, opened 2023-06-25
create waku error


himself65:
```shell
yarn create waku
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0000: └ Completed
➤ YN0000: Done in 0s 69ms

node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: connect ECONNREFUSED 0.0.0.0:443
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16)
Emitted 'error' event on ClientRequest instance at:
    at TLSSocket.socketErrorListener (node:_http_client:502:9)
    at TLSSocket.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '0.0.0.0',
  port: 443
}

Node.js v18.16.0
```

dai-shi:
It's working for me:
```bash
$ yarn create waku                       
yarn create v1.22.19
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...

success Installed "create-waku@0.4.7" with binaries:
      - create-waku
[####] 4/4Done! Change directory "waku-example"
✨  Done in 6.09s.
```

himself65:
I think this happened because the HTTP proxy, since we need access internet through VPN

dai-shi:
Has it ever worked before then? What would be a solution?
(Related: https://github.com/dai-shi/waku/pull/52#issuecomment-1566325047 )

childrentime:
@dai-shi Can we avoid fetching template code from Github and instead pull it directly from the file system and install its dependencies, just like create-next-app does?

dai-shi:
Yeah, this is still a hack which was slightly better than `curl` (which was the first hack.)
Let's improve it after #64.

Meanwhile, you don't need to use it and just grab the example instead somehow.

PR #52, opened 2023-05-27
feat: support input dirName in create-waku

himself65:

PR #89, opened 2023-06-28
Add .gitignore to Simplify Project Creation from Template

JiantingFeng:
I noticed that in creating a new project from the template, some files (like local environment files, build files, etc.) that should not be pushed to the repo are being included. This causes additional steps for the users to remove or update these files to suit their specific project.

I have added a `.gitignore` file to the template to streamline this process and make it more user-friendly. This file includes standard patterns for ignoring files that are typically unnecessary in the repo.

This will make the project creation process more straightforward and help maintain the cleanliness and relevance of the repository files.

You can see the changes I have made in this pull request. 

I appreciate your time and consideration. Kindly review these changes and let me know if there are any suggestions or issues.

Issue #64, opened 2023-06-03
use monorepo packages stucture


himself65:
```
- packages
  - waku
  - create-waku
  - website
- examples
  - 01_xxx
  - 02_xxx
```

dai-shi:
I have been generally against monorepo for my projects (especially `jotai`), but it actually makes sense for this project. In fact, I considered it once when I added `create-waku`, put it in `contrib` for the moment, and thought eventually move to monorepo.

I'm not very experienced with monorepo setup. Would you be able to help on it? I assume we use pnpm workspace.

himself65:
I'm busy recently, but I will work on this if I have time.

nickytonline:
Would the project want to go with [npm workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces), or something like [Nx](https://nx.dev/)?

himself65:
Dont use nx :(

himself65:
Pnpm workspace is enough. I have idea but no time to setup

PR #115, opened 2023-08-03
feature: better frameworks like cli

vasucp1207:
Method like other frameworks use(mainly **vue**).

### Cli including,
- Dynamically project name.
- Choose from different default templates.
- Overriding the directory if it exits after confirmation.
- Validates package name.

Should try to commit in https://github.com/dai-shi/waku/pull/52 but I messed up and also thought it's a completely new strategy so I raised this PR instead.


vasucp1207:
Regarding https://github.com/dai-shi/waku/pull/89 , templates only contain necessary files it does not include folders like node_modules etc. Well if you run ``npm run build`` a resulted ``out.cjs`` file is generated, but I can't figure out now how to publish it to npm to actually check it. Do you have any idea about this.

dai-shi:
> Regarding #89 , templates only contain necessary files it does not include folders like node_modules etc.

As I understand, #89's request is just to put `.gitignore` file in those templates.
But, on second thought, I'm not sure what should be in there.
Does create-vue generate `.gitignore` file?

contrib/create-waku/cli.js
```diff
@@ -1,46 +1,148 @@
 #!/usr/bin/env node
 
-const path = require("node:path");
-const fs = require("node:fs");
-const https = require("node:https");
+import fs from 'fs'
+import path from 'path'
```
dai-shi, line 4:
```suggestion
import fs from 'node:fs'
import path from 'node:path'
```

contrib/create-waku/cli.js
```diff
@@ -1,46 +1,148 @@
...
   }
-});
+  console.log(`${bold(green(commandsMap.install[packageManager]))}`)
+  console.log(`${bold(green(commandsMap.dev[packageManager]))}`)
+  console.log()
+}
+
+init().catch((e) => {
+  console.error(e)
+})
```
dai-shi, line 148: Let's refactor those files in TypeScript after #64. cc @himself65

contrib/create-waku/package.json
```diff
@@ -1,5 +1,15 @@
 {
   "name": "create-waku",
   "version": "0.4.10",
-  "bin": "./cli.js"
+  "bin": "./cli.js",
+  "type": "module",
+  "scripts": {
+    "build": "./node_modules/.bin/esbuild cli.js --bundle --platform=node --outfile=out.cjs",
```
dai-shi, line 7: wait, you want to build it before publish? we don't have CD yet.

contrib/create-waku/package.json
```diff
@@ -1,5 +1,15 @@
...
+    "test": "node ./tests/generateTemplates.test.js"
+  },
+  "dependencies": {
+    "esbuild": "0.18.17",
```
dai-shi, line 11: it should be in devDependencies.

dai-shi (changes requested):
One big issue for me is now we have to maintain both examples and templates.
As we are adding the "build" process, we should copy files from examples folder, instead of copying them.

contrib/create-waku/cli.js
dai-shi: It seems this adds the build process anyway, so let's make code in TypeScript.

contrib/create-waku/package.json
dai-shi: Can we use swc instead of esbuild?

contrib/playground/README.md
```diff
@@ -0,0 +1 @@
+## For testing purpose
```
dai-shi, line 1: please remove this.

contrib/create-waku/package.json
vasucp1207: Remove esbuild

vasucp1207:
> > Regarding #89 , templates only contain necessary files it does not include folders like node_modules etc.
> 
> As I understand, #89's request is just to put `.gitignore` file in those templates. But, on second thought, I'm not sure what should be in there. Does create-vue generate `.gitignore` file?

Yes, create-vue generates a `.gitignore` file, I think a minimal file is okay for this like [create-vue .gitignore](https://github.com/vuejs/create-vue/blob/9bba10e2ff2dba6acfc130752b74addc0f99b1e6/template/base/_gitignore#L4)

contrib/create-waku/cli.js
```diff
@@ -1,46 +1,148 @@
...
+  return !fs.existsSync(dir) || fs.readdirSync(dir).length === 0
+}
+
+async function init() {
+  const cwd = process.cwd();
+
+  let targetDir;
+  let defaultProjectName = 'waku-project'
+
+  const CHOICES = fs.readdirSync(path.resolve(cwd, '../../examples'))
```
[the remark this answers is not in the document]
vasucp1207: > This doesn't work as we don't publish examples dir.
> 
> We have two options.
> 
> 1. add build step to copy examples dir
> 2. download from github
> 
> Eventually, we should go with 1 because it fixes #81, but for now 2 is also fine. If we were to go with 1, we would like to make the code in TypeScript and transpile with SWC in the build step.

Summary,
- If we ``npm run build:esbuild`` an ``out.js`` file generated.
- Running ``npm run build:estemplate`` generates an ``copy.js`` file.
- If we run ``node copy.js`` templates copied from ``example`` dir and a ``template`` folder generated.
- Finally, running ``node cli.js`` works fine.

Problem with transpiling with ``swr``,
- If we transpiling files with ``swr`` then the ``imports`` in output files are not working ex,
```js
import { renderTemplate } from './renderTemplate'
```
This import statement would not be able to resolve the file.

Is, I'm missing something about this ``swr``.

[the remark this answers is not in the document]
dai-shi: Okay, I don't mind using esbuild for now. I just thought SWC might work better and we also use it in `waku`. We can try that later.

Can you move TS files into `./src` folder please?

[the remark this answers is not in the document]
vasucp1207: Yeah, sure we can look around it in the near future.

contrib/create-waku/package.json
```diff
@@ -1,5 +1,22 @@
 {
   "name": "create-waku",
   "version": "0.4.10",
-  "bin": "./cli.js"
+  "bin": "out.js",
+  "type": "commonjs",
+  "scripts": {
```
dai-shi, line 6: Let's add this:

```suggestion
  "files": [
    "src",
    "dist"
  ],
  "scripts": {
```

contrib/create-waku/package.json
```diff
@@ -1,5 +1,22 @@
 {
   "name": "create-waku",
   "version": "0.4.10",
-  "bin": "./cli.js"
+  "bin": "out.js",
+  "type": "commonjs",
```
dai-shi, line 5: I thought your first proposition was "module". Which do you prefer?
Ah, I see it's because esbuild outputs CJS.

contrib/create-waku/package.json
```diff
@@ -1,5 +1,22 @@
 {
   "name": "create-waku",
   "version": "0.4.10",
-  "bin": "./cli.js"
+  "bin": "out.js",
```
dai-shi, line 4: can you make it `dist/cli.js`?

contrib/create-waku/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "target": "ESNext",
+    "moduleResolution": "node",
+    "allowSyntheticDefaultImports": true,
+    "outDir": "dist",
+    "rootDir": "./",
+    "strict": true,
+  },
+  "include": ["./"],
+  "exclude": ["node_modules", "dist"]
+}
```
dai-shi, line 12: Let's make it consistent with the project config. we use module=nodenext for example.
Or, can we simply remove this file?

contrib/create-waku/src/renderTemplate.ts
```diff
@@ -0,0 +1,34 @@
...
+      renderTemplate(path.resolve(src, file), path.resolve(dest, file));
+    }
+    return;
+  }
+
+  const filename = path.basename(src);
+
+  if (filename === 'package.json' && fs.existsSync(dest)) {
+    // merge instead of overwriting
+    const pkg = deepMerge(
```
dai-shi, line 20: why do we need deep merge here?

contrib/create-waku/src/renderTemplate.ts
```diff
@@ -0,0 +1,34 @@
...
+      // @ts-expect-error
+      JSON.parse(fs.readFileSync(dest)), JSON.parse(fs.readFileSync(src))
+    );
+    fs.writeFileSync(dest, JSON.stringify(pkg, null, 2) + '\n');
+    return;
+  }
+
+  if (filename.startsWith('_')) {
+    // rename `_file` to `.file`
```
dai-shi, line 29: Can we simply have dot files?

dai-shi (commented):
Thanks for working on it. Here are some reviews.

contrib/create-waku/package.json
vasucp1207: Should we also include ``template`` in this?

contrib/create-waku/tsconfig.json
vasucp1207: Yeah, removing this is fine for now.

contrib/create-waku/src/renderTemplate.ts
vasucp1207: For using ``projectName`` as ``name`` key in ``package.json`` we created the ``package.json`` already, so here we have to deep merge rest of the fields.

contrib/create-waku/package.json
dai-shi: Oh, good point. We need that too.

contrib/create-waku/src/renderTemplate.ts
dai-shi: Can't it be shallow merge??

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env node
+
+import fs from 'node:fs'
+import path from 'node:path'
+import prompts from 'prompts'
+import { red, green, bold } from 'kolorist'
+
+import { emptyDir } from './emptyDir'
```
dai-shi, line 8: We need an extension for module=nodenext. `.js` not `.ts`.
```suggestion
import { emptyDir } from './emptyDir.js'
```

contrib/create-waku/src/renderTemplate.ts
vasucp1207: Yes, its working.

contrib/create-waku/src/cli.ts
vasucp1207: Do you mean adding ``.js`` to all importing files?

dai-shi: https://www.typescriptlang.org/docs/handbook/esm-node.html

> As a result, it will have to be rewritten to use the extension of the output of foo.ts - so bar.ts will instead have to import from ./foo.js.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,153 @@
...
+        ],
+      }
+    ], {
+      onCancel: () => {
+        throw new Error(red('✖') + ' Operation cancelled')
+      }
+    })
+  } catch (cancelled) {
+    if(cancelled instanceof Error) 
+      console.log(cancelled.message)
```
dai-shi, line 91: Just a preference:
```suggestion
    if(cancelled instanceof Error) {
      console.log(cancelled.message)
    }
```

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  const pkg = { name: packageName?? toValidPackageName(targetDir), version: '0.0.0' }
+
+  fs.writeFileSync(
+    path.resolve(root, 'package.json'),
+    JSON.stringify(pkg, null, 2),
+  )
+
+  console.log("Setting up project...")
+  const templateRoot = path.resolve(cwd, 'template')
```
dai-shi, line 113: I think `path.resolve('template')` is the same. But, not super important anyway.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,153 @@
...
+    console.log(`${bold(green(`cd ${path.relative(cwd, root)}`))}`)
+  }
+  console.log(`${bold(green(commandsMap.install[packageManager]))}`)
+  console.log(`${bold(green(commandsMap.dev[packageManager]))}`)
+  console.log()
+}
+
+init().catch((e) => {
+  console.error(e)
+})
```
dai-shi, line 153: This file feels a little bit complicated for me. Maybe, we look for something in the future.

contrib/create-waku/src/copyTemplateDir.ts
```diff
@@ -0,0 +1,23 @@
...
+  let root = path.join(cwd, '/template')
+  
+  if(!fs.existsSync(root)) {
+    fs.mkdirSync(root)
+  }
+  
+  const CHOICES = fs.readdirSync(path.resolve(cwd, '../../examples'))
+  const templateRoot = path.resolve(cwd, '../../examples')
+  
+  for(let i = 0; i < 3; i++) {
```
dai-shi, line 16: Can `3` be `CHOICES.length`?

contrib/create-waku/src/renderTemplate.ts
```diff
@@ -0,0 +1,29 @@
...
+      renderTemplate(path.resolve(src, file), path.resolve(dest, file));
+    }
+    return;
+  }
+
+  const filename = path.basename(src);
+
+  if (filename === 'package.json' && fs.existsSync(dest)) {
+    // merge instead of overwriting
+    const pkg = shallowMerge(
```
dai-shi, line 20: Can shallow merging simply be `Object.assign({}, obj1, obj2)`?

contrib/create-waku/src/shallowMerge.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function shallowMerge(target: any, obj: any) {
+  for (const key of Object.keys(obj)) {
+    const oldVal = target[key]
+    const newVal = obj[key]
+
+    if(key === 'name' || key === 'version') continue
+
+    if (Array.isArray(oldVal) && Array.isArray(newVal)) {
+      target[key] = mergeArrayWithDedupe(oldVal, newVal)
```
dai-shi, line 11: ah, you want to deep merge arrays.

dai-shi (commented):
tbh, there are many things I'd like to change, like coding practice, naming, structuring.
I'd like to move this forward, but in order for me to refactor things, tests might be important.
If you have any ideas to simplify things for now, please let me know.

contrib/create-waku/src/renderTemplate.ts
```diff
@@ -0,0 +1,25 @@
...
+    fs.mkdirSync(dest, { recursive: true });
+    for (const file of fs.readdirSync(src)) {
+      renderTemplate(path.resolve(src, file), path.resolve(dest, file));
+    }
+    return;
+  }
+
+  const filename = path.basename(src);
+
+  if (filename === 'package.json' && fs.existsSync(dest)) {
```
dai-shi, line 17: Can you explain again when this happens?

dai-shi (commented):
Getting good. I still have many requests. Please be patient.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  if (shouldOverwrite) {
+    emptyDir(root)
+  } else if (!fs.existsSync(root)) {
+    fs.mkdirSync(root)
+  }
+
+  const pkg = { name: packageName?? toValidPackageName(targetDir), version: '0.0.0' }
+
+  fs.writeFileSync(
```
vasucp1207, line 108: Here, ``package.json`` already created with the input project name.

contrib/create-waku/src/renderTemplate.ts
vasucp1207: Maybe ``fs.existsSync(dest)``is not much relevant here.

contrib/create-waku/src/cli.ts
dai-shi: I mean `copyTemplateDir.ts` implementation can be much simplified because it doesn't need to "render".

dai-shi: So, I don't think we need `renderTempalte.ts` at all (it doesn't mean to just move code).
copyTemplateDir is basically `cp -r src dest` and we could implement it easily, or can use some npm packages.
In cli.ts, we copy from a template and "edit" `package.json`. We don't need to do it during copy. I guess it simplifies the logic. As we start using some dependencies, we could use something for it too.

Basically, we want to reduce our code.

dai-shi: Thanks, that's cool. Feels more maintainable.

contrib/create-waku/package.json
```diff
@@ -1,5 +1,27 @@
...
+  "files": [
+    "src",
+    "dist",
+    "template"
+  ],
+  "type": "commonjs",
+  "scripts": {
+    "start": "node dist/cli.js",
+    "template": "cp -r ../../examples template/",
+    "test": "node dist/generateTemplates.test.js",
```
dai-shi, line 14: How and when do I use this?

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,174 @@
...
+    const newPackageJson = JSON.parse(fs.readFileSync(newPackageJsonPath, 'utf-8'))
+
+    const mergedPackageJson = {
+      ...newPackageJson,
+      ...existingPackageJson
+    }
+
+    const result = spawnSync('cp', ['-r', `${templateDir}/`, root])
+
+    // Write the merged package.json back to the target directory
```
dai-shi, line 132: Nice.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,174 @@
...
+  return !fs.existsSync(dir) || fs.readdirSync(dir).length === 0
+}
+
+async function init() {
+  const cwd = process.cwd()
+
+  let targetDir = '';
+  let defaultProjectName = 'waku-project'
+
+  const CHOICES = fs.readdirSync(path.resolve(cwd, 'template'))
```
dai-shi, line 35: I'm curious if we can simplify this, because cwd is always `process.cwd()`.

```suggestion
  const CHOICES = fs.readdirSync('template')
```

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,174 @@
...
+    // Read new package.json from the template directory
+    const newPackageJsonPath = path.join(templateDir, 'package.json')
+    const newPackageJson = JSON.parse(fs.readFileSync(newPackageJsonPath, 'utf-8'))
+
+    const mergedPackageJson = {
+      ...newPackageJson,
+      ...existingPackageJson
+    }
+
+    const result = spawnSync('cp', ['-r', `${templateDir}/`, root])
```
dai-shi, line 130: This probably doesn't work on Windows.
Let's use some libraries. I often  use https://www.npmjs.com/package/shelljs, but there can be some others that fit well with this use case.

dai-shi (commented):
I like the new approach.

dai-shi:
Please also take care of CI errors.

contrib/create-waku/package.json
vasucp1207: Copy templates with ``npm run template`` before publishing the package, ``npm run start`` is just for running the cli locally.

dai-shi: should we run `npm test` after `npm run template`?

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+    process.exit(1);
+  }
+
+  const { packageName, shouldOverwrite, chooseProject } = result;
+
+  const root = path.join(cwd, targetDir);
+
+  if (shouldOverwrite) {
+    spawn.sync("rm", ["-rf", `${root}/*`]);
```
dai-shi, line 104: Interesting. So, cross-spawn translates it for windows? (or does windows support `rm -rf`?)

dai-shi (changes requested):
I'll take another look later. Please check CI errors meanwhile.

contrib/create-waku/src/cli.ts
vasucp1207: Updated with ``fs-extra`` node also have ``fs.cpSync`` but currently it's experimental so I'm using ``fs-extra``.
Can we have anyone who can test the cli on windows?

dai-shi: Maybe @Sharlottes @hyf0

dai-shi (commented):
Some more comments. I think we are getting close.

contrib/create-waku/src/cli.ts
hyf0: I'm back to mac world now. :-). Probably not able to be of help.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,184 @@
...
+  } catch (cancelled) {
+    if (cancelled instanceof Error) {
+      console.log(cancelled.message);
+    }
+    process.exit(1);
+  }
+
+  const { packageName, shouldOverwrite, chooseProject } = result;
+
+  const root = path.join(cwd, targetDir);
```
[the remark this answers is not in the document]
vasucp1207: For local development, cwd here is for the ``create-waku`` dir, ``template`` folder is copied under this.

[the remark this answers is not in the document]
dai-shi: We may want to try how it works with `npx`, but my assumption is cwd is where people run the command. So, it has to be something like this:
```js
const template = path.join(__dirname, "template");
const root = path.resolve(targetDir);
```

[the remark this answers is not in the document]
vasucp1207: Yeah, I get the point of using the ``__dirname`` this is relative to the script location.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,191 @@
...
+      pnpm: "pnpm dev",
+      yarn: "yarn dev",
+      npm: "npm run dev",
+    },
+  };
+
+  console.log(`\nDone. Now run:\n`);
+  if (root !== cwd) {
+    console.log(`${bold(green(`cd ${path.relative(cwd, root)}`))}`);
+  }
```
dai-shi, line 183: I don't think the condition becomes `false` in general:
```suggestion
  console.log(`${bold(green(`cd ${path.relative(cwd, root)}`))}`);
```

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,191 @@
...
+      existingPackageJsonPath,
+      JSON.stringify({
+        ...existingPackageJson,
+        name: packageName ?? toValidPackageName(targetDir),
+        version: "0.0.0",
+      }, null, 2)
+    );
+  };
+
+  render(chooseProject as string);
```
dai-shi, line 158: It's probably inferred correctly.
```suggestion
  render(chooseProject);
```

contrib/create-waku/src/generateTemplates.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    true,
+    console.log(bold(green(`Template generated successfully.`)))
+  );
+
+  // Clean up the generated file after the test
+  fse.removeSync(targetDir);
+}
+
+// Run the test
+testTemplateGeneration();
```
dai-shi, line 30: I don't think this test makes much sense. Let's remove it for now.
(I think we should add tests with Vitest, to cover each function, but it can be a separate PR.)

dai-shi (changes requested):
Please check CI errors too.

contrib/create-waku/src/generateTemplates.test.ts
vasucp1207: Yeah, I actually do not decide how to test this, so a separate PR in future is a good idea.

vasucp1207:
> Please check CI errors too.

I also run prettier, don't know why it fails

dai-shi:
Maybe we need to add the empty config line
https://github.com/dai-shi/waku/blob/33c41546aac3be3d706f22a055589ef82398b0f4/package.json#L93
in create-waku/package.json too.

contrib/create-waku/src/cli.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  const pkg = {
+    name: packageName ?? toValidPackageName(targetDir),
+    version: "0.0.0",
+  };
+
+  console.log("Setting up project...");
+  const templateRoot = path.join(__dirname, "../template");
+
+  const render = function render(templateName: string) {
```
dai-shi, line 117: Can we avoid the function? I don't necessarily prefer a long function (`init` in this case), and we generally should split into smaller functions (better for testing), but this doesn't seem necessary and `render` sounds confusing (we just copy and patch json).

dai-shi (commented):
Probably the last review comment.

contrib/create-waku/src/cli.ts
vasucp1207: Removed `render` as it is not necessary, splitting `init` in the same file or creating another file for the functions.

dai-shi (approved):
The code looks good to me.
I think we want to make the project monorepo at this point. I need to learn how to do it. Give me some time.

dai-shi:
@vasucp1207 #122 is merged. Please follow the new structure.

dai-shi:
Please check CI errors.

packages/create-waku/package.json
```diff
@@ -1,10 +1,29 @@
 {
   "name": "create-waku",
   "version": "0.4.10",
-  "author": "Daishi Kato",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/dai-shi/waku.git"
+  "bin": "dist/cli.js",
```
dai-shi, line 4:
```suggestion
  "author": "Daishi Kato",
  "contributors": [
    "Vasu Singh"
  ],
  "repository": {
    "type": "git",
    "url": "https://github.com/dai-shi/waku.git"
  },
  "bin": "./dist/cli.js",
```

packages/create-waku/package.json
```diff
@@ -2,9 +2,36 @@
...
+    "prompts": "^2.4.2"
+  },
+  "devDependencies": {
+    "@types/fs-extra": "^11.0.1",
+    "@types/node": "^20.4.7",
+    "@types/prompts": "^2.4.4",
+    "esbuild": "0.18.17",
+    "typescript": "^5.1.6"
+  },
+  "prettier": {}
```
dai-shi, line 36: Maybe we don't need it because we have in the repo root?
```suggestion
  }
```

packages/create-waku/src/cli.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  const { packageName, shouldOverwrite, chooseProject } = result;
+
+  const root = path.resolve(targetDir);
+
+  if (shouldOverwrite) {
+    fse.emptyDirSync(root);
+  } else if (!fs.existsSync(root)) {
+    fs.mkdirSync(root);
```
dai-shi, line 107: Curious if we should have the recursive option:
https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fsmkdirsyncpath-options

packages/create-waku/package.json
```diff
@@ -2,9 +2,36 @@
...
+  "bin": "./dist/cli.js",
+  "files": [
+    "src",
+    "dist",
+    "template"
+  ],
+  "type": "commonjs",
+  "scripts": {
+    "start": "node dist/cli.js",
+    "template": "cp -r ../../examples template/",
```
dai-shi, line 21: We should add `/packages/create-waku/template` in the .gitignore file.

dai-shi (changes requested):
Please check CI errors again.

packages/create-waku/src/cli.ts
vasucp1207: Yes, good for creating nested dir.

packages/create-waku/package.json
```diff
@@ -2,9 +2,35 @@
...
     "url": "https://github.com/dai-shi/waku.git"
   },
-  "bin": "./cli.js"
+  "bin": "./dist/cli.js",
+  "files": [
+    "src",
+    "dist",
+    "template"
+  ],
+  "type": "commonjs",
```
dai-shi, line 18: Future TODO: make type=module and use swc instead of esbuild.

packages/create-waku/src/cli.ts
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env node
+
+// FIXME the following is temporarily and will remove it soon
+/* eslint-disable */
+
```
dai-shi, line 5: My intention is to remove this in this PR.

```suggestion
```

packages/create-waku/package.json
vasucp1207: Sure, I will take look into this.

dai-shi (approved):
Let's merge it. Feel free to work on improvements later.

dai-shi:
https://www.npmjs.com/package/create-waku/v/0.5.1 is published.
@vasucp1207 Can you please check if it works as expected?

PR #117, opened 2023-08-05
fix: readable stream type issue workaround

PerfectPan:
Since fetch api will be used both in browser and node runtime in waku, it is better to just use type cast rather than shim it in global.d.ts.


dai-shi (approved):
Thanks. It's at least better than any. Nice to have those links in comment.

PR #118, opened 2023-08-06
experimental: add another router example

dai-shi:
Highly experimental. Not for merge.


dai-shi:
Our latest take is #293.

We also explore TanStack Router integration: #272

PR #120, opened 2023-08-17
fix: package deps to run examples from root

gabrielcsapo:
running pnpm install and running the examples I ran into 

Screenshot 2023-08-17 at 2 45 35 PM

Fixing the deps and running pnpm install I get a clean pnpm install and working demo

Screenshot 2023-08-17 at 2 48 15 PM
Screenshot 2023-08-17 at 2 48 27 PM
Screenshot 2023-08-17 at 2 50 31 PM

### DIFF
--- a/examples/01_counter/package.json
+++ b/examples/01_counter/package.json
@@ -13,7 +13,8 @@
     "react": "18.3.0-canary-7118f5dd7-20230705",
     "react-dom": "18.3.0-canary-7118f5dd7-20230705",
     "react-server-dom-webpack": "18.3.0-canary-7118f5dd7-20230705",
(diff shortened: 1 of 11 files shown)


dai-shi (commented):
Is this a new issue with latest pnpm? It's been working fine for me. 

The example is designed to be work as standalone, so the proposed change is not acceptable. Is there any workaround? We plan to make the project monorepo in the future. 
Not adding webpack is intentional. If it causes a real issue, we should look for a different solution.

dai-shi:
#122 is merged. Please give it a try. Feel free to open a new issue if something goes wrong.

PR #124, opened 2023-09-01
breaking: multiple elements in a single response

### LINKED ISSUE #84: how to get current pathname in server component?
```tsx
'use server';

const App = () => {
  // Does this possible
  const pathname = usePathname()
}
```

### LINKED ISSUE #85: unexpected cached result
```ts
export default defineRouter(
  async id => {
    switch (id) {
      case 'index': {
        const { default: AppCreator } = await import('./src/app.js');
        return AppCreator({
          name: 'this is index'
        });
      }
      case 'about': {
        const { default: AppCreator } = await import('./src/app.js');
        return AppCreator({
          name: 'this is about'
        });
      }
      default:
        return null;
    }
  },
  async () => {
    return ['index', 'about'];
  }
);
```

```tsx
const App = ({ name = "Anonymous" }) => () => {
  return (
    <div style={{ border: "3px red dashed", margin: "1em", padding: "1em" }}>
      <h1>Hello {name}!!</h1>
      <h3>This is a server component.</h3>
    </div>
  );
};

export default App;
```

image

image

### LINKED ISSUE #104: Ssr doesn't work if use getContext in entries.ts 's defineRouter/getComponent
# What's the issue

I use ssr with ``routes`` and ``getContext``

As the code show, it doesn't work if use ``getContext`` in defineRouter/getComponent.

https://github.com/EasonYou/waku-example/blob/71308df89396c7fa6b5c9ce54ff46d8fafcdf7aa/src/entries.ts#L1-L21

image

# What do I find
The error tells that ctx was empty, it cause by the ``getComponent`` so that waku can't get element by ``getSsrConfig`` because of the component import error. Then the ssr will not work. But no error on the client side, just like a csr app.

https://github.com/dai-shi/waku/blob/5af1c3461df9ac02472775cd16ae6d382d1726c9/src/router/server.ts#L155-L171

I don't know if it is a purposeful design that user can't use ``getSsrConfig`` or it is a bug.

### LINKED ISSUE #119: Redesign `serve()` and `mutate()` apis, and maybe `getEntry` too
There are a few things of my concern:
- `serve()` caches too much #84
- refetch on rerender might not be the best
- props serialization is tricky
- `mutate()` can only return one tree
  - Just noticed https://github.com/facebook/react/blob/997f52fbb30ec4b777b66edcdd75f594d9fe0c55/fixtures/flight/src/index.js#L19 returning multiple values
- `getEntry()` might not be the best
  - Should we expect to use `lazy`?

dai-shi:
There are many changes, but what's important is
- Supports multiple elements in one RSC payload
  - One single input string
- `entries.tsx` is changed
  - `renderEntries` instead of `getEntry`
- Explicit refetch on client
- Refactor `waku/router`
  - Do not cache on client
  - Pass `path` as a prop

Close #119
Close #84
Close #104
Close #85

Check list:
- [x] examples/01
- [x] examples/02
- [x] examples/03
- [x] examples/04
- [x] examples/05
- [x] examples/06
- [x] examples/07
- [x] examples/08
- [x] examples/09


dai-shi:
Hmm, I wonder if we should have client based caching option...

dai-shi (commented):
There is still room for improvement, but I think I need some feedback. Let's merge it and ship it.

PR #127, opened 2023-09-11
fix(cli): scan template error

vasucp1207:
Running `npx create-waku` resulting the error,
```
Error: ENOENT: no such file or directory, scandir 'template'
```

Changing the code manually in `node_modules` on my system in the package `create-waku/dist/cli.js` works,
```diff
- const CHOICES = fs.readdirSync("template");
+ const templateRoot = path.join(__dirname, "../template");
+ const CHOICES = fs.readdirSync(templateRoot);
```

Because it checks the template folder in `cwd` on our system rather than in `create-waku` package we have to change this.

### DIFF
--- a/packages/create-waku/src/cli.ts
+++ b/packages/create-waku/src/cli.ts
@@ -35,7 +35,8 @@ async function init() {
   let targetDir = "";
   const defaultProjectName = "waku-project";
 
-  const CHOICES = fs.readdirSync("template");
+  const templateRoot = path.join(__dirname, "../template");
+  const CHOICES = fs.readdirSync(templateRoot);
   let result: {
     packageName: string;
     shouldOverwrite: string;
@@ -116,7 +117,6 @@ async function init() {
 
   console.log("Setting up project...");
 
-  const templateRoot = path.join(__dirname, "../template");
   const templateDir = path.resolve(templateRoot, chooseProject);
 
   // Read existing package.json from the root directory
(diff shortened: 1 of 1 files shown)


packages/create-waku/src/cli.ts
```diff
@@ -35,7 +35,8 @@ async function init() {
   let targetDir = "";
   const defaultProjectName = "waku-project";
 
-  const CHOICES = fs.readdirSync("template");
+  const templateRoot = path.join(__dirname, "../template");
+  const CHOICES = fs.readdirSync(templateRoot);
```
dai-shi, line 39: Oh, nice catch. I overlooked that.

packages/create-waku/src/cli.ts
```diff
@@ -116,7 +117,6 @@ async function init() {
 
   console.log("Setting up project...");
 
-  const templateRoot = path.join(__dirname, "../template");
   const templateDir = path.resolve(templateRoot, chooseProject);
```
dai-shi, line 120: This can be `path.join`. Not a big deal though.
```suggestion
  const templateDir = path.join(templateRoot, chooseProject);
```

dai-shi (commented):
I wonder if `pnpm create ...` allows to use csb build somehow...

dai-shi (approved):
LGTM! Thanks for your work!

PR #132, opened 2023-09-19
refactor: use tsconfig monorepo

himself65:
Add top-level ts config that includes all submodules, which restricts the type and isolated type context. If some files are missing or wrong-typed, it will throw the error.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,21 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
+  type-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: pnpm/action-setup@v2
+        with:
+          version: 8.2.0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+          cache: 'pnpm'
+          cache-dependency-path: '**/pnpm-lock.yaml'
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm run typecheck
+
   e2e:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,21 +20,6 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
-  type-check:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - uses: pnpm/action-setup@v2
-        with:
-          version: 8.2.0
-      - uses: actions/setup-node@v3
-        with:
-          node-version: 18
-          cache: 'pnpm'
-          cache-dependency-path: '**/pnpm-lock.yaml'
-      - run: pnpm install --frozen-lockfile
-      - run: pnpm run typecheck
-
   e2e:
     runs-on: ubuntu-latest
     steps:
--- a/examples/01_counter/tsconfig.json
+++ b/examples/01_counter/tsconfig.json
@@ -11,5 +11,5 @@
     "jsx": "react-jsx",
     "outDir": "./dist",
     "rootDir": "./src"
-  },
+  }
 }
--- a/examples/01_counter/tsconfig.json
+++ b/examples/01_counter/tsconfig.json
@@ -8,6 +8,8 @@
     "skipLibCheck": true,
     "noUncheckedIndexedAccess": true,
     "exactOptionalPropertyTypes": true,
-    "jsx": "react-jsx"
-  }
+    "jsx": "react-jsx",
+    "outDir": "./dist",
+    "rootDir": "./src"
+  },
 }
--- a/examples/02_async/tsconfig.json
+++ b/examples/02_async/tsconfig.json
@@ -12,13 +12,9 @@
     "outDir": "./dist",
     "tsBuildInfoFile": "./dist/.tsbuildinfo",
     "composite": true,
-    "types": [
-      "react/experimental"
-    ]
+    "types": ["react/experimental"]
   },
-  "include": [
-    "./src"
-  ],
+  "include": ["./src"],
   "references": [
     {
       "path": "./tsconfig.node.json"
--- a/examples/02_async/tsconfig.json
+++ b/examples/02_async/tsconfig.json
@@ -12,7 +12,10 @@
     "outDir": "./dist",
     "tsBuildInfoFile": "./dist/.tsbuildinfo",
     "composite": true,
-    "types": ["react/experimental"]
+    "types": [
+      // React Server Async Component type
+      "react/experimental"
+    ]
   },
   "include": ["./src"],
   "references": [
--- a/examples/02_async/tsconfig.json
+++ b/examples/02_async/tsconfig.json
@@ -8,6 +8,20 @@
     "skipLibCheck": true,
     "noUncheckedIndexedAccess": true,
     "exactOptionalPropertyTypes": true,
-    "jsx": "react-jsx"
-  }
+    "jsx": "react-jsx",
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
+    "composite": true,
+    "types": [
+      "react/experimental"
+    ]
+  },
+  "include": [
+    "./src"
+  ],
+  "references": [
+    {
+      "path": "./tsconfig.node.json"
+    }
+  ]
 }
--- a/examples/02_async/tsconfig.json
+++ b/examples/02_async/tsconfig.json
@@ -11,11 +11,7 @@
     "jsx": "react-jsx",
     "outDir": "./dist",
     "tsBuildInfoFile": "./dist/.tsbuildinfo",
-    "composite": true,
-    "types": [
-      // React Server Async Component type
-      "react/experimental"
-    ]
+    "composite": true
   },
   "include": ["./src"],
   "references": [
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "compile": "pnpm -r --filter='./packages/*' run compile",
     "csb-install-FIXME": "pnpm install --no-frozen-lockfile",
     "postinstall": "test -d ./packages/waku/dist || (pnpm -r --filter='./packages/waku' run compile && pnpm install)",
-    "test": "prettier -c . && eslint . && tsc -b tsconfig.json --diagnostics",
+    "test": "prettier -c . && eslint . && tsc -b --diagnostics",
     "e2e": "playwright test",
     "examples:dev": "(cd ./examples/${NAME} && pnpm run dev)",
     "examples:dev:01_counter": "NAME=01_counter pnpm run examples:dev",
--- a/package.json
+++ b/package.json
@@ -4,12 +4,11 @@
   "type": "module",
   "private": true,
   "scripts": {
-    "typecheck": "tsc -b tsconfig.json --diagnostics",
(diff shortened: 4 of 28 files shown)


packages/waku/src/types.d.ts
```diff
@@ -3,11 +3,3 @@ declare module "react-server-dom-webpack/server";
...
 declare module "react-server-dom-webpack/client";
 declare module "react-server-dom-webpack/client.node.unbundled";
-
-// FIXME Is this too naive? How can we avoid the eslint warning?
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-namespace JSX {
-  export type ElementType =
-    | string
-    | ((props: any) => React.ReactNode | Promise<React.ReactNode>);
-}
```
himself65, line 13: in `react/experimental`

packages/waku/src/config.ts
```diff
@@ -64,12 +64,12 @@ export interface ExtendedUserConfig extends UserConfig {
...
-export function defineConfig(
-  config:
+export function defineConfig<
+  T extends
     | ExtendedUserConfig
     | Promise<ExtendedUserConfig>
     | ((env: ConfigEnv) => ExtendedUserConfig)
     | ((env: ConfigEnv) => Promise<ExtendedUserConfig>),
-) {
+>(config: T): T {
```
himself65, line 73: otherwise, tsc will cause an error

packages/create-waku/src/cli.ts
```diff
@@ -2,13 +2,8 @@
 
 import fs from "node:fs";
 import path from "node:path";
-// FIXME what's the proper fix?
-// eslint-disable-next-line import/no-named-as-default
```
himself65, line 6: change `moduleResolution` to `bundler`

.github/workflows/cd.yml
```diff
@@ -11,8 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
-        with:
-          version: 8.2.0
```
himself65, line 15: add `packageManager` into `package.json` will lock the version

dai-shi (changes requested):
I'm sorry, I had pending comments that I thought I submitted, but it's actually not.

dai-shi:
Please make changes minimal as I'm working on #136 too.

himself65:
I will continue work on this today

.github/workflows/ci.yml
```diff
@@ -20,6 +20,21 @@ jobs:
       - run: pnpm install --frozen-lockfile
       - run: pnpm test
 
+  type-check:
```
dai-shi, line 23: We run tsc in `pnpm test`, so it's not necessary here?

examples/01_counter/tsconfig.json
```diff
@@ -8,6 +8,8 @@
     "skipLibCheck": true,
     "noUncheckedIndexedAccess": true,
     "exactOptionalPropertyTypes": true,
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "outDir": "./dist",
+    "rootDir": "./src"
```
dai-shi, line 13: Not sure why they are required, but maybe they are?

examples/02_async/tsconfig.json
```diff
@@ -8,6 +8,16 @@
...
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
+    "composite": true,
+    "types": ["react/experimental"]
+  },
+  "include": ["./src"],
+  "references": [
+    {
+      "path": "./tsconfig.node.json"
+    }
+  ]
```
dai-shi, line 22: Why do we need this in 02_async, not in 01_counter?
I do not understand why two tsconfig files are required, instead of one.

packages/create-waku/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "emitDeclarationOnly": true,
```
dai-shi, line 5: Can these be in the base tsconfig?

packages/waku/src/types.d.ts
```diff
@@ -3,11 +3,3 @@ declare module "react-server-dom-webpack/server";
 declare module "react-server-dom-webpack/server.node.unbundled";
 declare module "react-server-dom-webpack/client";
 declare module "react-server-dom-webpack/client.node.unbundled";
-
-// FIXME Is this too naive? How can we avoid the eslint warning?
```
dai-shi, line 7: Nice! How did you do it? Which change does this job?

packages/create-waku/tsconfig.json
dai-shi: maybe because it's build specific.

packages/waku/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "types": ["react/experimental"],
```
dai-shi, line 8: I prefer inline declaration instead of having it here.
Also, we shouldn't use `experimental`.

tsconfig.json
```diff
@@ -12,9 +12,57 @@
...
+    {
+      "path": "./packages/waku"
+    },
+    {
+      "path": "./packages/create-waku"
+    },
+    {
+      "path": "./packages/website"
+    },
+    // Examples
```
dai-shi, line 35: We have 10_ and 11_ now.

Do we need all references??

dai-shi (commented):
This looks very nice.
I haven't used "references" before.
My preference is to have things as minimal as possible.
Some of tsconfig properties are new to me, and I can't judge, but please keep that in mind.

tsconfig.json
himself65: yep

packages/waku/src/types.d.ts
himself65: `react/experimental`

examples/02_async/tsconfig.json
himself65: because `vite.config.ts` is actually in a different environment with files in `./src`.

For example, you cannot include src files in config because of two ts configs with different files included

[tsconfig.json](https://github.com/dai-shi/waku/pull/132/files/5ee1aed268049e10f0007917dfb5cd7c66480414#diff-d77d626902deb144fd543b56199c8719c93b7f43b024d6d92f5fce182cc97551)

This is good for mono repo because sometimes you will mistakenly include files across the packages

Also it will be compiled to the different outputDir based on tsconfig.

01_counter doesn't have two configs because there is no vite.config.ts

examples/01_counter/tsconfig.json
himself65: it is required otherwise the output file structure will be
```
- dist
  - src
    - xxx.js
```

instead of

```
- dist
- xxx.js
```

It's not needed here since `waku build` is real output files. But just some habit for me.

But it's necessary for `packages/waku` since we're monorepo tsconfig now

himself65:
You can read the document here: https://www.typescriptlang.org/docs/handbook/project-references.html

examples/01_counter/tsconfig.json
dai-shi: hm, that makes sense.

examples/02_async/tsconfig.json
dai-shi: Thanks for the explanation. We will revisit this later.

examples/02_async/tsconfig.json
```diff
@@ -8,6 +8,19 @@
...
     "exactOptionalPropertyTypes": true,
-    "jsx": "react-jsx"
-  }
+    "jsx": "react-jsx",
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
+    "composite": true,
+    "types": [
+      // React Server Async Component type
+      "react/experimental"
```
dai-shi, line 17: Can we put this also in the source code as inline comment? Just a preference, but it looks more explicit.

packages/waku/src/types.d.ts
dai-shi: Okay, yeah, I'd love to have in the source code instead of tsconfig.

packages/create-waku/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./dist",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
```
dai-shi, line 7: I don't think this is correct for cli tools. I think it should be `nodenext`, no?

dai-shi (commented):
Looking good. I feel like I need to learn some more later.

packages/create-waku/tsconfig.json
himself65: bundler is better? Since you are using esbuild here

https://www.typescriptlang.org/tsconfig#moduleResolution

himself65: For now i think change it to nodenext wont affect anything

dai-shi: Oh, I forgot that we use esbuild in cli. Yeah, let's use nodenext if we can. Using esbuild is only a temporary solution anyway.

himself65: Oh I see a error and its seems correct
image

himself65: In this case, create-waku is cjs module but we import a ESM module, I'm not sure the purpose of you're importing this

dai-shi: importing fs-extra/esm ?
@vasucp1207 introduced it.
Maybe we should refactor it. My intention is everything should be ESM and no CJS in this repo. (and our plan was to avoid esbuild eventually.)

examples/02_async/tsconfig.json
dai-shi: This is remaining. Also in 10_dynamicroute.

package.json
```diff
@@ -9,7 +9,7 @@
     "compile": "pnpm -r --filter='./packages/*' run compile",
     "csb-install-FIXME": "pnpm install --no-frozen-lockfile",
     "postinstall": "test -d ./packages/waku/dist || (pnpm -r --filter='./packages/waku' run compile && pnpm install)",
-    "test": "prettier -c . && eslint . && tsc --project . --noEmit",
+    "test": "prettier -c . && eslint . && tsc -b tsconfig.json --diagnostics",
```
dai-shi, line 12: Curious if we need to pass the default tsconfig?
`tsc -b --diagnostics`

(but, there was `--project .`, so it can't work without passing `tsconfig.json`?)

dai-shi (commented):
we are almost there.

packages/create-waku/tsconfig.json
himself65: I see, i can do that but it's too far away from this PR

package.json
himself65: we can

packages/create-waku/tsconfig.json
dai-shi: That would be great. Feel free to open a new PR.

dai-shi (approved):
LGTM
Thanks for your contribution!

packages/waku/tsconfig.json
dai-shi: https://github.com/dai-shi/waku/pull/332 I changed my mind for examples/templates.

PR #133, opened 2023-09-30
feat(router): slug support

dai-shi:
slug support isn't in waku/router, but in userland (entries.tsx).
but it requires some tweaks.
ssr is troublesome and we need to revisit it later.


dai-shi:
There are some issues (not only with this PR change) and room for improvement, but it's a progress. Let's merge this and continue development.

PR #138, opened 2023-10-17
update react canary version

dai-shi:
- [x] update react canary version
- [x] fix using `decodeReply` outside "react-server" condition
- [x] add 11_form example


dai-shi:
Oh, no. https://github.com/facebook/react/pull/27436 prevents to use `decodeReply` outside "react-server" condition. 😱

dai-shi:
It was far bigger change than I expected. Merging.

dai-shi:
18147fa

> this is confusing

ah, the form status is only available in the form tree. makes sense.

PR #139, opened 2023-10-21
fix: resolve multiple port occupancy conflict

lesenelir:
When I start multiple services in `/examples`, I notice that the server only uses port 3000. There seems to be a better option, to choose the next port number when the current one is occupied.


dai-shi:
Thanks for the suggestion.
Can you check CI errors?

lesenelir:
I've adjusted the code style, and all local tests have passed. 
If there are any further issues, please bear with me...

dai-shi (approved):
We may change the implementation in the future, but it looks good meanwhile.

PR #142, opened 2023-10-23
chore: lock pnpm version in `package.json`

himself65:
`packageVersion` is a new built-in API from node.js that could lock the package manager and its version from the top level.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -11,8 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
-        with:
-          version: 8.2.0
       - uses: actions/setup-node@v3
         with:
           node-version: 18
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
-        with:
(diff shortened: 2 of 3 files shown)


dai-shi:
Can you please add PR descriptions (for all PRs) mentioning the background of the change and so forth?

dai-shi:
Hm, this is interesting.
- https://nodejs.org/dist/latest-v18.x/docs/api/all.html#all_packages_packagemanager
- https://nodejs.org/dist/latest-v18.x/docs/api/all.html#all_corepack_supported-package-managers

.github/workflows/cd.yml
```diff
@@ -11,8 +11,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: pnpm/action-setup@v2
-        with:
-          version: 8.2.0
```
dai-shi, line 15: Are we sure if corepack is enabled in the CI/CD?

himself65: See

https://github.com/pnpm/action-setup/blob/d2613e087f2e0aa841925861c5a5f7395d552177/README.md?plain=1#L11

dai-shi: My concern was about enabling corepack, but it seems to be working anyway. https://github.com/dai-shi/waku/actions/runs/6620491644/job/17982999317

dai-shi (approved):
This is new to me, but it feels good. will see how it goes.

PR #144, opened 2023-10-25
refactor: remove esbuild

himself65:
https://github.com/dai-shi/waku/pull/132#discussion_r1370887758

### DIFF
--- a/packages/create-waku/tsconfig.json
+++ b/packages/create-waku/tsconfig.json
@@ -5,8 +5,10 @@
     "outDir": "./dist",
     "module": "NodeNext",
     "moduleResolution": "nodenext",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"],
-    "tsBuildInfoFile": "./dist/.tsbuildinfo"
+    "types": [],
+    "lib": ["ESNext"],
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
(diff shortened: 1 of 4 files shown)


packages/create-waku/tsconfig.json
```diff
@@ -5,8 +5,10 @@
     "outDir": "./dist",
     "module": "NodeNext",
     "moduleResolution": "nodenext",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"],
-    "tsBuildInfoFile": "./dist/.tsbuildinfo"
+    "types": [],
```
dai-shi, line 8:
```suggestion
```

dai-shi (commented):
Nice, we thought about using `swc` but `tsc` would work too. Would you be interested in doing it with `swc`, like `packages/waku`?

himself65:
> Would you be interested in doing it with swc, like packages/waku?

I tried, that is impossible if the project both output cjs and esm, since tsconfig will check `require` in cjs, `import.meta.url` in esm. they cannot both together

dai-shi:
Interesting. I wasn't aware of it.

Did you try the cli locally to see if works as expected? Can I merge and just release it?

himself65:
I tried it, works

dai-shi (approved):
Thanks so much.

PR #146, opened 2023-10-26
fix: type on load hook

himself65:
Upstream: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65490


dai-shi (approved):
Very nice. Thanks!

PR #154, opened 2023-11-06
feat: add website

sandren:
Renew the website based on #152.


packages/website/.npmrc
```diff
@@ -0,0 +1 @@
+strict-peer-dependencies=false
```
dai-shi, line 1: Is this file necessary in the repo? If we can remove it, let's remove it for now.

packages/website/.nvmrc
```diff
@@ -0,0 +1 @@
+18
```
dai-shi, line 1: Please remove this file.

packages/website/package.json
```diff
@@ -1,6 +1,11 @@
 {
   "name": "waku-website",
-  "version": "0.1.0",
+  "version": "0.0.0",
```
dai-shi, line 3: We don't need to change. It's private anyway though.
```suggestion
  "version": "0.1.0",
```

packages/website/package.json
```diff
@@ -1,6 +1,11 @@
...
+  "description": "⛩️ The minimal React framework",
+  "homepage": "https://github.com/dai-shi/waku",
+  "bugs": "https://github.com/dai-shi/waku/issues",
+  "license": "MIT",
```
dai-shi, line 7: Not required here.

packages/website/package.json
```diff
@@ -9,14 +14,19 @@
...
-    "autoprefixer": "^10.4.16",
-    "bright": "^0.8.4",
     "express": "^4.18.2",
-    "glob": "^10.3.10",
     "react": "18.3.0-canary-a41957507-20231017",
     "react-dom": "18.3.0-canary-a41957507-20231017",
     "react-server-dom-webpack": "18.3.0-canary-a41957507-20231017",
-    "tailwindcss": "^3.3.3",
-    "waku": "workspace:*"
+    "waku": "0.16.0"
```
dai-shi, line 21:
```suggestion
    "waku": "workspace:*"
```
We use the workspace protocol in monorepo.

packages/website/prettier.config.cjs
```diff
@@ -0,0 +1,16 @@
...
+  jsxSingleQuote: false,
+  printWidth: 100,
+  quoteProps: 'consistent',
+  semi: true,
+  singleQuote: true,
+  tabWidth: 2,
+  trailingComma: 'all',
+  tailwindConfig: './tailwind.config.js',
+  tailwindFunctions: ['cx'],
+};
```
dai-shi, line 16: I think the prettier config should be the same across the monorepo. It may require to install prettier-plugin-tailwindcss in root, but that can't be helped, I guess.

packages/website/vite.config.ts
```diff
@@ -1,23 +1,23 @@
...
+        output: {
+          preserveModules: true,
+          preserveModulesRoot: modulesRoot,
+        },
       },
     },
-  },
-});
+  });
+};
```
dai-shi, line 23: We can probably remove this file, because now the website doesn't use waku/router.

dai-shi (changes requested):
We need slight changes for setup/env/etc.

packages/website/tailwind.config.js
```diff
@@ -1,14 +1,39 @@
+const colors = require("tailwindcss/colors");
+
 /** @type {import('tailwindcss').Config} */
-export default {
-  content: ["./index.html", "./{src,routes}/**/*.{ts,tsx}"],
+module.exports = {
```
dai-shi, line 4: Does this work? Everything should be ESM.
```suggestion
export default {
```

packages/website/tsconfig.json
```diff
@@ -1,17 +1,7 @@
...
     "outDir": "./dist",
-    "tsBuildInfoFile": "./dist/.tsbuildinfo",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"]
-  },
-  "include": ["src"],
-  "references": [
-    {
-      "path": "../waku"
-    }
-  ]
```
dai-shi, line 16: I'm not 100% sure, but we should probably keep "include" and "references".

packages/website/tsconfig.json
```diff
@@ -1,12 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "declaration": true,
-    "emitDeclarationOnly": true,
-    "rootDir": "./src",
     "outDir": "./dist",
-    "tsBuildInfoFile": "./dist/.tsbuildinfo",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"]
+    "rootDir": "./src"
```
dai-shi, line 5: Please move this before "outDir" to make the diff minimal.

packages/website/src/components/credits.tsx
```diff
@@ -0,0 +1,56 @@
+export const Credits = () => {
+  if (process.env.VITE_SHOW_CREDITS !== "1") return null;
```
dai-shi, line 2:
```suggestion
  if (process.env.VITE_SHOW_CREDITS !== "YES") return null;
```

packages/website/tsconfig.json
```diff
@@ -1,12 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "declaration": true,
-    "emitDeclarationOnly": true,
     "rootDir": "./src",
-    "outDir": "./dist",
-    "tsBuildInfoFile": "./dist/.tsbuildinfo",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"]
+    "outDir": "./dist"
```
dai-shi, line 5: It feels good to me, but let me mention @himself65 to check if this can cause any problems.

dai-shi (approved):
LGTM

packages/website/tsconfig.json
himself65: I think we should keep `tsBuildInfoFile` otherwise it will generate to the package dir

himself65: Like this 
image

dai-shi: I see. Is it for incremental build? I'd put it outside `dist`. Let me try.

himself65: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

We could put it anywhere. It's just a cache file for incremental build.

dai-shi: Yeah, I don't think it's necessary in dist: https://unpkg.com/browse/waku@0.16.0/dist/.tsbuildinfo

PR #163, opened 2023-11-14
chore: revise Prettier config

sandren:


dai-shi (approved):

dai-shi:
Just a tiny preference, but let's go with this.

PR #164, opened 2023-11-14
fix(create-waku): fix binary definition

gregberge:


dai-shi:
Does it cause a real issue? I thought it's a valid shorthand.

gregberge:
Actually I think it is. I can't run any create command that works and the binary file are not present in ˋnode_modules/.bin` when I install it with ˋnpm install`. Does the create command work for you with npm and pnpm?

dai-shi:
`create-waku` isn't expected to be used with `npm install`.

https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin
image

That said, I can merge this, if it helps to avoid any confusion.

dai-shi:
As it probably doesn't hurt, let's merge it.

PR #165, opened 2023-11-15
switch to hono from express (still depends on node)

dai-shi:
From express/node to hono/node. It still depends on Node.

This PR does the job with `wrap` **hack**. I will continue working on refactors to make it more straightforward.


Aslemammad:
Why not hattip or nitro? these just came to my mind.
https://github.com/hattipjs/hattip
https://nitro.unjs.io

dai-shi:
Note that Waku middleware isn't tied to a specific server implementation. (It depends on Node though.)
People could develop adapters for various server implementations. I will make Waku middleware as portable as possible in upcoming PRs. It will take time.
For Waku cli, we just need to pick one and I picked hono.
For example, 08_cookies keeps using express.

PR #166, opened 2023-11-16
feat(website): add readme

sandren:


packages/website/src/components/button.tsx
```diff
@@ -0,0 +1,32 @@
+// FIXME we should be able to remove this directive
```
dai-shi, line 1: The first issue.

packages/website/src/main.tsx
```diff
@@ -10,4 +10,17 @@ const rootElement = (
   </StrictMode>
 );
 
-createRoot(document.getElementById('root')!).render(rootElement);
+// FIXME temporary fix, doesn't feel ideal.
```
dai-shi, line 13: The second issue.

dai-shi (approved):
There are two remaining issues, but let's tackle them separately.

Issue #168, opened 2023-11-17
Missing `dist/cli.js` from `create-waku`


Asuka109:
It seems the last release is broken.

![image](https://github.com/dai-shi/waku/assets/18379948/96e92620-4940-40c6-ade8-560ff0f8d864)

![image](https://github.com/dai-shi/waku/assets/18379948/b8745f1a-1eb3-42b9-b726-cb37f2842c6e)

dai-shi:
Thanks for reporting. 
Seems like I broke something around build. 


See #167

dai-shi:
create-waku v0.6.2 is published.

Issue #167, opened 2023-11-16
Creating a new project fails


mauricedb:
Try to create a new project using `npm create waku@latest`.

Observed result:
```
PS C:\demos> npm create waku@latest
'create-waku' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! code 1
npm ERR! path C:\demos
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c create-waku

npm ERR! A complete log of this run can be found in: C:\Users\mauri\AppData\Local\npm-cache\_logs\2023-11-16T12_43_09_734Z-debug-0.log
PS C:\demos>
```

Content of the log file created:
```
0 verbose cli C:\Program Files\nodejs\node.exe C:\Users\mauri\AppData\Roaming\npm\node_modules\npm\bin\npm-cli.js
1 info using npm@10.2.4
2 info using node@v20.9.0
3 timing npm:load:whichnode Completed in 2ms
4 timing config:load:defaults Completed in 2ms
5 timing config:load:file:C:\Users\mauri\AppData\Roaming\npm\node_modules\npm\npmrc Completed in 3ms
6 timing config:load:builtin Completed in 3ms
7 timing config:load:cli Completed in 1ms
8 timing config:load:env Completed in 0ms
9 timing config:load:file:C:\demos\.npmrc Completed in 1ms
10 timing config:load:project Completed in 2ms
11 timing config:load:file:C:\Users\mauri\.npmrc Completed in 0ms
12 timing config:load:user Completed in 0ms
13 timing config:load:file:C:\Users\mauri\AppData\Roaming\npm\etc\npmrc Completed in 0ms
14 timing config:load:global Completed in 0ms
15 timing config:load:setEnvs Completed in 1ms
16 timing config:load Completed in 10ms
17 timing npm:load:configload Completed in 10ms
18 timing config:load:flatten Completed in 2ms
19 timing npm:load:mkdirpcache Completed in 0ms
20 timing npm:load:mkdirplogs Completed in 1ms
21 verbose title npm create waku@latest
22 verbose argv "create" "waku@latest"
23 timing npm:load:setTitle Completed in 0ms
24 timing npm:load:display Completed in 1ms
25 verbose logfile logs-max:10 dir:C:\Users\mauri\AppData\Local\npm-cache\_logs\2023-11-16T12_43_09_734Z-
[...]

PR #170, opened 2023-11-19
chore: fix build output disappear

himself65:
`.tsbuildinfo` is a cache file, if you didn't remove it correctly. tsc will treat the output still remain in this case.

Fixes: https://github.com/dai-shi/waku/issues/167
Fixes: https://github.com/dai-shi/waku/issues/168

### DIFF
--- a/packages/create-waku/package.json
+++ b/packages/create-waku/package.json
@@ -18,7 +18,7 @@
   ],
   "scripts": {
     "start": "node dist/cli.js",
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Thanks for the investigation. The diff looks good to me.

dai-shi:
This indeed fixes the build issue. Thanks so much!

PR #171, opened 2023-11-20
fix: vanilla extract compatibility and better css handling

### LINKED ISSUE #59: not compatible with `@vanilla-extract/css`
See https://github.com/Himself65/affine-waku/tree/css-issue

Aslemammad:
Resolves #59 

- rscHmrPlugin now does not inject to every file
- importing CSS files did not work properly, this would fix it
- in waku's plugins, vite.config.ts plugins were not being used, now they're being imported using vite's resolveConfig
- - For now, we only get the plugins, it's better that in the future, we read the whole config and use it using mergeConfig. I tried doing this but it did not work properly.


examples/12_css/.temp-1443d91a2418a076.js
```diff
@@ -0,0 +1,4 @@
+
+import { loadServerFile } from '/Users/alibaba/oss/waku/packages/waku/dist/lib/middleware/rsc/ssr.js';
+const { Counter } = await loadServerFile('/Users/alibaba/oss/waku/examples/12_css/src/components/Counter.tsx', 'dev');
+export { Counter }
```
dai-shi, line 4: Please remove `.temp-*` files.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -1,7 +1,7 @@
 import path from 'node:path';
 import fsPromises from 'node:fs/promises';
 import type { IncomingMessage, ServerResponse } from 'node:http';
-import type { ViteDevServer } from 'vite';
+import { resolveConfig as resolveViteConfig, mergeConfig, type ViteDevServer } from 'vite';
```
dai-shi, line 4: Just stylistic preference.
```suggestion
import { resolveConfig as resolveViteConfig, mergeConfig } from 'vite';
import type { ViteDevServer } from 'vite';
```

dai-shi: Actually, `rsc` middleware should work without vite in production.
So, these vite function should be imported dynamically `getViteServer`.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -150,6 +168,19 @@ const shutdown = async () => {
   parentPort!.close();
 };
 
+const collectCssUrls = (mods: Set<ModuleNode>, styles: Map<string, string>) => {
+  for (const mod of mods) {
+    if (mod.ssrModule && mod.file && mod.id) {
+      if (mod.file.endsWith('.css') || /\?vue&type=style/.test(mod.id)) {
```
himself65, line 174: why is there a `vue`

Aslemammad: This whole function is not used, I just copied it for an experiment from https://github.com/vitejs/vite/issues/2282#issuecomment-787049601

I'll remove it when I get this done!

dai-shi:
> * in waku's plugins, vite.config.ts plugins were not being used, now they're being imported using vite's resolveConfig

Let me grab this idea and open a PR separately.

dai-shi:
> > * in waku's plugins, vite.config.ts plugins were not being used, now they're being imported using vite's resolveConfig
> 
> Let me grab this idea and open a PR separately.

After trying it, I don't feel using `mergeConfig` selectively is a good idea. I'll recover configFileConfig in v0.16.0.

Aslemammad:
> After trying it, I don't feel using mergeConfig selectively is a good idea. I'll recover configFileConfig in v0.16.0.

Agreed, It's not the correct approach, but we had to do it, if you have a  better idea that brings the whole user's config, that's so much better and feel free to push it so I rebase.

dai-shi:
> Agreed, It's not the correct approach, but we had to do it, if you have a better idea that brings the whole user's config, that's so much better and feel free to push it so I rebase.

https://github.com/dai-shi/waku/pull/176/files#r1404458905

Aslemammad:
I'll continue in another PR.

dai-shi:
btw, this brought #228, so much appreciated.

PR #176, opened 2023-11-23
fix: ssr and bundling

dai-shi:
This fixes various issues, notably waku/router bundling and ssr.


packages/waku/src/lib/config.ts
```diff
@@ -51,3 +51,15 @@ export async function resolveConfig() {
...
+  if (process.env.CONFIG_FILE) {
+    return { configFile: path.resolve(process.env.CONFIG_FILE) };
+  }
+  for (const file of ['vite.config.ts', 'vite.config.js']) {
+    if (fs.existsSync(file)) {
+      return { configFile: path.resolve(file) };
+    }
+  }
+  return {};
+};
```
dai-shi, line 65: I'm not super comfortable with fs.existsSync, but this is what I came up with.

Aslemammad: what's the problem of vite resolveConfig? Can't we use that?

dai-shi: I tried that, but it didn't work very well, because we have two `root` configs. There might be my oversight though.
Maybe we should re-consider changing `root`. Let's try that after merging this PR.

Aslemammad: Yes, I tried the root on where the vite.config is and it worked well! Let me know your findings, because having multiple roots makes loading configs really hard (and trying to keep them consistent would be so hard in the future)

dai-shi: Yeah, we want `src` and `dist` at the same level.
Can you open a PR that only changes it? (=not changing "root")

Aslemammad: Yes, I can work on it tomorrow if you're ok with it. Or later tonight.

dai-shi: Yes, please go ahead. But, note that it may not work as we want. So, I mean a draft PR to show your idea. We can then consider to move forward.

PR #177, opened 2023-11-23
build: fix postinstall error on windows

himself65:
Only fixes: https://github.com/dai-shi/waku/issues/175

There are too many compatibility issues on Windows. I want to make it step by step.

First is to fix install pnpm


dai-shi (approved):
The change looks good. Thanks!

PR #178, opened 2023-11-23
ci: cover more versions

himself65:


himself65:
```shell
[WebServer] You did not run Node.js with the `--conditions react-server` flag. Any "react-server" override will only work with ESM imports.
[WebServer] You did not run Node.js with the `--conditions react-server` flag. Any "react-server" override will only work with ESM imports.
```

himself65:
> ```shell
> [WebServer] You did not run Node.js with the `--conditions react-server` flag. Any "react-server" override will only work with ESM imports.
> [WebServer] You did not run Node.js with the `--conditions react-server` flag. Any "react-server" override will only work with ESM imports.
> ```

This is a bug from node.js source code. I will open an issue

dai-shi (approved):
This is very nice!!

PR #179, opened 2023-11-23
fix: type on `worker-impl.ts`

himself65:


packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -24,7 +24,9 @@ const { renderToPipeableStream, decodeReply, decodeReplyFromBusboy } =
 
 const IS_NODE_20 = Number(process.versions.node.split('.')[0]) >= 20;
 if (IS_NODE_20) {
-  const { register } = (await import('node:module')) as any;
+  const {
+    default: { register },
+  } = await import('node:module');
```
dai-shi, line 29: Hm, okay if we use the default export, the types work.

dai-shi (approved):

PR #181, opened 2023-11-26
fix: unify root paths across different phases

Aslemammad:


examples/09_cssmodules/src/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <title>Waku example</title>
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <style>
+    <!-- <style>
```
Aslemammad, line 7: Weirdly enough, having styles in this html would give me a weird error from babel. 

```
Cannot render RSC SyntaxError: /home/aslemammad/oss/waku/examples/09_cssmodules/src/components/Counter.tsx?html-proxy&direct&index=0.css: Support for the experimental syntax 'decorators' isn't currently enabled (2:7):

  1 |
> 2 |       @keyframes spinner {
    |       ^
  3 |         to {
  4 |           transform: rotate(360deg);
  5 |         }

Add @babel/plugin-proposal-decorators (https://github.com/babel/babel/tree/main/packages/babel-plugin-proposal-decorators) to the 'plugins' section of your Babel config to enable transformation.
If you want to leave it as-is, add @babel/plugin-syntax-decorators (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-decorators) to the 'plugins' section to enable parsing.
```

examples/09_cssmodules/src/index.html
```diff
@@ -22,15 +22,15 @@
...
+    </style> -->
   </head>
   <body>
     <!--placeholder1-->
     <div id="root">
       <div class="spinner"></div>
     </div>
     <!--/placeholder1-->
-    <script src="/main.tsx" async type="module"></script>
+    <script src="/src/main.tsx" async type="module"></script>
```
dai-shi, line 33: I see what you mean. This is probably unexpected to normal vite users.
How about doing
```suggestion
    <script src="main.tsx" async type="module"></script>
```
and make it an absolute path with a new vite plugin for both DEV and PRD?

But, actually, I don't think it's super important. People don't often modify this file from the template.
Yeah, I think `"/src/main.tsx"` is acceptable on second thought.

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -291,16 +291,15 @@ export const renderHtml = async <Context>(
   const moduleMap = new Proxy(
     {},
     {
-      get(_target, filePath: string) {
+      get(_target, _filePath: string) {
```
dai-shi, line 294: Please don't use `_` prefix other than for "unused parameters".

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -213,17 +214,8 @@ const resolveClientEntry = (
...
-    command === 'dev' ? config.srcDir : config.distDir,
-  );
-  if (path.sep !== '/') {
-    // HACK to support windows filesystem
-    root = root.replaceAll(path.sep, '/');
-    if (filePath[0] === '/') {
-      filePath = filePath.slice(1);
-    }
-  }
+  let root = path.join(config.rootDir, command === 'dev' ? '' : config.distDir);
```
dai-shi, line 217: Hmm, root is not `srcDir` in DEV? I would like to see `srcDir` and `distDir` symmetrically.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -213,17 +214,8 @@ const resolveClientEntry = (
...
+  root = normalizePath(root);
```
dai-shi, line 218: Wow, this looks promising.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -234,7 +226,11 @@ const resolveClientEntry = (
       );
     }
   }
-  return config.basePath + path.relative(root, filePath);
+  return (
+    (command === 'dev' ? '/@id' : '') +
```
dai-shi, line 230: What's `@id`? It must be different from `@fs`.

packages/waku/src/lib/middleware/rsc/ssr.ts
Aslemammad: Yes sure, I was just trying to do a prototype. I'll fix it.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
Aslemammad: [Yes, they're so different](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/constants.ts#L57C1-L65)

packages/waku/src/lib/middleware/rsc/worker-impl.ts
Aslemammad: I'm not sure how to achieve this to be honest. Do you have any idea? As we talked the purpose was to make something like reading files working, and It's working now (at least in the entries.ts). 

If you have any other reasons for that, it'd help understand the motivation and hopefully find a solution. If you have a solution also let me know.

packages/waku/src/lib/builder.ts
```diff
@@ -312,6 +312,7 @@ const emitHtmlFiles = async (
     config.rootDir,
     config.distDir,
     config.publicDir,
+    config.srcDir,
```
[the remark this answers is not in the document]
dai-shi: Try this:
```bash
(cd ./examples/01_counter && NODE_ENV=production npm run build)
npx serve examples/01_counter/dist/public
```

packages/waku/src/lib/middleware/rsc/worker-impl.ts
dai-shi: Interesting. I might be missing something, but what are not valid browser import specifiers? Any example in our case? I mean, we didn't need this before.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
Aslemammad: As we talked in discord, can we skip this and face it when we have a problem? Let me know.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
Aslemammad: Yes, when we changed roots, vite tried to identify /src/components/Coutner.tsx in its own way which caused it to create some virtual modules that load the index.html css styles in those virtual modules and when we loaded them in the server we faced the error that I shared above.  
https://github.com/dai-shi/waku/pull/181#discussion_r1405399116

https://github.com/vitejs/vite/blob/2fce647edae3c93915f43c2a2971b87aa2db53c3/packages/vite/src/node/server/middlewares/indexHtml.ts#L178C2-L191

https://github.com/vitejs/vite/blob/2fce647edae3c93915f43c2a2971b87aa2db53c3/packages/vite/src/node/server/middlewares/indexHtml.ts#L333C1-L358


I just added /@id/ and it fixed the issue which led vite to skip this process.

packages/waku/src/lib/builder.ts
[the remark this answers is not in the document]
Aslemammad: I'll try this tomorrow.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
dai-shi: Yes, please leave some comments:
```suggestion
  // FIXME We want to make `root` symmetric for 'srcDir' and 'distDir', but we don't know how to achieve it.
  let root = path.join(config.rootDir, command === 'dev' ? '' : config.distDir);
```

packages/waku/src/lib/middleware/rsc/worker-impl.ts
dai-shi: Ah, because we changed root.
Please add the source comment too.

I was hoping to make DEV and PRD as similar as possible in the code base, but in the future, we may revisit the design anyway.

packages/waku/src/lib/builder.ts
```diff
@@ -312,11 +312,14 @@ const emitHtmlFiles = async (
...
     config.indexHtml,
   );
-  const publicIndexHtml = fs.readFileSync(publicIndexHtmlFile, {
+  const publicIndexHtml = await fs.promises.readFile(publicIndexHtmlFile, {
```
dai-shi, line 318: I'm trying to avoid mixing fs and fsPromises. Let's import from 'fs/promises'

packages/waku/src/lib/builder.ts
```diff
@@ -312,11 +312,14 @@ const emitHtmlFiles = async (
...
     encoding: 'utf8',
   });
+
+  await fs.promises.unlink(publicIndexHtmlFile);
```
dai-shi, line 322: Does this mean it creates two index html files?

dai-shi (changes requested):
Can you update all examples and resolve conflicts with main?

packages/waku/src/lib/builder.ts
Aslemammad: It creates one, gets it path, then emits it. That's how it was done before, but since it was the same path, it was getting overwritten as far as I know.

dai-shi (commented):
The change looks generally good, but I will review it again.

Can you separate a PR for fsPromises refactor? I would like to merge it in advance.

packages/waku/src/lib/middleware/rsc/utils.ts
```diff
@@ -66,7 +66,7 @@ export const endStream = (stream: WritableStream, message?: string) => {
   const writer = stream.getWriter();
   if (message) {
     new TextEncoder().encode(message).forEach((chunk) => {
-      writer.ready.then(() => writer.write(chunk));
+      writer.ready.then(() => writer.write(new Uint8Array([chunk])));
```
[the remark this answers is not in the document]
dai-shi: Oh, this is my bad.
I took the code from https://developer.mozilla.org/en-US/docs/Web/API/WritableStream/getWriter, but ti seems wrong.
We should simply do this:
```js
writer.ready.then(() => writer.write(new TextEncoder().encode(message)))
```

I'll open a PR.

packages/waku/src/lib/utils/path.ts
```diff
@@ -90,3 +90,8 @@ export const extname = (filePath: string) => {
   const index = filePath.lastIndexOf('.');
   return index > 0 ? filePath.slice(index) : '';
 };
+
+export const dirname = (filePath: string) => {
+  const index = filePath.lastIndexOf('/');
+  return index > 0 ? filePath.slice(0, index) : '.';
+};
```
dai-shi, line 97: Instead of adding this, I used `joinPath(..., '..')`.

packages/website/src/index.html
```diff
@@ -56,8 +56,7 @@
     <!--placeholder1-->
     <div id="root"></div>
     <!--/placeholder1-->
-    <!-- FIXME we should use async instead of defer -->
-    <script src="/main.tsx" defer type="module"></script>
+    <script src="/src/main.tsx" async type="module"></script>
```
dai-shi, line 59: Please revert this.

Aslemammad: This would make vercel fail though!

packages/waku/src/lib/builder.ts
```diff
@@ -323,11 +327,20 @@ const emitHtmlFiles = async (
...
+    config.srcDir,
     config.indexHtml,
   );
-  const publicIndexHtml = fs.readFileSync(publicIndexHtmlFile, {
+  const publicIndexHtml = await fs.readFile(publicIndexHtmlFile, {
     encoding: 'utf8',
   });
+
+  await fs.unlink(publicIndexHtmlFile);
+  clientBuildOutput.output.splice(
```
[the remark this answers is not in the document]
dai-shi: Still effective.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -193,6 +190,13 @@ export function rsc<
         return;
       }
     } else {
+      if (pathStr === '/') {
+        const htmlStr = await getHtmlStr(pathStr);
+        res.setHeader('Content-Type', 'text/html');
+        res.setStatus(200);
+        endStream(res.stream, htmlStr!);
+        return;
+      }
```
dai-shi, line 199: Can you explain why you need this?

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -392,16 +392,17 @@ export const renderHtml = async <Context>(
       >
     >,
     {
-      get(_target, filePath: string) {
+      get(_target, resolvedFilePath: string) {
         return new Proxy(
           {},
           {
             get(_target, name: string) {
-              const file = filePath.slice(config.basePath.length);
```
dai-shi, line 400: Please keep this. basePath is configurable.

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -392,16 +392,17 @@ export const renderHtml = async <Context>(
       >
     >,
     {
-      get(_target, filePath: string) {
+      get(_target, resolvedFilePath: string) {
```
dai-shi, line 395: Let's keep this name here, and change the name in the internal function.

packages/waku/src/lib/rsc/renderer.ts
```diff
@@ -50,7 +50,7 @@ const resolveClientEntry = (
   isDev: boolean,
 ) => {
   let filePath = file.startsWith('file://') ? fileURLToFilePath(file) : file;
-  const root = joinPath(config.rootDir, isDev ? config.srcDir : config.distDir);
+  const root = joinPath(config.rootDir, !isDev ? config.distDir : '');
```
dai-shi, line 53:
```suggestion
  const root = joinPath(config.rootDir, isDev ? '' : config.distDir);
```

packages/waku/src/lib/rsc/renderer.ts
```diff
@@ -65,7 +65,7 @@ const resolveClientEntry = (
       );
     }
   }
-  return config.basePath + relativePath(root, filePath);
+  return (isDev ? '/@id' : '') + config.basePath + relativePath(root, filePath);
```
dai-shi, line 68:
```suggestion
  return  config.basePath + (isDev ? '@id/' : '') + relativePath(root, filePath);
```

dai-shi: But, I'm not sure if that works. Maybe, we need to tell basePath to viteServer?

packages/website/src/index.html
```diff
@@ -56,8 +56,7 @@
     <!--placeholder1-->
     <div id="root"></div>
     <!--/placeholder1-->
-    <!-- FIXME we should use async instead of defer -->
-    <script src="/main.tsx" defer type="module"></script>
+    <script src="/main.tsx" async type="module"></script>
```
dai-shi, line 59: Sorry, I meant this:
```suggestion
    <!-- FIXME we should use async instead of defer -->
    <script src="/src/main.tsx" defer type="module"></script>
```

dai-shi (commented):
I'm not exactly sure how it works. Let me ask some questions.

packages/waku/src/lib/middleware/rsc.ts
Aslemammad: In dev mode, vite expects index.html to be in / and not in /src/, since we don't want to move it to /, we need to handle it manually like the way we handle it in ssr.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -55,7 +53,6 @@ export function rsc<
     ]);
     const viteServer = await viteCreateServer({
       ...(await viteInlineConfig()),
-      root: joinPath(config.rootDir, config.srcDir),
       optimizeDeps: {
```
dai-shi, line 56: We need this. Let's add it in this PR.
```suggestion
      base: config.basePath,
      optimizeDeps: {
```

dai-shi:
Please remove `viteInlineConfig` in lib/config.ts entirely, because it is a hack to have different roots.

packages/waku/src/lib/middleware/rsc.ts
dai-shi: Maybe we need to this in all vite configs in the code. We can do that in a separate PR.

packages/waku/src/lib/rsc/renderer.ts
dai-shi: https://github.com/dai-shi/waku/pull/181#discussion_r1415395109

packages/waku/src/lib/middleware/rsc.ts
Aslemammad: And the obvious solution (which is a breaking change) that's aligned with the PR, is to move index.html to / rather than having it in /src/

Any other idea?

packages/website/src/index.html
dai-shi: sorry, I was talking about the FIXME comment and "defer".

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -35,27 +35,22 @@ export function rsc<
 
   let lastViteServer: ViteDevServer | undefined;
   const getViteServer = async (): Promise<ViteDevServer> => {
+    const config = await configPromise;
     if (lastViteServer) {
       return lastViteServer;
     }
     const [
-      config,
```
dai-shi, line 42: Please keep this instead of L38.

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -399,20 +397,22 @@ export const renderHtml = async <Context>(
             get(_target, name: string) {
               const file = filePath.slice(config.basePath.length);
               if (command === 'dev') {
-                const filePath = file.startsWith('@fs/')
-                  ? decodeFilePathFromAbsolute(file.slice('@fs'.length))
-                  : joinPath(config.rootDir, config.srcDir, file);
+                const resolvedFilePath = file.startsWith('/@fs/')
+                  ? decodeFilePathFromAbsolute(file.slice('/@fs'.length))
```
dai-shi, line 401: Should be this?
```suggestion
                const resolvedFilePath = file.startsWith('@fs/')
                  ? decodeFilePathFromAbsolute(file.slice('@fs'.length))
```

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -399,20 +397,22 @@ export const renderHtml = async <Context>(
...
+                  : file;
```
dai-shi, line 402: This doesn't seem correct, but now sure.

packages/waku/src/lib/rsc/renderer.ts
dai-shi: Still effective.
I guess you reverted while merging??

dai-shi (commented):
I guess, this approach isn't that bad.
But, we will see. If another approach is super simple and doesn't break any capabilities, that would work too.

packages/waku/src/lib/middleware/rsc.ts
dai-shi: I'm not against to it. Can you open a new PR for that idea?

dai-shi: > we'd have /src/foo/index.html at some point?

Yes, we have `/foo` and now we have such an e2e test, so the test fails.

dai-shi (commented):
It almost works, except for this.

dai-shi:
Closing in favor of #228.

PR #183, opened 2023-11-27
fix: reduce deprecated api and node warnings

himself65:


dai-shi:
I understand the Node 18 warning, but how the Node 20 warning happens? Are we not use ESM imports?

himself65:
> I understand the Node 18 warning, but how the Node 20 warning happens? Are we not use ESM imports?

This is complex. i read the nodejs source code. Reason is conditions will not pass to the esm module in worker thread. BUT, cjs will. So in the first time, require(react) is in correct condition(used in webpack source code). Second time when you import react-server-webpack, it’s not react-server condition actually

dai-shi (commented):
The change looks good to me. My hesitation is that I will not be able to remember that this change is just about reducing warnings.
Can you add source code comments which include the link to this PR?

e2e/counter.spec.ts
```diff
@@ -9,12 +10,3 @@ test('counter', async ({ page }) => {
   await page.click('text=Increment');
   await expect(page.locator('text=Count: 3')).toBeVisible();
 });
-
-test('first render', async ({ browser }) => {
```
himself65, line 13: I think that this is a bug. In dev mode, SSR is enforced, and no spinner. But in prod mode, there's a spinner

dai-shi: > In dev mode, SSR is enforced

That's not expected. SSR will only be enables with `--with-ssr`.
Can you try it again? Maybe it's a timing issue.

himself65: My mistake. Everything good

e2e/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+    const unexpectedErrors: RegExp[] = [
+      /^You did not run Node.js with the `--conditions react-server` flag./,
+      /^\(node:14372\)/,
+    ];
+    const callback = (msg: ConsoleMessage) => {
+      if (unexpectedErrors.some((re) => re.test(msg.text()))) {
+        throw new Error(msg.text());
+      }
+      console.log(`${msg.type()}: ${msg.text()}`);
+    };
```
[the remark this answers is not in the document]
dai-shi: well, what I meant was the other way around. When there is no longer an extra warning issue upstream, we should theoretically revert the change.

dai-shi:
https://github.com/dai-shi/waku/actions/runs/7012521326/job/19077187950?pr=183
Do you know why CI timeouts sometimes?

himself65:
> [dai-shi/waku/actions/runs/7012521326/job/19077187950?pr=183](https://github.com/dai-shi/waku/actions/runs/7012521326/job/19077187950?pr=183) Do you know why CI timeouts sometimes?

There might be some flaky. or in some cases, CI is slower than usual cause the build is overdue. Let me debug it

himself65:
could you run that again with debug log?

I will go to dinner now. do this later

himself65:
Will rebase this later

dai-shi (approved):
LGTM

PR #184, opened 2023-11-27
fix: error log

himself65:


dai-shi:
Hmm, I hoped that the `Error` object is somehow serializable. https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#error_types
I would prefer avoiding doing this kind of stuff, but if needed we should probably create a utility function for (de)serialize Error object.

himself65:
Closing, doing this in another way

PR #190, opened 2023-11-29
refactor: use web stream internally (step 1)

dai-shi:
Follow-up #165 

It will use web streams internally. We also change the middleware signature and provide adapters for Express and Hono.


dai-shi:
I would like to continue the refactor, but it seems challenging more than expected.
So, I will merge this as the first step of the refactor. It seems pretty stable.

dai-shi (commented):
Should be okay. We might revisit the ReqObject and ResObject structures in the future.

PR #191, opened 2023-11-29
test: put test cases into fixtures

himself65:
Move the test code into the fixtures directory. Since examples are the showcases to the client user, not good for test


dai-shi (approved):
LGTM at a glance

Issue #192, opened 2023-11-29
server action not working on Windows


himself65:
Please try:

https://github.com/dai-shi/waku/blob/d688b4cf76e234992b37b136bac824784b469d4a/e2e/rsc-basic.spec.ts#L72-L75

PR #194, opened 2023-12-01
fix: server action & ssr on windows

himself65:
Fixes: https://github.com/dai-shi/waku/issues/192 and add SSR test cases

### DIFF
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -21,7 +21,7 @@ const commands = [
 const cwd = fileURLToPath(new URL('./fixtures/ssr-basic', import.meta.url));
 
 for (const { build, command } of commands) {
-  test.describe(`rsc-basic: ${command}`, () => {
+  test.describe(`ssr-basic: ${command}`, () => {
     let cp: ChildProcess;
     let port: number;
     test.beforeAll(async () => {
@@ -64,17 +64,18 @@ for (const { build, command } of commands) {
       await expect(page.getByTestId('count')).toHaveText('3');
     });
 
-    const noJSTest = test.extend({});
-    noJSTest.use({
-      javaScriptEnabled: false,
-    });
-
-    noJSTest('no js environment should have first screen', async ({ page }) => {
+    test('no js environment should have first screen', async ({ browser }) => {
+      const context = await browser.newContext({
+        javaScriptEnabled: false,
+      });
+      const page = await context.newPage();
       await page.goto(`http://localhost:${port}/`);
       await expect(page.getByTestId('app-name')).toHaveText('Waku');
       await expect(page.getByTestId('count')).toHaveText('0');
       await page.getByTestId('increment').click();
       await expect(page.getByTestId('count')).toHaveText('0');
+      await page.close();
+      await context.close();
     });
   });
 }
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -65,7 +65,6 @@ for (const { build, command } of commands) {
     });
 
     test('no js environment should have first screen', async ({ browser }) => {
-      test.fixme();
       const noJSContext = await browser.newContext({
         javaScriptEnabled: false,
       });
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -2,6 +2,7 @@ import { expect } from '@playwright/test';
 import { execSync, exec, ChildProcess } from 'node:child_process';
 import { fileURLToPath } from 'node:url';
 import waitPort from 'wait-port';
+import { rm } from 'node:fs/promises';
 import { test } from './utils.js';
 
 const waku = fileURLToPath(
@@ -49,6 +50,10 @@ for (const { build, command } of commands) {
     });
 
     test.afterAll(async () => {
+      await rm(`${cwd}/dist`, {
+        recursive: true,
+        force: true,
+      });
       cp.kill();
     });
 
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -10,11 +10,11 @@ const waku = fileURLToPath(
 
 const commands = [
   {
-    command: 'dev',
+    command: 'dev --with-ssr',
   },
   {
-    build: 'build',
-    command: 'start',
+    build: 'build --with-ssr',
+    command: 'start --with-ssr',
   },
 ];
 
@@ -70,7 +70,6 @@ for (const { build, command } of commands) {
     });
 
     noJSTest('no js environment should have first screen', async ({ page }) => {
-      noJSTest.fixme();
       await page.goto(`http://localhost:${port}/`);
       await expect(page.getByTestId('app-name')).toHaveText('Waku');
       await expect(page.getByTestId('count')).toHaveText('0');
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -1,0 +1,78 @@
+import { expect } from "@playwright/test";
+import { execSync, exec, ChildProcess } from "node:child_process";
+import { fileURLToPath } from "node:url";
+import waitPort from "wait-port";
+import { test } from "./utils.js";
+
+const waku = fileURLToPath(
+  new URL("../packages/waku/dist/cli.js", import.meta.url)
+);
+
+const commands = [
+  {
+    command: "dev"
+  },
+  {
+    build: "build",
+    command: "start"
+  }
+];
+
+const cwd = fileURLToPath(new URL("./fixtures/ssr-basic", import.meta.url));
+
+for (const { build, command } of commands) {
+  test.describe(`rsc-basic: ${command}`, () => {
+    let cp: ChildProcess;
+    let port: number;
+    test.beforeAll(async () => {
+      if (build) {
+        execSync(`node ${waku} ${build}`, {
+          cwd
+        });
+      }
+      port = Math.floor(Math.random() * 10000) + 10000;
+      console.log(`node ${waku} ${command}`);
+      console.log("cwd: ", cwd);
+      cp = exec(`node ${waku} ${command}`, {
+        cwd,
+        env: {
+          ...process.env,
+          PORT: `${port}`
+        }
+      });
+      cp.on("message", (message) => {
+        console.log("cp message: ", message);
+      });
+      await waitPort({
+        port
+      });
+    });
+
+    test.afterAll(async () => {
+      cp.kill();
+    });
+
+    test("increase counter", async ({ page }) => {
+      await page.goto(`http://localhost:${port}/`);
+
+      await expect(page.getByTestId("app-name")).toHaveText("Waku");
+
+      await expect(page.getByTestId("count")).toHaveText("0");
+      await page.getByTestId("increment").click();
+      await page.getByTestId("increment").click();
+      await page.getByTestId("increment").click();
+      await expect(page.getByTestId("count")).toHaveText("3");
+    });
+
+    test("no js environment should have first screen", async ({ browser }) => {
+      test.fixme();
+      const noJSContext = await browser.newContext({
+        javaScriptEnabled: false
+      });
+      const noJSPage = await noJSContext.newPage();
+      await noJSPage.goto(`http://localhost:${port}/`);
+      await expect(noJSPage.getByTestId("app-name")).toHaveText("Waku");
+      await expect(noJSPage.getByTestId("count")).toHaveText("0");
+    });
+  });
+}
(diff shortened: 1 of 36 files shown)


packages/waku/src/client.ts
```diff
@@ -50,7 +50,7 @@ export const fetchRSC = cache(
     const options = {
       async callServer(actionId: string, args: unknown[]) {
         const response = fetch(
-          basePath + encodeInput(encodeURIComponent(actionId)),
+          basePath + encodeURI(encodeInput(encodeURIComponent(actionId))),
```
dai-shi, line 53: This seems too ad-hoc. `input` is used for both components and actions.
I was hoping `encodeURIComponent` works, but if it doesn't escape enough, maybe we should use base64.
We can of course re-consider `encodeInput` implementation, but encoding too much is bad in DX.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -294,7 +297,10 @@ async function renderRSC(rr: RenderRequest): Promise<PipeableStream> {
     {},
     {
       get(_target, encodedId: string) {
-        const [filePath, name] = encodedId.split('#') as [string, string];
+        const [filePath, name] = normalizePath(encodedId).split('#') as [
```
dai-shi, line 300: Shouldn't we normalize only the `filePath` part?

himself65:
there is only windows prod mode now

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -252,7 +244,7 @@ const transformRsfId = (prefixToRemove: string) =>
       for (let i = 0; i < lines.length; ++i) {
         const match = lines[i].match(
           new RegExp(
-            `^([0-9]+):{"id":"(?:file://)?${prefixToRemove}(.*?)"(.*)$`,
+            `^([0-9]+):{"id":"(?:file:\/\/\/)?${prefixToRemove}(.*?)"(.*)$`,
```
himself65, line 247: where the bug comes from on Windows

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -249,8 +244,8 @@ const transformRsfId = (prefixToRemove: string) =>
       for (let i = 0; i < lines.length; ++i) {
         const match = lines[i].match(
           new RegExp(
-            `^([0-9]+):{"id":"(?:file://)?${prefixToRemove}(.*?)"(.*)$`,
-          ),
+            `^([0-9]+):{"id":"(?:file://?/)?${prefixToRemove}(.*?)"(.*)$`
```
himself65, line 247: On windows, there is one more extra `/`

himself65:
@dai-shi I added the SSR test case, seems like no js environment doesn't work

e2e/ssr-basic.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    test('no js environment should have first screen', async ({ browser }) => {
+      test.fixme();
+      const noJSContext = await browser.newContext({
+        javaScriptEnabled: false,
+      });
+      const noJSPage = await noJSContext.newPage();
+      await noJSPage.goto(`http://localhost:${port}/`);
+      await expect(noJSPage.getByTestId('app-name')).toHaveText('Waku');
+      await expect(noJSPage.getByTestId('count')).toHaveText('0');
```
himself65, line 75: todo: fix this

packages/waku/src/client.ts
```diff
@@ -49,13 +49,11 @@ export const fetchRSC = cache(
...
     const options = {
       async callServer(actionId: string, args: unknown[]) {
-        const response = fetch(
-          basePath + encodeInput(encodeURIComponent(actionId)),
-          {
-            method: 'POST',
-            body: await encodeReply(args),
-          },
-        );
+        const input = basePath + encodeInput(encodeURIComponent(actionId));
```
dai-shi, line 52: In my terminology, `input` is a string passed to a function in `entries.ts`.
So, let's rename it:
```suggestion
        const reqPath = basePath + encodeInput(encodeURIComponent(actionId));
```

packages/waku/src/lib/builder.ts
```diff
@@ -176,15 +176,19 @@ const buildServerBundle = async (
   const code = `export const resolveClientPath = (filePath, invert) => (invert ? ${JSON.stringify(
     Object.fromEntries(
       Object.entries(clientEntryFiles).map(([key, val]) => [
-        path.posix.join(config.rootDir, config.distDir, 'assets', key + '.js'),
+        normalizePath(
+          path.join(config.rootDir, config.distDir, 'assets', key + '.js'),
+        ),
```
dai-shi, line 181: Nice.

packages/waku/src/lib/builder.ts
```diff
@@ -279,8 +283,7 @@ const emitRscFiles = async (
           config.distDir,
           config.publicDir,
           config.rscPath,
-          // HACK to support windows filesystem
-          encodeInput(input).replaceAll('/', path.sep),
+          encodeInput(normalizePath(input)),
```
dai-shi, line 286: Very nice.

packages/waku/src/lib/config.ts
```diff
@@ -2,6 +2,7 @@ import fs from 'node:fs';
 import path from 'node:path';
 
 import type { Config } from '../config.js';
+import { normalizePath } from 'vite';
```
dai-shi, line 5: This file is used in production. So, we can't depend on `vite`.

Maybe, we need to take `command` argument in `resolveConfig` function below, and only if it's `'dev'` , do `import('vite')`. We do this in various places.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -121,7 +121,10 @@ export function rsc<Context>(options: {
   return async (req, res, next) => {
     const config = await configPromise;
     const basePrefix = config.basePath + config.rscPath + '/';
-    const pathStr = req.url || '';
+    let pathStr = req.url || '';
+    if (pathStr.startsWith('/@fs/')) {
+      pathStr = pathStr.slice('/@fs/'.length);
+    }
```
dai-shi, line 127: This is DEV-only, so only do it if `command==="dev"`.
Though, I'm not 100% sure if it's the right way to trim `/@fs` here.

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -18,6 +18,7 @@ import { defineEntries } from '../../../server.js';
 import { ServerRoot } from '../../../client.js';
 import { renderRSC } from './worker-api.js';
 import { hasStatusCode } from './utils.js';
+import { normalizePath } from 'vite';
```
dai-shi, line 21: Same here. We can't depend vite in production mode.

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -303,16 +304,20 @@ export const renderHtml = async <Context>(
             get(_target, name: string) {
               const file = filePath.slice(config.basePath.length);
               if (command === 'dev') {
-                const filePath = file.startsWith('@fs/')
-                  ? file.slice(3)
-                  : path.join(config.rootDir, config.srcDir, file);
+                const filePath = normalizePath(
+                  file.startsWith('@fs/')
+                    ? file.slice('@fs/'.length)
```
dai-shi, line 309: Is it intentional? `3 !== '@fs/'.length`

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -18,6 +18,7 @@ import {
   defineEntries,
   runWithAsyncLocalStorage as runWithAsyncLocalStorageOrig,
 } from '../../../server.js';
+import { normalizePath } from 'vite';
```
dai-shi, line 21: Same here again. If we need such a functionality in production, we need to create one for ourselves.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -249,7 +244,7 @@ const transformRsfId = (prefixToRemove: string) =>
       for (let i = 0; i < lines.length; ++i) {
         const match = lines[i].match(
           new RegExp(
-            `^([0-9]+):{"id":"(?:file://)?${prefixToRemove}(.*?)"(.*)$`,
+            `^([0-9]+):{"id":"(?:file://?/)?${prefixToRemove}(.*?)"(.*)$`,
```
dai-shi, line 247: Hm, what was the case that this regex didn't work?

dai-shi: And, can it be this?
```suggestion
            `^([0-9]+):{"id":"(?:file:///?)?${prefixToRemove}(.*?)"(.*)$`,
```

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -296,7 +291,9 @@ async function renderRSC(rr: RenderRequest): Promise<PipeableStream> {
       get(_target, encodedId: string) {
         const [filePath, name] = encodedId.split('#') as [string, string];
         const id = resolveClientEntry(
-          filePath,
+          filePath.startsWith('file:///')
```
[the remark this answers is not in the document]
dai-shi: I wonder if we can create a utility function. Maybe a separate pr.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -328,8 +325,10 @@ async function renderRSC(rr: RenderRequest): Promise<PipeableStream> {
         args = await decodeReply(body);
       }
     }
-    const [fileId, name] = actionId.split('#');
-    const filePath = path.join(config.rootDir, fileId!);
+    const [fileId, name] = actionId.split('#') as [string, string];
+    const filePath = normalizePath(
+      path.join(config.rootDir, normalizePath(fileId)),
+    );
```
dai-shi, line 331: You need to normalize twice?

dai-shi (changes requested):
some of comments should be addressed before merging.

packages/waku/src/lib/middleware/rsc/ssr.ts
himself65: mistake I think, auto-complete by copilot 🤔

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -109,6 +109,9 @@ export function rsc<
     } catch (e) {
       // does not exist
     }
+    if (pathStr.startsWith('/@fs')) {
+      return null;
+    }
```
himself65, line 114: I dont know why pathStr could be `'/@fs/C://xxx/Counter.tsx` on Windows

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -317,16 +317,21 @@ export const renderHtml = async <Context>(
...
+                    : path.join(config.rootDir, config.srcDir, file),
+                );
+                console.log('filePath', filePath);
```
dai-shi, line 325:
```suggestion
```

packages/waku/src/lib/middleware/rsc/utils.ts
```diff
@@ -1,5 +1,14 @@
 // This file should not include Node specific code.
 
+// copy from Node.js source code
+const win32Sep = '\\';
+const posixSep = '/';
+
+// Refs: https://github.com/rollup/plugins/blob/d49bbe8dc5ec41157de5787c72c858f73be107ff/packages/pluginutils/src/normalizePath.ts
+export function normalizePath(filePath: string) {
+  return filePath.split(win32Sep).join(posixSep);
```
dai-shi, line 9: That's simple.
I'd inline them, as it's fairly readable.
```suggestion
  return filePath.split('\\').join('/');
```

This doesn't work if posix path contains backslashes. Can't be helped.

dai-shi (commented):
Looks good. Please see comments.

dai-shi (approved):

packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -319,7 +319,9 @@ export const renderHtml = async <Context>(
               if (command === 'dev') {
                 const filePath = normalizePath(
                   file.startsWith('@fs/')
-                    ? file.slice('@fs/'.length)
+                    // FIXME This is ugly. We need to refactor it.
+                    // remove '@fs'(3) on Unix and '@fs/'(4) on Windows
+                    ? file.slice(2 + path.sep.length)
```
himself65, line 324: I just found that `@fs/User/path/to/file` on Unix, but `@fs/C://path/to/file` on Windows

Is this kind of **feature** on Vite?

dai-shi (commented):
Now, I need to resolve conflicts in my PRs.

dai-shi:
Ah, this PR broke `npm run examples:dev:03_promise` and `05` too. And also `07`.

PR #189, opened 2023-11-28
fix: ssr in dev mode

dai-shi:
Fix #188

There was an issue with `waku dev --with-ssr`. It worked within the repo, so it was hard to notice.

The error is now reproduced within the repo, but CI doesn't have `--with-ssr` yet. #186

PR #195, opened 2023-12-01
fix: ssr in dev mode again

dai-shi:
I just noticed #189 didn't fix the issue fully. `ChildrenContext` is isolated.


dai-shi (commented):
Now, I'm getting to understand how things work. We need a big refactor in the near future.

PR #196, opened 2023-12-01
refactor: use web stream internally (step 2)

dai-shi:
Continuing from #190.

This converts mainly `ssr.ts`.


dai-shi:
It's already quite a big change. Merging this and will continue with another PR.

Issue #202, opened 2023-12-03
Hydration error with examples/03_promise


dai-shi:
Since #147, there has been a hydration error like this:
<img width="531" alt="image" src="https://github.com/dai-shi/waku/assets/490574/bb7799ae-dbc2-4709-87cb-0ee9da0283c6">

I thought, it's the timing issue of HTML and RSC payload streaming, but I don't think that solves it.
For now, it's a limitation. Don't know if it's something we can do.

Feel free to dig into it, if you are familiar with hydration with streaming.

PR #203, opened 2023-12-03
fix: hydration error on promise example

himself65:
This will fix the hydration issue. Not sure if this is correct

Fixes: https://github.com/dai-shi/waku/issues/202

### DIFF
--- a/examples/03_promise/src/components/Counter.tsx
+++ b/examples/03_promise/src/components/Counter.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { Suspense, useState } from 'react';
+import { Suspense, useState, use } from 'react';
 
 export const Counter = ({
   delayedMessage,
@@ -26,9 +26,12 @@ const Message = ({
 }: {
   count: number;
   delayedMessage: Promise<string>;
-}) => (
-  <ul>
-    <li>count: {count}</li>
-    <li>delayedMessage: {delayedMessage as any}</li>
-  </ul>
-);
+}) => {
+  const message = use(delayedMessage);
+  return (
+    <ul>
+      <li>count: {count}</li>
+      <li>delayedMessage: {message}</li>
+    </ul>
+  );
+};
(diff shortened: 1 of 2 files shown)


examples/03_promise/src/components/Counter.tsx
```diff
@@ -26,9 +26,12 @@ const Message = ({
...
-    <li>count: {count}</li>
-    <li>delayedMessage: {delayedMessage as any}</li>
-  </ul>
-);
+}) => {
+  const message = use(delayedMessage);
+  return (
+    <ul>
+      <li>count: {count}</li>
+      <li>delayedMessage: {message}</li>
```
dai-shi, line 34: Hm, ARAIU, React should internally `use` a promise which is directly put in JSX.

examples/03_promise/tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "esModuleInterop": true,
     "module": "nodenext",
     "skipLibCheck": true,
+    "types": ["react/canary"],
```
dai-shi, line 9: Let's prefer `///` reference in the source code.

examples/03_promise/src/components/Counter.tsx
himself65: I'm not sure why...🤔 maybe ask React member to get some help?

dai-shi:
> This will fix the hydration issue. Not sure if this is correct

If using `use` fixes the issue, I think it's React's issue.

examples/03_promise/src/components/Counter.tsx
dai-shi, line 34: I would like to merge this with something like this:
```suggestion
     {/* We want show the usage without `use` but it causes a hydration error. https://github.com/dai-shi/waku/issues/202 */}
      <li>delayedMessage: {use(delayedMessage)}</li>
```

dai-shi (approved):

PR #205, opened 2023-12-03
fix: error on waku examples

himself65:
I typed the wrong number

https://github.com/dai-shi/waku/commit/827b9c8a81e98eca4d8d7f5d761e510fe6d6dde2#diff-4b94963be63fc6422d66747d2480378c76469ec0c8d7b039aa709aefc16e58b3R324


packages/waku/src/lib/middleware/rsc/ssr.ts
```diff
@@ -299,7 +299,7 @@ export const renderHtml = async <Context>(
                   file.startsWith('@fs/')
                     ? // FIXME This is ugly. We need to refactor it.
                       // remove '@fs'(3) on Unix and '@fs/'(4) on Windows
-                      file.slice(2 + path.sep === '/' ? 3 : 4)
+                      file.slice(2 + (path.sep === '/' ? 1 : 2))
```
dai-shi, line 302: Is `2 +` meaningful? Otherwise:
```suggestion
                      file.slice(path.sep === '/' ? 3 : 4)
```

packages/waku/src/lib/middleware/rsc/worker-impl.ts
```diff
@@ -208,6 +208,9 @@ const resolveClientEntry = (
   command: 'dev' | 'build' | 'start',
   resolveClientPath: Entries['resolveClientPath'],
 ) => {
+  filePath = filePath.startsWith('file:///')
```
dai-shi, line 211: Does triple slash mean it's only for posix?

dai-shi (approved):
The behavior is fine and the change looks okay.

packages/waku/src/lib/middleware/rsc/worker-impl.ts
himself65: This is just a check, won't remove the prefix `/`

dai-shi: yeah, but I mean for windows, is it like `file://d/...`?

himself65: I bet there only have file:///C://

let me try it tomorrow

dai-shi: oh i see.

PR #207, opened 2023-12-03
feat: use vite5

himself65:
Just wanna see if it works


dai-shi:
No breaking changes for us between v4 and v5?

himself65:
I go through the migration doc. I think there is no

himself65:
https://vitejs.dev/guide/migration.html

PR #209, opened 2023-12-03
chore: fs promises

Aslemammad:
...and prefer for-of to forEach.


dai-shi:
Thanks for working on the refactor. Let me add some commits.

packages/waku/src/lib/middleware/rsc/utils.node.ts
```diff
@@ -0,0 +1,3 @@
+import fsPromises from 'node:fs/promises';
+export const fileExists = async (path: string) =>
+  !!(await fsPromises.stat(path).catch(() => false));
```
dai-shi, line 3: Let's use sync version for exists.

I need to restructure the code base, but I'd like to avoid "utils".

PR #213, opened 2023-12-04
breaking: no async local storage

dai-shi:
We can't get context dynamically, but this simplifies things.
Hopefully, we revisit it in the future.


dai-shi:
Let's revisit this after we have our own compiler.

Issue #216, opened 2023-12-04
"TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed"


himself65:
Next.js also have same issue, seems like a React bug

https://github.com/vercel/next.js/issues/56919

dai-shi:
fyi, we start seeing this error after #165 or #190.

dai-shi:
#229 doesn't seem to fix it.

PR #217, opened 2023-12-04
fix: page aborted but rsc still rendering

himself65:
Fixes: https://github.com/dai-shi/waku/issues/216


packages/waku/src/lib/middleware/connectWrapper.ts
```diff
@@ -5,7 +5,7 @@ import type { BaseReq, BaseRes, Middleware } from './types.js';
 export function connectWrapper(
   m: Middleware<
     BaseReq & { orig: IncomingMessage },
-    BaseRes & { orig: ServerResponse }
+    BaseRes & { orig: ServerResponse; signal: AbortSignal }
```
dai-shi, line 8: BaseRes already has signal prop, right?

dai-shi:
#304

PR #220, opened 2023-12-04
Refactor: custom path utils

### LINKED ISSUE #211: Refactor: less dependent on node:path
For example:
- Create `pathJoin(items: string[], sep = '/')` in src/lib/utils/path.ts
- Do not use path.dirname, use pathJoin instead
- Do not use path.extname, use regex instead

dai-shi:
Close #211


dai-shi (commented):
should be good.

PR #228, opened 2023-12-06
breaking: move index.html to parent

Aslemammad:


e2e/fixtures/rsc-router/index.html
```diff
@@ -0,0 +1,37 @@
+<!doctype html>
```
Aslemammad: Wow, I thought the router one was doing things without index.html. Thank you so much.

dai-shi (approved):

PR #233, opened 2023-12-09
feat: separate dev and prd apis

dai-shi:
for dev, it needs cwd (now using vite's root), for prd, it needs loaded entries.


Aslemammad:
What's the motivation behind this? Just curious.

PR #237, opened 2023-12-09
chore: update dependencies

dai-shi:


dai-shi:
https://github.com/honojs/hono/pull/1761 requires the content-type header. we should have had it anyway.

PR #240, opened 2023-12-09
feat: css support

Aslemammad:
Follow up #171


packages/waku/src/lib/rsc/worker-impl.ts
```diff
@@ -3,7 +3,11 @@
 import url from 'node:url';
 import { parentPort } from 'node:worker_threads';
 import { Server } from 'node:http';
-import { createServer as viteCreateServer } from 'vite';
+import {
+  createServer as viteCreateServer,
+  resolveConfig as resolveViteConfig,
+  mergeConfig as mergeViteConfig,
```
dai-shi, line 9: Let's make our naming convention consistent.
Do you prefer `resolveViteConfig` to `viteResolveConfig`?

dai-shi (commented):
wow, this look like a great work. i'll try it later.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -1,22 +1,33 @@
...
+
   import.meta.hot.on('hot-import', (data) => import(/* @vite-ignore */ data));
+
+  import.meta.hot.on('module', (data) => {
+    const code = data.code
+    const script = document.createElement('script')
+    script.type = 'module'
+    script.text = code
+    document.head.appendChild(script)
+  });
```
himself65, line 18: code format inconstant. semi

packages/waku/src/lib/rsc/worker-impl.ts
Aslemammad: The first one is better I guess.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
Aslemammad: You're correct, I'll fix it after Daishi reviews the code. Thank you.

packages/waku/src/lib/rsc/worker-impl.ts
dai-shi: Fine then. We should change everything. If it's small, it can be done in this PR.

dai-shi:
When I run `npm run examples:dev:12_css`, there's a warning on the server console:

> The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.


and, a warning on the bowser console:

> Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/html". Strict MIME type checking is enforced for module scripts per HTML spec.

Are these expected?

packages/waku/src/lib/plugins/vite-plugin-rsc-delegate.ts
```diff
@@ -1,25 +1,27 @@
...
+  let server: ViteDevServer;
   return {
     name: 'rsc-delegate-plugin',
     configResolved(config) {
       mode = config.mode;
       base = config.base;
     },
-    transform(code, id) {
+    configureServer(_server) {
+      server = _server;
```
dai-shi, line 22: We use `_` prefix specifically for "unused" values.

```suggestion
    configureServer(s) {
      server = s;
```

dai-shi: Or
```suggestion
    configureServer(instance) {
      server = instance;
```

```suggestion
    configureServer(serverInstance) {
      server = serverInstance;
```

packages/waku/src/lib/plugins/vite-plugin-rsc-delegate.ts
```diff
@@ -36,10 +38,17 @@ export function rscDelegatePlugin(
...
+                item.source.value,
+                id,
+                { ssr: true },
+              );
+              if (resolvedSource?.id) {
+                const transformedResult = await server.transformRequest(
+                  resolvedSource.id,
+                );
+                transformedResult && importCallback(transformedResult);
+              }
```
dai-shi, line 51: I do not fully understand this, but if works, it would be fine. I wonder if @himself65 has some experience.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -37,3 +48,22 @@ export function hotImport(vite: ViteDevServer, source: string) {
   sourceSet.add(source);
   vite.ws.send({ type: 'custom', event: 'hot-import', data: source });
 }
+
+const modulePendingMap = new WeakMap<ViteDevServer, Set<TransformResult>>();
+export function moduleImport(
```
dai-shi, line 53:
```suggestion

export function moduleImport(
```

packages/waku/src/lib/rsc/handler-dev.ts
```diff
@@ -36,27 +45,40 @@ export function createHandler<
...
-        exclude: ['waku'],
+    const resolvedViteConfig = await resolveViteConfig({}, 'serve');
+    const mergedViteConfig = await mergeViteConfig(
+      {
+        ...resolvedViteConfig,
+
+        plugins: resolvedViteConfig.plugins.filter(
+          (plugin) => !plugin.name.startsWith('vite:'),
+        ),
       },
```
dai-shi, line 57: What happens if we don't have this? Why do we need to remove `vite:` plugins?

packages/waku/src/lib/rsc/worker-api.ts
```diff
@@ -30,6 +31,7 @@ export type MessageReq =
 export type MessageRes =
   | { type: 'full-reload' }
   | { type: 'hot-import'; source: string }
+  | { type: 'module'; result: TransformResult }
```
dai-shi, line 34: It's ambiguous, can we call it `module-import` instead of `module`?

packages/waku/src/lib/rsc/worker-api.ts
```diff
@@ -43,27 +45,23 @@ const getWorker = () => {
...
-          ],
-        });
-        worker.on('message', (mesg: MessageRes) => {
-          if ('id' in mesg) {
-            messageCallbacks.get(mesg.id)?.(mesg);
-          }
-        });
-        resolve(worker);
-      })
-      .catch((e) => reject(e));
```
dai-shi, line 66: Please don't remove this.

packages/waku/src/lib/rsc/worker-impl.ts
```diff
@@ -87,30 +91,47 @@ const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
...
-      conditions: ['react-server', 'workerd'],
-      externalConditions: ['react-server', 'workerd'],
+
+const resolvedViteConfig = await resolveViteConfig({}, 'serve');
+const mergedViteConfig = await mergeViteConfig(
+  {
+    ...resolvedViteConfig,
+
+    plugins: resolvedViteConfig.plugins.filter(
+      (plugin) => !plugin.name.startsWith('vite:'),
```
dai-shi, line 101: Again, I wonder why we need this and if we need both in two places.

dai-shi (commented):
Sorry for some questions due to my limited understanding of Vite.

packages/waku/src/lib/plugins/vite-plugin-rsc-delegate.ts
himself65: `pluginContainer` can run the vite to a specific file

packages/waku/src/lib/rsc/worker-impl.ts
Aslemammad: It's because we don't vite default plugins to be mentioned two times. For instance react refresh would fail because there would two instances of it.

packages/waku/src/lib/rsc/worker-api.ts
Aslemammad: Why not, let me try.

packages/waku/src/lib/rsc/handler-dev.ts
Aslemammad: React refresh would fail

packages/waku/src/lib/plugins/vite-plugin-rsc-delegate.ts
Aslemammad: First, the previous implementation always assumes there's a filesystem import, which is wrong. 

Two, resolveId would run all of the plugins' resolveId function to resolve the right path for the import (whether local or npm module). When we get the right path for that, we transform the request so we get a vite-compatible module result. For instance, if we transformRequest a CSS file, in this case, vite would give us a hmr js module code specific to that CSS file, so when we use that code, we enable CSS in hmr and it'd also be linked to the document (CSS would get applied) 

Let me know if there's any thing vague.

dai-shi: thanks!

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
dai-shi: please fix it.

packages/waku/src/lib/rsc/worker-impl.ts
dai-shi: Please check all `import { doSomethign as doViteSomething} from 'vite'` throughout the code base. My previous code did `viteDoSomething`.

packages/waku/src/lib/rsc/worker-impl.ts
dai-shi: https://github.com/dai-shi/waku/pull/240#discussion_r1424492002

> React refresh would fail

I see. I will play with it later. Does it mean there's only one vite server that keep vite plugins? (Can't guess which from the diff, and I don't remember how many they are...)

Please extract this patch/merge logic in a separate file and reuse it. something like `packages/waku/src/lib/plugins/patch-react-refresh.ts`.

dai-shi (commented):
Looking good! Please see comments.

packages/waku/src/lib/rsc/worker-impl.ts
dai-shi: I don't exactly mean patching plugin. It should be a reusable logic and preferably add some comments for the intention.

Aslemammad:
I don't see the second warning, but for the first one, to be honest I don't have any idea. Can we tackle it in another PR later?

examples/12_css/vite.config.ts
```diff
@@ -0,0 +1,9 @@
+import url from 'node:url';
+import path from 'node:path';
+import { defineConfig } from 'vite';
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+
+export default defineConfig({
+  root: path.dirname(url.fileURLToPath(import.meta.url)),
```
dai-shi, line 7: I think we can omit this. (Can be a separate PR, as I might be wrong, and there are other places like this.)

examples/12_css/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "build": "waku build --with-ssr",
+    "start": "waku start --with-ssr"
+  },
+  "dependencies": {
+    "@hono/node-server": "^1.3.1",
+    "hono": "^3.11.4",
+    "react": "18.3.0-canary-5bcade5fc-20231208",
+    "react-dom": "18.3.0-canary-5bcade5fc-20231208",
+    "react-server-dom-webpack": "18.3.0-canary-5bcade5fc-20231208",
+    "waku": "0.17.1"
```
dai-shi, line 17:
```suggestion
    "waku": "0.18.0"
```

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -1,22 +1,33 @@
...
-      const ext = path.extname(id);
-      if (['.ts', '.tsx', '.js', '.jsx'].includes(ext)) {
-        return code + customCode;
-      }
-      return code;
+    enforce: 'post',
+    transformIndexHtml(html) {
+      return html.replace(
+        '</body>',
+        `<script type="module">${customCode}</script></body>`,
```
dai-shi, line 29:
```suggestion
        `<script type="module" async>${customCode}</script></body>`,
```
should be more consistent and safer, but i'm not sure. will see..

packages/waku/src/lib/utils/merge-vite-config.ts
```diff
@@ -0,0 +1,24 @@
+import type { UserConfig } from 'vite';
+import {
+  resolveConfig as resolveViteConfig,
+  mergeConfig as mergeViteConfig,
+} from 'vite';
+
+export async function mergeUserViteConfig(config: UserConfig) {
```
dai-shi, line 7: It would be nice to add a comment about this workaround is required only (?) for hmr. (Can be a separate PR.)

dai-shi (approved):
thanks for your work!

PR #244, opened 2023-12-11
fix: getClientId error

dai-shi:
The error happens when we run waku build normally (outside of this repo).

I wish there's a way to reproduce the behavior within the repo...


packages/waku/src/lib/plugins/vite-plugin-nonjs-resolve.ts
```diff
@@ -18,7 +18,7 @@ export function nonjsResolvePlugin(): Plugin {
         const resolved = await this.resolve(
           id.slice(0, -path.extname(id).length) + ext,
           importer,
-          { ...options, skipSelf: true },
+          options,
```
dai-shi, line 21: https://vitejs.dev/guide/migration#rollup-4 says:
> For Vite plugins, this.resolve skipSelf option is now true by default.

himself65:
> I wish there's a way to reproduce the behavior within the repo...

I can do that

PR #252, opened 2023-12-13
fix: website issue with safari

dai-shi:
This is tricky...


packages/waku/src/lib/builder.ts
```diff
@@ -418,7 +418,7 @@ export function loadHtml(pathStr) {
           // HACK is this too naive to inject script code?
           htmlStr = htmlStr.replace(
             /<\/head>/,
-            `<script>${code}</script></head>`,
+            `<script type="module" async>${code}</script></head>`,
```
dai-shi, line 421: This is the major fix. The other two fixes are sort of small improvements.

Issue #253, opened 2023-12-13
url with `?a=b` will crash vercel serverless function


himself65:
```shell
Unhandled Promise Rejection 	{"errorType":"Runtime.UnhandledPromiseRejection","errorMessage":"Error: not handled","reason":{"errorType":"Error","errorMessage":"not handled","stack":["Error: not handled","    at file:///var/task/serve.js:7:11","    at file:///var/task/node_modules/waku/dist/lib/rsc/handler-prd.js:93:9","    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"]},"promise":{},"stack":["Runtime.UnhandledPromiseRejection: Error: not handled","    at process.<anonymous> (file:///var/runtime/index.mjs:1276:17)","    at process.emit (node:events:529:35)","    at emit (node:internal/process/promises:149:20)","    at processPromiseRejections (node:internal/process/promises:283:27)","    at process.processTicksAndRejections (node:internal/process/task_queues:96:32)"]}
Unknown application error occurred
Runtime.Unknown
```

himself65:
workaround:
![image](https://github.com/dai-shi/waku/assets/14026360/f826a431-0953-44e5-999a-4fd812611657)

dai-shi:
So, I wanted to support RSC static sites with query params, but that was the wrong decision. We should just ignore query params for static sites regardless of with or without SSR(SSG).

dai-shi:
#256 fixed the "crash", so the issue was fixed at that point. Reopening this because we need to avoid 404 too.

PR #254, opened 2023-12-13
fix: try to debug error

himself65:
Fixes: https://github.com/dai-shi/waku/issues/253


packages/waku/src/lib/rsc/handler-prd.ts
```diff
@@ -42,7 +42,7 @@ export function createHandler<
   return async (req, res, next) => {
     const config = await configPromise;
     const basePrefix = config.basePath + config.rscPath + '/';
-    const pathStr = req.url.slice(new URL(req.url).origin.length);
+    const pathStr = new URL(req.url).pathname;
```
dai-shi, line 45: This isn't our design. `pathStr` means `pathname` + `query`. We don't have examples that use query yet, but we can't drop it here.

himself65:
I think we need a better logic of fallback. I will fix this in another way

PR #256, opened 2023-12-13
fix: avoid 500 when it should not be

### LINKED ISSUE #253: url with `?a=b` will crash vercel serverless function

dai-shi:
close #253 
supersede #254


himself65 (commented):
things interesting. In 01 example, visit `http://localhost:3000/?a=b`. Website will be at loading spinner

dai-shi:
Can't repro. Is it vite's 504 error? How can it be fixed?

himself65 (changes requested):
I think we should return 404 only if getSsrConfig returns null?

dai-shi:
> I think we should return 404 only if getSsrConfig returns null?

When getSsrConfig returns `null`, it doesn't mean 404. It goes to the next middleware and static files might be served. Consider, for example, `/favicon.ico`.

PR #262, opened 2023-12-14
fix: css

Aslemammad:
This reverts commit f32f338fe9138456d68b3bbd2b078633565646c7 which is originally #240.

The only issue was that I added an extra merge config hack which I *assume* it didn't let vercel generate enough assets for deployment. I removed it from that stage since it was already working before, and now it's fixed.


packages/waku/src/lib/rsc/html-renderer.ts
```diff
@@ -48,16 +48,14 @@ const getViteServer = async () => {
...
-    await mergeUserViteConfig({
-      plugins: [nonjsResolvePlugin()],
-      ssr: {
-        external: ['waku'],
-      },
-      appType: 'custom',
-      server: { middlewareMode: true, hmr: { server: dummyServer } },
-    }),
-  );
+  const viteServer = await createViteServer({
```
dai-shi: Oh, I forgot this code runs on build, even though it doesn't on PRD. 😓

Aslemammad: haha, i just thought I forgot to put it so I put it here, but it was the reason for failure.

PR #264, opened 2023-12-14
fix: extra rsc

dai-shi:
There was a regression in #252.


packages/waku/src/lib/rsc/html-renderer.ts
```diff
@@ -101,7 +101,7 @@ const injectRscPayload = (readable: ReadableStream, input: string) => {
   const modifyHead = (data: string) => {
     const matchPrefetched = data.match(
       // HACK This is very brittle
-      /(.*)<script>\nglobalThis\.__WAKU_PREFETCHED__ = {\n(.*?)\n};(.*)/s,
+      /(.*)<script[^>]*>\nglobalThis\.__WAKU_PREFETCHED__ = {\n(.*?)\n};(.*)/s,
```
dai-shi, line 104: > very brittle

Yes, indeed.

PR #270, opened 2023-12-14
fix: suppress terserOptions warning

### LINKED ISSUE #266: Terser warning after css support in server components
> @Mohammad Bagher I forgot to check the warning. It's still there. Can you investigate it please?
build.terserOptions is specified but build.minify is not set to use Terser. Note Vite now defaults to use esbuild for minification. If you still prefer Terser, set build.minify to "terser".

To reproduce, run examples:dev:01_counter.

Aslemammad:
Resolves #266 

Just set the options to null when there's nothing there.


himself65 (approved):

dai-shi (approved):
Thanks!

PR #274, opened 2023-12-15
fix: Update engine spec and some refactoring on cli.ts

idiglove:
- Update node engine spec
- Add error.message on startServer error
- refactor some code on cli.ts - applied DRY principle


dai-shi (approved):
Thanks for your contribution!

PR #282, opened 2023-12-17
docs: npm install to pnpm install

ojj1123:
Hi. I'm interested in waku. So I've forked and played with some examples.
When I install the dependencies using `npm`, it had an error below:

스크린샷 2023-12-17 23 13 15

So I revised README.md [here](https://github.com/dai-shi/waku?tab=readme-ov-file#how-to-try-it-1) (Need to install dependencies using `pnpm`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ the following commands:
 ```sh
 git clone https://github.com/dai-shi/waku.git
 cd waku
-npm install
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
Oh, nice catch! Why did we miss it for such a long time. 😓

Issue #110, opened 2023-07-28
Cannot import commonjs module from server component


tom-sherman:
When trying to import a CJS module from a server component I get an error in dev & production

```
Error when evaluating SSR module /src/components/App.tsx: failed to import "/src/components/foo.cjs"
|- ReferenceError: require is not defined
    at eval (/Users/tom.sherman/code/waku-example/src/components/foo.cjs:3:31)
    at instantiateModule (file:///Users/tom.sherman/code/waku-example/node_modules/.pnpm/vite@4.4.7/node_modules/vite/dist/node/chunks/dep-3b8eb186.js:55920:15)

Cannot render RSC ReferenceError: require is not defined
    at eval (/Users/tom.sherman/code/waku-example/src/components/foo.cjs:3:31)
    at instantiateModule (file:///Users/tom.sherman/code/waku-example/node_modules/.pnpm/vite@4.4.7/node_modules/vite/dist/node/chunks/dep-3b8eb186.js:55920:15)
```

Minimal reproduction: https://github.com/tom-sherman/waku-example/commit/77cf3501920bacdd2fd9a873d01abc4e848b373f

dai-shi:
Hmm, I thought it works, but it doesn't. (btw, `.cts` nor `.mts` doesn't work and we need to deal with it.) The error is caused with Vite server file loading.
It's either a) there's a misconfiguration for Vite, b) some Vite plugins do something wrong, c) Vite doesn't support it (unlikely?) or d) there's a misunderstanding of mine.

tom-sherman:
> btw, .cts nor .mts doesn't work and we need to deal with it

Fair point, this was the easiest way to trigger the error for me but I also was able to trigger the error by importing a CJS package from node_modules

dai-shi:
@PerfectPan @nksaraf 
Do you happen to know if ssrLoadModule supports `.cjs` ? (More precisely, import .cjs file from an ESM file loaded with `ssrLoadModule`.)
Or, if we are missing something.

nksaraf:
No I think within your project you shouldn't really have cjs.. that way vite is ESM only.. it just knows what to do about cjs files in node_modules. But also I could be wrong to pinging @patak-dev

tom-sherman:
Note, although the example uses CJS inside my project, this error also appears for CJS inside node_modules. 

IMO both of these behaviours are strange, in a Node.js ESM environment you are able to import a CJS module without issue.

dai-shi:
[...]

PR #286, opened 2023-12-17
feat: add `@stylexjs/stylex` example

himself65:


himself65 (commented):
First. stylex doesn't support esm module, so I use vite cjs plugin to transform it to esm. But I think in dev mode, server inline the code twice so make the error happened

himself65 (commented):
First. stylex doesn't support esm module, so I use vite cjs plugin to transform it to esm. But I think in dev mode, server inline the code twice so make the error happened

dai-shi:
> stylex doesn't support esm module

So, is it essentially the same as #110?

himself65:
> > stylex doesn't support esm module
> 
> So, is it essentially the same as #110?

CJS not supported is the vital issue. They suggest using `vite-plugin-commonjs`. However, doesn't work in waku dev mode.

himself65 (commented):
Ok, I investigated a while and I believe is plugin stylex bug with SSR

HorusGoul:
The plugin mostly works (https://github.com/HorusGoul/waku-stylex-demo), but I encountered an issue that prevented me from installing a Waku demo on the mono repo of the plugin to add tests to it.

I believe it has to do with having multiple react versions in the same mono repo, for some reason, Waku is trying to pick the one from the top-level node_modules, see this PR: https://github.com/HorusGoul/vite-plugin-stylex/pull/22

himself65:
> I believe it has to do with having multiple react versions in the same mono repo, for some reason, Waku is trying to pick the one from the top-level node_modules, see this PR: https://github.com/HorusGoul/vite-plugin-stylex/pull/22

/cc @dai-shi any idea? I think this is package manager behavior

dai-shi:
I'm not quite following. So, is it like this PR is working for @himself65 without dual package issue, but @HorusGoul has the dual package issue?

dai-shi:
> Ok, I investigated a while and I believe is plugin stylex bug with SSR

What does it mean? It feels like this example has a limitation, or no?

himself65:
> > Ok, I investigated a while and I believe is plugin stylex bug with SSR
> 
> What does it mean? It feels like this example has a limitation, or no?

Just we have to add `@stylex stylesheet;` in one CSS stylesheet to make things work. Not a limitation

himself65:
> I'm not quite following. So, is it like this PR is working for @himself65 without dual package issue, but @HorusGoul has the dual package issue?

No. dual package issue happens on his repo. Not waku

examples/12_css/vite.config.ts
```diff
@@ -1,6 +1,10 @@
 import { defineConfig } from 'vite';
 import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+import styleX from 'vite-plugin-stylex';
 
 export default defineConfig({
-  plugins: [vanillaExtractPlugin({ emitCssInSsr: true })],
+  ssr: {
+    external: ['@stylexjs/stylex'],
+  },
```
dai-shi, line 8: Does it work as expected?
Anyway, I would like to merge #289 first. Please be patient.

himself65: yes

examples/12_css/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@hono/node-server": "1.3.3",
+    "@stylexjs/stylex": "^0.3.0",
```
dai-shi, line 13: We remove `^` in dependencies in examples.

package.json
```diff
@@ -75,5 +75,10 @@
     "typescript": "^5.3.3",
     "wait-port": "^1.1.0",
     "waku": "workspace:*"
+  },
+  "pnpm": {
+    "overrides": {
+      "vite": "5.0.9"
+    }
```
dai-shi, line 82: Even with this?

examples/12_css/package.json
```diff
@@ -24,6 +25,7 @@
     "@vanilla-extract/vite-plugin": "3.9.3",
     "@vitejs/plugin-react": "4.2.1",
     "typescript": "5.3.3",
-    "vite": "5.0.9"
+    "vite": "5.0.9",
+    "vite-plugin-stylex": "^0.4.0"
```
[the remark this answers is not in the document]
himself65: this is my mistake

package.json
himself65: I think so, some upstream package use different version of vite

PR #291, opened 2023-12-18
fix: dual version packages

himself65:
Refs: https://github.com/pnpm/pnpm/issues/2713#issuecomment-1141000426


.pnpmfile.cjs
```diff
@@ -0,0 +1,49 @@
+// make sure every example has the same version of these packages
```
dai-shi, line 1:
```suggestion
// make sure every example has the same version of these packages
// Refs: https://github.com/pnpm/pnpm/issues/2713#issuecomment-1141000426
```

dai-shi (approved):
I'm not sure the need for this, but if you think it helps, it would be file.
Do think we will eventually remove this once we are sure that nothing is wrong?

PR #297, opened 2023-12-20
docs: update README.md

himself65:


README.md
```diff
@@ -297,13 +297,20 @@ folder in the repository.
...
 
-```sh
-vercel
+Add `.vercel.json` in the project root directory with the following:
+
+```json
+{
+  "buildCommand": "npm run build --with-vercel",
+  "devCommand": "npm run start"
+}
```
dai-shi, line 306: Nice. Does it work?
We can omit `--with-vercel` as it detects from an env var.
I'm not sure how `devCommand` is used.
Don't we need "output directory"?

himself65: I'm still figuring it out. I think the best solution is vercel add waku commands into their document

dai-shi: Yeah, we can send them a PR. But, I hope to have a meantime solution too.

dai-shi:
Let's reopen this if there's a progress.

PR #298, opened 2023-12-20
refactor(build, cli): generate vercel bundle automatically

himself65:
https://vercel.com/docs/projects/environment-variables/system-environment-variables#system-environment-variables

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -50,21 +50,21 @@ if (values.version) {
 } else if (values.help) {
   displayUsage();
 } else {
-  const withSsr = !!values['with-ssr'];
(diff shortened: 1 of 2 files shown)


packages/waku/src/cli.ts
```diff
@@ -58,7 +58,7 @@ if (values.version) {
     case 'build':
       runBuild({
         ssr: withSsr,
-        vercel: values['with-vercel'],
+        vercel: values['with-vercel'] ?? !!process.env.VERCEL,
```
dai-shi, line 61: We already have it in `.../builder/build.ts`.
Should we move it here? Yeah, that makes sense probably.

dai-shi (approved):
It's basically a refactor.

PR #289, opened 2023-12-18
breaking: no index html

dai-shi:
- [x] examples
- [ ] website
- [ ] fixtures

PR #300, opened 2023-12-22
breaking: server to allow fallback to static

dai-shi:
This is something I wanted for some time.
Now, we can skip rendering RSC and/or HTML on the server.
In addition to it, with router, we can let the client to skip requesting for some/all route components.

It's working as expected, but now I fee the API and the core implementation is full of hacks. It's not easy to understand and thus not very maintainable for the future.

We should revisit it after getting some feedbacks.

### DIFF
--- a/packages/website/vite.config.ts
+++ b/packages/website/vite.config.ts
@@ -4,6 +4,11 @@ export default defineConfig(({ mode }) => {
   process.env = { ...process.env, ...loadEnv(mode, process.cwd()) };
   if (mode === 'development') {
     return {
+      optimizeDeps: {
+        include: [
+          // '@uidotdev/usehooks',
+        ],
+      },
       ssr: {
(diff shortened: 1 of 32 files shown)


packages/website/vite.config.ts
```diff
@@ -4,6 +4,11 @@ export default defineConfig(({ mode }) => {
   process.env = { ...process.env, ...loadEnv(mode, process.cwd()) };
   if (mode === 'development') {
     return {
+      optimizeDeps: {
+        include: [
+          // '@uidotdev/usehooks',
```
dai-shi, line 9: I expected this would improve `npm run website:dev`, but it makes it worse (errors) if we remove `--with-ssr`.

dai-shi:
It was the issue of #289. We should solve it, but let's move on this pr.

PR #311, opened 2023-12-25
fix(website): possibly reduce hydration mismatch

dai-shi:


packages/website/src/main.tsx
```diff
@@ -12,17 +12,8 @@ const rootElement = (
...
-// FIXME temporary fix, doesn't feel ideal.
-function init() {
-  try {
-    if ((globalThis as any).__WAKU_SSR_ENABLED__) {
-      hydrateRoot(document.body, rootElement);
-    } else {
-      createRoot(document.body).render(rootElement);
-    }
-  } catch (e) {
-    console.log('retrying as an error is caught: ' + e);
```
dai-shi, line 24: We don't see this very often, so it's probably safe to remove this.

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -99,9 +99,9 @@ const injectRscPayload = (
         if (!(chunk instanceof Uint8Array)) {
           throw new Error('Unknown chunk type');
         }
+        controller.enqueue(chunk);
```
dai-shi, line 102: I expected this would help, but couldn't confirm it.

PR #313, opened 2023-12-26
fix: ensure sending scripts after html chunks

dai-shi:
Continuing from #312.

### DIFF
--- a/packages/waku/src/lib/renderers/html-renderer.ts
+++ b/packages/waku/src/lib/renderers/html-renderer.ts
@@ -92,23 +92,17 @@ const injectRscPayload = (
 ) => {
   const chunks: Uint8Array[] = [];
   let closed = false;
-  let notify: (() => void) | undefined;
   const copied = readable.pipeThrough(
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -92,23 +92,17 @@ const injectRscPayload = (
...
     new TransformStream({
       transform(chunk, controller) {
         if (!(chunk instanceof Uint8Array)) {
           throw new Error('Unknown chunk type');
         }
-        controller.enqueue(chunk);
         chunks.push(chunk);
-        // HACK delay a little bit to avoid hydration mismatch
-        setTimeout(() => {
-          notify?.();
```
dai-shi, line 106: I thought it's better to push as quickly as possible, but hydration process seems sensitive with the order, so it's safer to always send it after html. Not sure if it causes undesirable behavior in edge cases.

dai-shi:
_Hopefully_, this fixes the website issue.

PR #314, opened 2023-12-26
website: disable --with-vercel-static for now

dai-shi:
even #313 didn't work.

The problem is generated html that only happens Vercel production build. It doesn't happen with Vercel preview build.

[index.html.zip](https://github.com/dai-shi/waku/files/13767782/index.html.zip)

Assuming it's static only problem, disabling it for now.


dai-shi:
Oooops, this doesn't help because `/` is a static path. 😢

PR #317, opened 2023-12-26
feat: css hmr

Aslemammad:
In `handleHotUpdate` we re-inject/re-transform the module and send an update to the hmr listener (`import.meta.hot.on('module', ...)`). 

To avoid multiple hmr script tags, we add an `id` to the `TransformResult` (now `ModuleImportResult`) so we can identify any script tag with the same `waku-module-id` so we remove it from the DOM and replace it with a new one that has the new content.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
@@ -1,4 +1,5 @@
-import type { Plugin, TransformResult, ViteDevServer } from 'vite';
+import type { Plugin, ViteDevServer } from 'vite';
+import type { ModuleImportResult } from '../handlers/types.js';
 
 const customCode = `
 import { createHotContext as __vite__createHotContext } from "/@vite/client";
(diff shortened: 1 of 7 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -8,10 +9,17 @@ if (import.meta.hot && !globalThis.__WAKU_HMR_CONFIGURED__) {
   globalThis.__WAKU_HMR_CONFIGURED__ = true;
   import.meta.hot.on('hot-import', (data) => import(/* @vite-ignore */ data));
   import.meta.hot.on('module', (data) => {
+    // remove element with the same 'waku-module-id'
+    let script = document.querySelector(
+      'script[waku-module-id="' + data.id + '"]',
+    );
+    script?.remove()
```
dai-shi, line 16:
```suggestion
    script?.remove();
```

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -8,10 +9,17 @@ if (import.meta.hot && !globalThis.__WAKU_HMR_CONFIGURED__) {
...
+
     const code = data.code;
-    const script = document.createElement('script');
+    script = document.createElement('script');
     script.type = 'module';
     script.text = code;
+    script.setAttribute('waku-module-id', data.id)
```
dai-shi, line 22:
```suggestion
    script.setAttribute('waku-module-id', data.id);
```

dai-shi (approved):
wow, I don't fully understand how it fixes it, but the diff seems good.

PR #321, opened 2023-12-26
feat: environment variables

### LINKED ISSUE #292: environment variables improvement
treat `WAKU_PUBLIC_` prefix as public env vars that can only be used in client side rendering (or ssr?)

others can be used in RSC

dai-shi:
close #292 

- [x] core
- [x] website
- [x] `.env.local`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - breaking: no index html #289
 - fix(vercel): option for static build #310
 - feat(router): createPages #293
+- feat: environment variables #321
 
 ## [0.18.1] - 2023-12-15
 ### Changed
--- a/e2e/fixtures/ssr-basic/src/main.tsx
+++ b/e2e/fixtures/ssr-basic/src/main.tsx
@@ -10,7 +10,7 @@ const rootElement = (
   </StrictMode>
 );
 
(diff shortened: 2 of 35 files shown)


dai-shi:
I thought about using `import.meta.env` like Vite (and maybe use Vite feature), but it may not work well with server components. Thus, a global variable. We may revisit before v1.

Also, thought about using `dotenv`, but it's a naive one for now. We can consider depending on `dotenv` for more capabilities.

himself65 (commented):
Can we re-use exsitingreact api?

https://react.dev/reference/react/experimental_taintUniqueValue

dai-shi:
We can use it for safety when it's in `canary`.

himself65 (approved):

himself65:
I found a related vite issue: 

https://github.com/vitejs/vite/issues/9171

dai-shi:
I was wondering if we could avoid optimized deps changes from the ground.

dai-shi (commented):
Should be good.

himself65:
How you do that

dai-shi:
reviewed the breaking commit dbc3d29 and reverted some changes, and applied a new approach.

PR #324, opened 2023-12-27
chore(worker-api/impl): pass stream to parent

### LINKED ISSUE #308: refactor idea: worker-api/impl
I thought we needed to stream data through message port, but looking at
https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Transferable_objects
can we actually transfer `ReadableStream`. Does it work on Node too?

Aslemammad:
Resolves #308 

We implemented our messaging method, but streams are already transferrable between the worker and the main thread. The only issue is that the types are not supporting them although the implementation does support them, that's why I used `as unknown as TransferListItem`

### DIFF
--- a/packages/waku/src/lib/handlers/dev-worker-api.ts
+++ b/packages/waku/src/lib/handlers/dev-worker-api.ts
@@ -21,10 +21,10 @@ export type BuildOutput = {
 
 export type MessageReq =
   | ({
-    id: number;
-    type: 'render';
-    hasModuleIdCallback: boolean;
-  } & Omit<RenderRequest, 'stream' | 'moduleIdCallback'>)
+      id: number;
+      type: 'render';
+      hasModuleIdCallback: boolean;
+    } & Omit<RenderRequest, 'stream' | 'moduleIdCallback'>)
   | { id: number; type: 'pipe'; stream: ReadableStream };
 
 export type MessageRes =
@@ -122,7 +122,7 @@ export async function renderRscWithWorker<Context>(
   const id = nextId++;
   const pipe = async () => {
     if (rr.stream) {
-      worker.postMessage({ id, type: 'pipe', stream: rr.stream } as MessageReq)
+      worker.postMessage({ id, type: 'pipe', stream: rr.stream } as MessageReq);
     }
   };
   let started = false;
--- a/packages/waku/src/lib/handlers/dev-worker-api.ts
+++ b/packages/waku/src/lib/handlers/dev-worker-api.ts
@@ -1,6 +1,6 @@
 import type {
   TransferListItem,
-  Worker as WorkerOrig,
+  Worker as WorkerType,
 } from 'node:worker_threads';
 
 import type { ResolvedConfig } from '../config.js';
@@ -22,13 +22,11 @@ export type BuildOutput = {
   htmlFiles: string[];
 };
 
-export type MessageReq =
-  | ({
-      id: number;
-      type: 'render';
-      hasModuleIdCallback: boolean;
-    } & Omit<RenderRequest, 'stream' | 'moduleIdCallback'>)
-  | { id: number; type: 'pipe'; stream: ReadableStream };
+export type MessageReq = {
+  id: number;
+  type: 'render';
+  hasModuleIdCallback: boolean;
+} & Omit<RenderRequest, 'moduleIdCallback'>;
 
 export type MessageRes =
   | { type: 'full-reload' }
@@ -40,12 +38,12 @@ export type MessageRes =
 
 const messageCallbacks = new Map<number, (mesg: MessageRes) => void>();
 
-let lastWorker: Promise<WorkerOrig> | undefined;
+let lastWorker: Promise<WorkerType> | undefined;
 const getWorker = () => {
   if (lastWorker) {
     return lastWorker;
   }
-  return (lastWorker = new Promise<WorkerOrig>((resolve, reject) => {
+  return (lastWorker = new Promise<WorkerType>((resolve, reject) => {
     Promise.all([
       import('node:worker_threads').catch((e) => {
         throw e;
@@ -123,15 +121,6 @@ export async function renderRscWithWorker<Context>(
 ): Promise<readonly [ReadableStream, Context]> {
   const worker = await getWorker();
   const id = nextId++;
-  const pipe = async () => {
-    if (rr.stream) {
-      const mesg: MessageReq = { id, type: 'pipe', stream: rr.stream };
-      worker.postMessage(
-        mesg,
-        [rr.stream as unknown as TransferListItem],
-      );
-    }
-  };
   let started = false;
   return new Promise((resolve, reject) => {
     messageCallbacks.set(id, (mesg) => {
@@ -143,7 +132,6 @@ export async function renderRscWithWorker<Context>(
               messageCallbacks.delete(id);
             },
           });
-          mesg.stream.pipeThrough(bridge);
           resolve([mesg.stream.pipeThrough(bridge), mesg.context as Context]);
         } else {
           throw new Error('already started');
@@ -170,9 +158,9 @@ export async function renderRscWithWorker<Context>(
       id,
       type: 'render',
       hasModuleIdCallback: !!rr.moduleIdCallback,
+      ...(rr.stream ? { stream: rr.stream } : {}),
       ...copied,
     };
-    worker.postMessage(mesg);
-    pipe();
+    worker.postMessage(mesg, [rr.stream as unknown as TransferListItem]);
   });
 }
--- a/packages/waku/src/lib/handlers/dev-worker-api.ts
+++ b/packages/waku/src/lib/handlers/dev-worker-api.ts
@@ -21,13 +21,11 @@ export type BuildOutput = {
 
 export type MessageReq =
   | ({
-      id: number;
-      type: 'render';
-      hasModuleIdCallback: boolean;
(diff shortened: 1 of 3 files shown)


himself65:
Upstream issue: https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/67980

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -5,10 +5,11 @@ import { parentPort } from 'node:worker_threads';
 import { Server } from 'node:http';
 import { createServer as createViteServer } from 'vite';
 
+import type { TransferListItem } from 'node:worker_threads';
```
dai-shi, line 8:
```suggestion
import type { TransferListItem } from 'node:worker_threads';
import { createServer as createViteServer } from 'vite';

```

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -57,29 +58,13 @@ const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
       customImport: loadServerFile,
       entries: await loadEntries(rr.config),
     });
-    const mesg: MessageRes = { id, type: 'start', context: rr.context };
-    parentPort!.postMessage(mesg);
-    deepFreeze(rr.context);
```
dai-shi, line 62: We should keep this after L67.

dai-shi (commented):
Looks good so far!

Can we do the same thing for main -> worker too?

packages/waku/src/lib/handlers/dev-worker-api.ts
```diff
@@ -1,4 +1,7 @@
-import type { Worker as WorkerOrig } from 'node:worker_threads';
+import type {
+  TransferListItem,
+  Worker as WorkerOrig,
```
dai-shi, line 3: btw, I recently use `WorkerType` for the naming convention instead of `WorkerOrig`. What do you think?

packages/waku/src/lib/handlers/dev-worker-api.ts
```diff
@@ -126,67 +125,30 @@ export async function renderRscWithWorker<Context>(
...
-          }
-        } while (!result.done);
-      } catch (err) {
-        const mesg: MessageReq = { id, type: 'err', err };
-        worker.postMessage(mesg);
-      }
+      worker.postMessage(
+        { id, type: 'pipe', stream: rr.stream } as MessageReq,
+        [rr.stream as unknown as TransferListItem],
+      );
```
dai-shi, line 131: Can you follow this style and add type assertion only for the necessary part?
```suggestion
      const mesg: MessageReq = { id, type: 'pipe', stream: rr.stream };
      worker.postMessage(
        mesg,
        [rr.stream as unknown as TransferListItem],
      );
```

packages/waku/src/lib/handlers/dev-worker-api.ts
```diff
@@ -126,67 +125,30 @@ export async function renderRscWithWorker<Context>(
...
         if (!started) {
           started = true;
-          resolve([stream, mesg.context as Context]);
+          const bridge = new TransformStream({
+            flush() {
+              messageCallbacks.delete(id);
+            },
+          });
+          mesg.stream.pipeThrough(bridge);
+          resolve([bridge.readable, mesg.context as Context]);
```
dai-shi, line 146: I didn't know this `.readable` usage.
I would have done it like this, but either should be fine:
```suggestion
          resolve([mesg.stream.pipeThrough(bridge), mesg.context as Context]);
```

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -142,16 +128,21 @@ const loadEntries = async (config: ResolvedConfig) => {
...
-  } else if (mesg.type === 'end') {
-    const controller = controllerMap.get(mesg.id)!;
-    controller.close();
-  } else if (mesg.type === 'err') {
+  } else if (mesg.type === 'pipe') {
     const controller = controllerMap.get(mesg.id)!;
-    const err =
-      mesg.err instanceof Error ? mesg.err : new Error(String(mesg.err));
-    controller.error(err);
+    mesg.stream.pipeTo(
```
dai-shi, line 133: I don't think this is necessary.
So, we don't need mesg.type=pipe at all, and attach the `stream` for mesg.type=render.

dai-shi (commented):
Hi, this looks great. I found a way to make it much simpler.

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -142,16 +128,21 @@ const loadEntries = async (config: ResolvedConfig) => {
...
     handleRender(mesg);
-  } else if (mesg.type === 'buf') {
-    const controller = controllerMap.get(mesg.id)!;
-    controller.enqueue(new Uint8Array(mesg.buf, mesg.offset, mesg.len));
-  } else if (mesg.type === 'end') {
-    const controller = controllerMap.get(mesg.id)!;
-    controller.close();
-  } else if (mesg.type === 'err') {
+
     const controller = controllerMap.get(mesg.id)!;
```
dai-shi, line 132: I don't think we need `controllerMap`.

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -26,17 +27,14 @@ const HAS_MODULE_REGISTER = typeof module.register === 'function';
 if (HAS_MODULE_REGISTER) {
   module.register('waku/node-loader', pathToFileURL('./'));
 }
-const controllerMap = new Map<number, ReadableStreamDefaultController>();
 
-const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
+const handleRender = async (
+  mesg: MessageReq & { type: 'render' },
+  stream: ReadableStream,
```
dai-shi, line 33: I was thinking we can simply use `rr.stream`. Any issue with it?

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -141,17 +124,19 @@ const loadEntries = async (config: ResolvedConfig) => {
...
+          controller.enqueue(chunk);
+        } else if (chunk instanceof ArrayBuffer) {
+          controller.enqueue(new Uint8Array(chunk, 0, chunk.byteLength));
+        } else {
+          controller.error(new Error('Unexepected buffer type'));
+        }
+      },
+    });
+
+    mesg.stream?.pipeThrough(bridge);
```
dai-shi, line 139: This bridge is almost no-op, no?

dai-shi (commented):
Sorry, if my previous comment was unclear.

packages/waku/src/lib/handlers/dev-worker-impl.ts
Aslemammad: you mean just use what's coming from the message?

packages/waku/src/lib/handlers/dev-worker-impl.ts
Aslemammad: Could you clarify what a no-op is?

packages/waku/src/lib/handlers/dev-worker-impl.ts
dai-shi: Yes, can we do it?

packages/waku/src/lib/handlers/dev-worker-impl.ts
dai-shi: I mean it doesn't do anything. "no operation".

packages/waku/src/lib/handlers/dev-worker-impl.ts
Aslemammad: I'll try my best!

packages/waku/src/lib/handlers/dev-worker-impl.ts
Aslemammad: isn't it converting array buffer also to Uint8Array?

dai-shi: As we no longer use post message for stream, we shouldn't need any conversion.

Aslemammad:
Nice, is there any resource that shows why postMessage needs the
conversion! Just curious.

And sure, I'll handle this.

dai-shi:
I don't know if it's related with conversion. Never mind.

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -50,36 +44,21 @@ const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
       searchParams: new URLSearchParams(rr.searchParamsString),
       method: rr.method,
       context: rr.context,
-      body: rr.stream,
+      body: rr.stream!,
```
dai-shi, line 47: How can we assure that it's not undefined?

Aslemammad: To be honest, I thought about this, and I could not find a solution, it required so many types gymnastics which I then could not get it right. Let me know if you have an idea.

dai-shi: Ah, yeah, it's optional. 204a34e should do for now. We can probably accept `undefined` so that we don't need such syntax.

dai-shi (approved):
This looks great! Thanks for your work.

PR #325, opened 2023-12-28
fix: dual package `waku/client`

himself65:
I don't know why


himself65:
https://github.com/dai-shi/waku/pull/328

PR #330, opened 2023-12-28
feat(create-waku): bundle into single file

himself65:


packages/create-waku/tsconfig.json
```diff
@@ -1,12 +1,12 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "./src",
+    "rootDir": ".",
     "outDir": "./dist",
```
dai-shi, line 5: Is it a convention or a standard?
I'm not super comfortable with root being `.` and out being `./dist`.

himself65:
I have to rename the file into `index.ts`, cause the output entry will always be `index.js`.

packages/create-waku/src/index.ts
```diff
@@ -1,12 +1,11 @@
 #!/usr/bin/env node
-
 import { existsSync, readdirSync } from 'node:fs';
 import fsPromises from 'node:fs/promises';
-import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import path, { dirname } from 'node:path';
```
dai-shi, line 4: Not a big fan of mixing default and named. Can you go with all named?

packages/create-waku/src/index.ts
```diff
@@ -1,12 +1,11 @@
...
 import { default as prompts } from 'prompts';
 import { red, green, bold } from 'kolorist';
 import fse from 'fs-extra/esm';
+import { fileURLToPath } from 'node:url';
```
dai-shi, line 8: move up to the original position.

packages/create-waku/src/index.ts
```diff
@@ -31,9 +30,12 @@ function canSafelyOverwrite(dir: string) {
 async function init() {
   let targetDir = '';
   const defaultProjectName = 'waku-project';
-
-  const templateRoot = fileURLToPath(new URL('../template', import.meta.url));
-  const CHOICES = await fsPromises.readdir(templateRoot);
+  const __dirname = dirname(fileURLToPath(import.meta.url));
+  const templateRoot = path.resolve(__dirname, '../template');
```
dai-shi, line 34: Can we simplify like this?
```suggestion
  const templateRoot = path.join(fileURLToPath(import.meta.url), '../../template');
```

dai-shi (commented):
Looks good. Some refactor ideas.

dai-shi (approved):
Not familiar with https://github.com/vercel/ncc, but looks okay.

PR #331, opened 2023-12-28
ci: support npm provenance

himself65:
https://docs.npmjs.com/generating-provenance-statements

https://github.blog/2023-04-19-introducing-npm-package-provenance/


dai-shi (approved):
I'm not sure about some details, but it seems good.

PR #334, opened 2024-01-01
feat(create-waku): notify cli update

himself65:


packages/create-waku/src/index.ts
```diff
@@ -27,6 +28,19 @@ function canSafelyOverwrite(dir: string) {
   return !existsSync(dir) || readdirSync(dir).length === 0;
 }
 
+async function notifyUpdate() {
+  const nodeRequire = createRequire(import.meta.url);
+  const checkForUpdate = nodeRequire('update-check');
```
himself65, line 33: `update-check` is a CJS module, so

dai-shi:
```suggestion
  // `update-check` is a CSJ module.
  const checkForUpdate = nodeRequire('update-check');
```

dai-shi (approved):
LGTM

PR #335, opened 2024-01-01
fix: hydration mismatch on Windows

himself65:
Fixes https://github.com/dai-shi/waku/issues/337

### DIFF
--- a/e2e/utils.ts
+++ b/e2e/utils.ts
@@ -1,7 +1,13 @@
 import net from 'node:net';
-import { test as basicTest } from '@playwright/test';
+import { expect, test as basicTest } from '@playwright/test';
 import type { ConsoleMessage } from '@playwright/test';
 import type { ChildProcess } from 'node:child_process';
+
+const unexpectedErrors: RegExp[] = [
+  /^You did not run Node.js with the `--conditions react-server` flag/,
+  /^\(node:14372\)/,
+  /^Warning: Expected server HTML to contain a matching/,
+];
 
 export async function getFreePort(): Promise<number> {
   return new Promise<number>((res) => {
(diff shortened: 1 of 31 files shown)


himself65:
CI flaky might related to: https://github.com/actions/download-artifact/issues/249

e2e/utils.ts
```diff
@@ -1,8 +1,14 @@
 import net from 'node:net';
-import { test as basicTest } from '@playwright/test';
+import { expect, test as basicTest } from '@playwright/test';
 import type { ConsoleMessage } from '@playwright/test';
 import type { ChildProcess } from 'node:child_process';
 
+const unexpectedErrors: RegExp[] = [
+  /^You did not run Node.js with the `--conditions react-server` flag/,
+  /^\(node:14372\)/,
+  /^Warning: Expected server HTML to contain a matching/,
```
himself65, line 9: I noticed that there are some hydration mismatch in current CI

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -68,6 +68,9 @@ const getViteServer = async () => {
 };
 
 const loadServerFileDev = async (fileURL: string) => {
+  if (fileURL.endsWith('waku/dist/client.js')) {
+    return import(WAKU_CLIENT_MODULE_VALUE);
+  }
```
dai-shi, line 73: Is this only for our monorepo setup?

himself65: I think so, but cannot find why

himself65:
I guess this only happened in Monorepo. So, if it possible we change the way of writing examples?

Or this is still some corner case on Windows

himself65:
Ok, the real bug is `/` on prefix again

himself65:
image
image

packages/waku/src/lib/utils/path.ts
```diff
@@ -36,8 +36,17 @@ export const fileURLToFilePath = (fileURL: string) => {
 };
 
 // for filePath
```
dai-shi, line 38: joinPath is only for "filePath" style. Please do not modify for multiple path styles. We should adjust it in advance. This makes things more maintainable.

dai-shi (changes requested):

dai-shi (approved):
LGTM!

PR #339, opened 2024-01-01
refactor(waku): prettify cli using commander.js

himself65:
now the CLI has growing complex, I think we need a standard library to handle the thing

### DIFF
--- a/e2e/__snapshots__/waku-cli.spec.ts/waku---help
+++ b/e2e/__snapshots__/waku-cli.spec.ts/waku---help
@@ -1,0 +1,11 @@
+Usage: waku [options] [command]
+
+Options:
+  -V, --version    output the version number
+  -h, --help       display help for command
+
+Commands:
+  build [options]  Build the application for production
+  dev [options]    Start the development server
+  start [options]  Start the production server
+  help [command]   display help for command
--- a/e2e/__snapshots__/waku-cli.spec.ts/waku-build---help
+++ b/e2e/__snapshots__/waku-cli.spec.ts/waku-build---help
@@ -1,0 +1,11 @@
+Usage: waku build [options]
+
+Build the application for production
+
+Options:
+  --with-ssr            Use opt-in SSR
+  --with-vercel         Output for Vercel on build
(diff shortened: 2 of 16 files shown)


dai-shi:
> now the CLI has growing complex, I think we need a standard library to handle the thing

Yeah, I was thinking about it too. But, I have a few concerns:
- The diff doesn't seem to improve so much. Maybe it's too early at this point.
- I used commander.js before, and found it somewhat traditional. Is it still widely used?
- It must be an optional peer dependency and it requires to add a dependency in the templates. That's a huge motivation to avoid using libraries unless it's very helpful.

himself65:
> > now the CLI has growing complex, I think we need a standard library to handle the thing
> 
> Yeah, I was thinking about it too. But, I have a few concerns:
> 
> * The diff doesn't seem to improve so much. Maybe it's too early at this point.
> * I used commander.js before, and found it somewhat traditional. Is it still widely used?
> * It must be an optional peer dependency and it requires to add a dependency in the templates. That's a huge motivation to avoid using libraries unless it's very helpful.

There are some options:

arg: https://github.com/huozhi/bunchee/blob/main/src/bin/index.ts
commander.js: https://github.com/vercel/next.js/blob/canary/packages/create-next-app/index.ts
clipanion: https://github.com/napi-rs/napi-rs/blob/main/cli/src/cli.ts

himself65:
https://www.npmjs.com/package/meow

dai-shi:
Thanks.

I would like to hold this for the moment, and revisit it in the future. There might be more hidden requirements in cli.ts and we can't decide it now.

dai-shi:
> It must be an optional peer dependency and it requires to add a dependency in the templates. That's a huge motivation to avoid using libraries unless it's very helpful.

This limitation is gone with #355.

dai-shi:
I'm still interested in using commander.js or some alternatives (maybe commander.js is fine.)

Maybe some time later, maybe before v1-alpha. Let's start over anyway.

dai-shi:
#1319 should be a follow-up.

Issue #338, opened 2024-01-01
transform code run twice


himself65:
```diff
Index: examples/12_css/vite.config.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/12_css/vite.config.ts b/examples/12_css/vite.config.ts
--- a/examples/12_css/vite.config.ts	(revision ebd005ff88a3236ec02d21a466f0dd5f086aa507)
+++ b/examples/12_css/vite.config.ts	(date 1704119642367)
@@ -1,10 +1,23 @@
 import { defineConfig } from 'vite';
 import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
 import styleX from 'vite-plugin-stylex';
-
+let counter = 0;
 export default defineConfig({
   ssr: {
     external: ['@stylexjs/stylex'],
   },
-  plugins: [vanillaExtractPlugin({ emitCssInSsr: true }), styleX()],
+  plugins: [
+    vanillaExtractPlugin({ emitCssInSsr: true }),
+    styleX(),
+    {
+      name: 'test-plugin',
+      transform(code) {
+        const c = `${code};${counter}`
+        console.log('c', c)
+        return {
+          code: c
+        }
+      }
+    }
+  ],
 });

```

```shell
/Users/himself65/Library/Caches/fnm_multishells/70636_1703645957566/bin/pnpm run dev

> waku-example@0.1.0 dev /Users/himself65/Code/waku/examples/12_css
> waku dev --with-ssr

ready: Listening on http://localhost:3000/
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
c import { jsxDEV } from "react/jsx-dev-runtime";
import { lazy } from "react";
import { defineEntries } from "waku/server";
const App = lazy(() => import("./components/App.js"));
export default defineEntries(
  // renderEntries
  async (input) => {
    return {
      App: /* @__PURE__ */ jsxDEV(App, { name: input || "Waku" }, void 0, false, {
        fileName: "/Users/himself65/Code/waku/examples/12_css/src/entries.tsx",
        lineNumber: 10,
        columnNumber: 12
      }, this)
    };
  },
  // getBuildConfig
  async () => [{ pathname: "/", entries: [{ input: "" }] }],
  // getSsrConfig
  async (pathname) => {
[...]

PR #341, opened 2024-01-02
fix: vite plugin run twice

himself65:
Fixes: https://github.com/dai-shi/waku/issues/338

Upstream: https://github.com/vitejs/vite/issues/15482


packages/waku/src/lib/utils/merge-vite-config.ts
```diff
@@ -11,6 +11,7 @@ export async function mergeUserViteConfig(config: UserConfig) {
   const mergedViteConfig = mergeViteConfig(
     {
       ...resolvedViteConfig,
+      configFile: false,
```
dai-shi, line 14: So, config file is already resolved in `resolvedViteConfig`, right?
I mean we still support custom `vite.config.ts`. cc @Aslemammad

dai-shi (approved):
Thanks. Let's merge.

Issue #344, opened 2024-01-03
website: local development issue


dai-shi:
- [x] ~I notice a flash of unstyled content (FOUC) on save. (reported by @sandren)~ #414 mitigates it. 👉 #427
- [x] ~https://github.com/dai-shi/waku/pull/345/files#r1440484445~ #364 fixes it.
- [x] ~prop mismatch error with ssr https://github.com/dai-shi/waku/pull/364#issuecomment-1888505712~ #426
- [ ] initial page load is very slow (~10s)
- [ ] event handler sometimes fails with HMR https://github.com/dai-shi/waku/issues/344#issuecomment-1890323113

dai-shi:
@Aslemammad you might be interested in taking this.

Aslemammad:
Yep, I might have some ideas.

sandren:
Event handlers are sometimes failing after page reload or HMR until the dev server is restarted.

ojj1123:
> I notice a flash of unstyled content (FOUC) on save.

I would like to try to track this issue

dai-shi:
https://github.com/dai-shi/waku/issues/113#issuecomment-1902528306

@ojj1123 In the mean time, how about you taking a look at this? 
> initial page load is very slow (~10s)

ojj1123:
> @ojj1123 In the mean time, how about you taking a look at this?

Okay. This warning message was shown also to me.

PR #345, opened 2024-01-03
chore: avoid website issues

dai-shi:
and minor fixes.

### DIFF
--- a/packages/website/vite.config.ts
+++ b/packages/website/vite.config.ts
@@ -4,7 +4,8 @@ export default defineConfig(({ mode }) => {
   if (mode === 'development') {
     return {
       optimizeDeps: {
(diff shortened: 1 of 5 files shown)


packages/website/vite.config.ts
```diff
@@ -4,7 +4,8 @@ export default defineConfig(({ mode }) => {
   if (mode === 'development') {
     return {
       optimizeDeps: {
-        include: ['@uidotdev/usehooks'],
+        // FIXME this causes an error with `--with-ssr`
+        // include: ['@uidotdev/usehooks'],
```
dai-shi, line 8: I want to make website work in development in both with and without `--with-ssr`.

dai-shi:
This PR doesn't fix anything. Follow-up in #344.

Issue #346, opened 2024-01-04
create-waku: Error: Cannot find module 'update-check'


dai-shi:
### version

v0.19.0-alpha.0

### how to reproduce

```
npm create waku@next
```

### possible solution

`update-check` should be in "dependencies" instead of "devDependencies"

himself65:
I'm working on this

PR #347, opened 2024-01-04
fix: `update-check` not found

himself65:
Fixes: https://github.com/dai-shi/waku/issues/346

Just noticed that we are using `@vercel/ncc` which is a bundler, so we can mix cjs and esm


dai-shi (approved):
hm, okay, so ncc bundles `update-check`.

himself65:
> hm, okay, so ncc bundles `update-check`.

Yeah, ncc is using webpack actually

PR #352, opened 2024-01-05
fix(router): path handling createPages

dai-shi:
(we need unit tests...)


dai-shi:
`/` is very exceptional...

PR #354, opened 2024-01-05
refactor: move `__WAKU_PREFETCHED__` to top

himself65:


packages/waku/src/lib/renderers/utils.ts
```diff
@@ -32,6 +32,7 @@ export const hasStatusCode = (x: unknown): x is { statusCode: number } =>
   typeof (x as any)?.statusCode === 'number';
 
 export const codeToInject = `
+globalThis.__WAKU_PREFETCHED__ = {};
 globalThis.__waku_module_cache__ = new Map();
 globalThis.__webpack_chunk_load__ = (id) => import(id).then((m) => globalThis.__waku_module_cache__.set(id, m));
 globalThis.__webpack_require__ = (id) => globalThis.__waku_module_cache__.get(id);`;
```
dai-shi, line 38: I first thought this PR is a good idea, but on second thought, I would like to remove this `codeToInject` completely once we migrate to `react-server-dom-esm` from `react-server-dom-webpack`. So, I rather think we should avoid this change. What do you think?

himself65: I think we will have some initial variables. I'm thinking to move such logic into `transformIndexHtml` of vite plugin for now?

dai-shi: Are we not using transformIndexHtml now?
Anyway, I don't think we need to initialize always. The current lazy approach seems okay for now.

PR #356, opened 2024-01-05
waku: remove cjs build

dai-shi:
There was a reason to have CJS, but I think the need is already gone. So, let's try.


himself65 (commented):
if so, we don't need swc either

dai-shi:
why so?

himself65:
> why so?

https://github.com/dai-shi/waku/pull/144#pullrequestreview-1696248663

dai-shi:
Oh, you mean to use `tsc` instead of `swc`.
I'm afraid tsc is to slow for my development flow.

My desire was to use `swc` in `create-waku`. Did we miscommunicate there?

PR #360, opened 2024-01-07
fix(waku): link props

himself65:
@sandren


packages/waku/src/router/client.ts
```diff
@@ -11,7 +11,15 @@ import {
...
   Fragment,
 } from 'react';
-import type { ComponentProps, FunctionComponent, ReactNode } from 'react';
+import type {
+  ComponentProps,
+  FunctionComponent,
+  ReactNode,
+  AnchorHTMLAttributes,
+  ReactElement,
+  PropsWithChildren,
```
dai-shi, line 20: I slightly prefer `children: ReactNode` for explicitness.

packages/waku/src/router/client.ts
```diff
@@ -69,19 +77,23 @@ export function useLocation() {
   return value.loc;
 }
 
+export type LinkProps = PropsWithChildren<
+  {
+    href: string;
```
dai-shi, line 82: Should we change this to `to`? v0.19.0 would be a good timing for such a breaking change.

dai-shi (commented):
`packages/waku/src/router/client.ts` looks good to me.

himself65:
hey we need rerun the error CI

sandren (approved):
> @sandren

Thanks! 🙂

Yes, let's use `to` for the Waku Router API since it's more or less the React community standard (React Router, Remix, Gatsby, TanStack Router, etc).

PR #361, opened 2024-01-08
chore: switch `vite-plugin-stylex`  to `vite-plugin-stylex-dev`

nonzzz:
# Background

`vite-plugin-stylex-dev` has more features than `vite-plugin-stylex`, and then `vite-plugin-stylex-dev` can convert most of scense. Such as `path alias` and etc. And it can work well with vite’s internal css plugin. So i think using `vite-plugin-stylex-dev` is better.

### DIFF
--- a/examples/12_css/package.json
+++ b/examples/12_css/package.json
@@ -23,6 +23,6 @@
     "typescript": "5.3.3",
     "vite": "5.0.10",
     "vite-plugin-commonjs": "0.10.1",
-    "vite-plugin-stylex": "0.4.1"
+    "vite-plugin-stylex-dev": "^0.2.3"
   }
 }
--- a/examples/12_css/package.json
+++ b/examples/12_css/package.json
@@ -23,6 +23,6 @@
     "typescript": "5.3.3",
     "vite": "5.0.10",
     "vite-plugin-commonjs": "0.10.1",
-    "vite-plugin-stylex-dev": "^0.2.3"
+    "vite-plugin-stylex-dev": "0.2.3"
   }
 }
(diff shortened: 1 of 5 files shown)


dai-shi:
@himself65

himself65:
yeah, this guy has messaged me privately. I will check if it indeed better like say so. For now I have no particular preference for this PR since examples are just show cases, does not have better or worse

dai-shi:
are these packages developed by different authors?

nonzzz:
> are these packages developed by different authors?

yes,I'm one of them, They are not official plugins, but their development concepts are different, I think it's very troublesome to adapt each framework or library, So i made `vite-plugin-stylex-dev` to solve it.As the stylex team said, the official version won't be provided until the community is stable.

dai-shi:
There will be `vite-plugin-stylex-dev2` in the future.
I think unless `vite-plugin-stylex` is really broken, we don't need to switch, and wait for the official version. Though, it might be interesting, if the new one provides a meaningful capability for Waku's show case. I'll leave it to @himself65 .

himself65:
We have one reason to use this plugin; this plugin supports the latest stylex with the ESM module that won't break Vite

https://github.com/facebook/stylex/pull/328

himself65:
BTW, I want to have a way to include as many plugin into our CI as possible so we can see the ecosystem compatibility

dai-shi:
> a way to include as many plugin into our CI

You mean without explicitly adding them in examples? Would be great.

dai-shi:
As I tried, this behaves better. Previously, the style isn't applied on the initial load (dev-only). So, I think we should merge this.

@nonzzz Would you please resolve conflicts?

nonzzz:
@dai-shi I will resolve them :)

nonzzz:
Resolve done.

PR #364, opened 2024-01-11
fix(website): dev experience and use Link

dai-shi:
- [x] remove unused packages
- [x] fix optimizeDeps
- [x] use `<Link>`
- [x] fix `--with-ssr`

### DIFF
--- a/packages/waku/src/lib/handlers/handler-dev.ts
+++ b/packages/waku/src/lib/handlers/handler-dev.ts
@@ -49,6 +49,9 @@ export function createHandler<
       optimizeDeps: {
         include: ['react-server-dom-webpack/client', 'react-dom'],
         exclude: ['waku'],
+        entries: [
+          // `${config.srcDir}/${config.entriesJs}`.replace(/\.js$/, '.*'),
+        ],
       },
       plugins: [
         patchReactRefresh(viteReact()),
--- a/packages/waku/src/lib/handlers/handler-dev.ts
+++ b/packages/waku/src/lib/handlers/handler-dev.ts
@@ -49,9 +49,6 @@ export function createHandler<
       optimizeDeps: {
         include: ['react-server-dom-webpack/client', 'react-dom'],
         exclude: ['waku'],
-        entries: [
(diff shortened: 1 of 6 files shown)


packages/waku/src/lib/handlers/handler-dev.ts
```diff
@@ -49,6 +49,9 @@ export function createHandler<
       optimizeDeps: {
         include: ['react-server-dom-webpack/client', 'react-dom'],
         exclude: ['waku'],
+        entries: [
+          // `${config.srcDir}/${config.entriesJs}`.replace(/\.js$/, '.*'),
+        ],
```
dai-shi, line 54:
```suggestion
```

I will try this in a follow-up PR.

packages/website/src/components/menu.tsx
```diff
@@ -70,15 +69,16 @@ type MenuLinkProps = {
...
+  let Element: any = Link;
+  const props: any = {};
 
   if (link.href.startsWith('http')) {
     Element = 'a';
     props.target = '_blank';
     props.rel = 'noopener noreferrer';
+    props.href = link.href;
+  } else {
+    props.to = link.href;
```
himself65, line 81: `to` is a required parameter, and this will cause error if it's empty

https://github.com/dai-shi/waku/blob/59a2c606f8d2adf98475f25d382f65e5644cb1a0/packages/waku/src/router/client.ts#L95-L97

himself65 (changes requested):

packages/website/src/components/menu.tsx
himself65: No never mind, I missed your `Element = 'a';` on the top

himself65: But reduce `any` types would help for reading

dai-shi: Yeah, I agree. It's not very readable.

dai-shi:
Oh, no. I didn't notice this error...

image

PR #367, opened 2024-01-13
fix: reduce unnecessary ts config from examples

himself65:
It might be ok to remove this. Don't be too strict with the user.

Also remove examples from monorepo. E2e smoke check will build the project and check the type


dai-shi (approved):
This looks niceee!

PR #387, opened 2024-01-16
docs: add delpy button

himself65:


dai-shi:
Is this still work in progress?

himself65:
> Is this still work in progress?

yeah might be, I have one concern should this README.md and button included in the templates by `create-waku`

dai-shi:
i don't think it's important to add it. @sandren ?

sandren:
> i don't think it's important to add it. @sandren ?

I don't think it's necessary.

Issue #24, opened 2023-05-05
v1 Roadmap


dai-shi:
Here are tasks that I would like to finish before v1.

- [ ] Server bundling #21 
- [ ] RSC-only SSR (or opt-in SSR)
- [ ] New API as middleware #23
- [ ] Explore RSC-only forms (optional?)
- [ ] Explore running RSCs on client (optional?)

Aslemammad:
How about wakuwork actions? I was thinking about developing a plugin for vite a while ago, Since I think it's something similar to telefunc while developing this tool with Rom Brillout.

dai-shi:
Actions (what I call "React Server Functions") are already implemented. I wouldn't say it's complete though.

Aslemammad:
Nice!

EasonYou:
I found that router ssr doesn't work at present, is that any plane to do it?
(dig into the code, seems like some thing wrong with ``getComponent`` in ``getSsrConfig``, but I have no idea why)

tom-sherman:
Anecdotally there seems to be a fairly large gap between waku's server actions and Next.js' implementation - I suppose this is expected because server actions are completely experimental in React right now.

IMO it makes sense to mark them as such (as an opt-in config option) before releasing v1.

dai-shi:
I don't think our RSF (react server function) isn't that experimental, meaning it's the same level as react server component. I suppose Next's server action does more experimental things.

That said, I would like to revisit our approach anyway, which will require some feedbacks from collaborators/contributors/users.

In any case, unless we have an official doc or RFC from React, we wouldn't be able to finalize our API, which means v1 probably won't land until then.

tom-sherman:
The problem with shipping server actions now (or before they're stable in React) is that the semantics are not specified yet. For example Next.js server actions are queued, it's impossible to execute them concurrently - not the case in Waku. There is also the integration with transitions, not sure if this is implemented in Waku.

IMO need to wait for a React RFC.

dai-shi:
Yeah, if we would release v1 before the callserver RFC, we should hide it behind a flag. For now, just noting it's not finalized would be enough. But in that sense, everything isn't finalized.

childrentime:
@dai-shi How about SSG? And change docs website to static files.

dai-shi:
docs website is already static files.
we generate static files on build (SSG is a vague word, as people expect to generate HTML files.)

dai-shi:
[...]

PR #389, opened 2024-01-17
docs: updates

sandren:


README.md
```diff
@@ -90,7 +94,9 @@ To learn more about the modern React architecture, we recommend [Making Sense of
 
 The entry point for routing in Waku projects is `./src/entries.tsx`. Export the `createPages` function to create your layouts and pages programatically.
 
-Both `createLayout` and `createPage` accept a configuration object to specify the route path, React component, and render method (`'static'` for SSG or `'dynamic'` for SSR). Layout components must accept a `children` prop.
+Both `createLayout` and `createPage` accept a configuration object to specify the route path, React component, and render method. Waku currently supports two options: `'static'` for static prerendering (SSG) or `'dynamic'` for server-side rendering (SSR).
```
dai-shi, line 97: What's not describe but technically interesting is that it works without SSR.
In that case, `'static'` means static RSC and `'dynamic'` means dynamic RSC. Even with SSR, the subsequent updates are RSC only and static/dynamic matter.

It's a hard concept and may not fit in README though.

docs/minimal-api.mdx
```diff
@@ -0,0 +1,90 @@
+---
+slug: minimal-api
+title: Minimal API
```
dai-shi, line 3: @ojj1123 
> https://github.com/dai-shi/waku/issues/24#issuecomment-1894091550

PR #392, opened 2024-01-17
fix: `getEnv` for cross-platform

himself65:
main idea is use `conditional exports` to handle cross-platform cases

### DIFF
--- a/packages/waku/src/environment/index.worker.ts
+++ b/packages/waku/src/environment/index.worker.ts
@@ -1,0 +1,3 @@
+export function getEnv(key: string): string | undefined {
+  return (globalThis as any).__WAKU_PRIVATE_ENV__[key]
+}
--- a/packages/waku/src/environment/index.worker.ts
+++ b/packages/waku/src/environment/index.worker.ts
@@ -1,3 +1,6 @@
 export function getEnv(key: string): string | undefined {
-  return (globalThis as any).__WAKU_PRIVATE_ENV__[key]
+  if (!(globalThis as any).__WAKU_PRIVATE_ENV__) {
+    return undefined;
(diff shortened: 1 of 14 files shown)


packages/waku/src/environment/index.worker.ts
```diff
@@ -0,0 +1,6 @@
+export function getEnv(key: string): string | undefined {
+  if (!(globalThis as any).__WAKU_PRIVATE_ENV__) {
+    return undefined;
+  }
```
himself65, line 4: This shouldn't return `undefined` as expected, but waku is using `workerd` instead of `worker` now

https://github.com/dai-shi/waku/pull/394

packages/waku/src/lib/builder/output-cloudflare.ts
```diff
@@ -38,6 +36,12 @@ app.use('*', serveStatic({ root: './' }));
 export default {
   async fetch(request, env, ctx) {
     if (!serveWaku) {
+      const entries = await import('./${entriesFile}');
+      globalThis.__WAKU_PRIVATE_ENV__ = Object.assign(
+        Object.create(null),
+        env,
+      );
+      const { honoMiddleware } = await entries;
```
himself65, line 44:
```suggestion
      globalThis.__WAKU_PRIVATE_ENV__ = Object.assign(
        Object.create(null),
        env,
      );
      const entries = await import('./${entriesFile}');
      const { honoMiddleware } = await entries;
```

dai-shi:
Thanks for working on it, but this is not what I expect to fix the `getEnv` issue.
Maybe in the future, this approach might be useful for some features.

PR #397, opened 2024-01-19
feat(website): use npm builds

sandren:


packages/website/package.json
```diff
@@ -17,7 +17,7 @@
     "next-mdx-remote": "^4.4.1",
     "react": "18.3.0-canary-c5b937576-20231219",
     "react-dom": "18.3.0-canary-c5b937576-20231219",
-    "waku": "workspace:*"
+    "waku": "0.19.0"
```
himself65, line 20: I'm not sure how pnpm will deal with this in monorepo, you might set `npm:0.19.0` would be better?

```
By default, pnpm will link packages from the workspace if the available packages match the declared ranges. For instance, foo@1.0.0 is linked into bar if bar has "foo": "^1.0.0" in its dependencies and foo@1.0.0 is in the workspace.
```

https://pnpm.io/workspaces#workspace-protocol-workspace

packages/website/package.json
```diff
@@ -17,7 +17,7 @@
     "next-mdx-remote": "^4.4.1",
     "react": "18.3.0-canary-b30030471-20240117",
     "react-dom": "18.3.0-canary-b30030471-20240117",
-    "waku": "workspace:*"
+    "waku": "https://pkg.csb.dev/dai-shi/waku/commit/e5791fa5/waku"
```
dai-shi, line 20: for now, let's use csb version. once we release a new (alpha) version, we change change this to an old (stable) version.

PR #407, opened 2024-01-21
refactor: avoid @fs naming

dai-shi:
It looks like vite 5 made some changes.


dai-shi:
No, that was a misunderstanding. something is still depending on `@fs`.

PR #411, opened 2024-01-22
fix: hmr plugin with async component (with ssr, dev only)

dai-shi:


examples/09_cssmodules/src/components/Layout.tsx
```diff
@@ -0,0 +1,22 @@
...
+      {children}
+      <Suspense fallback="Pending...">
+        <ServerMessage />
+      </Suspense>
+    </div>
+  );
+};
+
+const ServerMessage = async () => {
+  await new Promise((resolve) => setTimeout(resolve, 2000));
```
dai-shi, line 18: We need this to reproduce the bug.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -33,7 +33,7 @@ export function rscHmrPlugin(): Plugin {
       return [
         {
           tag: 'script',
-          attrs: { type: 'module' },
+          attrs: { type: 'module', async: true },
```
dai-shi, line 36: By default it's "defer" and the script isn't evaluated during HTML streaming.
cc @Aslemammad

Aslemammad: niceeee, i love that solution.

PR #419, opened 2024-01-23
feat: output AWS Lambda handler

aheissenberger:
**experimental!**

- uses the Hono wrapper for AWS Lambda
- bundles all dependencies except node into one file

**suggested AWS setup:**
- copy /public/** to S3 bucket
- CloudFront
  - /assets/*, /images/* from S3 bucket
  - /* from API Gateway with Lambda Function
 - API Gateway / Lambda Function
  - serve all dynamic content
  - serve static content which is not coved by the url pattern forwarded to the S3 bucket

## Demo Repo

https://github.com/aheissenberger/waku-aws-lambda-demo

(needs to be linked to my fork to work)
 

**Please provide some feedback:**
1) the current implementation tries to keep the needed transformations of the output in [output-aws-lambda.ts](https://github.com/dai-shi/waku/compare/main...aheissenberger:feature/aws_lambda?expand=1#diff-c1ae7ed7cd73a1b05c7d9813cd3b85d933691c4396cc303fdacba67d401bd322) which requires the build process to repeate to include the external dependencies
2) when doing the second build step rollup is not able to find the `hono` package. This package needs to be added to the WAKU project to avoid the build to fail.


packages/waku/src/lib/builder/serve-aws-lambda.ts
```diff
@@ -0,0 +1,20 @@
...
+const publicDir = import.meta.env.WAKU_CONFIG_PUBLIC_DIR;
+const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
+
+const env = process.env as Record<string, string>;
+
+const app = new Hono();
+app.use('*', honoMiddleware({ loadEntries, ssr, env }));
+app.use('*', serveStatic({ root: `${distDir}/${publicDir}` }));
+
+export const handler = handle(app);
```
himself65, line 20: Ending newline?

aheissenberger: sorry - missed the option to run the tests - all styling errors are fixed

packages/waku/src/cli.ts
```diff
@@ -155,6 +159,7 @@ Options:
   --with-vercel         Output for Vercel on build
   --with-cloudflare     Output for Cloudflare on build
   --with-deno           Output for Deno on build
+  --with-aws-lambda     Output for AWS Lambda on build
```
dai-shi, line 162: Just wondering if `--with-lambda` might work or not. It's too unclear maybe.

dai-shi:
Hi, nice work!

> **experimental!**

Let's make it a draft PR for the time.

> CloudFront Function which intercepts all requests to Lambda Function and redirects to S3 bucket based on a list of static routes - optional to reduce the calls to the lambda function

I'm working on some (big) changes to make this possible. Stay tuned.

> 1. the current implementation tries to keep the needed transformations of the output in [output-aws-lambda.ts](https://github.com/dai-shi/waku/compare/main...aheissenberger:feature/aws_lambda?expand=1#diff-c1ae7ed7cd73a1b05c7d9813cd3b85d933691c4396cc303fdacba67d401bd322) which requires the build process to repeate to include the external dependencies

What does that mean? Ah, you are using vite build. Yeah, we want to avoid it if possible.

> 2. when doing the second build step rollup is not able to find the `hono` package. This package needs to be added to the WAKU project to avoid the build to fail.

I'll work on it for output-vercel later. Please wait for that.

aheissenberger:
@dai-shi any idea why switching to **Lambda response streaming** does not work?
https://hono.dev/getting-started/aws-lambda#lambda-response-streaming

I switched the handler to `streamHandle` and added `RESPONSE_STREAM` in `serverless.yml`
```yml
functions:
  ssr:
    handler: dist/serve.handler
    events:
      - httpApi: "*"
    url:
      invokeMode: RESPONSE_STREAM
```

dai-shi:
Unfortunately, I'm not very familiar with Lambda.

packages/waku/src/cli.ts
dai-shi: On second thought, I prefer `--with-aws-lambda`.

dai-shi:
> I'll work on it for output-vercel later. Please wait for that.

Still work in progress, but #442 will be ready eventually.

packages/waku/src/cli.ts
aheissenberger: I agree - this is more clear

dai-shi:
thanks. readme should be kept short. we may have `docs/*` if we need detailed explanation.

as for examples, we already have many examples and i'm trying to reduce them. maybe, we should create a new repository for examples, eventually.

dai-shi:
#442 is already merged and #443 is ready. I think it helps you to finish up this PR.

packages/waku/src/lib/builder/build.ts
```diff
@@ -203,6 +204,7 @@ const buildServerBundle = async (
           ...serverEntryFiles,
         },
         output: {
+          format: serve === 'aws-lambda' ? 'esm' : 'es',
```
aheissenberger, line 207: @dai-shi please check if 'es' is the correct default value for all other build targets

himself65: es and esm are same, and esm is not the valid option

dai-shi: What's the default format if we don't specify anything?

README.md
```diff
@@ -661,6 +661,25 @@ npm run build -- --with-deno
 DENO_DEPLOY_TOKEN=... deployctl deploy --project=... --prod dist/serve.js --exclude node_modules
 ```
 
+### AWS Lambda (experimental)
+
+you have to add the [`hono`](https://hono.dev/getting-started/aws-lambda) package to your project:
```
dai-shi, line 666: hono is already a dependency, so we don't need to install it, or no?

aheissenberger: you are right - I tested again and it works without - did remove the info.

packages/waku/src/lib/builder/build.ts
aheissenberger: sorry  for this :-( I removed it and I still was able to bundle a working app for aws lambda

packages/waku/src/lib/builder/build.ts
```diff
@@ -48,6 +48,7 @@ import { rscServePlugin } from '../plugins/vite-plugin-rsc-serve.js';
 import { rscEnvPlugin } from '../plugins/vite-plugin-rsc-env.js';
 import { emitVercelOutput } from './output-vercel.js';
 import { emitCloudflareOutput } from './output-cloudflare.js';
+import { emitAwsLambdaOutput } from './output-aws-lambda.js';
 import { emitNetlifyOutput } from './output-netlify.js';
```
dai-shi, line 52:
```suggestion
import { emitNetlifyOutput } from './output-netlify.js';
import { emitAwsLambdaOutput } from './output-aws-lambda.js';
```

packages/waku/src/lib/builder/serve-aws-lambda.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint import/no-unresolved: off */
+
+import { Hono } from 'hono';
+import { handle } from 'hono/aws-lambda';
+
+import { honoMiddleware } from '../middleware/hono-prd.js';
+import { serveStatic } from '@hono/node-server/serve-static';
```
dai-shi, line 7:
```suggestion
import { serveStatic } from '@hono/node-server/serve-static';

import { honoMiddleware } from '../middleware/hono-prd.js';
```

packages/waku/src/lib/builder/serve-aws-lambda.ts
```diff
@@ -0,0 +1,20 @@
+/* eslint import/no-unresolved: off */
```
dai-shi, line 1: Do we still need this?

aheissenberger: no - removed

README.md
```diff
@@ -668,6 +668,20 @@ npm run build -- --with-netlify
...
+### AWS Lambda (experimental)
+
+```
+npm run build -- --with-aws-lambda
+```
+
+The generated output format is compatible with [AWS Lambda](https://aws.amazon.com/lambda/).
+The output entrypoint `dist/serve.js` is compatible with [AWS Lambda format](https://docs.aws.amazon.com/lex/latest/dg/lambda-input-response-format.html).
+
+You can zip and upload the `dist` folder [manual to AWS console](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-package.html) or use one of the tools to deploy:
```
dai-shi, line 680: I don't try it, but just to confirm, it works as described, right?

dai-shi (approved):
The code looks good to me.
Is this ready to be megerd?

README.md
aheissenberger: yes - I do not manual upload, I use the [serverless framework](https://www.serverless.com) - you can find my demo here https://github.com/aheissenberger/waku-aws-lambda-demo

aheissenberger: Here is a LIVE Demo
https://d1pvniolnanw17.cloudfront.net
(will be removed soon)

README.md
```diff
@@ -668,6 +668,20 @@ npm run build -- --with-netlify
...
+
+- [serverless framework](https://www.serverless.com)
+- [AWS CDK](https://aws.amazon.com/de/cdk/)
+
```
dai-shi, line 684: I would like to keep README simple, like just a few lines as others.
It's like to demonstrate that it is possible.

If we need some detailed instructions, my suggestion is to add `./docs/*.mdx`, and add a link in README.

```suggestion
You can zip and upload the `dist` folder [manual to AWS console](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-package.html) with the entrypoint `dist/serve.js`.

```

aheissenberger: I understand but the I would suggest to only point to the [Hono AWS Lambda Deploy Docs] - have look on my latest version.

PR #420, opened 2024-01-23
test: update standalone test

himself65:
Upstream: https://github.com/dai-shi/waku/pull/374

Real standalone test:

1. isolated environment
2. install waku by file package instead of copy files

PR #425, opened 2024-01-24
test: improve GitHub action annotation

himself65:
emit such error:
 
image


himself65:
moved to https://github.com/dai-shi/waku/pull/420

PR #427, opened 2024-01-26
fix: FOUC

Aslemammad:
part of #344 

Since the style scripts are modules, they run after the HTML parse phase, which causes the styles to flash. 

What I do in this PR, is I get the styles and I inject them as normal style tags in the first load and then for HMR, later they'll be removed automatically.

### Why not script tags?
The module type of the script tags and them importing things is limiting and won't let me to use blocking script tags.
blocking:
```html
<script />
```
non-blocking:
```html
<script type="module" />
```

And since the client-transformed CSS (that supports HMR) in vite is a JS module, we cannot use normal script tags. What can be done is server-transforming CSS, so we get the transformed plain CSS and then inject it in the style tags.

I also remove the spinner.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
@@ -97,8 +97,12 @@ async function generateInitialScripts(
 ): Promise<HtmlTagDescriptor[]> {
   const sourceSet = modulePendingMap.get(viteServer);
 
+  if (!sourceSet) {
+    return []
+  }
+
   const scripts: HtmlTagDescriptor[] = [];
-  for (const result of sourceSet!) {
+  for (const result of sourceSet) {
     scripts.push({
       tag: 'style',
       attrs: { type: 'text/css', 'waku-module-id': result.id },
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
@@ -98,7 +98,7 @@ async function generateInitialScripts(
   const sourceSet = modulePendingMap.get(viteServer);
 
   if (!sourceSet) {
-    return []
+    return [];
   }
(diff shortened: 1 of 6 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -13,80 +19,44 @@ import.meta.hot = __vite__createHotContext(import.meta.url);
...
     script = document.createElement('script');
     script.type = 'module';
     script.text = code;
     script.setAttribute('waku-module-id', data.id);
     document.head.appendChild(script);
-    if (data.css) removeSpinner();
+    // avoid HMR flash by first applying the new and removing the old styles 
+    if (style) {
+      queueMicrotask(style.remove);
+    }
```
ojj1123, line 36: I was wondering why you queued `style.remove` function?

Aslemammad: So we do not have a FOUC in HMR, if we don't do this, styles will be removed, for a small bit of time, there are no styles until the new script tag applies the new styles. By queuing, we first apply the new script styles and then after that we remove the old styles.

dai-shi (approved):
The code looks good.

`npm run examples:dev:01_template` works well.
Do you notice `npm run website:dev` still has FOUC at first?

Aslemammad:
Thank you

> Do you notice npm run website:dev still has FOUC at first?

I don't on my system! Is it still there?

dai-shi:
@Aslemammad Yeah, as I tried, the `<style>` tag isn't present in the HTML.

Can anyone else try?

Aslemammad:
> @Aslemammad Yeah, as I tried, the `<style>` tag isn't present in the HTML.
> 
> Can anyone else try?

Have u tried linking waku or updating it in the website directory? since the version is fixed there! Let me know

dai-shi:
Ah, that's it. I forgot.

PR #431, opened 2024-01-28
fix(website): restore vendor prefixes and resolve hydration mismatch

sandren:


dai-shi (approved):
LGTM

@Aslemammad Would you like to review?

packages/website/src/components/code.tsx
```diff
@@ -1,11 +1,14 @@
...
+import { ClientOnly } from './client-only.js';
 import theme from '../theme.json';
 
 type CodeProps = {
   code: string;
 };
 
 export const Code = ({ code, ...rest }: CodeProps) => (
-  <BrightCode lang="tsx" theme={theme} code={code.trim()} {...rest} />
+  <ClientOnly>
```
Aslemammad, line 11: Nice, I didn't think about that since I thought the code SSR is important! But this makes sense, the code does not play a huge role. 

Ideally, we'd have a PR in the bright repo itself. https://github.com/code-hike/bright/issues/39

Aslemammad (approved):

packages/website/src/components/code.tsx
dai-shi: Okay, let's hope it will be resolved upstream. Merging this for now.

PR #432, opened 2024-01-28
test: use same build artifacts for test cases

himself65:
https://github.com/dai-shi/waku/actions/runs/7655254723/job/20860921312#step:10:689

```shell
[tsconfig-paths] An error occurred while parsing "D:\a\waku\waku\e2e\.cache\35dcc6d8-f1a9-4278-9694-5803ef866815\create-waku\tsconfig.json".
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,32 @@ jobs:
           path: packages/create-waku/*.tgz
           if-no-files-found: error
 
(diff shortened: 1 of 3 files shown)


himself65 (commented):
```
"template": "rsync --prune-empty-dirs --exclude node_modules --exclude dist -r ../../examples/* template/",
```

This is better but not supported in CI?

PR #440, opened 2024-01-29
build: remove postinstall script & fix `ENOENT` error

himself65:
I saw tons of ENOENT in many CIs and my local device, and we should let the maintainer or CI manually link it.

Before:
image

After:
image

Refs: https://github.com/pnpm/pnpm/issues/1801, https://www.webpro.nl/scraps/compiled-bin-in-typescript-monorepo


packages/create-waku/cli.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+// eslint-disable-next-line import/no-unresolved
+import './dist/index.js';
```
himself65, line 3: placeholder cli file to make pnpm happy

dai-shi (approved):
Having `./cli.js` seems a little less ideal, but it looks good generally.

dai-shi:
@himself65 Thanks for your contribution as always!

PR #446, opened 2024-02-04
fix: css modules 

### LINKED ISSUE #439: CSS modules broken in DEV after #427
`npm run examples:dev:09_cssmodules`

### Before #427 

image

### After #427 

image

Aslemammad:
Resolves #439 

The #427 PR treated all the CSS files the same way. CSS modules in vite act like real scripts and there's no way to ssr transform them and get the actual CSS, instead, we get the CSS keys that we should use in our general files.

What I do in this PR is revert the script injecting way of loading CSS, only for CSS modules, and for fixing FOUC, I make that script blocking, which is not ideal at all, but works for now. 

I see right now also that hmr with CSS modules in server components is broken since I guess hmr does not work well in server components.


dai-shi (approved):
Works great.

Not a big fan of all those hacks, but no idea from me. Definitely worth revisiting.

PR #449, opened 2024-02-05
fix: client components default exports

### LINKED ISSUE #434: Error: Element type is invalid. Received a promise that resolves to: undefined. Lazy element type must resolve to a class or function.
hi there 
i was just exploring waku and i got  an error when i tried to import client component into server component here is my code 

```jsx
"use client";
function SubmitButton({ addTodo }: { addTodo: () => void }) {
  return (
    <div>
      <button onClick={() => addTodo()} type="button">
        add
      </button>
    </div>
  );
}

export default SubmitButton;

```

```js 

"use server";
import { type RenderContext } from "waku/server";
export const todos: Array<{ name: string; completed: boolean }> = [
  { name: "initial todo", completed: false },
];

addTodo const addTodo = (ctx: RenderContext) => {
  todos.push({name : 'new todo", completed:false})
  ctx.rerender("waku");
};
```

```jsx
import { addTodo, todos } from "./actions.js";
import SubmitButton from "./submitButton.js";

export const Todos = () => {
  return (
      <div>
        {todos.map(({ name, completed }) => {
          return (
            <div>
              {name} <input type="checkbox" defaultChecked={completed} />
            </div>
          );
        })}
      </div>
      <input type="text" name="name" />
      <SubmitButton addTodo={addTodo} />
    </section>
  );
};

```
Upon running this code, I receive the error: "Element type is invalid. Received a promise that resolves to: undefined. Lazy element type must resolve to a class or function.

Aslemammad:
Resolves #434 

TIL: 
```ts
export * from 'url'
```
does not re-export the `default` export.

### DIFF
--- a/packages/waku/src/lib/handlers/handler-dev.ts
+++ b/packages/waku/src/lib/handlers/handler-dev.ts
@@ -228,9 +228,15 @@ export function createHandler<
         item.url !== viteUrl &&
         !item.url.includes('?html-proxy')
       ) {
+        const { code } = (await vite.transformRequest(item.url))!;
         res.setHeader('Content-Type', 'application/javascript');
         res.setStatus(200);
-        endStream(res.stream, `export * from "${item.url}";`);
+        let exports = `export * from "${item.url}";`;
+        // `export *` does not re-export `default`
+        if (code.includes('export default')) {
+          exports += `export { default } from "${item.url}";`;
+        }
+        endStream(res.stream, exports);
         return;
       }
     }
--- a/examples/05_actions/src/components/App.tsx
+++ b/examples/05_actions/src/components/App.tsx
@@ -1,6 +1,6 @@
 import { Balancer } from 'react-wrap-balancer';
 
-import { Counter } from './Counter.js';
+import Counter from './Counter.js';
 import { greet, getCounter, increment } from './funcs.js';
 
 type ServerFunction<T> = T extends (...args: infer A) => infer R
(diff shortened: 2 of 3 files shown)


packages/waku/src/lib/handlers/handler-dev.ts
```diff
@@ -228,9 +228,15 @@ export function createHandler<
         item.url !== viteUrl &&
         !item.url.includes('?html-proxy')
       ) {
+        const { code } = (await vite.transformRequest(item.url))!;
         res.setHeader('Content-Type', 'application/javascript');
         res.setStatus(200);
-        endStream(res.stream, `export * from "${item.url}";`);
```
dai-shi, line 233: It might not follow ESM, but does this cause an error?
```ts
endStream(res.stream, `export * from "${item.url}"; export { default } from "${item.url}";`);
```

Aslemammad: Yep, I haven't tried it, but the reason I guess is that the default export does not exist in many cases.

Aslemammad: And exporting what is not exported would cause an error I assume.

dai-shi: yeah, that's what I've concerned. If you tried it, it's good.

packages/waku/src/lib/handlers/handler-dev.ts
```diff
@@ -228,9 +228,15 @@ export function createHandler<
         item.url !== viteUrl &&
         !item.url.includes('?html-proxy')
       ) {
+        const { code } = (await vite.transformRequest(item.url))!;
```
dai-shi, line 231: I personally didn't like my hack from L218, and this makes me concerned more.
Let's look for a better solution (I'll open a new issue), and merge this as a temporary workaround for now.

dai-shi (approved):

PR #458, opened 2024-02-06
fix: output AWS Lambda / AWS HTTP API V2 breaks RSC POST requests url needed for server actions

aheissenberger:
This fix is hack to revert the decoded url in the rawPath property provided by the AWS API Gateway with a HTTP API V2 Event.

The hack will only apply to POST request which start with `/RSC/`.

There is no way to fix this in the Hono layer as an arbitrary decoded Path cannot be encoded back to the same original!

see https://github.com/honojs/hono/issues/2156

Without this fix, the `url: new URL(c.req.url),` will ignore everything after the `#` on the decoded url.

**default url (e.g. node environment):**
`http://localhost/RSC/%40id%2Fassets%2Frsf0.js%23save.txt`

**url (AWS API Gateway):**
`http://localhost/RSC/@id/assets/rsf0.js#save.txt`

`new URL(http://localhost/RSC/@id/assets/rsf0.js#save.txt').pathname` => `/RSC/@id/assets/rsf0.js` 
**missing:** `#save.txt`

https://github.com/dai-shi/waku/blob/32d52242c1450b5f5965860e671ff73c42da8bd0/packages/waku/src/lib/middleware/hono-utils.ts#L49-L63

### DIFF
--- a/packages/waku/src/lib/builder/serve-aws-lambda.ts
+++ b/packages/waku/src/lib/builder/serve-aws-lambda.ts
@@ -15,7 +15,7 @@ app.use('*', serveStatic({ root: `${distDir}/${publicDir}` }));
 app.use('*', honoMiddleware({ loadEntries, ssr, env }));
 
 /*
- FIX: AWS HTTP API V2 breaks RSC POST requests needed for server actions
+ FIX: AWS HTTP API V2 breaks RSC POST requests url needed for server actions
     - AWS HTTP API V2 sets rawPath to a decoded path
     - We need to encode the string after /RSC/ and replace the orginal rawPath
     - Hono bug: https://github.com/honojs/hono/issues/2156
--- a/packages/waku/src/lib/builder/serve-aws-lambda.ts
+++ b/packages/waku/src/lib/builder/serve-aws-lambda.ts
@@ -1,8 +1,7 @@
 import { Hono } from 'hono';
-import { handle } from 'hono/aws-lambda';
-import { serveStatic } from '@hono/node-server/serve-static';
-
 import { honoMiddleware } from '../middleware/hono-prd.js';
+import { handle, type LambdaContext, type LambdaEvent } from 'hono/aws-lambda';
+import { serveStatic } from '@hono/node-server/serve-static';
 
 const ssr = !!import.meta.env.WAKU_BUILD_SSR;
 const distDir = import.meta.env.WAKU_CONFIG_DIST_DIR;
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/builder/serve-aws-lambda.ts
```diff
@@ -15,4 +14,32 @@ const app = new Hono();
...
+    - Hono bug: https://github.com/honojs/hono/issues/2156
+*/
+
+const fixHTTPAPIV2 = (event: LambdaEvent) => {
+  if ('version' in event && event?.version === '2.0') {
+    if (
+      'http' in event.requestContext &&
+      event?.requestContext?.http?.method === 'POST' &&
+      'rawPath' in event &&
+      event?.rawPath?.startsWith('/RSC/')
```
dai-shi, line 30: `/RSC/` is configurable (theoretically, we've never tried.)
Can you make it `config.basePath + config.rscPath`? We need to get them through `import.meta.env.WAKU_CONFIG_*`.

dai-shi (changes requested):
Hmm, it seems like the hono issue will be resolved before we make a new release.
Can we wait for the upstream fix?

dai-shi:
> Is adding the parameters as url encoded path segment after `/RSC/` the official protocol for react server actions?

It's Waku's protocol.

dai-shi:
For server components, the current approach should be mostly fine. But, I'm not 100% happy with protocol for server actions. We can revisit it.

aheissenberger:
> Hmm, it seems like the hono issue will be resolved before we make a new release. Can we wait for the upstream fix?

There is no way to resolve this issue - only the team at AWS API Gateway can fix this and I know no way to get this reported.

packages/waku/src/lib/builder/serve-aws-lambda.ts
aheissenberger: I do not need to change the prefix path `/RSC/` - I would need an other transportation - e.g. by custom header to fix this problem. The problem ist the url encoded data after `/RSC/` - this is decoded by the API Gateway and Waku expects it to be encoded as it is the standard for `c.req.url` from the Hono API.

dai-shi: In that case, you can remove L30 condition and encode the entire path.

dai-shi:
If you can wait, I would like to fix it by avoiding URL encode in Waku core, because I wasn't very happy with it. Please give me some time.

dai-shi:
@aheissenberger Can you confirm if #463 fixes it?

aheissenberger:
> @aheissenberger Can you confirm if #463 fixes it?

yes - changing the id in the url fixed the problem :-)

PR #462, opened 2024-02-06
feat: server context & middleware

aheissenberger:
This is a proof of concept:
* define middleware in the `waku.config.ts`
* no need to change the build processes

ToDo:
* [ ] avoid to remove the middleware from the config `handler-dev.ts:56` - maybe using a different config file
  https://github.com/aheissenberger/waku/blob/fe2584c7766ac15e0a6e3b87ce3d0c780bd76c45/packages/waku/src/lib/handlers/handler-dev.ts#L53-L59  
* [X] access Server Context in any Server Component or Server Action- e.g. `useServerContext()`
* [ ] support typescript for middleware `examples/14_middleware/src/middleware/index.js` - maybe different loading of `waku.config.ts` in `loadConfig()`
* [ ] support an array of middleware functions 
* [ ] allow posthook without prehook
* [ ] rename hooks to `onBeforeRequest` and `onBeforeResponse`
* [ ] hot reload on code changes

### DIFF
--- a/examples/14_middleware/src/entries.tsx
+++ b/examples/14_middleware/src/entries.tsx
@@ -1,9 +1,8 @@
-import path from 'node:path';
-import { fileURLToPath } from 'node:url';
-import fsPromises from 'node:fs/promises';
 import { lazy } from 'react';
 import { defineEntries } from 'waku/server';
 import { Slot } from 'waku/client';
+//import { setServerContext } from './server-context.js';
+import { ServerContextWrapper } from './components/ServerContextWrapper.js';
 
 const App = lazy(() => import('./components/App.js'));
 
@@ -13,7 +12,8 @@ export default defineEntries(
     const ctx = this.context as { count: number };
     ++ctx.count;
     return {
-      App: <App name={input || 'Waku'} count={ctx.count}  />,
+      App: <ServerContextWrapper context={ctx}><App name={input || 'Waku'} count={ctx.count} /></ServerContextWrapper>
+
     };
   },
   // getBuildConfig
--- a/examples/14_middleware/src/entries.tsx
+++ b/examples/14_middleware/src/entries.tsx
@@ -1,0 +1,35 @@
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import fsPromises from 'node:fs/promises';
+import { lazy } from 'react';
+import { defineEntries } from 'waku/server';
+import { Slot } from 'waku/client';
+
+const App = lazy(() => import('./components/App.js'));
+
+export default defineEntries(
+  // renderEntries
+  async function (input) {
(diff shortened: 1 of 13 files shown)


dai-shi:
As answered in #461, it's in the plan for v0.20.0. We had several issues in v0.19 and it's taking time. Sorry for the delay and please be patient.

You can try anything, but my plan is a little bit different, so this PR won't be merged.
As a hint, I want to keep waku config serializable (so, no functions) for now.

examples/14_middleware/src/entries.tsx
```diff
@@ -13,7 +12,8 @@ export default defineEntries(
     const ctx = this.context as { count: number };
     ++ctx.count;
     return {
-      App: <App name={input || 'Waku'} count={ctx.count}  />,
+      App: <ServerContextWrapper context={ctx}><App name={input || 'Waku'} count={ctx.count} /></ServerContextWrapper>
```
dai-shi, line 15: Yay, nice try!

examples/14_middleware/src/use-server-provider.tsx
```diff
@@ -0,0 +1,20 @@
+import 'server-only';
+import { cache } from 'react';
```
dai-shi, line 2: Yes.

examples/14_middleware/src/use-server-provider.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const serverContext = cache(() => new Map());
+
+export const useServerProvider = <T,>(
```
dai-shi, line 6: I wouldn't name it `use*` because it's not a hook.

dai-shi (commented):
Not sure if you expect any feedback at the moment, but just had a look.

examples/14_middleware/src/entries.tsx
aheissenberger: I was not able to set the React.cache in entries.tsx - only by using the wrapper the context worked.

aheissenberger:
> You can try anything, but my plan is a little bit different, so this PR won't be merged. As a hint, I want to keep waku config serializable (so, no functions) for now.

1. Starting with the config was the easiest way but I will change this to ist own file which should fix the current problems with not supporting typescript and hot module replacement on changes.
2. please give me feedback on my other open tasks

examples/14_middleware/src/entries.tsx
dai-shi: Yeah, I will set it higher in the tree when I implement it.

dai-shi:
1: Just note that I will almost re-architect something completely around core in v0.20.
2: Which are they? Let me know if I seem to miss something.

dai-shi:
I'll work on React.cache based getContext for the next minor release.

Issue #450, opened 2024-02-05
Refactor to avoid dual module hacks


dai-shi:
follow up https://github.com/dai-shi/waku/pull/449#discussion_r1477724444

We should avoid this hack completely:
https://github.com/dai-shi/waku/blob/274dc34ef2516908542672213e19c78365a5e5ed/packages/waku/src/lib/handlers/handler-dev.ts#L218-L236

AFAIR, the reason we need it is because of our hack:
https://github.com/dai-shi/waku/blob/274dc34ef2516908542672213e19c78365a5e5ed/packages/waku/src/lib/renderers/rsc-renderer.ts#L26-L27

We can grab the Vite's module graph here (or I'd just pass a DEV-only resolving function).

dai-shi:
A trial failed: https://github.com/dai-shi/waku/pull/464#issuecomment-1936976554

I think it requires a decent knowledge with Vite's module graph and multiple Vite instance behavior.

PR #464, opened 2024-02-07
refactor resolveClientEntry

dai-shi:
- [x] split resolveClientEntry for DEV and PRD
- [ ] ~use module graph~


dai-shi:
I was hoping to fix #450 with 5fb24e1, but it didn't work.

Merging this as a refactor and let's tackle #450 separately.

PR #471, opened 2024-02-10
update dependencies

dai-shi:


packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -8,7 +8,7 @@ const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
 let serveWaku: ReturnType<typeof honoMiddleware> | undefined;
 
 const app = new Hono();
-app.use('*', serveStatic({ root: './' }));
+app.use('*', serveStatic({ root: './' } as any)); // FIXME we manifest should be optional
```
dai-shi, line 11: @yusukebe Hi, `manifest` can be optional, right? (Seems so according to the source code.)

e2e/07_router_standalone.spec.ts
```diff
@@ -97,4 +102,6 @@ test.describe('07_router standalone', () => {
     await testRouterExample(page, port);
     cp.kill();
   });
+
+  // FIXME shouldn't we also test `--with-ssr`?
```
dai-shi, line 106: @himself65 How does that sound?
Related: https://github.com/dai-shi/waku/pull/420

dai-shi:
oops, reverting doesn't fix...

packages/waku/src/lib/builder/serve-cloudflare.ts
dai-shi: resolved.

PR #479, opened 2024-02-12
feat: add screenshot matching to smoke test "12_css"

aheissenberger:
* add the option to add a snapshot of the screen to a test
* activate screenshots for "12_css"
* add the reference screenshots for "12_css"


dai-shi:
@himself65 Can you review this?

aheissenberger:
I have no idea on how to add the missing reference screenshots for the other platforms :-(

himself65:
> I have no idea on how to add the missing reference screenshots for the other platforms :-(

Maybe not possible or we will have to write a complex CI

himself65 (changes requested):
there is a tricky thing that if we skip `windows` and `linux`, the future contributors who don't have a Mac will hard to upgrade the snapshot.

1. custom visual test in GitHub Action, to support update snapshots with all platforms.
2. using [storybook](https://storybook.js.org/docs/writing-tests/visual-testing) + [chromatic](https://www.chromatic.com/). But since they don't support testing in the framework level, so also need to make some hack.

himself65:
For 2, I would say we separately add new fixtures and use storybook + chromatic would that be better?

dai-shi:
Yeah, in any case, this should be a new fixture, not a smoke test on an example.
Please open a new PR.

PR #481, opened 2024-02-12
feat: Add Example / SSR Target selection based deployment target

### LINKED ISSUE #478: ReferenceError: document is not defined / SSR Production Build / wrong target
WAKU Version Commit: 7a8609c190e68cf70563a07d9d4bfdff65666101 (Main / latest)

**Error:**
```
ReferenceError: document is not defined
    at file:///Users/ah/SVN-Checkouts/TST/waku-mantine/dist/ssr/assets/rsc271-59485b3e0.js:250:12
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:323:24)
```

The bundled `rsc271-59485b3e0.js` running on the server contains the code from the library only working in a browser.

* `pnpm waku dev` 👍 
* `pnpm waku dev --with-ssr` 👍 
* `pnpm waku build --with-ssr` 👍 
* `pnpm waku start --with-ssr` 👎  - ** ReferenceError: document is not defined**

This problem is created by `target: 'webworker',` in `packages/waku/src/lib/builder/build.ts` in buildVite for Server and SSR.

**FIX:** Switching all three in `build.ts` to `target: 'node20',` which is my local runtime and in case of a `--with-aws-lambda` the correct environment for the lambda! For other deployment targets e.g. Vercel this might be 'webworker'.

I suggest to set this based on the deployment target and when no target is set by detecting the current node version.

This is the same Example as used in #448

**Example:**
1. add [`react-textarea-autosize`](https://www.npmjs.com/package/react-textarea-autosize?activeTab=code) to `package.json` 
2. add this component to a page
  ```tsx
  'use client';
  import TextareaAutosize from 'react-textarea-autosize';

  export const MyTextarea = () => {
      return (<div><TextareaAutosize/></div>);
  }
  ```
3. run `pnpm waku build --with-ssr`

**Error:** `ReferenceError: document is not defined`

The reason is that Waku bundles the wrong module [`node_modules/react-textarea-autosize/dist/react-textarea-autosize.browser.esm.js`](https://unpkg.com/browse/react-textarea-autosize@8.5.3/dist/react-textarea-autosize.browser.esm.js) into `rsc2-fcfef706c.js` which is optimized for the browser and fails in nodejs when accessing the global `document`.

The correct file would be [`node_modules/react-textarea-autosize/dist/react-textarea-autosize.esm.js`](https://unpkg.com/browse/react-textarea-autosize@8.5.3/dist/react-textarea-autosize.esm.js) based on the `"exports"` section if the [`package.json`](https://unpkg.com/browse/react-textarea-autosize@8.5.3/package.json).

aheissenberger:
An Example to detect broken selection of the server side modules from external packages.
Should help to fix #478


dai-shi (changes requested):
Thanks for working on this.
I would like to avoid adding examples for now. Some of examples are showcases and others are for development. I want to reduce the number of them.

Can you turn this into `e2e/fixtures/lib-compat` or something and add e2e spec?
fixtures should be very minimal.

aheissenberger:
> Thanks for working on this. I would like to avoid adding examples for now. Some of examples are showcases and others are for development. I want to reduce the number of them.
> 
> Can you turn this into `e2e/fixtures/lib-compat` or something and add e2e spec? fixtures should be very minimal.

I will convert it in a fixture and replace the screenshot with specific test checks

aheissenberger:
replaced by #484

PR #484, opened 2024-02-13
feat: Test bundling packages which provide environment depended code parts

aheissenberger:
As requested #481 this is a simpler Version which does not use screen shots to test results.

Choosing the right code based on the `export` section of the `react-textarea-autosize` module based on the backend server deployment target ('node' | 'webworker') and the client browser. No browser only code should be bundled with server only code.

You can simple test this fixture by setting `isNodeCompatible` to false

### DIFF
--- a/e2e/fixtures/ssr-target-bundle/README.md
+++ b/e2e/fixtures/ssr-target-bundle/README.md
@@ -1,0 +1,3 @@
+# SSR Bundling for NPM packages with environment depended code parts
+
+Choosing the right code based on the `export` section of the `react-textarea-autosize` module based on the backend server deployment target ('node' | 'webworker') and the client browser. No browser only code should be bundled with server only code.
--- a/e2e/fixtures/ssr-target-bundle/package.json
+++ b/e2e/fixtures/ssr-target-bundle/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "ssr-basic",
+  "name": "ssr-target-bundle",
   "version": "0.1.0",
   "type": "module",
   "private": true,
--- a/e2e/fixtures/ssr-target-bundle/package.json
+++ b/e2e/fixtures/ssr-target-bundle/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "ssr-basic",
+  "version": "0.1.0",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "dev": "waku dev --with-ssr",
+    "build": "waku build --with-ssr",
+    "start": "waku start --with-ssr"
+  },
+  "dependencies": {
+    "react": "18.3.0-canary-4b2a1115a-20240202",
+    "react-dom": "18.3.0-canary-4b2a1115a-20240202",
+    "react-server-dom-webpack": "18.3.0-canary-4b2a1115a-20240202",
+    "react-textarea-autosize": "^8.5.3",
+    "waku": "0.19.2"
+  },
+  "devDependencies": {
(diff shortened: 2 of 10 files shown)


e2e/fixtures/ssr-target-bundle/README.md
```diff
@@ -0,0 +1,3 @@
+# SSR Bundling for NPM packages with environment depended code parts
+
+Choosing the right code based on the `export` section of the `react-textarea-autosize` module based on the backend server deployment target ('node' | 'webworker') and the client browser. No browser only code should be bundled with server only code.
```
dai-shi, line 3: To me, it's not our fault that when target=webworker, the bundler picks the browser version, instead of worker version. So, it's our temporarily workaround.

e2e/fixtures/ssr-target-bundle/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "ssr-basic",
```
dai-shi, line 2:
```suggestion
  "name": "ssr-target-bundle",
```

dai-shi:
@aheissenberger do you know ci is failing?

@himself65 can you do a quick review?

e2e/fixtures/ssr-target-bundle/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "dev": "waku dev --with-ssr",
+    "build": "waku build --with-ssr",
+    "start": "waku start --with-ssr"
+  },
+  "dependencies": {
+    "react": "18.3.0-canary-4b2a1115a-20240202",
+    "react-dom": "18.3.0-canary-4b2a1115a-20240202",
+    "react-server-dom-webpack": "18.3.0-canary-4b2a1115a-20240202",
+    "react-textarea-autosize": "^8.5.3",
+    "waku": "0.19.2"
```
himself65, line 16:
```suggestion
    "waku": "workspace:*"
```

e2e/fixtures/ssr-target-bundle/src/entries.tsx
```diff
@@ -0,0 +1,28 @@
+import { lazy } from 'react';
+import { defineEntries } from 'waku/server';
+import { Slot } from 'waku/client';
```
himself65, line 3: error here because you import a package from monorepo.

Solution is add it to `references` on `tsconfig.json`

e2e/fixtures/ssr-target-bundle/package.json
himself65: Oh I think I should change other fixtures also, because they use monorepo actually

himself65 (changes requested):

e2e/fixtures/ssr-target-bundle/package.json
dai-shi: It's actually fine, if the version is the same, it will use the workspace version. (And, there's no way to opt-out this behavior, which is actually troublesome.)

e2e/fixtures/ssr-target-bundle/README.md
aheissenberger: What would you suggest?

dai-shi: Nothing. Sorry for the confusion. It's just a note so that we are on the same page.

dai-shi (approved):
LGTM
Thanks for your contribution!
As you may notice, we should cover more cases in e2e.

PR #494, opened 2024-02-16
feat: output partykit

threepointone:
This adds experimental support for deploying to PartyKit (https://www.partykit.io/)

### DIFF
--- a/packages/waku/src/lib/builder/output-partykit.ts
+++ b/packages/waku/src/lib/builder/output-partykit.ts
@@ -1,0 +1,27 @@
+import path from 'node:path';
+import { existsSync, writeFileSync } from 'node:fs';
+
+import type { ResolvedConfig } from '../config.js';
+
+// XXX this can be very limited. FIXME if anyone has better knowledge.
+export const emitPartyKitOutput = async (
+  rootDir: string,
+  config: ResolvedConfig,
+) => {
+  const partykitJsonFile = path.join(rootDir, 'partykit.json');
+  if (!existsSync(partykitJsonFile)) {
+    writeFileSync(
+      partykitJsonFile,
+      JSON.stringify(
+        {
+          name: 'waku-project',
+          main: `${config.distDir}/${config.serveJs}`,
+          compatibilityDate: '2023-02-16',
+          assets: `./${config.distDir}/${config.publicDir}`,
+        },
+        null,
+        2,
+      ) + '\n',
+    );
+  }
+};
--- a/packages/waku/src/lib/builder/serve-partykit.ts
+++ b/packages/waku/src/lib/builder/serve-partykit.ts
@@ -1,0 +1,19 @@
+import { Hono } from 'hono';
+
+import { honoMiddleware } from '../middleware/hono-prd.js';
+
+const ssr = !!import.meta.env.WAKU_BUILD_SSR;
+const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
+let serveWaku: ReturnType<typeof honoMiddleware> | undefined;
+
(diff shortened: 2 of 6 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-serve.ts
```diff
@@ -44,6 +50,17 @@ export function rscServePlugin(opts: {
             'Unsupported: build.rollupOptions.external is not an array',
           );
         }
+      } else if (opts.serve === 'partykit') {
```
dai-shi, line 53: I think this can be merged with the previous condition. I'll make a follow-up PR later.

packages/waku/src/lib/builder/output-partykit.ts
```diff
@@ -0,0 +1,27 @@
+import path from 'node:path';
+import { existsSync, writeFileSync } from 'node:fs';
+
+import type { ResolvedConfig } from '../config.js';
+
+// XXX this can be very limited. FIXME if anyone has better knowledge.
```
dai-shi, line 6: I think you would have better knowledge on partykit.

packages/waku/src/lib/builder/serve-partykit.ts
```diff
@@ -0,0 +1,19 @@
...
+const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
+let serveWaku: ReturnType<typeof honoMiddleware> | undefined;
+
+const app = new Hono();
+
+app.use('*', (c, next) => serveWaku!(c, next));
+export default {
+  onFetch(request: Request, lobby: any, ctx: Parameters<typeof app.fetch>[2]) {
+    if (!serveWaku) {
+      serveWaku = honoMiddleware({ loadEntries, ssr, env: lobby });
```
dai-shi, line 15: btw, this was the only workaround I came up with for cloudflare.
How do you think this is feasible in the long run?

dai-shi (approved):
LGTM
Thanks for your contribution.

I don't know, but we may consider making those deploy adapters more pluggable somehow. Maybe in the future, maybe before reaching v1.

packages/waku/src/lib/builder/serve-partykit.ts
threepointone: This is fine. Another option would be to have a helper function that generates the export. Yet another option would be to use AsyncLocalStorage. I'll send a PR later if you'd like to clean this up.

packages/waku/src/lib/builder/output-partykit.ts
threepointone: a starter kit that already has this config would also work

dai-shi: I mean, we might not need this comment in this file, which I added in output-cloudflare.ts.

packages/waku/src/lib/builder/serve-partykit.ts
dai-shi: If such a solution works for both --with-cloudflare and --with-partykit, I would be interested. I don't think AsyncLocalStorage would work, because it's a Node.js API.

threepointone: AsyncLocalStorage is now available in cloudflare with their node compat mode (which is turned on by default in partykit) https://developers.cloudflare.com/workers/runtime-apis/nodejs/

dai-shi: currently, we don't use node compat mode for cloudflare deploy.

what's your second option? > a helper function that generates the export.

Issue #329, opened 2023-12-28
support for non-React API endpoints


himself65:
I want to extend the Vercel serverless function to support non-react functions like AI stream(text stream), JSON response.

vercel edge function example:

```ts
import { connectMiddleware } from 'waku'
import { OpenAIStream, StreamingTextResponse } from 'ai';
import OpenAI from 'openai';

export default async function handler(req: Request, res: Response) {
  if (req.url === '/api/ai') {
    const { messages } = await req.json();

    const response = await openai.chat.completions.create({
      model: 'gpt-3.5-turbo',
      stream: true,
      messages,
    });

    const stream = OpenAIStream(response);

    return new StreamingTextResponse(stream);
 } else {
    connectMiddleware(/* ... */)
 }
}
```

Related discussion: https://github.com/dai-shi/waku/discussions/290

dai-shi:
Please provide description.

dai-shi:
> I want to extend the Vercel serverless function to support non-react functions

I would create another serverless function and configure routes. It's true that there is no direct support for such use cases.

himself65:
I used next.js days ago. I found that they support multiple runtimes on the same platform. edge and runtime. just one const

```ts
export const runtime = "edge";
```

I'm not sure if waku provide createPage with runtime would be a good option

himself65:
Full JS environment people are using 

- Node.js (18/20)
- Edge
- Deno
- Bun
- Cloudflare worker

dai-shi:
We are taking a different approach. Waku core is designed completely with web standard, and provides wrappers for each runtime. (And I'm not a big fan of `export const ...` for configuring compiler behavior.)

That said, we care DX for supporting different runtimes. Will see.

dai-shi:
I want try something soon, but can't tell what it's going to be like yet.

dai-shi:
Thanks. It's probably similar to something I have in mind. I will refactor things. Please wait for it. (Lots of bug fixes are required before adding features. Please see "help wanted" issues.)
Your review and so on will be important when I finish my draft.

Meanwhile, enjoy experiments.

dai-shi:
I moved a task related to this to #513.

dai-shi:
And, reverted to the original issue.

Here's my first question.
Do people want a separate serverless function for API routes, or embed it in our framework (meaning a combined serverless function).
In the case of former, it's probably out of the scope, because it's cloud environment specific.

For the second question,
[...]

PR #497, opened 2024-02-17
support custom 404 (static html only)

dai-shi:
Jut put `./public/404.html`.


dai-shi:
@aheissenberger Can you work on AWS Lambada?
@threepointone Can you work on Partykit?

If this PR is still open, send a PR to this branch, otherwise to `main`.

examples/01_template/public/404.html
```diff
@@ -0,0 +1 @@
+<h1>Custom Not Found Page</h1>
```
himself65, line 1: What if I want use javascript here?

himself65: This looks like a manual setup not related to Waku itself

dai-shi: Think it like a static image asset file. Up until now, it's very hard coded. So, this at least makes it configurable. And, it works without SSR.

I plan to do full customization in #329. But, if we want a solution "with React" we need another solution. I'll think about it, but it requires SSR.

A question for you: How should we deal with customizing 404 without SSR? What's the typical way with Vite + React SPA?

> not related to Waku itself

Yeah, it's not related to RSC. But, we need do something and this is a minimal setup.

PR #499, opened 2024-02-17
feat: add fix and test for asset handling of imported images

### LINKED ISSUE #498: build / image import broken
an image imported in a server component is missing in the `dist/public/assets` folder.
It exists in the `dist/assets/sample-Xox1zmov.jpg` which is not served by the static server.

Waku Commit: e85380c519a899244c548c460de1d0152efdabbc

```js
import SampleImage from '../assets/sample.jpg';

export const MantineCorePage = async () => {
   return (<img src={SampleImage} />)
}
```

aheissenberger:
This a test case to fix #498


packages/waku/src/lib/builder/build.ts
```diff
@@ -355,9 +355,10 @@ const buildClientBundle = async (
   ssr: boolean,
 ) => {
   const mainJsFile = joinPath(rootDir, config.srcDir, config.mainJs);
-  const cssAssets = serverBuildOutput.output.flatMap(({ type, fileName }) =>
-    type === 'asset' && fileName.endsWith('.css') ? [fileName] : [],
+  const nonJsAssets = serverBuildOutput.output.flatMap(({ type, fileName }) =>
```
dai-shi, line 358: Can you please do the same for the SSR bundle? We need it right?

aheissenberger: Copying assets to ssr folder is not needed - have a look at my new tests - the leak of "private" JSON assets is tested and works as expected.

aheissenberger: @dai-shi can you look at the failed  typescript error - it is not shown when I run `pnpm test` local but I think it is related to the fact that I have changed the `App` component to an async function in `./e2e/fixtures/ssr-target-bundle/src/components/App.tsx` - should I disable the error in `entries.tsx`?

> Type 'Promise<typeof import("/waku/e2e/fixtures/ssr-target-bundle/src/components/App", { with: { "resolution-mode": "import" } })>' is not assignable to type 'Promise<{ default: ComponentType<any>; }>'.

dai-shi: > Copying assets to ssr folder is not needed - have a look at my new tests - the leak of "private" JSON assets is tested and works as expected.

Oh, because we don't need styles for SSR? Yeah, that's my oversight.

aheissenberger: I have to clarify - the "private" JSON is not leaked to the public directory in your current implementation :-)

dai-shi:
> @dai-shi can you look at the failed typescript error - it is not shown when I run `pnpm test` local but I think it is related to the fact that I have changed the `App` component to an async function in `./e2e/fixtures/ssr-target-bundle/src/components/App.tsx` - should I disable the error in `entries.tsx`?
> 
> > Type 'Promise<typeof import("/waku/e2e/fixtures/ssr-target-bundle/src/components/App", { with: { "resolution-mode": "import" } })>' is not assignable to type 'Promise<{ default: ComponentType; }>'.

`pnpm test` shows the error for me.

dai-shi (approved):
LGTM. Thanks for your work!

PR #510, opened 2024-02-19
fix: improve handler order to support catch-all route

dai-shi:
Thanks to #442, we can now do it. Also refactored router examples.


dai-shi:
I'll open a follow up PR later.

PR #515, opened 2024-02-20
feat: better initial speed by preimporting entries

Aslemammad:
While trying to fix the initial load speed issue for #344 I scoped the code blocks to tsx and tried to pre-transform the mdx in codeToHtml since shiki was slowing the initial page load, but that didn't work and I found out that waku does not start importing the entries file unless we send an initial request.

This does not cause huge harm, but it's not ideal since we need to pre-import the entries the same way vite pre-transform some files, so we avoid the module imports and other side-effects (reading initial markdown files) overhead

For the initial page load issue, I could not find an easy solution for pre-transform the code blocks, but if we implement it at some point (cc @sandren), this PR is the initial step for it.


dai-shi:
Added a refactor commit.

tbh, this doesn't make any behavioral change for me, because the other issue make it too slow.
I'm fine to merge this as the first step because it's harmless especially in the worker.
But, my understanding was Vite prefers lazy loading and avoids eager loading when possible.

Aslemammad:
Yea, but in this case, it's better to preload since everything is ready and
at the same time, the user expects lower cold start overhead.

PR #518, opened 2024-02-21
fix: document body undefined

### LINKED ISSUE #344: website: local development issue

Aslemammad:
Partially resolves #344 

When refreshing a waku page (website), the hydration failed and as result, hmr would not work. 

The reason behind the failure, is that for the next refreshes, the `/main.tsx` file is optimized and warm so it gets resolved before the `<body>` is getting streamed to the browser since it was in the `<head>` and that loads before the `<body>`, what that mean is the `document.body` would be undefined during the hydration and React would throw an error since that is undefined. One solution was to restart the server so we get `/main.tsx` cold again.

What this PR does is that it puts the script of `/main.tsx` in the body so it gets resolved after the body is being streamed and as a result, document.body would be never be `undefined`. 

This also fixed the hmr issue after the page manual refresh.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
@@ -15,9 +15,9 @@ export function rscIndexPlugin(config: {
 <html>
   <head>
 ${config.htmlHead}
-<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>
   </head>
   <body>
+    <script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>
   </body>
 </html>
 `;
@@ -26,9 +26,9 @@ ${config.htmlHead}
     configureServer(server) {
       return () => {
         server.middlewares.use((req, res) => {
-          res.statusCode = 200;
-          res.setHeader('content-type', 'text/html; charset=utf-8');
           server.transformIndexHtml(req.url || '', html).then((content) => {
+            res.statusCode = 200;
+            res.setHeader('content-type', 'text/html; charset=utf-8');
             res.end(content);
           });
         });
--- a/packages/waku/src/lib/renderers/html-renderer.ts
+++ b/packages/waku/src/lib/renderers/html-renderer.ts
@@ -103,8 +103,12 @@ globalThis.__WAKU_PREFETCHED__ = {
       if (scriptsClosed) {
         return;
       }
-      // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded 
-      controller.enqueue(encoder.encode(`<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>`));
+      // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded
+      controller.enqueue(
+        encoder.encode(
+          `<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>`,
+        ),
+      );
 
       const scripts = chunks.splice(0).map(
         (chunk) =>
--- a/packages/waku/src/lib/renderers/html-renderer.ts
+++ b/packages/waku/src/lib/renderers/html-renderer.ts
@@ -349,7 +349,7 @@ export const renderHtml = async (
   );
   const [copied, interleave] = injectRscPayload(
     stream,
-    config,
+    `${config.basePath}${config.srcDir}/${config.mainJs}`,
     config.basePath + config.rscPath + '/' + encodeInput(ssrConfig.input),
   );
   const elements: Promise<Record<string, ReactNode>> = createFromReadableStream(
--- a/packages/waku/src/lib/renderers/html-renderer.ts
+++ b/packages/waku/src/lib/renderers/html-renderer.ts
@@ -108,7 +108,7 @@ globalThis.__WAKU_PREFETCHED__ = {
         // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded
         controller.enqueue(
           encoder.encode(
-            `<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>`,
+            `<script src="${mainJsPath}" async type="module"></script>`,
           ),
         );
         mainJsSent = true;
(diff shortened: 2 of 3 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
```diff
@@ -26,9 +26,9 @@ ${config.htmlHead}
     configureServer(server) {
       return () => {
         server.middlewares.use((req, res) => {
-          res.statusCode = 200;
-          res.setHeader('content-type', 'text/html; charset=utf-8');
           server.transformIndexHtml(req.url || '', html).then((content) => {
+            res.statusCode = 200;
```
Aslemammad, line 30: This is not related to the PR, but it's a minor change, since we used to set the status 200 even though the `.then` might never work and the promise might reject. 

So we just bring it when we're sure the promise is getting resolved.

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -102,6 +103,13 @@ globalThis.__WAKU_PREFETCHED__ = {
       if (scriptsClosed) {
         return;
       }
+      // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded
+      controller.enqueue(
+        encoder.encode(
+          `<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>`,
+        ),
+      );
+
```
dai-shi, line 112: Here, this adds `<script ...>` multiple times.
`npm run examples:dev:04_promise` can reproduce it.

dai-shi (changes requested):

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -44,6 +44,7 @@ Promise.resolve(new Response(new ReadableStream({
 
 const injectRscPayload = (
   readable: ReadableStream,
+  config: ResolvedConfig,
```
dai-shi, line 47: Let's change this to:
```suggestion
  mainJsPath: string,
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -337,6 +349,7 @@ export const renderHtml = async (
   );
   const [copied, interleave] = injectRscPayload(
     stream,
+    config,
```
dai-shi, line 352:
```suggestion
    `${config.basePath}${config.srcDir}/${config.mainJs}`,
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -102,6 +104,16 @@ globalThis.__WAKU_PREFETCHED__ = {
       if (scriptsClosed) {
         return;
       }
+      if (!mainJsSent) {
+        // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded
+        controller.enqueue(
+          encoder.encode(
+            `<script src="${config.basePath}${config.srcDir}/${config.mainJs}" async type="module"></script>`,
```
dai-shi, line 111:
```suggestion
            `<script src="${mainJsPath}" async type="module"></script>`,
```

dai-shi (approved):
should be good.

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -102,6 +104,16 @@ globalThis.__WAKU_PREFETCHED__ = {
       if (scriptsClosed) {
         return;
       }
+      if (!mainJsSent) {
+        // enqueue `mainJs` first in the body to avoid the document.body undefined error, before we used to inject it in the head which sometimes made it load before the document.body was loaded
```
dai-shi, line 108: My oversight, this should be DEV only, right?

`npm run examples:prd:04_promise`

image

Aslemammad: This should in general send the mainJs one time, weird error, do you know why that happens?

dai-shi: Here's the fix #528 

`main.tsx` doesn't exist on PRD, right?

Aslemammad: Yes, what exists then if that does not exist? because as far as I checked the website, there was an alternative. And we need that to be in the body too, so the behaviour is normalized between two environments.

Aslemammad: Oh, ok the reason we get 404 is that we give a path that's not valid for prd! And in prd we give a different value, correct?
if yes, and we control that value, are we sure it'd be passed in the body rather than the head? Or are we sure that the body is going to be returned? so we don't get the error.

dai-shi: See #528 for the fix. Yes, it's already in the head in PRD. It's already done in the build step, so we don't need to do anything in the runtime process.

Aslemammad:
Hmm, interesting, the goal is to have it in the body so we always sure that
document.body is available. wdyt?

dai-shi:
I think usually document.body is available with "async" script.
Can you try creating Vite+React app and see its built result? All scripts are in `<head>`, right?

packages/waku/src/lib/renderers/html-renderer.ts
Aslemammad: But when it's in the head, and we use document.body there, are we sure that document.body is going to be always defined?

dai-shi: https://github.com/dai-shi/waku/pull/518#issuecomment-1962238483

Aslemammad: But the error that this PR fixed was reproduced with async too, so I don't think that can always be true and we need to be sure about that.

dai-shi: My understanding is that the issue this PR fixed is for HMR in dev mode. I never see such an issue in PRD.

Aslemammad: Yes it fixed the HMR issue after the refresh in DEV, but I was just concerned how it'd be handled in PRD. That's it. sorry for the confusion.

Aslemammad: Yep, they're in the head! because the body is always sent in the initial template (index.html)

Issue #517, opened 2024-02-21
ws is deprecated


Aslemammad:
in the newer version of vite, ws is deprecated and we have `hot`. So we need to rely on hot too from now on.

```
    /**
     * web socket server with `send(payload)` method
     * @deprecated use `hot` instead
     */
    ws: WebSocketServer;
```

mjh316:
I pushed a PR attempting to perform this - would that be alright?

PR #521, opened 2024-02-22
feat: Migrate HMR api to `vite.hot` from `vite.ws`

mjh316:
Tackles issue #517 and migrates from the deprecated `vite.ws` API to the newer Vite 5.1 API, which is `vite.hot`.


Aslemammad (changes requested):
This would break on previous versions since there is no hot at that point. I suggest going something like this:

```
const key = server.hot ?? server.ws 
```

mjh316:
Just added fallback, not sure if it fixes the timeout errors in e2e

packages/waku/src/lib/handlers/dev-worker-impl.ts
```diff
@@ -158,7 +158,8 @@ const mergedViteConfig = await mergeUserViteConfig({
 });
 
 const vitePromise = createViteServer(mergedViteConfig).then(async (vite) => {
-  await vite.ws.close();
+  const HMRKEY = vite.hot ?? vite.ws;
```
Aslemammad, line 161: Great progress, this can be better.

```
const hot = viteHot(vite);
hot.close()
```

And viteHot would do the fallbacking. Same with other places.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -144,28 +145,30 @@ function hotImport(viteServer: ViteDevServer, source: string) {
     });
   }
   sourceSet.add(source);
-  viteServer.ws.send({ type: 'custom', event: 'hot-import', data: source });
+  HMRKEY.send({ type: 'custom', event: 'hot-import', data: source });
 }
 
 const modulePendingMap = new WeakMap<
-  ViteDevServer['ws'],
+  ViteDevServer['hot'],
```
Aslemammad, line 152: Can we have an `|` like `ViteDevServer['hot'] | ViteDevServer['ws']`?

Aslemammad (changes requested):

mjh316:
Exported the function for use - export or no?

dai-shi:
> Exported the function for use - export or no?

If not used now, don't export.

mjh316:
It is used in the two files, but not anywhere else, it seems. Should I just define it in both?

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -144,28 +149,30 @@ function hotImport(viteServer: ViteDevServer, source: string) {
...
+  const HMRKEY = viteServer.hot ?? viteServer.ws;
+  let sourceSet = modulePendingMap.get(HMRKEY);
   if (!sourceSet) {
     sourceSet = new Set();
-    modulePendingMap.set(viteServer.ws, sourceSet);
+    modulePendingMap.set(HMRKEY, sourceSet);
   }
   sourceSet.add(result);
-  viteServer.ws.send({ type: 'custom', event: 'module-import', data: result });
+  HMRKEY.send({ type: 'custom', event: 'module-import', data: result });
```
Aslemammad, line 168: please change this too

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -144,28 +149,30 @@ function hotImport(viteServer: ViteDevServer, source: string) {
...
 }
 
 async function generateInitialScripts(
   viteServer: ViteDevServer,
 ): Promise<HtmlTagDescriptor[]> {
-  const sourceSet = modulePendingMap.get(viteServer.ws);
+  const HMRKEY = viteServer.hot ?? viteServer.ws;
```
Aslemammad, line 174: same with this, `hot`

Aslemammad:
> If not used now, don't export.

I think since it's related to hmr, we can have it in the hmr file? up to you.

mjh316:
Are the specific failing tests non-deterministic? It seems like the formatting somehow broke a windows test somewhere in there.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -126,16 +126,24 @@ export function rscHmrPlugin(): Plugin {
   };
 }
 
-const pendingMap = new WeakMap<ViteDevServer['ws'], Set<string>>();
+const pendingMap = new WeakMap<
+  ViteDevServer['hot'] | ViteDevServer['ws'],
```
Aslemammad, line 130: Now I'm thinking, can this be `ReturnType<typeof viteHot>`? what do you think @mjh316

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -144,28 +152,30 @@ function hotImport(viteServer: ViteDevServer, source: string) {
     });
   }
   sourceSet.add(source);
-  viteServer.ws.send({ type: 'custom', event: 'hot-import', data: source });
+  hot.send({ type: 'custom', event: 'hot-import', data: source });
 }
 
 const modulePendingMap = new WeakMap<
-  ViteDevServer['ws'],
+  ViteDevServer['hot'] | ViteDevServer['ws'],
```
Aslemammad, line 159: same with this

Aslemammad:
>  It seems like the formatting somehow broke a windows test somewhere in there.
No, it's ok, a test failed because of formatting.

dai-shi:
> Are the specific failing tests non-deterministic?

Our tests are a little bit flaky (better than before, though.)
cc: @himself65

Aslemammad (approved):
Tried the hmr with the website and it's working. I think we can merge now.

mjh316:
Thank you so much for your support!

Issue #523, opened 2024-02-22
Example does not work


kamuiiiii:
<img width="1214" alt="image" src="https://github.com/dai-shi/waku/assets/25859424/3859590c-b595-425d-9b2b-c6e3e9a414e8">
<img width="1213" alt="image" src="https://github.com/dai-shi/waku/assets/25859424/c1e77d2e-016f-48b4-a310-be5fe5951ebe">

dai-shi:
How do you run examples?

kamuiiiii:
reinstall dependencies solved the problem, thanks!

PR #524, opened 2024-02-22
fix:handler issue

kamuiiiii:
fix https://github.com/dai-shi/waku/issues/523

### DIFF
--- a/packages/waku/src/lib/handlers/handler-dev.ts
+++ b/packages/waku/src/lib/handlers/handler-dev.ts
@@ -210,8 +210,6 @@ export function createHandler<
             .pipeTo(res.stream);
           return;
         }
-        next();
-        return;
       } catch (e) {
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/handlers/handler-dev.ts
```diff
@@ -210,8 +210,6 @@ export function createHandler<
             .pipeTo(res.stream);
           return;
         }
-        next();
-        return;
```
dai-shi, line 214: This is an intentional change in #510.
If this is causing some issues, I think `willBeHandledByVite` isn't working as expected.

dai-shi: It was intentional at that point, but this prevents skipping SSR conditionally even with `--with-ssr`, so I need to do this anyway.

PR #529, opened 2024-02-23
use rsc-html-stream

dai-shi:
use https://github.com/devongovett/rsc-html-stream which supports binary data.

I wasn't able to use their client code, because our SSR capability is opt-in and out of the bundler scope.


dai-shi:
https://github.com/devongovett/rsc-html-stream/issues/4

dai-shi:
🎉 The new version fixes for Deno Deploy.

PR #531, opened 2024-02-24
experimental: add openai example

himself65:


himself65:
closing as I think this is not stable enough

PR #533, opened 2024-02-25
fix: no analyzing after client boundary

Aslemammad:


packages/waku/src/lib/plugins/vite-plugin-rsc-analyze.ts
```diff
@@ -9,19 +9,12 @@ export function rscAnalyzePlugin(
   clientFileSet: Set<string>,
   serverFileSet: Set<string>,
 ): Plugin {
-  const rscTransform = rscTransformPlugin({ isBuild: false }).transform;
-  // TODO this is unused. a leftover in #490. remove it.
-  const dependencyMap = new Map<string, Set<string>>();
+  const rscTransform = rscTransformPlugin({ isBuild: false })
+    .transform as ReturnType<typeof rscAnalyzePlugin>['transform'] & {
+    handler: undefined;
+  };
```
dai-shi, line 15: nice hack. Can `rscAnalyzePlugin` be `rscTransformPlugin` at L13?

PR #534, opened 2024-02-25
feat: conditional ssr (minimal API only)

dai-shi:
- [x] add an example (in examples/06)
- [x] fix history api fallback in DEV
- [x] replace `WAKU_HYDRATE`
- [x] build html for non SSR html pages for PRD
- history api fallback 👉 #545


e2e/fixtures/ssr-basic/src/main.tsx
```diff
@@ -10,7 +10,7 @@ const rootElement = (
   </StrictMode>
 );
 
-if (import.meta.env.WAKU_HYDRATE) {
+if (document.body.dataset.hydrate) {
```
dai-shi, line 13: I know this isn't optimal.
We may revisit this in the future, or maybe it's okay with the Waku target.

dai-shi:
This is technically a breaking change, but the migration is easy.

Issue #536, opened 2024-02-26
`<a href="/[some_dynamic_page]"/>` will cause 404 in prod mode


himself65:
workaround is to use Link

himself65:
Seems like it's the issue on dynamic page

dai-shi:
Thanks for reporting. Can we reproduce it with examples/01_template with changing `render: 'dynamic'` for `path: '/'`?

himself65:
> Thanks for reporting. Can we reproduce it with examples/01_template with changing `render: 'dynamic'` for `path: '/'`?

im adding the test case now

himself65:
Oh, I think it's my custom app logic issue, here is fallback in waku start

```js
  if (!options.ssr) {
    // history api fallback
    app.use(
      '*',
      serveStatic({
        root: path.join(distDir, publicDir),
        rewriteRequestPath: () => '/',
      }),
    );
  }
```

dai-shi:
what's your custom app logic? is your `options.ssr` false?

himself65:
Yeah, I disabled SSR because it's a electron app, which run waku

dai-shi:
Is it specific to `/`? I wonder if it causes the same issue with `path: '/foo/[id]'`.
It might be related to #534, which i'm currently working on.

himself65:
Closing as this is not a bug in waku, but still need to update the document and cover the test

Solution: https://github.com/dai-shi/waku/issues/536#issuecomment-1963177880

dai-shi:
Okay, I think I understand the problem.
When `/` is dynamic, we don't create `dist/public/index.html`,
then history api fallback doesn't work.
(I will think about it.)

himself65:
> Okay, I think I understand the problem. When `/` is dynamic, we don't create `dist/public/index.html`, then history api fallback doesn't work. (I will think about it.)

Maybe we should fall back to SPA?

PR #537, opened 2024-02-26
test: add test cases for defineRouter

himself65:
Cover case: https://github.com/dai-shi/waku/issues/536


e2e/fixtures/waku-router/src/entries.tsx
```diff
@@ -0,0 +1,33 @@
...
+    component: AboutPage,
+  });
+
+  createPage({
+    render: 'dynamic',
+    path: '/[roomId]',
+    component: RoomPage,
+  });
+  // fixme: error TS2742
+}) as any;
```
himself65, line 33: IDK

tsconfig.json
```diff
@@ -16,11 +16,7 @@
     "declaration": true,
     "emitDeclarationOnly": true,
     "composite": true,
-    "incremental": true,
-    "paths": {
-      "waku": ["./packages/waku/src/main.ts"],
-      "waku/*": ["./packages/waku/src/*.ts"]
-    }
```
himself65, line 23: I found that examples are aliases to the src file, I think this is not good. we should type check from output

e2e/fixtures/waku-router/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "waku-router",
+  "version": "0.1.0",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "dev": "waku dev",
+    "build": "waku build",
+    "start": "waku start"
```
dai-shi, line 9: Is it intentional to not add `--with-ssr`?

himself65: It's okay for here since test case will run both cases

himself65:
```
 Locator: getByTestId('current-url')
    Expected string: "/"
    Received string: ""
```

It works perfectly on my PC, let me see tmr

e2e/fixtures/waku-router/package.json
dai-shi: I see.

PR #538, opened 2024-02-26
fix: stop eslint react warning

himself65:
https://github.com/DRD4-7R/eslint-config-7r-building/issues/1#issuecomment-779443161


dai-shi (approved):
LGTM

PR #540, opened 2024-02-26
fix: add option to define the root folder of a lambda deployment.

aheissenberger:
This change allows to handle the differences of bundling the deployment ZIP for the serverless framework and the AWS CDK.

### DIFF
--- a/packages/waku/src/lib/builder/serve-aws-lambda.ts
+++ b/packages/waku/src/lib/builder/serve-aws-lambda.ts
@@ -7,7 +7,7 @@ import path from 'node:path';
 import { existsSync, readFileSync } from 'node:fs';
 
 const ssr = !!import.meta.env.WAKU_BUILD_SSR;
-const distDir = import.meta.env.WAKU_CONFIG_DIST_DIR!;
+const distDir = process.env?.WAKU_BUILD_DIST_DIR ?? '';
 const publicDir = import.meta.env.WAKU_CONFIG_PUBLIC_DIR!;
 const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
 


packages/waku/src/lib/builder/serve-aws-lambda.ts
```diff
@@ -7,7 +7,7 @@ import path from 'node:path';
 import { existsSync, readFileSync } from 'node:fs';
 
 const ssr = !!import.meta.env.WAKU_BUILD_SSR;
-const distDir = import.meta.env.WAKU_CONFIG_DIST_DIR!;
+const distDir = process.env?.WAKU_BUILD_DIST_DIR ?? '';
```
dai-shi, line 10: Not exactly sure how it works, but should default be something other than `''`?

```js
const distDir = process.env?.WAKU_BUILD_DIST_DIR ?? import.meta.env.WAKU_CONFIG_DIST_DIR!;
```

```js
const distDir = process.env?.WAKU_BUILD_DIST_DIR ?? '.';
```

aheissenberger: This is a fix to handle the problem of the `Serverless framework` including the top folder `dist` in the deployed ZIP which is the reason for my problems #514  to implement the private folder in the build-adapter code.

Other deployment solution (e.g. AWS CDK) do not add this extra hierarchy and do not need to know the name of the `dist` folder.

This is not a perfect solution, but works for now with both deployment options described in the docs. I will have to optimize it in a next step to remove the currently fixed coded `dist` folder in deployment configs for the serverless framework and AWS CDK, to solution which can extract the value from `WAKU_CONFIG_DIST_DIR`.

dai-shi (approved):

PR #532, opened 2024-02-25
build/analyze: avoid walking after the client boundary

dai-shi:
Another take for #525 to resolve #418.

PR #301, opened 2023-12-23
feat: add `react-tweet` example

himself65:
Fixes: https://github.com/dai-shi/waku/issues/90

PR #541, opened 2024-02-26
feat: add react tweet example

### LINKED ISSUE #90: Support `react-tweet`

Aslemammad:
Resolves #90 

Prior art: #301 

With #532 we have this fixed now apparently! Which marks a huge milestone for waku as it shows waku is getting things right in terms of package exports and external library support after 8 months of work.


examples/14_react-tweet/vite.config.ts
```diff
@@ -0,0 +1,10 @@
+/** @type {import('vite').UserConfig} */
+export default {
+  optimizeDeps: {
+    include: ['react-tweet'],
+  },
+  ssr: {
+    external: ['use-sync-external-store'],
+    noExternal: ['react-tweet'],
```
Aslemammad, line 8: `use-sync-external-store` is cjs. 

`react-tweet` is externalized and that means vite will not resolve it using our plugins and will directly fallback to nodejs module system. What this means is that we'll `.module.css` error since it's not natively supported by nodejs.

examples/14_react-tweet/src/templates/home-page.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const HomePage = async () => {
+  const data = await getData();
+
+  return (
+    <div>
+      <title>{data.title}</title>
+      <h1 className="text-4xl font-bold tracking-tight">{data.headline}</h1>
+      <p>{data.body}</p>
+      <Tweet id="1735308967880823082" />
```
Aslemammad, line 12: image

That's legacy here 😄

dai-shi (approved):
I think we should generally reduce the number of examples, and make them better show cases. For testing `react-tweet`, an e2e fixture would be more appropriate.
That said, let's merge this for now, and migrate to a fixture eventually.

examples/14_react-tweet/vite.config.ts
```diff
@@ -2,10 +2,13 @@
 export default ({ mode }: { mode: string }) => ({
   ...(mode === 'development' && {
```
Aslemammad, line 3: I think this pattern might be confusing for some users :)

Aslemammad: so maybe we can remove it?

dai-shi: hmm, but I want to emphasize and confirm that this is dev-only workaround. any other idea?

Aslemammad: a comment maybe? // This is a dev-only solution.

dai-shi: Actually, we shouldn't externalize for production. Maybe we should use `if` statement to be more readable?
Like https://vitejs.dev/config/#conditional-config

Aslemammad: Yes, that can be good too, with a comment.

dai-shi: send a pr when you have time.

Issue #543, opened 2024-02-27
bug: index page will be always SSR


himself65:
See: https://github.com/himself65/waku/commit/5e8218f33fb1a7b9ec1ead5d9ce665f8199b75ac

```shell
/Users/himself65/Library/Caches/fnm_multishells/48575_1709013514281/bin/pnpm run build

> waku-starter@0.1.0 build /Users/himself65/Code/waku/examples/01_template
> DONT_LEAK_IN_BUILD_TIME=NO waku build

vite v5.1.4 building SSR bundle for production...
✓ 42 modules transformed.
✓ built in 374ms
vite v5.1.4 building SSR bundle for production...
✓ 43 modules transformed.
dist/assets/root-layout-BDx5PYJp.css              11.64 kB
dist/assets/rsc1-ddb75831a.js                      0.15 kB
dist/components/header.js                          0.45 kB
dist/components/counter.js                         0.48 kB
dist/components/error-boundary.js                  0.51 kB
dist/templates/contact-page.js                     0.53 kB
dist/components/footer.js                          0.71 kB
dist/templates/about-page.js                       1.06 kB
dist/templates/root-layout.js                      1.16 kB
dist/templates/home-page.js                        1.16 kB
dist/assets/jsx-runtime.react-server-DaQipseb.js   1.25 kB
dist/assets/rsc2-19056431e.js                      1.99 kB
dist/assets/rsc0-1f451937d.js                      2.67 kB
dist/assets/react.react-server-BRc6Y66K.js        10.76 kB
dist/entries.js                                   12.98 kB
dist/rsdw-server.js                               41.09 kB
✓ built in 106ms
vite v5.1.4 building for production...
✓ 40 modules transformed.
dist/public/index.html                        1.01 kB │ gzip:  0.45 kB
dist/public/assets/rsc0-1f451937d.js          0.21 kB │ gzip:  0.17 kB
dist/public/assets/rsc1-ddb75831a.js          0.42 kB │ gzip:  0.31 kB
dist/public/assets/jsx-runtime-CcapePxT.js    0.63 kB │ gzip:  0.42 kB
dist/public/assets/main-A3kPowAf.js           0.73 kB │ gzip:  0.46 kB
dist/public/assets/indexHtml-DVWfb-0O.js      0.86 kB │ gzip:  0.49 kB
dist/public/assets/rsc2-19056431e.js          4.24 kB │ gzip:  1.80 kB
[...]

PR #544, opened 2024-02-27
fix: enable no floating promise

himself65:
It looks like the floating promise caused this issue. I will enable this rule by default.

Discussion: https://github.com/dai-shi/waku/issues/543#issuecomment-1965866167

If you miss a promise, then there's an undefined behavior I guess (I cannot find the document about what if)


himself65:
Most of catch content auto-complete by AI

PR #545, opened 2024-02-27
step 1 for middleware architecture

dai-shi:
- [x] move old middleware to old-wrappers
- [x] new middleware and types
- [x] prd only rsc middleware
- [x] prd only ssr middleware


packages/waku/src/cli.ts
```diff
@@ -131,15 +132,27 @@ async function runStart(options: { ssr: boolean }) {
...
-  app.use(
-    '*',
-    honoPrdMiddleware({
-      ...options,
-      config,
-      loadEntries,
-      env: process.env as any,
-    }),
-  );
+  if (process.env.WAKU_WIP_MIDDLEWARE) {
```
dai-shi, line 135: It only works behind this flag.

PR #548, opened 2024-02-28
chore: pretty react tweet vite config 

Aslemammad:
https://github.com/dai-shi/waku/pull/541#discussion_r1503587633

### DIFF
--- a/examples/14_react-tweet/vite.config.ts
+++ b/examples/14_react-tweet/vite.config.ts
@@ -4,7 +4,6 @@ export default ({ mode }: { mode: string }) => {
     return {
       optimizeDeps: {
         include: [
-          'react-dom/client',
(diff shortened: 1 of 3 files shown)


examples/14_react-tweet/vite.config.ts
```diff
@@ -1,15 +1,18 @@
...
-      noExternal: ['react-tweet'],
-    },
-  }),
-});
+export default ({ mode }: { mode: string }) => {
+  if (mode === 'development') {
+    return {
+      optimizeDeps: {
+        include: [
+          'react-dom/client',
```
dai-shi, line 7: btw, I added this because I saw some warnings.
But, because this should be added by react plugin already, should we remove it?

Aslemammad: Yea, removing it also worked. done.

PR #456, opened 2024-02-06
reproduction of #452 in examples/09

dai-shi:
#452

PR #549, opened 2024-02-28
fix: client build css

### LINKED ISSUE #452: ssr build / use client / broken global CSS import from package
**Problem:**
The **imported CSS from a package** or **normal CSS file** is in the production build **but not loaded** in the browser.

**Example:**
```js
'use client';
import "@measured/puck/puck.css";
```
or
```js
'use client';
import "./puck.css";
```

* `pnpm waku dev` 👍 - CSS is loaded
* `pnpm waku dev --with-ssr` 👍  - CSS is loaded
* `pnpm waku build --with-ssr` 👍 
* `pnpm waku start --with-ssr` 👎  - **imported CSS is missing**

**No problem**, if the import is in a **server component** above the client component in the tree!

Aslemammad:
Resolves #452 

For Server build (not ssr), the determined css was being injected using cssAssets into the html when building the ssr/client output. That's why rsc css used to work in build.

The issue is when the client output was emitted, the css was generated too. But since vite does not find that css being generated by the index.html/main.tsx, rather by other entries, it didn't include it in the final index.html output.

The solution is to get that generated css and inject it manually into the final html file.

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -400,7 +400,7 @@ export const extractNonJsAssets = (buildOutput: RollupOutput) =>
     type === 'asset' && !fileName.endsWith('.js') ? [fileName] : [],
   );
 
-export const extracCssAssets = (assets: string[]) =>
+const extracCssAssets = (assets: string[]) =>
   assets.filter((asset) => asset.endsWith('.css'));
 
 const emitRscFiles = async (
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -395,7 +395,7 @@ const buildClientBundle = async (
   return clientBuildOutput;
 };
 
-export const extractNonJsAssets = (buildOutput: RollupOutput) =>
+const extractNonJsAssets = (buildOutput: RollupOutput) =>
   buildOutput.output.flatMap(({ type, fileName }) =>
     type === 'asset' && !fileName.endsWith('.js') ? [fileName] : [],
   );
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -4,7 +4,7 @@ import { pipeline } from 'node:stream/promises';
 
 import { build as buildVite, resolveConfig as resolveViteConfig } from 'vite';
 import viteReact from '@vitejs/plugin-react';
-import type { RollupLog, LoggingFunction } from 'rollup';
+import type { RollupLog, LoggingFunction, RollupOutput } from 'rollup';
 
 import type { Config } from '../../config.js';
 import type { EntriesPrd } from '../../server.js';
@@ -285,13 +285,14 @@ const buildSsrBundle = async (
   rootDir: string,
   config: ResolvedConfig,
   clientEntryFiles: Record<string, string>,
+  clientBuildCss: string[],
   serverBuildOutput: Awaited<ReturnType<typeof buildServerBundle>>,
   isNodeCompatible: boolean,
 ) => {
   const mainJsFile = joinPath(rootDir, config.srcDir, config.mainJs);
   const cssAssets = serverBuildOutput.output.flatMap(({ type, fileName }) =>
     type === 'asset' && fileName.endsWith('.css') ? [fileName] : [],
-  );
+  ).concat(clientBuildCss);
   await buildVite({
     base: config.basePath,
     plugins: [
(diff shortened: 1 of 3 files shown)


dai-shi (commented):
Do you mean #452 issue has already been resolved with previous changes?

Please check prettier format error.

dai-shi:
btw, I remember I did the similar one in #456

packages/waku/src/lib/builder/build.ts
```diff
@@ -394,6 +395,14 @@ const buildClientBundle = async (
   return clientBuildOutput;
 };
 
+export const extractNonJsAssets = (buildOutput: RollupOutput) =>
```
dai-shi, line 398: Please don't export.
```suggestion
const extractNonJsAssets = (buildOutput: RollupOutput) =>
```

packages/waku/src/lib/builder/build.ts
```diff
@@ -394,6 +395,14 @@ const buildClientBundle = async (
...
+  buildOutput.output.flatMap(({ type, fileName }) =>
+    type === 'asset' && !fileName.endsWith('.js') ? [fileName] : [],
+  );
+
+export const extracCssAssets = (assets: string[]) =>
```
dai-shi, line 403:
```suggestion
const extracCssAssets = (assets: string[]) =>
```

packages/waku/src/lib/builder/build.ts
```diff
@@ -474,18 +483,24 @@ const emitHtmlFiles = async (
...
     config.publicDir,
     config.indexHtml,
   );
-  const publicIndexHtml = await readFile(publicIndexHtmlFile, {
+  let publicIndexHtml = await readFile(publicIndexHtmlFile, {
     encoding: 'utf8',
   });
+  publicIndexHtml = publicIndexHtml.replace(/<\/head>/, `${css}</head>`);
+  await writeFile(publicIndexHtmlFile, publicIndexHtml);
+
```
dai-shi, line 503: This looks really bad. Too hacky. It should already be modified in advance. Can you do things in `rscIndexPlugin`?

packages/waku/src/lib/builder/build.ts
```diff
@@ -518,12 +533,10 @@ const emitHtmlFiles = async (
             moduleIdsForPrefetch,
           ) + (customCode || '');
         if (code) {
+          const str = `<script type="module" async>${code}</script>`;
```
dai-shi, line 536: `str` is too naive naming. `scriptTagStr`?

dai-shi (changes requested):
Please reconsider to do tasks in vite plugins.

packages/waku/src/lib/builder/build.ts
Aslemammad: Tried this, but could not, rollup should finish analyzing and creating an output. Then we'd know if there was any CSS emitted. What do you think? unless we create another build just for analyzing (which is too much as it's a complete build in itself)

packages/waku/src/lib/builder/build.ts
Aslemammad: yea thought the same

packages/waku/src/lib/builder/build.ts
dai-shi: Can we do it in rscAnalyzePlugin?

Aslemammad: let me see

Aslemammad: No:
1- We haven't found the clientEntries even yet. That's the plugin's job to do so.
2- We stop walking the client files anyway.

dai-shi: In the future, we need to 2-pass analyze anyway. So, let's wait this PR until then, or we do the 2-pass analyze in advance (in a different PR).

dai-shi (changes requested):
https://github.com/dai-shi/waku/pull/549#discussion_r1510104912

packages/waku/src/lib/builder/build.ts
Aslemammad: I'll do it in another PR after #552

dai-shi:
@Aslemammad How does #658 relate to this and #452?

dai-shi:
https://github.com/dai-shi/waku/pull/658#discussion_r1562005260 Yeah, I had the same comment. 😄

Aslemammad:
haha yea, we need to do the 2 pass thing as mentioned!

PR #551, opened 2024-02-29
fix: Do not process empty search params to avoid redundant question mark

MasterLambaster:
Link is always appends `?` to the end of the url even if there are no search params.

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -223,7 +223,7 @@ function InnerRouter() {
       if (typeof path === 'string') {
         url.pathname = path;
       }
-      if (searchParams) {
+      if (searchParams?.size) {
         url.search = '?' + searchParams.toString();
       }
       if (typeof hash === 'string') {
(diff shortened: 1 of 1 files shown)


packages/waku/src/router/client.ts
```diff
@@ -223,7 +223,7 @@ function InnerRouter() {
       if (typeof path === 'string') {
         url.pathname = path;
       }
-      if (searchParams) {
+      if (searchParams?.size) {
         url.search = '?' + searchParams.toString();
```
dai-shi, line 227: Nice catch. That looks like a good change.

But, if you are familiar with `url.search`, do you know if it's common or recommended to add a search string without prepending `?`?

Here's an idea:
```suggestion
      if (searchParams) {
        url.search = searchParams.toString();
```

MasterLambaster: Yes, that's a better one, updated

dai-shi: thanks.

dai-shi (approved):

PR #552, opened 2024-02-29
fix: avoid double module hazard 

### LINKED ISSUE #450: Refactor to avoid dual module hacks

Aslemammad:
Resolves #450

We warm up the vite module graph before serving the browser and the rsc tree, so at least for the initial module graph (main.tsx and its deps) we serve unified module paths (urls). 

This way, we avoid the double module hazard hack we had before.

Later, we need also to do this for each request, rather than only for the initial request, so we avoid hmr main.tsx potential bugs.

### DIFF
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -153,9 +153,9 @@ export const devServer: Middleware = (options) => {
 
   const willBeHandledLater = async (pathname: string) => {
     const vite = await vitePromise;
-    if (vite.moduleGraph.urlToModuleMap.has(pathname)) {
-      return true;
-    }
+    // if (vite.moduleGraph.urlToModuleMap.has(pathname)) {
+    //   return true;
+    // }
     try {
       const result = await vite.transformRequest(pathname);
       return !!result;
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -169,6 +169,7 @@ export const devServer: Middleware = (options) => {
     // and using the same modules if possible in the bundlerConfig in the stream
     const mainJs = `${config.srcDir}/${config.mainJs}`;
     const resolved = await vite.pluginContainer.resolveId(mainJs);
+    console.log('resolved', resolved)
     await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
     await Promise.all(
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -167,7 +167,6 @@ export const devServer: Middleware = (options) => {
     // and using the same modules if possible in the bundlerConfig in the stream
     const mainJs = `${config.srcDir}/${config.mainJs}`;
     const resolved = await vite.pluginContainer.resolveId(mainJs);
-    console.log('resolved', mainJs, resolved, vite.moduleGraph.idToModuleMap.keys())
     await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
     await Promise.all(
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -168,9 +168,10 @@ export const devServer: Middleware = (options) => {
     // pre-process the mainJs file to see which modules are being sent to the browser by vite
     // and using the same modules if possible in the bundlerConfig in the stream
     const mainJs = `${config.srcDir}/${config.mainJs}`;
+    
+    await vite.transformRequest(mainJs);
     const resolved = await vite.pluginContainer.resolveId(mainJs);
     console.log('resolved', resolved)
-    await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
     await Promise.all(
       [...resolvedModule.importedModules].map(({ id }) =>
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -171,11 +171,11 @@ export const devServer: Middleware = (options) => {
     const resolved = await vite.pluginContainer.resolveId(mainJs);
     await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
-    for (const { id } of resolvedModule.importedModules) {
-      if (id) {
-        await vite.warmupRequest(id);
-      }
-    }
+    await Promise.all(
+      [...resolvedModule.importedModules].map(({ id }) =>
+        id ? vite.warmupRequest(id) : null,
+      ),
+    );
 
     if (!initialModules) {
       initialModules = Array.from(vite.moduleGraph.idToModuleMap.values()).map(
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -167,7 +167,7 @@ export const devServer: Middleware = (options) => {
     // and using the same modules if possible in the bundlerConfig in the stream
     const mainJs = `${config.srcDir}/${config.mainJs}`;
     const resolved = await vite.pluginContainer.resolveId(mainJs);
-    console.log('resolved', resolved)
+    console.log('resolved', mainJs, resolved, vite.moduleGraph.idToModuleMap.keys())
     await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
     await Promise.all(
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -59,8 +59,6 @@ const createStreamPair = (): [Writable, Promise<ReadableStream | null>] => {
   return [writable, promise];
 };
 
-const externalizedWakuClientModules = ['waku/client', 'waku/router/client'];
-
 export const devServer: Middleware = (options) => {
   if (options.cmd !== 'dev') {
     // pass through if not dev command
@@ -89,14 +87,6 @@ export const devServer: Middleware = (options) => {
         exclude: ['waku'],
         entries: [
           `${config.srcDir}/${config.entriesJs}`.replace(/\.js$/, '.*'),
-        ],
-      },
-      ssr: {
-        external: [
-          'waku',
-          'waku/server',
-          'waku/router/server',
-          ...externalizedWakuClientModules,
         ],
       },
       server: { middlewareMode: true },
@@ -166,16 +156,18 @@ export const devServer: Middleware = (options) => {
       vitePromise,
     ]);
 
-    const mainJs = `${config.srcDir}/${config.mainJs}`;
     // pre-process the mainJs file to see which modules are being sent to the browser by vite
     // and using the same modules if possible in the bundlerConfig in the stream
-    for (const m of [mainJs, ...externalizedWakuClientModules]) {
-      const resolved = await vite.pluginContainer.resolveId(m);
-      if (resolved?.id) {
-        await vite.warmupRequest(resolved.id);
+    const mainJs = `${config.srcDir}/${config.mainJs}`;
+    const resolved = await vite.pluginContainer.resolveId(mainJs);
+    await vite.transformRequest(resolved!.id);
+    const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)! 
+    for (const { id } of resolvedModule.importedModules) {
+      if (id) {
+        await vite.warmupRequest(id);
       }
     }
-
+    
     if (!initialModules) {
       initialModules = Array.from(vite.moduleGraph.idToModuleMap.values()).map(
         (m) => ({ url: m.url, file: m.file! }),
@@ -184,7 +176,7 @@ export const devServer: Middleware = (options) => {
 
     ctx.devServer = {
       rootDir: vite.config.root,
-      initialModules: initialModules,
+      initialModules,
       renderRscWithWorker,
       getSsrConfigWithWorker,
       loadServerFile,
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -157,8 +157,6 @@ export const devServer: Middleware = (options) => {
     }
   };
 
-  let initialModules: ClonableModuleNode[];
-
   return async (ctx, next) => {
     const [{ middleware: _removed, ...config }, vite] = await Promise.all([
       configPromise,
@@ -177,11 +175,9 @@ export const devServer: Middleware = (options) => {
       ),
     );
 
-    if (!initialModules) {
-      initialModules = Array.from(vite.moduleGraph.idToModuleMap.values()).map(
-        (m) => ({ url: m.url, file: m.file! }),
-      );
-    }
+    const initialModules: ClonableModuleNode[] = Array.from(
+      vite.moduleGraph.idToModuleMap.values(),
+    ).map((m) => ({ url: m.url, file: m.file! }));
 
     ctx.devServer = {
       rootDir: vite.config.root,
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -167,6 +167,7 @@ export const devServer: Middleware = (options) => {
     // and using the same modules if possible in the bundlerConfig in the stream
     const mainJs = `${config.srcDir}/${config.mainJs}`;
     const resolved = await vite.pluginContainer.resolveId(mainJs);
+    console.log('resolved', resolved)
     await vite.transformRequest(resolved!.id);
     const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
     await Promise.all(
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -169,6 +169,7 @@ export const devServer: Middleware = (options) => {
     if (!initialModules) {
       // pre-process the mainJs file to see which modules are being sent to the browser by vite
       // and using the same modules if possible in the bundlerConfig in the stream
+
       const mainJs = `${config.basePath}${config.srcDir}/${config.mainJs}`;
       await vite.transformRequest(mainJs);
       const resolved = await vite.pluginContainer.resolveId(mainJs);
--- a/packages/waku/src/lib/middleware/dev-server.ts
+++ b/packages/waku/src/lib/middleware/dev-server.ts
@@ -158,7 +158,7 @@ export const devServer: Middleware = (options) => {
     }
   };
 
-  let initialModuleGraph: ClonableModuleNode[];
+  let initialModules: ClonableModuleNode[];
 
   return async (ctx, next) => {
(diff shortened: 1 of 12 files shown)


Aslemammad:
not ready for review yet, just wanted to debug ci.

dai-shi:
> not ready for review yet, just wanted to debug ci.

does draft not run the ci?

Aslemammad:
Not at that moment. I don't know why, but I'll get it back to draft if I
don't resolve it right now.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -42,7 +42,10 @@ export const rsc: Middleware = (options) => {
           contentType: headers['content-type'] || '',
         };
         const readable = await (ctx.devServer
-          ? ctx.devServer.renderRscWithWorker(args)
+          ? ctx.devServer.renderRscWithWorker({
+              ...args,
+              initialModuleGraph: ctx.devServer.initialModuleGraph,
+            })
```
dai-shi, line 48: We should make `args` for both for DEV and PRD:
```suggestion
          ? ctx.devServer.renderRscWithWorker(args, {
              initialModuleGraph: ctx.devServer.initialModuleGraph,
            })
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -19,6 +19,7 @@ import {
   fileURLToFilePath,
 } from '../utils/path.js';
 import { encodeInput, hasStatusCode } from './utils.js';
+import path from 'path';
```
dai-shi, line 22:
```diff
- import path from 'path';
+ import path from 'node:path';
```
But, this file should be independent from Node.js.
That's why we made `utils/path.ts`.

packages/waku/src/lib/middleware/types.ts
```diff
@@ -1,10 +1,13 @@
+import type { ModuleNode } from 'vite';
 import type { Config } from '../../config.js';
 import type { EntriesPrd } from '../../server.js';
 import type {
   renderRscWithWorker,
   getSsrConfigWithWorker,
 } from '../renderers/dev-worker-api.js';
 
+export type ClonableModuleNode = Pick<ModuleNode, 'url' | 'file'>;
```
dai-shi, line 9: Can you define type simply without Pick? Seems like the values are just strings.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -64,6 +59,8 @@ const createStreamPair = (): [Writable, Promise<ReadableStream | null>] => {
   return [writable, promise];
 };
 
+const externalizedWakuClientModules = ['waku/client', 'waku/router/client'];
```
dai-shi, line 62: I wonder if this limits third-party libraries.
What would happen without `waku/router/client` here?

packages/waku/src/lib/middleware/types.ts
```diff
@@ -26,6 +29,7 @@ export type HandlerContext = {
   readonly context: RscContext;
   devServer?: {
     rootDir: string;
+    initialModuleGraph: ClonableModuleNode[];
```
dai-shi, line 32: I wouldn't call it "Graph".

dai-shi (changes requested):
Thanks for working on it. Not exactly sure how it works/solves, but I think I get the flow.

Can you merge `main` branch too?

packages/waku/src/lib/middleware/dev-server.ts
dai-shi: This is really bad, I guess. We shouldn't depend on `waku/router` like this. Otherwise, other router integration like tanstack/router wouldn't be possible.
I guess the solution is to find a dependency from `main.tsx`.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -96,10 +94,9 @@ export const devServer: Middleware = (options) => {
       ssr: {
         external: [
           'waku',
-          'waku/client',
           'waku/server',
-          'waku/router/client',
           'waku/router/server',
+          ...externalizedWakuClientModules,
```
dai-shi, line 99: This is for different purpose (just for out monorepo setup), so we can revert.

packages/waku/src/lib/middleware/dev-server.ts
dai-shi: Even for `waku/client`, we know the dependency from `main.tsx`, no?

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -161,13 +158,33 @@ export const devServer: Middleware = (options) => {
...
   return async (ctx, next) => {
     const [{ middleware: _removed, ...config }, vite] = await Promise.all([
       configPromise,
       vitePromise,
     ]);
+
+    const mainJs = `${config.srcDir}/${config.mainJs}`;
+    // pre-process the mainJs file to see which modules are being sent to the browser by vite
+    // and using the same modules if possible in the bundlerConfig in the stream
+    for (const m of [mainJs, ...externalizedWakuClientModules]) {
```
dai-shi, line 172: So, start with `mainJs` and warm up all dependency modules in the graph.

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -42,7 +42,10 @@ export const rsc: Middleware = (options) => {
           contentType: headers['content-type'] || '',
         };
         const readable = await (ctx.devServer
-          ? ctx.devServer.renderRscWithWorker(args)
+          ? ctx.devServer.renderRscWithWorker({
+              ...args,
+              initialModules: ctx.devServer.initialModules,
+            })
```
dai-shi, line 48: Let's change the signature:
```suggestion
          ? ctx.devServer.renderRscWithWorker(args, {
              initialModules: ctx.devServer.initialModules,
            })
```

packages/waku/src/lib/middleware/ssr.ts
```diff
@@ -61,6 +61,7 @@ export const ssr: Middleware = (options) => {
                     config,
                     pathname,
                     searchParams,
+                    initialModules: devServer.initialModules,
```
dai-shi, line 64: This too:
```suggestion
                    }, {initialModules: devServer.initialModules,
```

packages/waku/src/lib/renderers/dev-worker-api.ts
```diff
@@ -125,7 +128,7 @@ export function registerHotUpdateCallback(
 let nextId = 1;
 
 export async function renderRscWithWorker(
-  args: RenderRscArgs,
+  args: RenderRscArgs & { initialModules: ClonableModuleNode[] },
```
dai-shi, line 131:
```suggestion
  args: RenderRscArgs,
  opts: { initialModules: ClonableModuleNode[] },
```

packages/waku/src/lib/renderers/dev-worker-api.ts
```diff
@@ -182,7 +186,9 @@ export async function renderRscWithWorker(
   });
 }
 
-export async function getSsrConfigWithWorker(args: GetSsrConfigArgs): Promise<{
+export async function getSsrConfigWithWorker(
+  args: GetSsrConfigArgs & { initialModules: ClonableModuleNode[] },
```
dai-shi, line 190:
```suggestion
  args: GetSsrConfigArgs,
  opts: { initialModules: ClonableModuleNode[] },
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -259,12 +259,14 @@ export const renderHtml = async (
           {},
           {
             get(_target, name: string) {
-              const file = filePath.slice(config.basePath.length);
+              const file =
+                '/' + filePath.slice(config.basePath.length).split('?')[0]!;
```
dai-shi, line 263: it smells something.

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -97,7 +99,14 @@ export async function renderRsc(
...
         const [file, name] = encodedId.split('#') as [string, string];
-        const id = resolveClientEntry(file, config);
+        let id = resolveClientEntry(file, config);
+        if (isDev) {
+          for (const moduleNode of opts.initialModules) {
+            if (moduleNode.file === file) {
+              id = moduleNode.url;
+            }
+          }
+        }
```
dai-shi, line 109: It's better to add DEV-only code here. Can you patch `resolveClientEntry` in the caller?

dai-shi (changes requested):
Please see comments.

packages/waku/src/lib/renderers/html-renderer.ts
Aslemammad: what's that smell?

dai-shi: that it doesn't work on edge cases, might be the case with windows. not sure.
why do we need to add `/` prefix?

Aslemammad: to make it a vite path. let's remove it, even before that, it was failing, so this has no effect.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -161,13 +157,35 @@ export const devServer: Middleware = (options) => {
...
+    // and using the same modules if possible in the bundlerConfig in the stream
+    const mainJs = `${config.srcDir}/${config.mainJs}`;
+    const resolved = await vite.pluginContainer.resolveId(mainJs);
+    await vite.transformRequest(resolved!.id);
+    const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
+    for (const { id } of resolvedModule.importedModules) {
+      if (id) {
+        await vite.warmupRequest(id);
+      }
+    }
```
dai-shi, line 178: Shouldn't we use `Promise.all`?

packages/waku/src/lib/middleware/rsc.ts
```diff
@@ -42,7 +42,14 @@ export const rsc: Middleware = (options) => {
           contentType: headers['content-type'] || '',
         };
         const readable = await (ctx.devServer
-          ? ctx.devServer.renderRscWithWorker(args)
+          ? ctx.devServer.renderRscWithWorker(
+              {
+                ...args,
+              },
```
dai-shi, line 48:
```suggestion
              args,
```

packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -49,7 +49,13 @@ const resolveClientEntryForDev = (id: string, config: { basePath: string }) => {
 };
 
 const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
-  const { id, type: _removed, hasModuleIdCallback, ...rest } = mesg;
+  const {
+    id,
+    type: _removed,
+    hasModuleIdCallback,
+    initialModules: initialModules,
```
dai-shi, line 56:
```suggestion
    initialModules,
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -259,7 +259,9 @@ export const renderHtml = async (
           {},
           {
             get(_target, name: string) {
-              const file = filePath.slice(config.basePath.length);
+              const file = filePath
+                .slice(config.basePath.length)
+                .split('?')[0]!;
```
dai-shi, line 264: Could removing after `?` be DEV-only if it's not required for PRD?

dai-shi (commented):
The diff looks good and clean. Much more readable.
Can you make a few adjustments?

packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -73,8 +79,15 @@ const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
         isDev: true,
         loadServerFile,
         loadServerModule,
-        resolveClientEntry: (id: string) =>
-          resolveClientEntryForDev(id, rest.config),
+        resolveClientEntry: (file: string) => {
+          let id = resolveClientEntryForDev(file, rest.config);
+          for (const moduleNode of initialModules) {
+            if (moduleNode.file === file) {
+              id = moduleNode.url;
```
dai-shi, line 86:
```suggestion
              id = moduleNode.url;
              break;
```

dai-shi: I think you probably have, but have you tried adding console.log here?
```suggestion
              console.log('replacing id', id, moduleNode.url);
              id = moduleNode.url;
              break;
```

Does it make any changes between ubuntu and windows?

Aslemammad: yea tried, let me try again

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -161,13 +157,35 @@ export const devServer: Middleware = (options) => {
...
+    const resolved = await vite.pluginContainer.resolveId(mainJs);
+    await vite.transformRequest(resolved!.id);
+    const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
+    await Promise.all(
+      [...resolvedModule.importedModules].map(({ id }) =>
+        id ? vite.warmupRequest(id) : null,
+      ),
+    );
+
+    if (!initialModules) {
```
dai-shi, line 180: Do we need this `if` condition? Can we or should we compute initialModules for each request? Otherwise, should we move this before L170? Or, can we even move this logic outside of this function? `vitePromise.then(...)`

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -259,9 +259,11 @@ export const renderHtml = async (
           {},
           {
             get(_target, name: string) {
-              const file = filePath.slice(config.basePath.length);
+              console.log('filePath', filePath);
```
dai-shi, line 262: Let's log `wakuDist` in L270 and see if L274 check works the same between ubuntu and windows.

packages/waku/src/lib/middleware/dev-server.ts
Aslemammad: I remember that calling in each request gave an issue that I don't remember, but let me try it again.

packages/waku/src/lib/renderers/html-renderer.ts
Aslemammad: sure, i'll try that soon.

packages/waku/src/lib/middleware/dev-server.ts
Aslemammad: The whole problem lies down to the fact that in the initial load, vite has not decided what to do for modules yet, and when things happen in parallel, it counts them as two different modules even if they are the same files.

But I wonder if that's going to cause an issue for HMR in the mainJS file.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -74,6 +69,7 @@ export const devServer: Middleware = (options) => {
   const configPromise = resolveConfig(options.config);
   const vitePromise = configPromise.then(async (config) => {
     const mergedViteConfig = await mergeUserViteConfig({
+      cacheDir: 'node_modules/.vite/waku-dev-server',
```
Aslemammad, line 72: Since we have multiple instances of vite, different ones might overwrite the others' cache. That's why we change it for this one.

packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -73,8 +89,8 @@ const handleRender = async (mesg: MessageReq & { type: 'render' }) => {
         isDev: true,
         loadServerFile,
         loadServerModule,
-        resolveClientEntry: (id: string) =>
-          resolveClientEntryForDev(id, rest.config),
+        resolveClientEntry: (file: string) =>
+          resolveClientEntryForDev(file, rest.config, initialModules),
```
dai-shi, line 93: If renaming to `file` is intentional,

packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -42,14 +43,29 @@ const configSrcDir = getEnvironmentData('CONFIG_SRC_DIR') as string;
 const configEntriesJs = getEnvironmentData('CONFIG_ENTRIES_JS') as string;
 const configPrivateDir = getEnvironmentData('CONFIG_PRIVATE_DIR') as string;
 
-const resolveClientEntryForDev = (id: string, config: { basePath: string }) => {
+const resolveClientEntryForDev = (
+  id: string,
```
dai-shi, line 47: this should be `file` too?

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -259,12 +260,13 @@ export const renderHtml = async (
           {},
           {
             get(_target, name: string) {
-              const file = filePath.slice(config.basePath.length);
+              let file = filePath.slice(config.basePath.length);
               // TODO too long, we need to refactor this logic
               if (isDev) {
+                file = file.split('?')[0]!;
```
dai-shi, line 266: Is this needed for both cases, a) startsWith('@fs/') and b) not?

dai-shi (commented):
looking good.

packages/waku/src/lib/middleware/dev-server.ts
dai-shi: Can you make it an inline comment in the code?

packages/waku/src/lib/renderers/dev-worker-impl.ts
Aslemammad: changed it for debugging, now i reverted it.

packages/waku/src/lib/renderers/dev-worker-impl.ts
Aslemammad: no need.

packages/waku/src/lib/renderers/html-renderer.ts
Aslemammad: rewrote it. It's for b only.

packages/waku/src/lib/middleware/dev-server.ts
Aslemammad: done

packages/waku/src/lib/middleware/dev-server.ts
dai-shi: Can we move this before L169?

dai-shi (approved):
Looks pretty good. Just one comment left.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -161,13 +158,35 @@ export const devServer: Middleware = (options) => {
...
+      const resolved = await vite.pluginContainer.resolveId(mainJs);
+      const resolvedModule = vite.moduleGraph.idToModuleMap.get(resolved!.id)!;
+      await Promise.all(
+        [...resolvedModule.importedModules].map(({ id }) =>
+          id ? vite.warmupRequest(id) : null,
+        ),
+      );
+
+      initialModules = Array.from(vite.moduleGraph.idToModuleMap.values()).map(
+        (m) => ({ url: m.url, file: m.file! }),
```
dai-shi, line 183: > `m.file!`

Is there a case when `m.file` is not defined? Just curious.

Aslemammad: I didn't find one!

dai-shi:
There might be some issues in some edge cases, but hope @Aslemammad can fix it soon in that case. Let's merge it.

Issue #554, opened 2024-03-01
"use server" inside server action doesn't work


himself65:
```
Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
  <... action={function sendSomething} children=...>
              ^^^^^^^^^^^^^^^^^^^^^^^^
```

```diff
Subject: [PATCH] fix: add catch
---
Index: examples/01_template/src/provider.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/01_template/src/provider.tsx b/examples/01_template/src/provider.tsx
new file mode 100644
--- /dev/null	(date 1709261769826)
+++ b/examples/01_template/src/provider.tsx	(date 1709261769826)
@@ -0,0 +1,17 @@
+'use client';
+import { createContext } from 'react';
+
+const AIContext = createContext<any>(null);
+
+export function Provider (
+  props: {
+    action: any
+    children: React.ReactNode
+  }
+) {
+  return (
+    <AIContext.Provider value={props.action}>
+      {props.children}
+    </AIContext.Provider>
+  );
+}
Index: examples/01_template/src/actions/index.ts
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/01_template/src/actions/index.ts b/examples/01_template/src/actions/index.ts
new file mode 100644
--- /dev/null	(date 1709261691100)
+++ b/examples/01_template/src/actions/index.ts	(date 1709261691100)
@@ -0,0 +1,5 @@
+export function sendSomething () {
+  'use server';
+  console.log('hello world!');
+  return 'hello world!';
+}
Index: examples/01_template/src/templates/root-layout.tsx
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/examples/01_template/src/templates/root-layout.tsx b/examples/01_template/src/templates/root-layout.tsx
--- a/examples/01_template/src/templates/root-layout.tsx	(revision 47846fe2b5ecdb9aa354f9725610aa2312cb98ba)
+++ b/examples/01_template/src/templates/root-layout.tsx	(date 1709261802086)
@@ -4,6 +4,8 @@
 
 import { Header } from '../components/header.js';
[...]

PR #555, opened 2024-03-01
fix: server action

himself65:
Fixes: https://github.com/dai-shi/waku/issues/554


dai-shi:
It's not that easy, we need to split server code and client code.
For now, the boundary is file based, so this PR alone doesn't work.

As noted in #554, our first focus is importing server files from client files. Please be patient.

Issue #455, opened 2024-02-06
Update @vanilla-extract/vite-plugin v4 in examples/12_css


dai-shi:
https://github.com/dai-shi/waku/pull/454/commits/454e77f820cc56836bfd6d7447fe701aa91a9c28  reverted because it doesn't pass CI.

dai-shi:
No rush, but @himself65 can you take this?

himself65:
OK

PR #559, opened 2024-03-02
chore: bump 12_css example

himself65:
I think 0.19.4 fix many compatibility issue

Fixes: https://github.com/dai-shi/waku/issues/455


himself65:
This is a plugin issue, working on it

PR #564, opened 2024-03-04
fix: rsc-transform-plugin logic

himself65:
- Use the SWC plugin to fix directive
- Remove redundant regex code using `resolveId`


packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -32,66 +35,49 @@ export function rscTransformPlugin(
...
+        //  to get the source code.
+        let source = code;
+        if (/\.[jt]sx?$/.test(id)) {
+          source = (
+            await transform(source, {
+              swcrc: false,
+              jsc: {
+                experimental: {
+                  disableBuiltinTransformsForInternalTesting: true,
+                  plugins: [[require.resolve('swc-plugin-react-server'), {}]],
```
himself65, line 67: https://github.com/swc-project/swc/issues/8688#event-11993095326

PR #567, opened 2024-03-04
fix: stop sending ssr hydrate flag when the `--with-ssr` flag is absent

Aslemammad:
While working on #552 

I found out that the ci was giving hydration error while the withSSR was false. The reason was that `data-hydrate` was always true.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -77,7 +77,7 @@ if (values.version) {
       await runDev({ ssr });
       break;
     case 'build':
-      await runBuild();
+      await runBuild({ ssr });
       break;
     case 'start':
       await runStart({ ssr });
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -95,7 +95,7 @@ async function runDev(options: { ssr: boolean }) {
   const config = await loadConfig();
   const app = new Hono();
   if (!process.env.WAKU_OLD_MIDDLEWARE) {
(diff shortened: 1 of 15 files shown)


dai-shi:
> The reason was that `data-hydrate` was always true.

That's kind of intentional. Can you help me understand the problem? How can I reproduce the bug?

Aslemammad:
> That's kind of intentional. Can you help me understand the problem? How can I reproduce the bug?

I see, well, in #552 the error is hydration error while withSSR is false. Another way to reproduce it is to put a console log in one main.tsx files, in the first condition (when the dataset.hydrate is true). And when running the project, remove the --with-ssr flag. At that point, we still hydrate and it will show the log.

If it's intentional, then for the 07_router ci test we need to consider another solution or remove the test completely since as it was commented there, the test was not supposed to work with ssr in standalone mode.

dai-shi:
> for the 07_router ci test we need to consider another solution

After #565 is merged, is the test breaking? sorry, if I miss some points.

Aslemammad:
> After https://github.com/dai-shi/waku/pull/565 is merged, is the test breaking? sorry, if I miss some points.

I haven't tried that PR yet, so let me try it and let you know.

dai-shi (changes requested):
> the test was not supposed to work with ssr in standalone mode.

let's remove it.

dai-shi:
Probably easier to create a new PR.

PR #568, opened 2024-03-05
feat: `getContext` & `rerender`

dai-shi:
Avoid `this`.

- [x] impl
- [x] refactor
- [x] modify ssr.ts to match the coding style

### DIFF
--- a/packages/waku/src/lib/renderers/rsc-renderer.ts
+++ b/packages/waku/src/lib/renderers/rsc-renderer.ts
@@ -1,15 +1,29 @@
-import type { ReactNode } from 'react';
+import type {
+  default as ReactType,
+  ComponentProps,
+  FunctionComponent,
+  ReactNode,
+} from 'react';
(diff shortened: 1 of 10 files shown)


packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -62,23 +72,92 @@ export async function renderRsc(
...
+    loadServerModule<{ default: typeof RSDWServerType }>('rsdw-server'),
+    loadServerModule<{ default: typeof RSDWClientType }>('rsdw-client'),
+    (isDev
+      ? opts.loadServerModule(SERVER_MODULE_MAP['waku-server'])
+      : loadModule!('waku-server')) as Promise<{
+      setRenderContext: typeof setRenderContextType;
+    }>,
+  ]);
+
+  const runWithRenderContext = async <T>(
```
dai-shi, line 103: I'm not super confident if this won't cause any unexpected behavior.

PR #571, opened 2024-03-06
fix: use web standard crypto api

himself65:


packages/waku/src/lib/builder/build.ts
```diff
@@ -1,46 +1,44 @@
-import { createHash } from 'node:crypto';
 import { Readable } from 'node:stream';
```
himself65, line 1: IDK why I ran `prettier --write .` and order totally changed

packages/waku/src/lib/builder/build.ts
```diff
@@ -76,17 +74,16 @@ const onwarn = (warning: RollupLog, defaultHandler: LoggingFunction) => {
...
-    const sha256 = createHash('sha256');
-    sha256.on('readable', () => {
-      const data = sha256.read();
-      if (data) {
-        resolve(data.toString('hex').slice(0, 9));
-      }
-    });
-    createReadStream(fname).pipe(sha256);
-  });
+const hash = async (fname: string): Promise<string> => {
```
himself65, line 77: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string

packages/waku/src/lib/builder/build.ts
dai-shi: You are probably using a different version or a different config.

packages/waku/src/lib/builder/build.ts
```diff
@@ -76,17 +74,16 @@ const onwarn = (warning: RollupLog, defaultHandler: LoggingFunction) => {
...
+  const encoder = new TextEncoder();
+  const data = encoder.encode(fname);
```
dai-shi, line 79: This seems wrong. We need to create a hash from the file content, not the file name.

packages/waku/src/lib/builder/build.ts
dai-shi: Hope you figure it out...

dai-shi (changes requested):

dai-shi (approved):
LGTM
Ready to merge?

himself65:
Yes

packages/waku/src/lib/builder/build.ts
```diff
@@ -98,6 +84,7 @@ const analyzeEntries = async (
   );
   const clientFileSet = new Set<string>([wakuClientDist]);
   const serverFileSet = new Set<string>();
+  const fileHashSet = new Map<string, string>();
```
dai-shi, line 87:
```suggestion
  const fileHashMap = new Map<string, string>();
```

dai-shi (changes requested):
No clue for prettier config?

dai-shi (approved):

PR #573, opened 2024-03-06
fix: do not SSR index page

himself65:
Fixes: https://github.com/dai-shi/waku/issues/543

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -9,7 +9,7 @@ import {
   getPathMapping,
 } from '../lib/utils/path.js';
 import type { PathSpec } from '../lib/utils/path.js';
-import type { defineEntries } from 'waku/server';
+import type { defineEntries } from '../server.js';
 
 // createPages API (a wrapper around unstable_defineRouter)
 
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -9,6 +9,7 @@ import {
   getPathMapping,
 } from '../lib/utils/path.js';
 import type { PathSpec } from '../lib/utils/path.js';
+import type { defineEntries } from 'waku/server';
 
 // createPages API (a wrapper around unstable_defineRouter)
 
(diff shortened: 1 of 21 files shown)


himself65 (commented):
oh no, git checkout delete some of my changes

packages/waku/src/router/create-pages.ts
```diff
@@ -9,6 +9,7 @@ import {
   getPathMapping,
 } from '../lib/utils/path.js';
 import type { PathSpec } from '../lib/utils/path.js';
+import type { defineEntries } from 'waku/server';
```
dai-shi, line 12: Prefer relative paths when possible:
```suggestion
import type { defineEntries } from '../server.js';
```

packages/waku/src/router/define-router.ts
```diff
@@ -110,14 +110,16 @@ export function unstable_defineRouter(
...
         continue;
       }
       const pathname = '/' + pathSpec.map(({ name }) => name).join('/');
       const input = getInputString(pathname);
-      const moduleIds = await unstable_collectClientModules(input);
-      path2moduleIds[pathname] = moduleIds;
+      if (isStatic) {
+        const moduleIds = await unstable_collectClientModules(input);
+        path2moduleIds[pathname] = moduleIds;
+      }
```
dai-shi, line 122: Hmm, I think we want to use `path2moduleIds` for dynamic pages too (even though, it's suboptimal).

If `unstable_collectClientModules` is only the place of the problem, it doesn't actually leak the information because it's thrown away.
That said, I wish there would be an alternative way to collect client modules without running the code. Can we statically analyze the code?

dai-shi (changes requested):
https://github.com/dai-shi/waku/pull/573#discussion_r1514091960

packages/waku/src/router/define-router.ts
himself65: Oh, yes we can. I read the React-server source code, we could collect client info during vite build

himself65: my current idea is to add a hook on the resolve module function and give id -> moduleId[] Map in the analysis plugin

dai-shi: that sounds nice. we may need some research, so this would come at v0.21 or later.

dai-shi:
https://github.com/dai-shi/waku/pull/573#discussion_r1514091960

Let's start it over.

PR #575, opened 2024-03-07
fix: fs-router for non-Node envs

dai-shi:
#572 depends on Node.js even in production build, which isn't ideal.

- [x] persist build config in entries
- [x] fix defineRouter
- [x] fix createPages
- [x] fix fsRouter


dai-shi:
That was a huge misunderstanding of mine. The approach I did for the full day doesn't work. Let me start over. 😓 😓 😓

dai-shi:
Hmm, maybe not fully wrong, but I wonder the current approach is ideal. Well, I've been going back and forth and it will happen again...

PR #587, opened 2024-03-10
fix: use AsyncLocalStorage for getContext

### LINKED ISSUE #586: Error: [Bug] No render context found (v0.20.0-alpha.0)
I am getting errors using server actions.
The error occurs when a server action function has `await` code.

```ts
'use server';

import { rerender } from 'waku/server';

const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));

export const actionWithFormData = async (formData: FormData) => {
  await sleep(0); // Getting error with await
  const title = formData.get('title');
  if (title === null) {
    throw new Error('Unexpected null title');
  }
  console.log(title);
  rerender('/');
};

export const actionWithoutFormData = async (title: string) => {
  await sleep(0); // Getting error with await
  console.log(title);
  rerender('/');
};
```

If you remove the await line, no errors will happen.

version: 0.20.0-alpha.0
repro: https://github.com/t6adev/waku-form-action-sample

dai-shi:
close #586 

It was my misunderstanding that `cache` does something magical behind the scene, but it uses ALS under the hood.

### DIFF
--- a/packages/waku/src/server.ts
+++ b/packages/waku/src/server.ts
@@ -1,4 +1,4 @@
-import { cache } from 'react';
+import type { AsyncLocalStorage as AsyncLocalStorageType } from 'node:async_hooks';
 import type { ReactNode } from 'react';
 
 import type { Config } from './config.js';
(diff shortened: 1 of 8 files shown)


packages/waku/src/server.ts
```diff
@@ -68,37 +68,63 @@ export function getEnv(key: string): string | undefined {
...
   rerender: (input: string, searchParams?: URLSearchParams) => void;
   context: RscContext;
 };
 
-const getRenderContextHolder = cache(() => [] as [RenderContext?]);
+const DO_NOT_BUNDLE = '';
+
+let renderStorage: AsyncLocalStorageType<RenderStore> | undefined;
+
+import(/* @vite-ignore */ DO_NOT_BUNDLE + 'node:async_hooks')
```
dai-shi, line 82: Actually, I think it needs to be bundled.

dai-shi: #590

Issue #585, opened 2024-03-08
Top-level await is not available


aheissenberger:
WAKU Commit (Main): f985f59b3e8ccb2a4f5ce1ee5a79afbfd882128f

Works with `pnpm dev` but fails in BUILD mode `waku build` - this file is only imported into a server action and there is no difference if there is an additional `'use server'` on top.

```
pnpm build

> waku-example@0.1.0 build /waku/examples/xx_auth
> waku build

vite v5.1.5 building SSR bundle for production...
src/func.ts (1:0) Error when using sourcemap for reporting an error: Can't resolve original location of error.
src/lib/db-lowdb.ts (1:0) Error when using sourcemap for reporting an error: Can't resolve original location of error.
✓ 71 modules transformed.
x Build failed in 341ms
/waku/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1651
  let error = new Error(text);
              ^

Error: [vite:esbuild-transpile] Transform failed with 2 errors:
assets/func-!~{005}~.js:571:16: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
assets/func-!~{005}~.js:572:19: ERROR: Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)

Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
569|  }
570|  
571|  const dbUsers = await JSONFilePreset("users.json", []);
   |                  ^
572|  const dbSessions = await JSONFilePreset("sessions.json", []);
573|  

Top-level await is not available in the configured target environment ("chrome87", "edge88", "es2020", "firefox78", "safari14" + 2 overrides)
570|  
571|  const dbUsers = await JSONFilePreset("users.json", []);
572|  const dbSessions = await JSONFilePreset("sessions.json", []);
   |                     ^
573|  
574|  const adapter = new S(dbSessions, dbUsers);

    at failureErrorWithLog (/waku/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:1651:15)
    at /waku/node_modules/.pnpm/esbuild@0.19.12/node_modules/esbuild/lib/main.js:849:29
[...]

PR #590, opened 2024-03-10
fix: React.cache to work correctly if AsyncLocalStorage is available

dai-shi:
Related with #587, `cache` hasn't been working as expected. This is to fix it.


packages/waku/src/lib/builder/build.ts
```diff
@@ -119,6 +119,7 @@ const analyzeEntries = async (
     build: {
       write: false,
       ssr: true,
+      target: 'node18',
```
dai-shi, line 122: This should fix #585 too. cc @Aslemammad

PR #581, opened 2024-03-08
feat(create-waku): `--example` option

ojj1123:
related https://github.com/dai-shi/waku/discussions/315

I wrote the prototype for `--example` option. I've referred to [create-next-app](https://github.com/vercel/next.js/tree/canary/packages/create-next-app)
I will be adding `automatically installing dependencies` in next PR.

## What features were added

```
pnpm create waku --example 14_react-tweet // will download https://github.com/dai-shi/waku/tree/main/examples/14_react-tweet

pnpm create waku --example https://github.com/dai-shi/waku/tree/main/examples/14_react-tweet

pnpm create waku --example https://github.com/himself65/waku-vercel-template
```

also in this PR, i've drop --choose-example as @dai-shi said
https://github.com/dai-shi/waku/discussions/315#discussioncomment-8708156

## What I want to discuss
In create-next-app, [they are dealing with the branch name or file path containing the slash.](https://arc.net/l/quote/dsgpummq) But I've thought this case is a bit rare one so I didn't implement this feature in create-waku. How about this case?

PR #599, opened 2024-03-13
fix(create-waku): do not download from main

dai-shi:
As I said in #581, downloading from `main` will break.

- [x] show help
- [x] download from the released branch
- [ ] refactor

### DIFF
--- a/packages/create-waku/src/helpers/example-option.ts
+++ b/packages/create-waku/src/helpers/example-option.ts
@@ -72,15 +72,16 @@ export function hasRepo({
   return isUrlOk(contentsUrl + packagePath + `?ref=${branch}`);
 }
 
-export function existsInRepo(nameOrUrl: string): Promise<boolean> {
+export function existsInRepo(nameOrUrl: string, ref: string): Promise<boolean> {
   try {
     const url = new URL(nameOrUrl);
     return isUrlOk(url.href);
   } catch {
+    const params = new URLSearchParams({ ref });
     return isUrlOk(
       `https://api.github.com/repos/dai-shi/waku/contents/examples/${encodeURIComponent(
         nameOrUrl,
-      )}`,
+      )}?${params}`,
     );
   }
 }
@@ -116,15 +117,20 @@ export async function downloadAndExtractRepo(
   );
 }
 
-export async function downloadAndExtractExample(root: string, name: string) {
+export async function downloadAndExtractExample(
+  root: string,
+  name: string,
+  ref: string,
+) {
   await pipeline(
     await downloadTarStream(
-      'https://codeload.github.com/dai-shi/waku/tar.gz/main',
+      `https://codeload.github.com/dai-shi/waku/tar.gz/${ref}`,
     ),
     tar.x({
       cwd: root,
       strip: 2 + name.split('/').length,
-      filter: (p) => p.includes(`waku-main/examples/${name}/`),
+      filter: (p) =>
+        p.replace(/waku-[^/]+\//, '').includes(`examples/${name}/`),
     }),
   );
 }
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -14,18 +14,23 @@ import {
   existsInRepo,
   getRepoInfo,
   hasRepo,
-  type RepoInfo,
-} from './helpers/example-option';
+} from './helpers/example-option.js';
+import type { RepoInfo } from './helpers/example-option.js';
+
+const DEFAULT_REF = 'v0.20.0-alpha.2';
 
 // FIXME is there a better way with prompts?
-const { tokens } = parseArgs({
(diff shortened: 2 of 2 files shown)


packages/create-waku/src/index.ts
```diff
@@ -185,13 +199,13 @@ async function init() {
     process.exit(1);
   }
   let repoInfo: RepoInfo | undefined;
-  const example = getExample(tokens);
 
-  if (example) {
+  // FIXME this is pretty hard to follow. Separate functions?
```
dai-shi, line 203: @ojj1123 I found the code pretty hard to read. There is a big room for improvement. I will factor out the functionality into the separate file, and you can consider some refactors there. Please work on refactors and adding tests before adding new capabilities.

packages/create-waku/src/helpers/example-option.ts
```diff
@@ -116,15 +118,130 @@ export async function downloadAndExtractRepo(
...
     await downloadTarStream(
-      'https://codeload.github.com/dai-shi/waku/tar.gz/main',
+      `https://codeload.github.com/dai-shi/waku/tar.gz/${ref}`,
     ),
     tar.x({
       cwd: root,
       strip: 2 + name.split('/').length,
-      filter: (p) => p.includes(`waku-main/examples/${name}/`),
+      filter: (p) =>
+        p.replace(/waku-[^/]+\//, '').includes(`examples/${name}/`),
```
dai-shi, line 134: This seems too hacky. Any idea to improve?

packages/create-waku/src/helpers/example-option.ts
```diff
@@ -116,15 +118,130 @@ export async function downloadAndExtractRepo(
...
+    } else {
+      console.log(
+        `Downloading files for example ${cyan(example)}. This might take a moment.`,
+      );
+      console.log();
+      await downloadAndExtractExample(root, example, defaultRef);
+    }
+  } catch (err) {
+    // download error
+    throw new Error(isErrorLike(err) ? err.message : err + '');
```
dai-shi, line 241: It doesn't seem doing anything good. If we want friendly error message we shouldn't throw anyway.

dai-shi:
> I will factor out the functionality into the separate file

I think I did more than that.

ojj1123 (commented):
thanks to you, the `index.ts` is simple and getting the high readability.
I will write test code first as you said

packages/create-waku/src/helpers/example-option.ts
ojj1123: nice your catch! I agree with you :)

packages/create-waku/src/index.ts
```diff
@@ -1,145 +1,57 @@
...
-  existsInRepo,
-  getRepoInfo,
-  hasRepo,
-  type RepoInfo,
-} from './helpers/example-option';
+  parseExampleOption,
+  downloadAndExtract,
+} from './helpers/example-option.js';
+
+const DEFAULT_REF = 'v0.20.0-alpha.2';
```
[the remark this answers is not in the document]
dai-shi: yes, manually.

packages/create-waku/src/index.ts
```diff
@@ -1,145 +1,57 @@
...
     example: {
       type: 'string',
     },
+    help: {
+      type: 'boolean',
+      short: 'h',
+    },
   },
-  tokens: true,
 });
```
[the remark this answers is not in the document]
dai-shi: Let's leave it. It's the same as what we do in `waku` too.

packages/create-waku/src/helpers/example-option.ts
```diff
@@ -116,15 +118,113 @@ export async function downloadAndExtractRepo(
...
+  repoInfo?: RepoInfo;
+};
+
+export async function parseExampleOption(
+  example: string | undefined,
+  defaultRef: string,
+): Promise<ParsedExampleOption | undefined> {
+  if (!example) {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
dai-shi: yeah, but let's leave it as type says `undefined` instead of `void`. or, we can return null.

packages/create-waku/src/helpers/example-option.ts
dai-shi: it has both replace and includes. `name` can contain `waku-foo`.

packages/create-waku/src/helpers/install-template.ts
```diff
@@ -0,0 +1,59 @@
...
+  packageName: string;
+}) {
+  const pkg = {
+    name: packageName,
+    version: '0.0.0',
+  };
+
+  const templateRoot = path.join(
+    fileURLToPath(import.meta.url),
+    '../../template',
```
dai-shi, line 21: It's very confusing that this isn't `../../../template`.

PR #603, opened 2024-03-18
fix: rsc hot reload

dai-shi:
it's a regression in #593.


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -75,11 +75,10 @@ export function rscHmrPlugin(): Plugin {
     async transform(code, id) {
       if (id.startsWith(wakuClientDist)) {
         // FIXME this is fragile. Can we do it better?
```
dai-shi, line 77: Indeed, it's fragile.

PR #607, opened 2024-03-19
fix: invalidating ssr module

Aslemammad:
This avoids hydration mismatch error in DEV.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
@@ -11,7 +11,6 @@ import {
   decodeFilePathFromAbsolute,
   filePathToFileURL,
 } from '../utils/path.js';
-import { moduleCache, moduleLoading } from '../utils/react-server-dom-webpack.js';
 
 type ModuleImportResult = TransformResult & {
   id: string;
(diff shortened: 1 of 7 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -41,7 +42,7 @@ if (import.meta.hot && !globalThis.__WAKU_HMR_CONFIGURED__) {
     document.head.appendChild(script);
     // avoid HMR flash by first applying the new and removing the old styles 
     if (style) {
-      queueMicrotask(style.remove);
+      queueMicrotask(() => style.parentElement?.removeChild(style));
```
dai-shi, line 45: I think you want this?
```suggestion
      queueMicrotask(() => style.remove());
```

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -122,6 +123,21 @@ export function rscHmrPlugin(): Plugin {
         );
       }
     },
+    handleHotUpdate({ file }) {
+      const moduleLoading = (globalThis as any).__webpack_module_loading__;
+      const moduleCache = (globalThis as any).__webpack_module_cache__;
+      const id = filePathToFileURL(file);
+      if (moduleLoading.has(id) && moduleCache.has(id)) {
```
dai-shi, line 130: Should we check only the first one?
```suggestion
      if (moduleLoading.has(id)) {
```

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
Aslemammad: Nope this errors out sometimes. Illegal invocation.

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
Aslemammad: I do not know about react's impl details, but I think checking the two might be safer?

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
dai-shi: I see...

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
dai-shi: It's not React impl, but yeah it can be a little safer. Let's leave it.

dai-shi (approved):
Thanks. This is a nice improvement.

PR #612, opened 2024-03-21
html-renderer: Simplify module loading for react

Tobbe:
In my testing I discovered that `react` and `react-dom-server-webpack/client.edge` could be imported using regular imports. No need to have a local build of those and import that build.

### DIFF
--- a/packages/waku/src/lib/renderers/html-renderer.ts
+++ b/packages/waku/src/lib/renderers/html-renderer.ts
@@ -4,7 +4,7 @@ import { createElement } from 'react';
 import { createFromReadableStream } from 'react-server-dom-webpack/client.edge';
 import { injectRSCPayload } from 'rsc-html-stream/server';
 
-import { ServerRoot } from '../../client.js';
+import type * as WakuClientType from '../../client.js';
 import type { EntriesPrd } from '../../server.js';
 import type { ResolvedConfig } from '../config.js';
 import { concatUint8Arrays } from '../utils/stream.js';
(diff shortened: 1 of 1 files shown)


dai-shi:
Nice finding. It must have been improved in some PRs like #457 or something.

I'd like to keep it as is for v0.20.0 and investigate more to understand why `rd-server` and `waku-client` can't be `import`ed. With that, we will be confident to merge this.

Marking a draft until then.

dai-shi:
I wonder if #762 makes any changes. Hm, but it's just for DEV. It shouldn't affect.

I still think this PR is a good move, so I wish we find the reason why `import` doesn't work for rd-server and waku-client.

dai-shi:
https://github.com/dai-shi/waku/pull/707#discussion_r1661267451

@Aslemammad This PR might be somewhat related.

dai-shi:
@Tobbe When you have a chance, can you merge main and try removing `loadClientModule` completely?

Tobbe:
Thanks for the reminder. I'll try to get to it today

Tobbe:
I'm too tired. I'll continue tomorrow

Tobbe:
Seems to be failing for at least `renderToReadableStream` still
image

I'll revert that one to keep using `loadClientModule` and see if that works

Tobbe:
Nope. Still broken
image

dai-shi:
Thanks so much for trying out. I'll have a closer look later.

dai-shi:
Actually, we are patching rsdw now since #707, it has to load through Vite.

dai-shi:
I think this is what we can do now.
@Aslemammad suggests to load everything through Vite from the start, and if it could be done, we might be able to simplify `html-renderer`.

dai-shi (approved):
Thanks for your suggestion anyway!

PR #629, opened 2024-03-28
feat: support omitting file extensions in import statements

himself65:


dai-shi (approved):
Nice. It works.

examples/12_css/src/components/App.tsx
```diff
@@ -1,9 +1,9 @@
 import classNames from 'classnames';
 import styles from './App.module.css';
 import './App.css';
-import { container } from './styles.css.js';
-import { Counter } from './Counter.js';
-import { Banner } from './Banner.js';
+import { container } from './styles.css';
```
dai-shi, line 4: This causes a temporary error like this in DEV, when you reload the browser.

Untitled

Adding back `.js` fixes it.

Do you have a better solution? @himself65

PR #632, opened 2024-03-29
chore(example): remove useless condition

szymonrybczak:
Hey! In this Pull Request, I have simply deleted a redundant condition as it is already checked a few lines above.


dai-shi (approved):
Thanks.

PR #633, opened 2024-03-30
feat: support js/jsx extensions (minimal api only)

dai-shi:
See: 21_minimal_js example


packages/waku/src/lib/builder/build.ts
```diff
@@ -295,6 +302,9 @@ const buildSsrBundle = async (
           },
           noExternal: /^(?!node:)/,
         },
+    esbuild: {
+      jsx: 'automatic',
+    },
```
dai-shi, line 307: I have no idea why it worked without this before.

PR #636, opened 2024-03-30
feat(router): support js/jsx extensions

dai-shi:
following #633
see 02_template_js example


packages/waku/src/router/fs-router.ts
```diff
@@ -33,20 +35,29 @@ export function fsRouter(
...
-          // HACK: replace "_slug_" to "[slug]"
+          // HACK: replace "_slug_" to "[slug]" for build
           file = file.replace(/(^|\/|\\)_(\w+)_(\/|\\|\.)/g, '$1[$2]$3');
-          if (sep === '/') {
-            return [file];
-          }
           // For Windows
-          return [file.replace(/\\/g, '/')];
+          file = sep === '/' ? file : file.replace(/\\/g, '/');
+          // HACK: resolve different extensions for build
```
dai-shi, line 47: I with there would be a nice way to avoid it. But, maybe it's not a priority.

PR #644, opened 2024-04-01
experimental: use serverless for website

dai-shi:


himself65 (changes requested):
I got a blank page

https://waku-git-website-with-serverless-daishis-projects.vercel.app/non-exist

dai-shi:
oh, that's interesting, not expected.

dai-shi:
Hmm, I don't know how it happens (to be `200` instead of `404`).
Help wanted.

himself65:
Add some log, found issue
image

himself65:
> What's the issue??

ctx.res === undefined

dai-shi:
> > What's the issue??
> 
> ctx.res === undefined

But, then it should be handled by `app.notFound`, no? At least, `waku start` on localhost works fine.

rmarscher:
> But, then it should be handled by `app.notFound`, no? At least, `waku start` on localhost works fine.

I'm having a similar 404 issue on my cloudflare pages deploy. I have a console.log in app.notFound and verified it runs. Something must modify the response or the ctx after that. https://fc7720e6.waku-test.pages.dev/asdf

If you build the project and then run the dist output instead of the vite dev server, you should be able to reproduce it locally. I'm going to see if I can set up a step debugger to figure it out.

dai-shi:
Fixed by #796.

PR #645, opened 2024-04-01
fix: double modules hazard by preferring relative paths

### LINKED ISSUE #631: [BUG] Context Provider invalid value
Description:
Context value seems to not be what the `Provider` has set, but always fallback to the default one, when using `use client`.

Reproducible sandbox:
https://codesandbox.io/p/devbox/waku-x-kendo-forked-yp6m3f?file=%2Fsrc%2Ftemplates%2Fgrid-page.tsx%3A11%2C11

Steps to reproduce:
- Start the dev server
- navigate to the `/grid` page
- Observe the context's value

Expected result:
- The value should equal what's set through the `Context.Provider`

Actual result:
- The value is always what's set as a default in the `.createContext()` call.

Aslemammad:
close #631


packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -45,20 +45,27 @@ const configPrivateDir = getEnvironmentData('CONFIG_PRIVATE_DIR') as string;
...
     if (moduleNode.file === id) {
       return moduleNode.url;
     }
   }
-  const filePath = id.startsWith('file://') ? fileURLToFilePath(id) : id;
-  // HACK this relies on Vite's internal implementation detail.
-  return config.basePath + '@fs' + encodeFilePathToAbsolute(filePath);
+  let filePath = id.startsWith('file://') ? fileURLToFilePath(id) : id;
+  console.log('startFilePath', filePath)
+  if (filePath.startsWith(config.rootDir)) {
```
dai-shi, line 58: You want to check if this works as expected on windows.

packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -45,20 +45,27 @@ const configPrivateDir = getEnvironmentData('CONFIG_PRIVATE_DIR') as string;
...
+    filePath = filePath.slice(config.rootDir.length, filePath.length);
```
dai-shi, line 59: Is it just removing a prefix?
```suggestion
    filePath = filePath.slice(config.rootDir.length);
```

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -262,18 +261,29 @@ export const renderHtml = async (
             get(_target, name: string) {
               if (isDev) {
                 // TODO too long, we need to refactor this logic
-                let file = filePath.slice(config.basePath.length);
-                file = file.split('?')[0]!;
-                file = file.startsWith('@fs/')
-                  ? file.slice('@fs'.length)
-                  : encodeFilePathToAbsolute(joinPath(opts.rootDir, file));
+                const file = (
+                  filePath.startsWith('/@fs/')
```
dai-shi, line 265:
```suggestion
                  filePath.slice(config.basePath.length).startsWith('@fs/')
```

(A note: we try to make it work with `basePath`, but as we don't have tests, there might be some bugs as of now.)

dai-shi (commented):
looks like we are almost there!

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -262,18 +262,28 @@ export const renderHtml = async (
...
                   '../../..',
                 );
-                if (file.startsWith(wakuDist)) {
+                const fileWithAbsolutePath =
+                  !file.startsWith(opts.rootDir) && !file.startsWith(wakuDist)
+                    ? encodeFilePathToAbsolute(joinPath(opts.rootDir, file))
+                    : file;
+                if (
+                  file.startsWith(wakuDist) ||
+                  fileWithAbsolutePath.startsWith(wakuDist)
```
dai-shi, line 280: This part looks a little bit complicated and has room for improvement.

dai-shi (commented):
Almost.

packages/waku/src/lib/renderers/html-renderer.ts
Aslemammad: it makes /user/project/src/comp.tsx into /src/comp.tsx

dai-shi: Yeah, but this entire block can be refactored for more readability, no?

dai-shi: Hm, I think we can remove L279.

dai-shi: > Hm, I think we can remove L279.


5e1ae0d

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -165,23 +165,21 @@ export const devServer: Middleware = (options) => {
       vitePromise,
     ]);
 
-    if (!initialModules) {
```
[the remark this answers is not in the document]
dai-shi: 651f04c reverted this. let's tackle it as #601 separately. feel free to submit a new PR, hopefully with a test case (e2e or an example).

dai-shi (approved):
should be good.

Aslemammad:
Everything's passing, congrats!

PR #641, opened 2024-03-31
New example with cssmodules and pages router

dai-shi:
- added `04_cssmodules` example
- removed `09_cssmodules` example
- changed `12_css` example to include layout in `09_cssmodules`

Remaining:
- [ ] Flash of unstyled content (FOUC) in DEV
- [ ] No styles in client components in PRD

PR #647, opened 2024-04-03
feat(create-waku): recover choose option

### LINKED ISSUE #625: Question: Optional Tools in Create Usage
Wouldn't it be ideal to include selections during creation? For instance, choosing whether to have the application with TypeScript or JavaScript, and even with or without Tailwind.

I believe that in many cases, especially concerning Tailwind, it's more a matter of personal preference. Of course, there may be projects that need to use ChakraUI or MaterialUI, depending on the company, the project, or the developer's preference. Making this functionality optional would be more practical than having Tailwind as the default.

dai-shi:
I had thought it would be nice to have a single `--example` option, but changed my mind. This introduces `--choose` option again.

https://github.com/dai-shi/waku/discussions/315#discussioncomment-8990718

closes #625
along with #641

### DIFF
--- a/packages/create-waku/package.json
+++ b/packages/create-waku/package.json
@@ -22,7 +22,7 @@
     "start": "node ./dist/index.js",
     "dev": "ncc build ./src/index.ts -w -o ./dist/",
     "compile": "rm -rf template dist *.tsbuildinfo && pnpm run template && pnpm run build",
-    "template": "cp -r ../../examples template && rm -rf template/*/dist && rm -rf template/*/node_modules && (for d in template/*; do mv $d/.gitignore $d/gitignore; done)",
+    "template": "cp -r ../../examples template && rm -rf template/*/dist && rm -rf template/*/node_modules && (for d in template/*; do mv -f $d/.gitignore $d/gitignore; done)",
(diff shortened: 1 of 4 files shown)


ojj1123:
> but it comes with a hack `DEFAULT_REF`.
> https://github.com/dai-shi/waku/discussions/315#discussioncomment-8990718

I agree with you.
But I don't think it is nessesery to use `--choose` option. Because even if we don't write `--choose` option, we can select the template **by default**. So, how about choosing one of the templates without `--choose` option?

I think it's easier **to choose the template without `--choose`** than **to choose it with `--choose`.**
Like this:
```shell
> pnpm create waku
✔ Project name: … waku-project
✔ Select a framework: › 
  > template-js
     template-ts
     tailwind-js
     tailwind-ts
     ...

Scaffolding project in /Users/jeongjin/GitHub/waku-project...
```

ojj1123:
And `--example` option is just used to download the other repo?

dai-shi:
> we can select the template by default.

We used to have it. It's intentional for now. This is an advanced option, because not all examples are well described.
Eventually, if we improve some templates so that we can recommend them, we can enable the selector again.

PR #650, opened 2024-04-05
Chapter 1 tutorial

dizzyjaguar:


dai-shi (commented):
Please move the file to `docs/tutorials/getting-started.mdx`

dai-shi:
@sandren any feedback?

sandren:
Thanks for working on this.

It's ok that it's still a work-in-progress, but we shouldn't merge any placeholder content. I notice a few pieces:

- (?possibly insert image here?)
- (insert a good example here)
- They will be...
- Insert Link To Next Chapter

It would be better to highlight what you _can_ do in the client components section (instead of saying what you _cannot_ do in server components). That will help balance those two sections as well.

Also I understand the routing section is scoped to what's present in the default starter, but it should still cover dynamic rendering, segment routes, etc.

Thanks! 🙂

dizzyjaguar:
@sandren  That all sounds good!
- I'll get those placeholders removed
- Great idea to balance the **RSC** and **Client Components**
- For the rest of the **routing** I was planning to go over those details in **chapter 2** when the user builds the **todo app**; since then we will have some more situations to use them. Do you think I should I still attempt to go over those features in **Chapter 1** as-well?  

Thanks for reviewing 🙏🏼 
Should hopefully have these changes pushed in the next few days.

sandren:
> * For the rest of the **routing** I was planning to go over those details in **Chapter 2** when the user builds the **todo app**; since then we will have some more situations to use them.

That's fine as long as the language of Chapter 1 doesn't imply that single-routes are all that are available. It should be clear that it's only covering those for now because that's what's present in the starter. 🙂

dizzyjaguar:
> > * For the rest of the **routing** I was planning to go over those details in **Chapter 2** when the user builds the **todo app**; since then we will have some more situations to use them.
> 
> That's fine as long as the language of Chapter 1 doesn't imply that single-routes are all that are available. It should be clear that it's only covering those for now because that's what's present in the starter. 🙂

Awesome, I'll add something in that section about going over the rest routing in the next chapter! 😎

dizzyjaguar:
@dai-shi  @sandren  

Just pushed those small updates 🙂

Let me know if theres anything else!~

sandren (approved):
I've made a few small revisions, but otherwise everything looks good to me. Thanks @dizzyjaguar! 🙂

dizzyjaguar:
@sandren Awesome! 🎆 
@dai-shi & @sandren  should I start working on **Chapter 2** after this gets merged in?

dai-shi:
Yeah!

PR #657, opened 2024-04-09
fix: css modules with hmr

### LINKED ISSUE #639: Suspense doesn't work in layout with css modules (DEV-only)
We have a tricky case in an example.

https://github.com/dai-shi/waku/blob/3b87654f39fc12592b1238e89e3423f105f41840/examples/09_cssmodules/src/components/Layout.tsx#L6-L20

It should show "Pending..." and then "Hello from server!".
However, it doesn't work as expected.

It seems like a regression in #446.
This is a dev-only issue.
@Aslemammad Can you investigate it please?

Here is a screencast that works before #446 commit.

https://github.com/dai-shi/waku/assets/490574/3e082901-5662-4360-84ef-085ae81c28ff

dai-shi:
close #639

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
@@ -194,38 +194,26 @@ async function generateInitialScripts(
   }
 
   const scripts: HtmlTagDescriptor[] = [];
-  let injectedBlockingViteClient = false;
 
   for (const result of sources.values()) {
-    // CSS modules do not support result.source (empty) since ssr-transforming them gives the css keys
-    // and client-transforming them gives the script tag for injecting them.
     if (result.id.endsWith('.module.css')) {
-      if (!injectedBlockingViteClient) {
-        // since we use the client-transformed script tag, we need to avoid FOUC by parse-blocking the vite client that the script imports
-        // this way we make sure to run the CSS modules script tag before everything
-        // blocking this way is not ideal but it works. It should be revisited.
-        scripts.push({
-          tag: 'script',
-          attrs: { type: 'module', blocking: 'render', src: '/@vite/client' },
-          injectTo: 'head-prepend',
-        });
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -194,38 +194,26 @@ async function generateInitialScripts(
...
-        injectedBlockingViteClient = true;
-      }
+      // CSS modules do not support result.source (empty) since ssr-transforming them gives the css keys
+      // and client-transforming them gives the script tag for injecting them.
       scripts.push({
         tag: 'script',
-        // tried render blocking this script tag by data url imports but it gives `/@vite/client: Invalid relative url or base scheme isn't hierarchical.` which could not find a way to fix.
-        attrs: { type: 'module', 'waku-module-id': result.id },
+        // async: true to wait vite/client to be loaded
+        attrs: { type: 'module', async: true, 'waku-module-id': result.id },
```
Aslemammad, line 205: I guess this would cause FOUC! Since css will potentially load later.

dai-shi: yeah, but better than waiting everything.

dai-shi: so, if there are a better solution, it's better.

Aslemammad (approved):

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
dai-shi: On another thought, I'm not sure if it really causes FOUC. Let's reconsider if it happens.

dai-shi: Yeah, it causes FOUC. I found a better fix: #659 
btw `async: true` makes the execution earlier.

PR #658, opened 2024-04-10
fix: No styles in client components in PRD

### LINKED ISSUE #651: Fix broken `04_cssmodules` example
#641 is incomplete and broken.

Remaining:
- [ ] Flash of unstyled content (FOUC) in DEV 👉 Technically impossible to fix to avoid FOUC for client components with the current architecture
- [ ] No styles in client components in PRD 👉 #658

Aslemammad:
Partially resolves #651

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -369,6 +369,7 @@ const buildClientBundle = async (
         onwarn,
         input: {
           main: mainJsFile,
+          // rollup will ouput the style files related to clientEntryFiles, but since it does not find any link to them in the index.html file, it will not inject them. They are only mentioned by the standalone `clientEntryFiles`
           ...clientEntryFiles,
         },
         preserveEntrySignatures: 'exports-only',
@@ -478,7 +479,12 @@ const emitHtmlFiles = async (
   distEntries: EntriesPrd,
   buildConfig: BuildConfig,
   getClientModules: (input: string) => string[],
+  clientBuildOutput: Awaited<ReturnType<typeof buildClientBundle>>,
 ) => {
+  const nonJsAssets = clientBuildOutput.output.flatMap(({ type, fileName }) =>
+    type === 'asset' && !fileName.endsWith('.js') ? [fileName] : [],
+  );
+  const cssAssets = nonJsAssets.filter((asset) => asset.endsWith('.css'));
   const basePrefix = config.basePath + config.rscPath + '/';
   const publicIndexHtmlFile = joinPath(
     rootDir,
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/builder/build.ts
```diff
@@ -502,6 +508,14 @@ const emitHtmlFiles = async (
           typeof pathname === 'string' ? pathname2pathSpec(pathname) : pathname;
         let htmlStr = publicIndexHtml;
         let htmlHead = publicIndexHtmlHead;
+        if (cssAssets.length) {
+          const cssStr = cssAssets
+            .map((asset) => `<link rel="stylesheet" href="${asset}">`)
+            .join('\n');
+          // HACK is this too naive to inject style code?
```
dai-shi, line 515: Yeah, I wish we could inject it in `rscIndexPlugin` for both server & client styles.
(hmm, we might consider about re-architecture at some point, for code-splitting styles, and (vite) plugins too.)

dai-shi (approved):
Other than that it looks good.

packages/waku/src/lib/builder/build.ts
Aslemammad: Yea, we haveto do code-splitting at some point! Shall I now work on it and get it working with rscIndexPlugin?

dai-shi: > Shall I now work on it and get it working with rscIndexPlugin?

Is it possible for the current architecture?
It would be nice if there's a less-hacky way.

Aslemammad: Sure I'll try my best, I'll get back to you soon! 
But if we want to do a release, I think let's merge it and tackle it later (maybe in the re-arch)

dai-shi: (re-arch won't be soon.)
Yeah, if there's a better way, we would like to have it now. I plan to merge it next week.

Aslemammad: Took a look, the emitted css files are emitted in the client bundle phase which is the last phase, while rscIndexPlugin happens at the same time or even before in the buildSsrBundle phase. 

I don't know if it's possible to bring the client bundle before ssr! But it's not ideal anyway. Wdyt? maybe we should skip it.

dai-shi: Yeah, it may require one more viteBuild step, which is not very good because it will make it slower.

I think the best is to send `<link>` tag in the RSC payload. But, it will require new technique (it's more work and different from re-arch I said above. It's more related with code-splitting.)

Thanks for your investigation. Let's merge this for now.

PR #663, opened 2024-04-19
feat: rsc-analyze-boundary

QDyanbing:
use client/server 判断有边界问题


dai-shi:
See #664 discussion.
Unless we observe a real problem, let's wait for a moment. Let us know if there's a real problem. Closing it for now, but thanks for your suggestion.

PR #664, opened 2024-04-19
fix: resolve use client/server boundary problem

jeasonnow:
Resolve the current issue with the ‘use client/server’ directive that involves ambiguity in the boundary of instruction resolution during compilation.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-analyze.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-analyze.ts
@@ -34,17 +34,38 @@ export function rscAnalyzePlugin(
           syntax: 'typescript',
           tsx: ext.endsWith('x'),
         });
+        // to fix the issue that the directive is not at the top of the file
+        let isDirective = true;
+        
         for (const item of mod.body) {
           if (
             item.type === 'ExpressionStatement' &&
             item.expression.type === 'StringLiteral'
           ) {
             if (item.expression.value === 'use client') {
-              clientEntryCallback(id);
-              fileHashMap.set(id, await hash(code));
+              if (!isDirective) {
+                const e = {
+                  messageText:
+                    'The `"use client"` directive must be put at the top of the file.',
+                }
+                throw e;
+              } else {
+                clientEntryCallback(id);
+                fileHashMap.set(id, await hash(code));
+              }
             } else if (item.expression.value === 'use server') {
-              serverEntryCallback(id);
+              if (!isDirective) {
+                const e = {
+                  messageText:
+                    'The `"use server"` directive must be put at the top of the file.',
+                }
+                throw e
+              } else {
+                serverEntryCallback(id); 
+              }
             }
+          } else {
+            isDirective = false;
           }
         }
       }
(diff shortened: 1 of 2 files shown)


jeasonnow:
https://github.com/dai-shi/waku/pull/663 
和这个是同样的问题

packages/waku/src/lib/plugins/vite-plugin-rsc-analyze.ts
```diff
@@ -34,17 +34,38 @@ export function rscAnalyzePlugin(
...
             item.type === 'ExpressionStatement' &&
             item.expression.type === 'StringLiteral'
           ) {
             if (item.expression.value === 'use client') {
-              clientEntryCallback(id);
-              fileHashMap.set(id, await hash(code));
+              if (!isDirective) {
+                const e = {
+                  messageText:
+                    'The `"use client"` directive must be put at the top of the file.',
```
dai-shi, line 49: Is there any requirement for this?

What's the exact problem you faced?

dai-shi:
Thanks for the suggestion. We will be working on a rewrite for the next version, so I won't be merging this.

But, we want to hear the problem.

packages/waku/src/lib/plugins/vite-plugin-rsc-analyze.ts
jeasonnow: I haven’t faced any issues related to this. I’m just doing this to make sure the compiled code follows React’s guidelines for the use-client directive: https://react.dev/reference/react/use-client#use-client.

In the current code, writing it like this also work like this:
```typescript
import { useState } from 'react';


export const Counter = () => {
  const [count, setCount] = useState(0);
  'use server';

  const handleIncrement = () => setCount((c) => c + 1);

  return (
      <div>Count: {count}</div>
      <button
        onClick={handleIncrement}
        className="rounded-sm bg-black px-2 py-0.5 text-sm text-white"
      >
        Increment
      </button>
    </section>
  );
};

```

dai-shi: Yeah, because inline directive aren't supported.
There's an issue that Vite adds import statements before the directive.

> 'use client' must be at the very beginning of a file, above any imports or other code (comments are OK). They must be written with single or double quotes, but not backticks.

Anyway, I will work on it later. Thanks for sharing your thought.

jeasonnow: Sorry, there’s a bit of a problem with my description. The way I’ve written it might cause the entire component to be mistakenly recognized as a server component and packaged. I hope this issue can be fixed in future versions, thx

dai-shi: > might cause the entire component to be mistakenly recognized as a server component

Currently, no, because we only check top-level directives.
But, in the future when we support inline directives, it might be a server action.

Anyway, if we only use beginning directive as you suggest, it should be fine. (I'm not sure if we error it, maybe eslint rule should warn it.)

dai-shi:
Closing this as it's not planned currently. Stay tuned for a rewrite.

PR #667, opened 2024-04-23
chore: set up vitest and ts-expect

pmelab:
Minimal setup of `vitest` and `ts-expect`.

### DIFF
--- a/packages/waku/tests/config.test.ts
+++ b/packages/waku/tests/config.test.ts
@@ -1,5 +1,5 @@
 import { test, expect } from 'vitest';
-import { type Config, defineConfig } from 'waku/config';
+import { type Config, defineConfig } from '../src/config.js';
 import { expectType } from 'ts-expect';
 
 // Absolutely meaningless unit and type test examples.
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    // Don't match "*.spec.ts" (like the vitest default does). They are for playwright.
+    include: ['**/*.test.?(c|m)[jt]s?(x)'],
+  },
+});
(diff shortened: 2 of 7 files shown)


vitest.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    // Don't match "*.spec.ts" (like the vitest default does). They are for playwright.
+    include: ['**/*.test.?(c|m)[jt]s?(x)'],
```
dai-shi, line 6: Can we put tests file in `./tests`? Also, what's `?` after `.` for?

dai-shi (commented):
Thanks for working on it. Just one comment.

vitest.config.ts
pmelab: `./tests` on the root level or in `packages/waku`?

And that brings another question:

I added `vitest` and `ts-expect` to the root level, since there is the `test` command, but I'm not sure if it would make more sense to have them in `packages/waku` only. Also not sure where the test fail comes from, but it smells like its related to directories.

dai-shi: > ./tests on the root level or in packages/waku?

I prefer `packages/waku/tests`.

> I added vitest and ts-expect to the root level, since there is the test command, but I'm not sure if it would make more sense to have them in packages/waku only.

Not sure either. @Aslemammad @himself65 thoughts?

himself65: packages/waku/tests seems good to me

packages/waku/tests/config.test.ts
himself65, line 1: Here are some of my ideas about the vitest test. Can we test each single Vite plugin (like rsc-transform-plugin)? Or we could snapshot the build output for fixture inputs

dai-shi (approved):
LGTM
Thanks!

PR #668, opened 2024-04-24
SSG: high volume performance

pmelab:
## The problem

HTML payloads of static pages grow linearly with the number of them. This is already evident in the `03_demo` example. Each pokemon has an entry in `__WAKU_ROUTER_PREFETCH__` that points to the identical component. The `/SHOULD_SKIP` entry also grows. With a larger number of pages, the build breaks due to memory issues or node simply not being able to `JSON.stringify` the build config in:
 https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/lib/builder/build.ts#L692

Also, when generating the map for `__WAKU_ROUTER_PREFETCH__`, pages are rendered non-concurrently, which means blocking operations can't run in parallel.

## Solutions

1. I switched router prefetching to a pattern instead of path-based model. That way each static page needs only one entry. Not one for each static path.
2. The `shouldSkipObj` grows with each page entry. To be completely honest, I don't know enough about the rendering process to judge if each page would need skip lists for each other page component. But that growth during rendering felt weird, so i just moved the object one scope level down to reset it for each page. It's a complete shot in the dark, but it did hit something.
https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/router/define-router.ts#L96
3. Collection of modules for `__WAKU_ROUTER_PREFETCH__` has been changed to run in parallel. This is also necessary to enable in-component optimizations, like database request batching or connection pooling.

## Notes

* The `ssg-performance` fixture renders 10k pages in about a minute on my M1. Before it would just break.
* The build config serialization might still break at some threshold, due to `JSON.stringify` limits.
* Rendering all static paths to collect modules for prefetching feels wasteful, but thats too big of a change for this PR. But this brings us into a working state.


dai-shi:
> The problem

Nice catch. I didn't know the exact problem, but I felt like the approach had some limits.
I haven't looked into the code, I will later. You know how CIs are failing?
Having a test is great, because eventually we would like to revisit `shouldSkip` mechanism.

pmelab:
> You know how CIs are failing?

Playwright browsers fail to install there:

```
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' is no longer signed.
```

> Having a test is great, because eventually we would like to revisit `shouldSkip` mechanism.

Absolutely. The change was more based on gut feeling rather than confident understanding.

I traced down the use of `/SHOULD_SKIP` into `getSkipList` ...

```
https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/router/client.ts#L213-L242
```

... which filters components of the current route path during fetching and pre-fetching of RSC payloads ...

https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/router/client.ts#L281-L297

https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/router/client.ts#L325-L332

I did not see how components of other routes could have an effect here, but I might be missing something. If you can help me to understand what it is actually supposed to **do**, I would be able to implement a test.

dai-shi:
> If you can help me to understand what it is actually supposed to do, I would be able to implement a test.

The current design is the client side doesn't (shouldn't) know about the server-side route configs. What the client knows is if it should skip certain route (or maybe tell what it has in its cache to the server, and the server decides what to respond).

But, e2e tests shouldn't depend on implementation details and what should be tested is `route: 'static'` and `route: 'dynamic'` behavior.

I really wish to refactor the internal mechanism. Hopefully some time soonish.

pmelab:
> But, e2e tests shouldn't depend on implementation details and what should be tested is route: 'static' and route: 'dynamic' behavior.

You mean `render: 'static'` and `render: 'dynamic'`? This is covered by a lot of integration tests already, no?

dai-shi:
> You mean `render: 'static'` and `render: 'dynamic'`?

You are right. My bad.

> This is covered by a lot of integration tests already, no?

I don't remember if we have such tests that they ensure static pages are really "static". I don't think so.

pmelab:
I found another issue: While generating the build config, all routes are rendered for collecting modules for prefetching:
https://github.com/dai-shi/waku/blob/6550a0d26e7830527c767fcda8eb906aeb0f1266/packages/waku/src/router/define-router.ts#L155-L163

This runs sequentially and currently blocks parallel rendering of pages. Thats probably not the best solution, since rendering all static instances of the same route to collect modules feels wrong, but fixing that would be a bigger endeavour. Parallelizing at least makes things work for now. Data requests should implement good caching though, since they run 3 times per page.

pmelab:
@dai-shi : Is that what you had in mind? https://github.com/dai-shi/waku/pull/668/commits/a4d88782b02712f42ca1901d1da65f27fc4478a2

dai-shi:
> @dai-shi : Is that what you had in mind? [a4d8878](https://github.com/dai-shi/waku/commit/a4d88782b02712f42ca1901d1da65f27fc4478a2)

Thanks. Hm, it looks good for the start, but what I meant with really "static" is that the static page has something like `Date.now()`, and if you reload the page, the date won't change.

To be honest, it's not fully implemented:
- SSR/SSG are fine
- RSC isn't static, so client routing may render static page again. (I hope to work on it after v0.21.0.)
  - Though, if client caches the rsc payload, it will skip it with `shouldSkip`. (Also, wanted to revisit in the future.)
- In DEV, static pages always rerender (I don't think it's solvable.)

dai-shi:
Re: https://github.com/dai-shi/waku/pull/668#issuecomment-2077905809

See also #543. We would like to change `unstable_collectClientModue` so that it statically analyzes the code, instead of executing it. https://github.com/dai-shi/waku/pull/573#discussion_r1514091960

pmelab:
> See also https://github.com/dai-shi/waku/issues/543. We would like to change unstable_collectClientModue so that it statically analyzes the code, instead of executing it. https://github.com/dai-shi/waku/pull/573#discussion_r1514091960

That would have been my next question. Thanks!

> Thanks. Hm, it looks good for the start, but what I meant with really "static" is that the static page has something like Date.now(), and if you reload the page, the date won't change.

So we want to test when components are executed. I changed the testcase to cover that (I hope 🤞 ).

Unfortunately I have no clue where the test fails come from. It only uses `tsconfig.e2e.json` for typechecks?

dai-shi:
> It only uses tsconfig.e2e.json for typechecks?

You need to add new tests in "references".

pmelab:
Ok, apparently using `createPages` in a fixture also leads to type errors. I switched it to file based routing instead.

dai-shi:
> Ok, apparently using `createPages` in a fixture also leads to type errors.

That sounds somewhat unexpected. Can you elaborate? (Sorry, I haven't had time to look into it yet.)

pmelab:
After adding the test package `tsconfig.json` files to references the type errors reduced, but some remained.

https://github.com/dai-shi/waku/actions/runs/8847605068/job/24295760378

Declaration files for `react-server-dom-webpack` could not be found. I noticed that no other fixtures use `createPages` and I wanted to know if there is a connection. The history is in my last 3 commits.

pmelab:
I feel a little dumb now 😬 Thank you for fixing it!

dai-shi:
> I feel a little dumb now 😬 Thank you for fixing it!

tbh, I don't know how it fixes it the problem.
`import { createPages } from 'waku'` should work in the app code.
There must be something different in e2e/fixtures setup. (cc @himself65 who should be more familiar with this.)

packages/waku/src/router/define-router.ts
```diff
@@ -152,20 +155,26 @@ export function unstable_defineRouter(
   ) => {
     const pathConfig = await getMyPathConfig();
     const path2moduleIds: Record<string, string[]> = {};
-    for (const { pathname: pathSpec } of pathConfig) {
+
+    const promises = pathConfig.map(async ({ pathname: pathSpec, pattern }) => {
```
dai-shi, line 159: Just a preference.
```suggestion
    await Promise.all(pathConfig.map(async ({ pathname: pathSpec, pattern }) => {
```

packages/waku/src/router/define-router.ts
```diff
@@ -152,20 +155,26 @@ export function unstable_defineRouter(
...
       if (pathSpec.some(({ type }) => type !== 'literal')) {
-        continue;
+        return;
       }
       const pathname = '/' + pathSpec.map(({ name }) => name).join('/');
       const input = getInputString(pathname);
-      const moduleIds = await unstable_collectClientModules(input);
-      path2moduleIds[pathname] = moduleIds;
-    }
+      path2moduleIds[pattern] = await unstable_collectClientModules(input);
```
dai-shi, line 165: Are we sure if we can run this concurrently?

dai-shi (commented):
Thanks for working on this.

This made me feel that we should revisit `shouldSkip` mechanism before reaching v1-alpha.

dai-shi:
Please check ci error.

By the way, would you be able to join my discord server? We have #waku channel there and we can communicate about the development. https://discord.gg/MrQdmzd

packages/waku/src/router/define-router.ts
pmelab: My [last commit](https://github.com/dai-shi/waku/pull/668/commits/ce07555acfbdb465d14b07ac695288f9e517d399) was **not** an answer to this comment. I saw it just now.

But maybe it does prove it? What are your concerns collecting client modules in parallel?

dai-shi: I'm not sure what my concerns are, tbh.
(When I coded it, I thought it's safer to make it sequential intentionally. But I don't remember why.)

pmelab: The test are green, and It is about to change to static analysis anyway right?

dai-shi: Exactly!

pmelab: And at that point it does not make sense to run it concurrently anyway 😅
I just need it for the POC I'm working on.

packages/waku/src/router/define-router.ts
```diff
@@ -152,20 +155,26 @@ export function unstable_defineRouter(
...
+        path2moduleIds[pattern] = await unstable_collectClientModules(input);
+      }),
+    );
+
     const customCode = `
 globalThis.__WAKU_ROUTER_PREFETCH__ = (path) => {
   const path2ids = ${JSON.stringify(path2moduleIds)};
-  for (const id of path2ids[path] || []) {
-    import(id);
+  const pattern = Object.keys(path2ids).find((key) => new RegExp(key).test(path));
```
dai-shi, line 173: I just wondered if `find` should be `filter`, but maybe it's fine.

packages/waku/tests/path.test.ts
```diff
@@ -0,0 +1,34 @@
+import { test, describe, expect } from 'vitest';
+import { parsePathWithSlug, path2regexp } from '../src/lib/utils/path.js';
+
+function matchPath(path: string, input: string) {
+  return new RegExp(path2regexp(parsePathWithSlug(path))).test(input);
+}
+
+describe('path2regexp', () => {
```
dai-shi, line 8: Nice!

e2e/fixtures/rsc-router/src/entries.tsx
```diff
@@ -6,6 +6,7 @@ export default unstable_defineRouter(
   // getPathConfig
   async () =>
     STATIC_PATHS.map((path) => ({
+      pattern: `/^${path}/`,
```
dai-shi, line 9: Is this correct? What's `^`?

dai-shi (commented):
2nd review. Looks good overall.

e2e/render-type.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  new URL('../packages/waku/dist/cli.js', import.meta.url),
+);
+
+const cwd = fileURLToPath(new URL('./fixtures/render-type', import.meta.url));
+
+test.describe(`render type`, () => {
+  test.skip(
+    ({ browserName }) => browserName !== 'chromium',
+    'Browsers are not relevant for this test. One is enough.',
+  );
```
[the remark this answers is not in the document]
dai-shi: It was my misunderstanding. CI matrix doesn't have browser matrix.
So, maybe this is fine. I just wonder how we could organize it overall.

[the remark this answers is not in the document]
pmelab: Yeah, browsers are controlled by playwright config. We could only split up test suites in "browser is relevant" and "browser is not relevant". This is the least intrusive solution I could come up with.

e2e/fixtures/render-type/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "build": "waku build",
+    "start:dynamic": "waku start",
+    "start:static": "serve dist/public"
+  },
+  "dependencies": {
+    "react": "19.0.0-beta-4508873393-20240430",
+    "react-dom": "19.0.0-beta-4508873393-20240430",
+    "react-server-dom-webpack": "19.0.0-beta-4508873393-20240430",
+    "waku": "workspace:*",
+    "serve": "^14.2.3"
```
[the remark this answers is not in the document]
pmelab: Not sure. Technically "serve" is the runtime part here, since its used to simulate a dumb http server. Does it make a difference?

e2e/fixtures/rsc-router/src/entries.tsx
pmelab: It is not! It should be `/^${path}$/`!

pmelab: Its a regular expression.

packages/waku/src/lib/builder/build.ts
```diff
@@ -686,11 +686,18 @@ export async function build(options: {
   );
 
   const distEntries = await import(filePathToFileURL(distEntriesFile));
+
+  // TODO: proper time measuring and logging
+  let timestamp = Date.now();
+  console.log('Generating build config');
```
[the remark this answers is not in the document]
pmelab: When the static builds take longer its nice to have at least some feedback. And there is the "TODO" but it felt out of scope for this PR. Should I address this? Either remove the logs or pull in a logging library?

e2e/fixtures/rsc-router/src/entries.tsx
pmelab: But it matches correctly for this case nonetheless.

dai-shi: Shouldn't it be this? I don't know why it works.

```suggestion
      pattern: `^${path}$`,
```

packages/waku/src/lib/builder/build.ts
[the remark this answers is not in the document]
dai-shi: Yeah, it might be useful in some cases.
But, we want to reduce logs by default.
However, not a big fan of adding a library at this point.
If you plan a follow-up PR, we can leave it. Otherwise, maybe remove it for now.

e2e/render-type.spec.ts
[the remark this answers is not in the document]
dai-shi: Fair enough.

I would like to ask @himself65 if he has any idea.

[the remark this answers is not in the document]
himself65: playwright has such config to separate test case, but for now it looks good

pmelab:
> tbh, I don't know how it fixes it the problem. `import { createPages } from 'waku'` should work in the app code. There must be something different in e2e/fixtures setup. (cc @himself65 who should be more familiar with this.)

I think the issues is related to this: https://github.com/dai-shi/waku/blob/c0eb2c9260a931d20d4087948ffa3ef9cd70d87e/tsconfig.json#L21-L22

The `waku` import only includes `main.ts` and the missing types are in `types.d.ts`. `waku/*` contains everything. In areal application this is not an issue because waku itself is not typechecked in this case.

e2e/fixtures/rsc-router/src/entries.tsx
pmelab: You are right ...

pmelab: It works because it has no effect in this case. Its just there to satisfy typescript.

packages/waku/src/lib/builder/build.ts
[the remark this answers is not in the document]
pmelab: If we don't add a logging library, there is no follow up PR. I will just leave a `TODO`.

e2e/fixtures/render-type/package.json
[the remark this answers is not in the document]
dai-shi: Not much.

dai-shi (approved):
Looks good to me!
Thanks so much for the improvement. Let's keep going.
(And, I really want to refactor many things. 💪 )

pmelab:
> Hm, isn't that types only? But, yeah something like that causes not resolving to `main.react-server.ts`.

The problem is just types. I appears as soon as a fixture imports something from `waku`, and not `waku/[whatever]`.
I will open another PR, along with another typechecking issue that I've seen.

PR #670, opened 2024-04-25
docs: update example (yarn pnp)

bysxx:
This example is an example of adding yarn's zero installation to the waku framework,
which is directly used by the company you work with in this same structure, and made PR to share.


dai-shi (approved):
Thanks! I love seeing the community evolve!

PR #672, opened 2024-04-26
prefetch on view (<Link />

bysxx:
This is the props I needed. If this component is exposed to the screen, prefetch target route


packages/waku/src/router/client.ts
```diff
@@ -144,6 +144,7 @@ export type LinkProps = {
   notPending?: ReactNode;
   children: ReactNode;
   unstable_prefetchOnEnter?: boolean;
+  unstable_prefetchOnView?: boolean;
```
dai-shi, line 147: As long as it's marked `unstable`, it feels safe to me. We may revisit "prefetching" thing later.

packages/waku/src/router/client.ts
```diff
@@ -166,6 +168,32 @@ export function Link({
...
+        { threshold: 0.1 },
+      );
+
+      observer.observe(ref.current);
+
+      return () => {
+        observer.disconnect();
+      };
+    }
+  }, [unstable_prefetchOnView, prefetchRoute, to, ref]);
```
dai-shi, line 196: `ref` isn't necessary.
`prefetchRoute` is not stable, you want just `router` for deps.

packages/waku/src/router/client.ts
```diff
@@ -166,6 +168,32 @@ export function Link({
         throw new Error('Missing Router');
       };
   const [isPending, startTransition] = useTransition();
+  const ref = useRef<HTMLAnchorElement>(null);
```
dai-shi, line 171: I'm not sure if we need `null` here.

dai-shi (commented):
Thanks for working on this. Added some comments.

bysxx:
@dai-shi I don't know if I fully understand what you said, but I applied it first! Please feedback!

dai-shi (approved):
LGTM
You got it correctly.

PR #675, opened 2024-04-26
fix: private dir hmr

### LINKED ISSUE #592: support listening to private directory for changes and trigger RSC hot reload
Right now changes to `./private` files requires a manual refresh to load the latest data in local development.

Aslemammad:
Resolves #592 

We did not react to changes in the private dir. 

Since we're sure they're server only, let's treat them as rsc updates, so the changes get reflected on the client.

### DIFF
--- a/packages/waku/src/lib/renderers/dev-worker-impl.ts
+++ b/packages/waku/src/lib/renderers/dev-worker-impl.ts
@@ -173,6 +173,7 @@ const hotUpdateCallback = (payload: HotUpdatePayload) => {
   const mesg: MessageRes = { type: 'hot-update', payload };
   parentPort!.postMessage(mesg);
 };
+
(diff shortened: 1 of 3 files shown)


packages/waku/src/lib/renderers/dev-worker-impl.ts
```diff
@@ -168,20 +169,21 @@ const handleGetSsrConfig = async (
 
 const dummyServer = new Server(); // FIXME we hope to avoid this hack
 
+const hotUpdateCallback = (payload: HotUpdatePayload) => {
+  const mesg: MessageRes = { type: 'hot-update', payload };
+  parentPort!.postMessage(mesg);
+};
```
dai-shi, line 175:
```suggestion
};

```

dai-shi (approved):
LGTM!

Issue #429, opened 2024-01-27
fail to resolve packages imported from another package in a monorepo


emilgpa:
file located in `monorepo/ui/components/Example/Example.tsx`
```tsx
import stylex from "@stylexjs/stylex"

const styles = stylex.create({
    base: {
        borderRadius: 40,
    },
})

export const Example = () => {
    return (
        <div {...stylex.props(styles.base)}>
            <h1>Example</h1>
        </div>
    )
}
```

file located in `monorepo/web/src/root-layout.tsx`
```tsx
import { Example } from "@my-org/ui/components/Example/Example"
import stylex from "@stylexjs/stylex"

const styles = stylex.create({
    base: {
        padding: 10,
    },
})

export const RootLayout = async ({ children }: RootLayoutProps) => {
    return (
        <div id="__waku">
            <Header />
            <main>
                  <Example />
                  <div {...stylex.props(styles.base)}>
                      {children}
                  </div>
            </main>
            <Footer />
        </div>
    )
}
```

When load the page in browser I get the follow error:
```
Error: Error: Failed to load url ../../@stylexjs/stylex (resolved id: ../../@stylexjs/stylex) in ...
```

In this example I use stylex but It occurs with any external packages. The external package in the same package works but if it is loaded from another package in the same monorepo then it fails with that message error.

dai-shi:
So, is this an issue that happens only with monorepo?
@himself65 do you have any insight?

himself65:
What package manager are you using? and what's your vite config

emilgpa:
@himself65 I'm using `pnpm` and the `vite.config.js` is:

```ts
import path from "node:path"
import url from "node:url"
import { defineConfig } from "vite"
import { stylexPlugin } from "vite-plugin-stylex-dev"
import tsconfigPaths from "vite-tsconfig-paths"

const workspacesRoot = path.resolve(
    path.dirname(url.fileURLToPath(import.meta.url)),
    "..",
)

export default defineConfig(({ mode }) => {
    return {
        plugins: [
            tsconfigPaths({
                projects: [
                    "./tsconfig.json",
                    "../browser-api/tsconfig.json",
                    "../browser-lib/tsconfig.json",
                    "../browser-ui/tsconfig.json",
                ],
            }),
            stylexPlugin({
                unstable_moduleResolution: {
                    type: "commonJS",
                    rootDir: workspacesRoot,
                },
            }),
        ],
    }
})
```

himself65:
[...]

PR #679, opened 2024-04-27
fix: build with customised entriesJs

nihgwu:
We always generate it as `entries.js` in dist folder, which makes sense if we have something like `entriesJs: /foo/bar/baz.start.ts`

### DIFF
--- a/packages/waku/src/config.ts
+++ b/packages/waku/src/config.ts
@@ -49,7 +49,8 @@ export interface Config {
    */
   mainJs?: string;
   /**
-   * The entries.js file relative to srcDir or distDir.
+   * The entries.js file relative to srcDir.
+   * Will be generated as `entries.js` to distDir.
    * The extension should be `.js`,
    * but resolved with other extensions in the development mode.
    * Defaults to "entries.js".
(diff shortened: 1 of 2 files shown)


dai-shi:
Good point. I didn't mean to support a path in `entriesJs`.
Looking at this, I wonder if we should support customizing `entriesJs`. 🤔 
I thought it's for some advanced users or library authors, but they should then create a vite plugin.

nihgwu:
I need it to run waku at background, not in user's code base, so this is exactly what I need, curious how it can be achieve with vite plugin

dai-shi:
That's interesting to hear.
So, do you want to have entries file somewhere else from `src`?
Let me see if I can do something.
As this is DEV/BUILD only, we could use environment variables.

nihgwu:
Yes, I don't need `pages` for routing, just `entries.js` is enough, and even is I just change the name like `entries.js` -> `start.js`, it still fails, so we do need to fix, I understand you point to just use it to change name, but I think my approach can serve more cases, and for those who want to generate it to a different file, just `mv` after built

dai-shi:
Yeah, it's never tested, and not intended for normal users. Glad to hear you have an interesting use case. I'll work on it and ask you to test it.

dai-shi:
#682 is still work in progress, but you can try it.

To customize `src/entries.tsx`, we need a custom vite plugin. However, I don't think if works if the files imported from there is outside the `src` folder. #429 is still to be solved.

dai-shi:
This PR will be superseded by #682, but let's keep it open for a while.

PR #682, opened 2024-04-28
Avoid exposing configs that are not configurable

dai-shi:
supersedes #679

Implementation-wise, there are many hacks. We should improve it later.


himself65 (commented):
image

PR #685, opened 2024-04-28
fix: default nodejs runtime to 20

himself65:
Vercel changed their default nodejs version to 20, waku will throw error when deploy to vercel if it's 18


dai-shi (approved):
LGTM

dai-shi:
When should we drop supporting Node 18 by Waku itself?

himself65:
> When should we drop supporting Node 18 by Waku itself?

when it's not LTS?

PR #671, opened 2024-04-25
update react beta

dai-shi:

PR #688, opened 2024-05-01
chore: bump to react@19-beta

himself65:
Closes: https://github.com/dai-shi/waku/pull/671
Upstream: https://github.com/facebook/react/pull/28921


himself65:
Whoops, async local storage is not working for some reasons


image

himself65:
```
var supportsRequestStorage = "function" === typeof AsyncLocalStorage, requestStorage = supportsRequestStorage ? new AsyncLocalStorage() : null;
```

I think the reason is, as I said, it edge / worked (Cloudflare worker) AsyncLocalStorage is under globalThis, so React will access it by only `AsyncLocalStorage`; that's why it's false here

https://github.com/facebook/react/blob/4508873393058e86bed308b56e49ec883ece59d1/packages/react-server/src/forks/ReactFlightServerConfig.dom-edge.js#L14-L17

I checked diff and there's no AsyncLocalStorage related changes, so I think there's some new timing issue now

/cc @dai-shi

himself65:
OK I found the issue, file name changed, but I think don't use workerd is a better way

packages/waku/src/lib/plugins/vite-plugin-rsc-entries.ts
```diff
@@ -48,7 +48,7 @@ export const loadConfig = async () => ({});
     transform(code, id) {
       if (
         // FIXME this is too hacky and not the right place to patch
-        id.endsWith('/react-server-dom-webpack-server.edge.production.min.js')
+        id.endsWith('/react-server-dom-webpack-server.edge.production.js')
```
dai-shi, line 51: Haha, of course this is brittle. Thanks for the fix!

dai-shi (approved):

PR #689, opened 2024-05-01
SSG: partial builds

pmelab:
Introduces a `--partial` flag that will skip rendering of already existing static RSC and HTML files.

## Use case

Content updates of larger static sites should not re-render every page.

## Notes

* Consciously avoids dealing with deleting files (cache invalidation). I think this can be solved much better with userland domain knowledge.
* Bundle builds should probably be skipped as well, but are too tangled at this point. At least they don't scale with the amount of static pages.


packages/waku/src/lib/builder/build.ts
```diff
@@ -696,6 +718,7 @@ export async function build(options: {
     config,
     distEntries,
     buildConfig,
+    options.partial || false,
```
dai-shi, line 721: For consistent coding style,
```suggestion
    !!options.partial,
```

packages/waku/src/lib/builder/build.ts
```diff
@@ -705,6 +728,7 @@ export async function build(options: {
     buildConfig,
     getClientModules,
     clientBuildOutput,
+    options.partial || false,
```
dai-shi, line 731:
```suggestion
    !!options.partial,
```

packages/waku/src/lib/builder/build.ts
```diff
@@ -432,6 +441,10 @@ const emitRscFiles = async (
           config.rscPath,
           encodeInput(input),
         );
+        // In partial mode, skip if the file already exists.
+        if (partial && existsSync(destRscFile)) {
```
dai-shi, line 445: I wonder if we need to check `partial` mode here. Can we simply skip if the file exists?

packages/waku/src/lib/builder/build.ts
```diff
@@ -364,6 +371,7 @@ const buildClientBundle = async (
       rscManagedPlugin(config),
     ],
     build: {
+      emptyOutDir: !partial,
```
dai-shi, line 374: I wonder if we can make this always `true` and empty `dist` folder at the beginning of `build()` function.

packages/waku/src/cli.ts
```diff
@@ -46,6 +46,9 @@ const { values, positionals } = parseArgs({
     'with-aws-lambda': {
       type: 'boolean',
     },
+    partial: {
```
dai-shi, line 49: I feel like I'd like to start with `--experimental-partial` first, and get some user feedback.

dai-shi (commented):
first review.
the approach seems reasonable (no notion of router).
there might be a tweak in the implementation as i commented.

packages/waku/src/lib/builder/build.ts
pmelab: It would make the diff simpler. I just did not want to potentially affect the existing behavior. If you feel comfortable trying that, I have no objections.

packages/waku/src/lib/builder/build.ts
pmelab: Good point. If not partial, everything is deleted anyway.

pmelab:
> > * Consciously avoids dealing with deleting files (cache invalidation). I think this can be solved much better with userland domain knowledge.
> 
> I see, so invalidation will be done outside waku build.

Yes, I heard that cache invalidation is a hard problem. I thought long about this, but if Waku should know what to clear, we inevitably go down the data-store route. Right now it is possible to implement a fetcher that collects calls during a render and associates them with the currently rendered path.

One thing that came into my mind is a `waku experimental-clear [path]` command that simply makes sure all the correct files (html + rsc) are deleted. But the value is not tremendous. What do you think?

packages/waku/src/lib/builder/build.ts
dai-shi: Yeah, I would have said so (=avoid potentially affect the current behavior).

I just checked the vite code around emptyOutDir, and it doesn't seem it's very straightfoward as I thought.

So, let's use `emptyOutDir: !partial` this time.

dai-shi:
> One thing that came into my mind is a `waku experimental-clear [path]` command that simply makes sure all the correct files (html + rsc) are deleted. But the value is not tremendous. What do you think?

Not sure I get it. Do you mean deleting rsc and html files that are outdated?

pmelab:
> > One thing that came into my mind is a `waku experimental-clear [path]` command that simply makes sure all the correct files (html + rsc) are deleted. But the value is not tremendous. What do you think?
> 
> Not sure I get it. Do you mean deleting rsc and html files that are outdated?

Yes. The system invoking the command has to know which paths are outdated, but waku knows which files belong to that path.

`waku experimental-clear /my/page`
runs effectively
`rm dist/public/my/page/index.html && rm dist/public/RSC/my/page.txt`

dai-shi:
I see `[path]` is a path of URL.

That would be only possible if it has the knowledge of `waku/router` which is technically outside `waku` core.

pmelab:
> I see `[path]` is a path of URL.
> 
> That would be only possible if it has the knowledge of `waku/router` which is technically outside `waku` core.

Let's not open pandora's box to save approximately 14 seconds in 3rd party tools. If the `dist` directory layout changes, they will have to adapt anyway.

dai-shi:
I'm not sure if I get your point, or if you get my point.
Some examples in Waku repository doesn't use `waku/router`. It's already the case that we don't know about the `dist` structure.
So, what you are suggesting would be `--experimental-router-clear`.

What's 14 seconds?

pmelab:
> I'm not sure if I get your point, or if you get my point. Some examples in Waku repository doesn't use `waku/router`. It's already the case that we don't know about the `dist` structure. So, what you are suggesting would be `--experimental-router-clear`.
> 
> What's 14 seconds?

Sorry. By that I meant that somebody implementing a invalidation mechanism would have to make the script delete two files instead of calling one command. Which is not a huge difference. And its not worth compromising the the separation between `waku/lib` and `waku/router` for that.

packages/waku/src/cli.ts
```diff
@@ -47,6 +47,9 @@ const { values, positionals } = parseArgs({
     'with-aws-lambda': {
       type: 'boolean',
     },
+    'experimental-partial': {
```
dai-shi, line 50: 👍 Feels comfortable.

packages/waku/src/lib/builder/build.ts
```diff
@@ -120,6 +124,7 @@ const analyzeEntries = async (rootDir: string, config: ResolvedConfig) => {
       noExternal: /^(?!node:)/,
     },
     build: {
+      emptyOutDir: !partial,
```
dai-shi, line 127: Do we need this with `write: false`?

dai-shi (commented):
Second review. Why CIs are failing?

packages/waku/src/lib/builder/build.ts
pmelab: This was there before I joined. I suppose because `analyzeEntries` does not have to create a bundle?

pmelab:
> fix: build breaks without that file (?)

I'm not sure. It happened after the merge of main and https://github.com/dai-shi/waku/pull/689/commits/920a4670bd560b50431c83fcfc7abea2d583ed9c mysteriously fixed it.

packages/waku/src/lib/builder/build.ts
dai-shi: Yes, I mean, as we have `write: false`, `emtpyOutDir: !partial` might be unnecessary. Can you try?

packages/waku/src/lib/builder/build.ts
```diff
@@ -310,6 +313,7 @@ const buildSsrBundle = async (
     },
     publicDir: false,
     build: {
+      emptyOutDir: !partial,
```
dai-shi, line 316: What I'm not sure is if `partial` is the right word for this.
Isn't this mean we keep garbage files?
Anyway, that's why I preferred "experimental partial".

dai-shi (approved):
Other than that, it looks good to me.

packages/waku/src/lib/builder/build.ts
pmelab: > Isn't this mean we keep garbage files?

We do. It wont delete anything, and if a page disappears from the router definition in a subsequent build, the static build artifacts will remain. Thats why its not `update` or `incremental`.
I _thought_ that i've seen this name for this behavior, but I can't remember where. `rsync` calls it `--ignore-existing`. 🤔

pmelab: Now I remember 😄 
I got it from the Drupal config import command: https://www.drush.org/12.x/commands/config_import/#options

But the behavior is not even the same, because in this case it does add and update, not delete 🤦‍♂️ 

> There are only two hard things in Computer Science: cache invalidation and naming things.

This PR has both!

dai-shi: > This PR has both!

lol 🤣

PR #692, opened 2024-05-02
test: unit and type tests for createPages

pmelab:
I have not yet found a way to add type-tests for `CreatePage` and `CreateLayout`. Since the generic is on the function and not on the argument, the `path` property always resolves to `never`. I'm not sure if a test is even possible in this case. Any suggestions are welcome.


dai-shi:
How about something like 403f4775 ?

pmelab:
> How about something like [403f477](https://github.com/dai-shi/waku/commit/403f4775800eda3d9b37bed892a0f388a44a5a9c) ?

Haha, yeah obviously. I was tunnel-visioned on `ts-expect`.

dai-shi:
Same. I thought about using `TypeEqual` first.

dai-shi (approved):
It looks good overall. Let me know when it's ready to merge.

pmelab:
> It looks good overall. Let me know when it's ready to merge.

Lets go!

PR #695, opened 2024-05-03
Guide/Docs: Documentation on Injecting attributes

JesseKoldewijn:
Since these three topics are fairly related to each other (or atleast in the way I've been using them), I've added the following three topics to the `docs/guides` and the link to my repo where I've been using these implementations in `docs/community`. 

The topics are as follows:

1. TailwindCSS v4 - Installation + configuration
2. TailwindCSS v4 (reference to guide 1 of this list) + Shadcn/ui 
3. Stream interception middleware to inject attributes into the stream content (example inject lang + classname into the html root element) 

In any changes are required, feel free to give me notice! You're time is appreciated 👍


dai-shi (commented):
The docs structure looks good to me. 👍 

@sandren Would you be able to to review this?

docs/guides/stream-intercept.mdx
```diff
@@ -0,0 +1,95 @@
...
+    }
+  };
+};
+
+export default streamInterceptor;
+```
+
+### Step 2: Register the middleware
+
+In your `waku.config.ts` file, import the middleware and add it to the `middlewares` array:
```
dai-shi, line 61: Isn't it just `middleware`?

docs/guides/stream-intercept.mdx
```diff
@@ -0,0 +1,95 @@
...
+  // ...
+  middlewares: (cmd: "dev" | "start") => [
+    import("./src/middleware/themePreload.js"), // + this is the one we just added
+    ...(cmd === "dev"
+      ? [
+          import(
+            /* @vite-ignore */ DO_NOT_BUNDLE + "waku/middleware/dev-server"
+          ),
+        ]
+      : []),
```
dai-shi, line 79: Note: This is unavoidable for now. I wish we find a better way to simply the user config.

docs/guides/stream-intercept.mdx
nihgwu: I'd expect it `middlewares` as well

JesseKoldewijn: Oh you can also use it by just having it exported from 'src/middleware'? I've just set it up according to some examples I've found. So if there's any technicality that is missing or different in my implementation, feel free to point me to the optimal direction 👌

sandren: They mean the config property is singular `middlware`:
https://github.com/dai-shi/waku/blob/main/packages/waku/src/config.ts#L63

sandren (changes requested):
@JesseKoldewijn Thanks for the PR! 🙂

Happy to add a link to your your example repo and merge the SSR stream interception doc. The Tailwind and Shadcn guides are also much appreciated, but feel more like blog articles than reference documentation. If you would like to publish those on your blog, we would be happy to link to them in a new `community/articles.mdx` page. Thanks!

JesseKoldewijn:
@sandren I've just removed the TailwindCSS v4 + Shadcn/ui guides and updated the community link to point to a stripped down version of the demo I had linked previously. 

If any other changes are required, let me know! 😃 👍

sandren (approved):
Made some small copyedits, but everything else looks good to me. Thanks! 🙂

PR #674, opened 2024-04-26
fix: remove <base />

Aslemammad:
Resolves #649

Issue #649, opened 2024-04-04
[BUG]: wrong anchor link in the website


PerfectPan:
The website's anchor links appear to be incorrect, missing the original path, such as /blog/introducing-pages-router, as shown in the picture.

![CleanShot 2024-04-04 at 23 03 40@2x](https://github.com/dai-shi/waku/assets/24316656/0bd1d3cb-e0f8-4816-84b8-1938227215a4)

sandren:
Thanks for reporting.

The root cause is an invalid `<base>` tag on nested routes:
https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts#L80

dai-shi:
https://github.com/dai-shi/waku/blob/e8ca9ac9646dfb6b108c94832b68fc838bfeb1a0/packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts#L78-L80

Yeah, I knew it wasn't the right solution. So, it's a framework issue.
It may require someone's help. Help wanted.

Aslemammad:
working on this!

dai-shi:
by #697

PR #697, opened 2024-05-04
fix: non js assets path should add config.basePath to be relative to root path

PerfectPan:
Solve https://github.com/dai-shi/waku/issues/649.

I did some research based on https://github.com/dai-shi/waku/pull/674. The root cause is that css path in html missing / so that when brower try to join the real path, it will use the current path instead of root path.

CleanShot 2024-05-05 at 00 44 55@2x

CleanShot 2024-05-05 at 00 45 27@2x

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -514,7 +514,7 @@ const emitHtmlFiles = async (
         let htmlHead = publicIndexHtmlHead;
         if (cssAssets.length) {
           const cssStr = cssAssets
-            .map((asset) => `<link rel="stylesheet" href="${asset}">`)
+            .map((asset) => `<link rel="stylesheet" href="/${asset}">`)
             .join('\n');
           // HACK is this too naive to inject style code?
           htmlStr = htmlStr.replace(/<\/head>/, cssStr);
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -514,7 +514,10 @@ const emitHtmlFiles = async (
         let htmlHead = publicIndexHtmlHead;
         if (cssAssets.length) {
           const cssStr = cssAssets
-            .map((asset) => `<link rel="stylesheet" href="/${asset}">`)
+            .map(
+              (asset) =>
+                `<link rel="stylesheet" href="${config.basePath}${asset}">`,
+            )
             .join('\n');
           // HACK is this too naive to inject style code?
           htmlStr = htmlStr.replace(/<\/head>/, cssStr);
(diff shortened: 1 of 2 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
```diff
@@ -79,17 +79,14 @@ ${opts.htmlHead}
...
-        { tag: 'base', attrs: { href: opts.basePath } },
         {
           tag: 'script',
           attrs: { type: 'module', async: true },
           children: codeToInject,
         },
         ...(opts.cssAssets || []).map((href) => ({
           tag: 'link',
-          attrs: { rel: 'stylesheet', href },
+          attrs: { rel: 'stylesheet', href: `/${href}` },
```
dai-shi, line 89:
```suggestion
          attrs: { rel: 'stylesheet', href: `${opts.basePath}${href}` },
```

packages/waku/src/lib/builder/build.ts
```diff
@@ -514,7 +514,7 @@ const emitHtmlFiles = async (
         let htmlHead = publicIndexHtmlHead;
         if (cssAssets.length) {
           const cssStr = cssAssets
-            .map((asset) => `<link rel="stylesheet" href="${asset}">`)
+            .map((asset) => `<link rel="stylesheet" href="/${asset}">`)
```
dai-shi, line 517:
```suggestion
            .map((asset) => `<link rel="stylesheet" href="${config.basePath}${asset}">`)
```

dai-shi (changes requested):
Thanks for investigating into it! It looks valid.
We use `basePath` instead of `/`. (Though, we don't have tests and changing basePath in config might not fully work, yet. Contributions are welcome.)

PerfectPan:
> Thanks for investigating into it! It looks valid. We use `basePath` instead of `/`. (Though, we don't have tests and changing basePath in config might not fully work, yet. Contributions are welcome.)

I totally ignore the basePath option lol. I have already changed the code.

For changing basePath in config, I tried and it seems not working well.

CleanShot 2024-05-05 at 10 58 35@2x

This may involve the paths of resources in HTML files, the paths of generated files, and the paths for server hosting static resources. It seems that something is not quite right with all of them.

Would you mind I open an another issue to follow this and try to add test cases and solution later?

dai-shi:
Yeah, we have the `basePath` config but it's never used in `cli.ts`.
(I guess it would be `app.use('/test/*', ...)`.)

Please open a PR and maybe our first target would be SSG.

dai-shi (approved):
The code looks okay.

I wonder if @Aslemammad can give a review too.

Aslemammad (approved):
Clean & Simple! I love the catch.

PR #699, opened 2024-05-05
fix: build target for vercel will throw error

PerfectPan:
Solve https://github.com/dai-shi/waku/issues/700

The bug was introduced in https://github.com/dai-shi/waku/pull/682, the reason is that the logic of https://github.com/dai-shi/waku/pull/682/files#diff-6de2cc9d9f6423e3f4faa0cb1a6200112f1bc057af0db1a9362fe6117c51d18aR79-R82 was adjusted.

I think the logic now assmues the entry file is endsWith .jsx while while the dynamic import in serve-vercel.js is endsWith .js

CleanShot 2024-05-05 at 17 02 38@2x

CleanShot 2024-05-05 at 17 03 30@2x

I am not so sure it is a correct way to remove .js extension, but I added the dry run testing, maybe it can help.

### DIFF
--- a/e2e/multi-platform.spec.ts
+++ b/e2e/multi-platform.spec.ts
@@ -4,7 +4,18 @@ import { test } from './utils.js';
 import { rm } from 'node:fs/promises';
 import { expect } from '@playwright/test';
 
-const cwd = fileURLToPath(new URL('../examples/01_template', import.meta.url));
+const dryRunList = [
+  // without entries.tsx
+  {
+    cwd: fileURLToPath(new URL('../examples/01_template', import.meta.url)),
+    project: '01_template',
+  },
+  // with entries.tsx
+  {
+    cwd: fileURLToPath(new URL('../examples/20_minimal', import.meta.url)),
+    project: '20_minimal',
+  },
+];
 
 const waku = fileURLToPath(
   new URL('../packages/waku/dist/cli.js', import.meta.url),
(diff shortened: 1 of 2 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-serve.ts
```diff
@@ -38,7 +38,7 @@ export function rscServePlugin(opts: {
     config(viteConfig) {
       // FIXME This seems too hacky (The use of viteConfig.root, '.', path.resolve and resolveFileName)
       const entriesFile = resolveFileName(
-        path.resolve(viteConfig.root || '.', opts.srcDir, SRC_ENTRIES + '.js'),
+        path.resolve(viteConfig.root || '.', opts.srcDir, SRC_ENTRIES),
```
dai-shi, line 41: Can you try this?
```suggestion
        path.resolve(viteConfig.root || '.', opts.srcDir, SRC_ENTRIES + '.jsx'),
```

PerfectPan: I tried, it seems ok as well.

PerfectPan:
All the windows tests are failed. Maybe there are another issues?

dai-shi:
probably, wrap with `normalizePath`.

PerfectPan:
> probably, wrap with `normalizePath`.

Ha, you are right

dai-shi (approved):
Great it works. Thanks for your contribution!
I really appreciate it. (As noted as "FIXME", I'm not super comfortable with the current solution. Hopefully, we will revisit it.)

PerfectPan:
I see the failed test stuck on `actions/setup-node` downloading, can you rerun the test?

e2e/multi-platform.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { execSync } from 'node:child_process';
+import { fileURLToPath } from 'node:url';
+import { test } from './utils.js';
+import { rm } from 'node:fs/promises';
+import { expect } from '@playwright/test';
+
+const cwd = fileURLToPath(new URL('../examples/01_template', import.meta.url));
```
dai-shi, line 7: Can you add test with `20_minimal` too?

dai-shi (changes requested):
Sorry, I didn't notice before. Can you try this please?

e2e/multi-platform.spec.ts
PerfectPan: OK

dai-shi (approved):
LGTM
Thanks for the adjustment.

PR #707, opened 2024-05-09
fix: avoid patching global webpack require

dai-shi:
close #579 

- [x] prd
- [x] dev


packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
```diff
@@ -79,10 +77,13 @@ ${opts.htmlHead}
     },
     transformIndexHtml() {
       return [
+        // HACK: vite won't inject __vite__injectQuery anymore
```
dai-shi, line 80: @Aslemammad What does it mean???

Aslemammad: Vite optimizes `import()` so it adds `?import` to imported urls. That'd cause double module hazard! 

This way, I hack it to use a global function so it does not get optimized.

packages/waku/src/lib/middleware/dev-server.ts
```diff
@@ -139,6 +139,11 @@ const createMainViteServer = (
   };
 
   const loadServerModuleMain = async (id: string) => {
+    if (id === 'waku' || id.startsWith('waku/')) {
+      // HACK I don't know why this is necessary.
+      // `external: ['waku']` doesn't somehow work?
+      return import(id);
```
Aslemammad, line 145: Tbh, it does not feel right! I'd rather figure out how we can move default handler to use ssrLoadModule, that'd be way more robust, I guess.

But this can work too.

Aslemammad (approved):

himself65 (approved):

himself65 (approved):

packages/waku/src/lib/middleware/dev-server.ts
dai-shi: > move default handler to use ssrLoadModule

Do you have time to try it?

PR #713, opened 2024-05-16
fix: do not overwrite vite react plugin

dai-shi:
Needed for testing React Compiler.


packages/waku/src/lib/utils/merge-vite-config.ts
```diff
@@ -8,18 +8,38 @@ import {
...
   const mergedViteConfig = mergeViteConfig(
     {
       ...resolvedViteConfig,
       configFile: false,
-      // weird error around plugin duplication when removed
+      // FIXME weird error around plugin duplication when removed
       plugins: resolvedViteConfig.plugins.filter(
-        (plugin) => !plugin.name.startsWith('vite:'),
+        (plugin) =>
+          !['vite:css-post', 'vite:import-analysis', 'vite:json'].includes(
```
dai-shi, line 32: For now, removing these three is working. But, we are far from confident.
We don't know why these causes issues, and why we need to remove.
Any insights are welcome. FIXME please.

Meanwhile, let's merge.

PR #714, opened 2024-05-17
feat: use server action from client modules

### LINKED ISSUE #403: rpc call in the client side
See https://github.com/himself65/waku-openai/tree/issue-2

```shell
cp .env.template .env
pnpm run dev
```

Error:
```shell
TypeError: Cannot read properties of undefined (reading 'A')
```

image

```tsx
"use client"
import { useFormState } from "react-dom"
import { generateResponse } from '../actions/chat.js'

export function Client () {
  const [state, dispatch] = useFormState(
    generateResponse,
    null
  )
  return (
    <div>
      {state}
      <button
        onClick={() => {
          dispatch()
        }}
      >
        Generate Response
      </button>
    </div>
  )
}
```

`chat.tsx`:

```tsx
'use server'
import { getEnv } from 'waku'

const apiKey = getEnv('A')!

export async function generateResponse () {
  return `apiKey: ${apiKey}`
}
```

`.env.local`:

```txt
A=Hello
```

dai-shi:
close #403

- [x] dev
- [x] prd

Remaining
- refactor
- tests
- docs


himself65 (commented):
this is awesome

packages/waku/src/lib/builder/build.ts
```diff
@@ -135,6 +139,25 @@ const analyzeEntries = async (rootDir: string, config: ResolvedConfig) => {
...
+      rscAnalyzePlugin({ isClient: true, serverFileSet }),
+      rscManagedPlugin(config),
+    ],
+    ssr: {
+      target: 'webworker',
+      noExternal: /^(?!node:)/,
+    },
+    build: {
+      write: false,
+      ssr: true,
```
dai-shi, line 153: on second look, i'm not sure if this should be "ssr". we will need to revisit this double build anyway.

PR #715, opened 2024-05-19
fix: merging vite plugins

dai-shi:
follow up #713


packages/waku/src/lib/utils/merge-vite-config.ts
```diff
@@ -29,9 +29,12 @@ export async function mergeUserViteConfig(config: UserConfig) {
       // FIXME weird error around plugin duplication when removed
```
dai-shi, line 29: I'm not confident with this approach at all. I hope someone with more Vite experience would help us. 🙏

PR #709, opened 2024-05-13
transform all files for rsc (DEV)

dai-shi:

PR #718, opened 2024-05-21
chore: without vite-plugin-commonjs

Aslemammad:


dai-shi:
#709 follows this approach.

PR #721, opened 2024-05-25
fix (create-waku): example GitHub urls can contain `www` too

Rec0iL99:
**Current Behaviour:**

```bash
pnpm start --example https://www.github.com/dai-shi/waku/tree/main/examples/02_template_js                                                                                                                 

> create-waku@0.9.2-alpha.0 start /Users/joelmathew/WebProjects/waku-dev/waku/packages/create-waku
> node ./dist/index.js "--example" "https://www.github.com/dai-shi/waku/tree/main/examples/02_template_js"

Invalid URL: "https://www.github.com/dai-shi/waku/tree/main/examples/02_template_js". Only GitHub repositories are supported. Please use a GitHub URL and try again.
```

The url is a valid GitHub url which should be accepted by the `create-waku` utility. This pr should fix this behaviour.

### DIFF
--- a/packages/create-waku/src/helpers/example-option.ts
+++ b/packages/create-waku/src/helpers/example-option.ts
@@ -140,7 +140,10 @@ export async function parseExampleOption(
   }
 
   // NOTE check github origin
-  if (repoUrl.origin !== 'https://github.com' && repoUrl.origin !== 'https://www.github.com') {
+  if (
+    repoUrl.origin !== 'https://github.com' &&
+    repoUrl.origin !== 'https://www.github.com'
+  ) {
     console.error(
       `Invalid URL: ${red(
         `"${example}"`,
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
Hi, thanks for the suggestion.
I wasn't sure if `www.github.com` was a valid URL, but this change seems harmless.
For curiosity, how can you get such an URL?

Please check CI error: you need to run prettier.

Rec0iL99:
Hi @dai-shi, most browsers automatically add `www` to urls if the user forgets to add it for better ux.

https://github.com/dai-shi/waku/tree/main/examples/02_template_js
https://www.github.com/dai-shi/waku/tree/main/examples/02_template_js

Both the above URLs resolve to the link except one has `www` and the other doesn't.

dai-shi (approved):
Yeah, I know both work. Just curious, when you get the `www` URL.

Rec0iL99:
Just tried appending `www` to the url out of curiosity and the CLI failed. Thanks for merging 👍

PR #724, opened 2024-05-27
fix(hmr): replace old module for SSR

### LINKED ISSUE #690: HMR with client component seems flakey
When updating a client component, either the client or server component seems to get stuck in its previous state. From what I've found it seems to get resolved when restarting the dev server, but nothing other than that.

Example code of client component:
```
"use client";

import { Link } from "../core/Link";
import { Button } from "../ui/button";

import { useRouter_UNSTABLE as useRouter } from "waku";

const routes = [
	{
		path: "/route-1",
		title: "hello",
	},
	{
		path: "/route-2",
		title: "world",
	},
];

const ReproductionComponent = () => {
	const router = useRouter();
	const pathname = router.path;

	return (
		<div className="flex items-center justify-center gap-2">
			{routes.map((route) => {
				if (route.path === pathname) return null;
				return (
					<Button key={route.title} asChild className="w-max">
						<Link to={route.path}>{route.title}</Link>
					</Button>
				);
			})}
		</div>
	);
};
export default ReproductionComponent;
```

In this component you only have to change a single className for instance to trigger a hydration error.

dai-shi:
close #690


dai-shi:
@Aslemammad Can you review it?

packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -126,10 +126,11 @@ export function rscHmrPlugin(): Plugin {
     handleHotUpdate({ file }) {
       const moduleLoading = (globalThis as any).__webpack_module_loading__;
       const moduleCache = (globalThis as any).__webpack_module_cache__;
+      if (file.startsWith(viteServer.config.root)) {
+        file = file.slice(viteServer.config.root.length);
+      }
       const id = filePathToFileURL(file);
-      if (moduleLoading.has(id) && moduleCache.has(id)) {
-        moduleLoading.delete(id);
```
Aslemammad, line 131: Why don't we delete anymore?

dai-shi: I thought checking `moduleCache.has(id)` is unnecessary. What do you think?

Aslemammad: I do not remember exactly where the condition might happen, but I think having it might be safer! So a new module gets loaded later anyway.

dai-shi: Hm? I think removing it might be safer.

PR #728, opened 2024-05-28
feat(create-waku): install dependencies automatically when creating a new waku

Rec0iL99:
Workflow:

```bash
> pnpm start                                                                                                                        

> create-waku@0.9.2-alpha.0 start /Users/joelmathew/WebProjects/waku-dev/waku/packages/create-waku
> node ./dist/index.js

✔ Project Name … waku-project
Setting up project...

Installing dependencies by running pnpm install...
Scope: all 32 workspace projects
Done in 1.6s

Done. Now run:

cd waku-project
pnpm dev
```

### DIFF
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -206,11 +206,14 @@ async function init() {
 
   // 1. check packageManager
   // 2. and then install dependencies
+  console.log();
+  console.log(`Installing dependencies by running ${commands.install}...`);
+
   process.chdir(targetDir);
   const [, installFlag] = commands.install.split(' ');
 
   // synchronously install all the deps
-  const installProcess = sync('npm', [installFlag!], {
+  const installProcess = sync(packageManager, [installFlag!], {
     stdio: 'inherit',
   });
   const error = installProcess.error;
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -16,7 +16,6 @@ import {
   parseExampleOption,
   downloadAndExtract,
 } from './helpers/example-option.js';
-import process from 'node:process';
 import { spawn } from 'node:child_process';
 
 const userAgent = process.env.npm_config_user_agent || '';
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -17,7 +17,7 @@ import {
   downloadAndExtract,
 } from './helpers/example-option.js';
 import process from 'node:process';
-import { sync } from 'cross-spawn';
+import { spawn } from 'node:child_process';
 
 const userAgent = process.env.npm_config_user_agent || '';
 const packageManager = /pnpm/.test(userAgent)
@@ -32,7 +32,7 @@ const commands = {
     create: 'pnpm create waku',
   },
   yarn: {
-    install: 'yarn install',
+    install: 'yarn',
     dev: 'yarn dev',
     create: 'yarn create waku',
   },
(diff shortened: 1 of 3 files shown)


packages/create-waku/src/index.ts
```diff
@@ -202,15 +204,32 @@ async function init() {
...
-  console.log(`${bold(green(commands.install))}`);
-  console.log(`${bold(green(commands.dev))}`);
   console.log();
+  console.log(`Installing dependencies by running ${commands.install}...`);
+
+  process.chdir(targetDir);
+  const [, installFlag] = commands.install.split(' ');
+
+  // synchronously install all the deps
+  const installProcess = sync(packageManager, [installFlag!], {
```
dai-shi, line 216: Let's try `exec(commands.install,`.

dai-shi (commented):
Hi, thanks for working on it! Please see comments.

cc @ojj1123 would you like to give a review?

packages/create-waku/src/index.ts
```diff
@@ -202,15 +204,32 @@ async function init() {
...
-
-  console.log(`\nDone. Now run:\n`);
-  console.log(`${bold(green(`cd ${targetDir}`))}`);
-  console.log(`${bold(green(commands.install))}`);
-  console.log(`${bold(green(commands.dev))}`);
   console.log();
+  console.log(`Installing dependencies by running ${commands.install}...`);
+
+  process.chdir(targetDir);
+  const [, installFlag] = commands.install.split(' ');
```
ojj1123, line 213: Why did you split the install command?
Is `installFlag` just `'install'`, right?

ojj1123 (commented):
Good works! I review for something.

packages/create-waku/src/index.ts
dai-shi: Okay, if that's case, let's use `spawn` from `child_process`.

dai-shi (changes requested):

dai-shi:
@Rec0iL99 Are you still around?

Rec0iL99:
> @Rec0iL99 Are you still around?

@dai-shi sorry to keep this open for a while now. I had intended to finish this last weekend but got caught up with some work. I'll finish this by the weekend.

packages/create-waku/src/index.ts
himself65, line 213: just use 'install'

himself65 (changes requested):

packages/create-waku/src/index.ts
```diff
@@ -16,6 +16,8 @@ import {
   parseExampleOption,
   downloadAndExtract,
 } from './helpers/example-option.js';
+import process from 'node:process';
```
himself65, line 19: `process` is a global environment, you don't need to mport it explicitly

himself65 (changes requested):

packages/create-waku/src/index.ts
```diff
@@ -202,15 +204,40 @@ async function init() {
...
-  console.log(`\nDone. Now run:\n`);
-  console.log(`${bold(green(`cd ${targetDir}`))}`);
-  console.log(`${bold(green(commands.install))}`);
-  console.log(`${bold(green(commands.dev))}`);
   console.log();
+  console.log(`Installing dependencies by running ${commands.install}...`);
+
+  process.chdir(targetDir);
+
+  const installProcess = spawn(packageManager, ['install']);
```
himself65, line 214:
```suggestion
  const installProcess = spawn(packageManager, ['install'], { encoding: 'utf8' });
```

himself65 (changes requested):

packages/create-waku/src/index.ts
```diff
@@ -202,15 +204,40 @@ async function init() {
...
+
+  installProcess.stdout.setEncoding('utf8');
+  installProcess.stdout.on('data', (data) => {
+    console.log(data);
```
dai-shi, line 218: https://github.com/dai-shi/waku/pull/728#discussion_r1620085498

Does it show the indicator as expected? cc @ojj1123

packages/create-waku/src/index.ts
Rec0iL99: Hi @himself65, `encoding` is not a property that exists on the options for `spawn`.

himself65: first to know this

himself65: Oh, I think if you just wanna bypass the output, use `stdio: 'inherit'`

himself65 (changes requested):

packages/create-waku/src/index.ts
ojj1123: Yes. I could see the indicator as expected!

ojj1123: [The docs is explaining a difference between `stdio: "inherit"` and `stdio: "pipe"`](https://nodejs.org/api/child_process.html#optionsstdio) but it's a bit confusing for me.
I've tried to use `stdio: "inherit"`, it could show the loading indicator.

### `stdio: "pipe"`(default)
- child stdios(stdin, stdout, stderr) and parent ones are mapped(piping) each other. But if we want to access the data, we have to use the events like this:
```js
childProcess.stdout.on("data", (data) => {
  console.log(data)
})
```

### `stdio: "inherit"`
- parent's stdios are handed off to the child process. That is, child process could use parent's stdios. **So we don't need to use the events.**

I think no matter we use `stdio: "pipe"` or `stdio: "inherit"`. But if we use "inherit", we don't need to use the events:

```diff
diff --git a/packages/create-waku/src/index.ts b/packages/create-waku/src/index.ts
index 1eb01d3..98ede50 100644
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -210,16 +210,8 @@ async function init() {

   process.chdir(targetDir);

-  const installProcess = spawn(packageManager, ['install']);
-
-  installProcess.stdout.setEncoding('utf8');
-  installProcess.stdout.on('data', (data) => {
-    console.log(data);
-  });
-
-  installProcess.stderr.setEncoding('utf8');
-  installProcess.stderr.on('data', (data) => {
-    console.log(data);
+  const installProcess = spawn(packageManager, ['install'], {
+    stdio: 'inherit',
   });

   installProcess.on('close', (code) => {
```

ojj1123: +) If we use `stdio: "inherit"`(or anything other than "pipe"), childProcess's `stdin`, `stdout` and `stderr` could be `null`:

스크린샷 2024-07-01 19 23 18

dai-shi:
@Rec0iL99 Are you still around?

ojj1123:
If @Rec0iL99 is busy so there is no response, it's okay to get this work.

dai-shi:
@ojj1123 Yeah, please go ahead.

ojj1123:
https://github.com/dai-shi/waku/pull/728#discussion_r1660850120

I think that it's okay to use `stdio: "pipe"`(by default).
And then I've checked other issues or reviews but I feel like there is nothing. Are there any reviews?

ojj1123:
> Yes. I could see the indicator as expected!
> https://github.com/dai-shi/waku/pull/728#discussion_r1656749444

Oh it do not show the loading indicator on installing. I got mistakes before. 😅
We should use `stdio: "inherit"`!

I couldn't push my commit to this PR, so I made a new PR 👉 https://github.com/dai-shi/waku/pull/808

dai-shi:
#808 is merged.

PR #738, opened 2024-06-04
fix: do not crash whole server when error

himself65:
The main logic is `ctx.res.body = stringToStream(`${err}`);`

Cleanup response stream when it raises an error because it's locked by react (how to release it?) otherwise hono server will crash


dai-shi (approved):
I'm not sure how to reproduce the problem, but the change seems reasonable and correct.

himself65:
> I'm not sure how to reproduce the problem

Just throw new Error in side root server component

dai-shi:
Well, I just count on you. Thanks for the fix!

PR #741, opened 2024-06-04
fix(client): avoid try-catch for use

dai-shi:
To avoid this error:
image


dai-shi:
I'm not very sure if there's a better implementation. Anyways, we need tests...

PR #743, opened 2024-06-06
add api example

dai-shi:
While we wait for #329 (which takes time to design and implement, and we believe most cases are covered by server actions), we can use the low-level middleware api.


himself65 (changes requested):
> most cases are covered by server actions

I noticed something recently that React only allows one server action running at the backend at the same time. So now I think this causes server action is not suitable for some cases if there're many users doing some operation in the same time.

> https://react.dev/reference/rsc/use-server
> Server Actions are designed for mutations that update server-side state; they are not recommended for data fetching. **Accordingly, frameworks implementing Server Actions typically process one action at a time** and do not have a way to cache the return value.

BTW, this is not same as the normal restful API, that processes multiple POST queries 
 simultaneously.

Some similar issue: https://github.com/vercel/next.js/issues/64396

dai-shi:
I wonder if it's a framework choice or a React rule.

Anyway, I believe #329 is unavoidable for us. I'll work on it, but it will take time. There are some other tasks with priority (not because of the use case, but the internal architecture.)

himself65:
> I wonder if it's a framework choice or a React rule.

I believe this is React behavior, which will prohibit race conditions in the React backend at all

PR #745, opened 2024-06-13
test: edge cases for server action

himself65:
I was trying to make vercel-ai + waku, but found some bugs, let me see if I can reproduce here


e2e/fixtures/rsc-basic/modules/ai/src/server.js
```diff
@@ -0,0 +1,28 @@
...
+async function innerAction({ action, options }, state, ...args) {
+  'use server';
+  const result = await action(...args);
+  // eslint-disable-next-line no-undef
+  console.log('wrapped action', result);
+  return result;
+}
+
+function wrapAction(action, options) {
+  return innerAction.bind(null, { action, options });
```
himself65, line 14: Wrap action will cause error for waku but okay for nextjs

@dai-shi

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -188,7 +198,46 @@ export async function renderRsc(
     ) {
       // XXX This doesn't support streaming unlike busboy
       const formData = parseFormData(bodyStr, contentType);
-      args = await decodeReply(formData);
```
himself65, line 191: There are something wrong, formData could include server references which need webpack runtime @dai-shi

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -188,7 +198,46 @@ export async function renderRsc(
...
+      const moduleMap = new Proxy({} as Record<string, ImportManifestEntry>, {
+        get(_target, rsfId: string): ImportManifestEntry {
+          const [fileId, name] = rsfId.split('#') as [string, string];
+          // fixme: race condition, server actions are not initialized in the first time
```
himself65, line 204: As said, I think we should initialize all server actions before running RSC otherwise it will crash for the first time

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -188,7 +198,46 @@ export async function renderRsc(
...
+              );
+            } else {
+              moduleLoading.set(
+                fileId,
+                opts
+                  .loadServerFile(filePathToFileURL(fileId))
+                  .then((m: any) => {
+                    moduleCache.set(
+                      fileId,
+                      Object.fromEntries(m['__waku_serverActions']),
```
himself65, line 227: Looks ugly

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -188,7 +198,46 @@ export async function renderRsc(
...
+                    );
+                  }),
+              );
+            }
+          }
+          return {
+            id: fileId,
+            chunks: [],
```
himself65, line 235: `chunks: [fileId]` is not necessary after reading the source code fo next.js

packages/waku/src/lib/renderers/rsc-renderer.ts
himself65: for next.js, they have a global server action map in the runtime

https://github.com/vercel/next.js/blob/86dbcbea431f2fd85d94a72a9c2d35781154a5d7/packages/next/src/server/app-render/encryption-utils.ts#L138-L158

packages/waku/src/lib/renderers/rsc-renderer.ts
himself65: https://github.com/facebook/react/blob/ddcecbbebf6af3fa32a323c0591dad1f63587aeb/packages/react-server/src/ReactFlightReplyServer.js#L927-L941

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -5,10 +5,25 @@ import { EXTENSIONS } from '../config.js';
 import { extname } from '../utils/path.js';
 import { parseOpts } from '../utils/swc.js';
 
-const collectExportNames = (mod: swc.Module) => {
+const collectExportNames = (
```
himself65, line 8: Here are other issues:

1. non export sever action could cause runtime error
2. nested server action could cause runtime error

```ts
const actionsMap = {
  foo: () => {
     "use server"
// ...
  } 
}
```

Those are exist in vercel/ai/rsc source so we need to support them

himself65 (commented):
This PR little mess up, but should separate into three parts: 

1. Test some RSC edge cases
2. RSC transform plugin
3. RSC module map resolution

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
dai-shi: Okay, I missed such use cases.
This pattern, inline "use server", is only allowed in the server files, right?

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -223,40 +238,46 @@ const transformServer = (
...
 import { registerClientReference } from 'react-server-dom-webpack/server';
 `;
-    for (const name of exportNames) {
+    for (const name of [...exports, ...internal]) {
       newCode += `
 export ${name === 'default' ? name : `const ${name} =`} registerClientReference(() => { throw new Error('It is not possible to invoke a client function from the server: ${getClientId(id)}#${name}'); }, '${getClientId(id)}', '${name}');
 `;
     }
     return newCode;
-  } else if (hasUseServer) {
```
dai-shi, line 236: I'm not sure if we can remove this branch. It's for another use case unless I'm mistaken.

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -223,40 +238,46 @@ const transformServer = (
...
+      if (name === 'default') {
+        continue;
+      }
       newCode += `
 if (typeof ${name} === 'function') {
   registerServerReference(${name}, '${getServerId(id)}', '${name}');
 }
 `;
+      if (internal.has(name)) {
+        newCode += `export { ${name} };\n`;
```
dai-shi, line 276: This doesn't feel correct. Possible name collisions.

packages/waku/src/lib/renderers/html-renderer.ts
```diff
@@ -260,17 +260,7 @@ export const renderHtml = async (
...
-      Record<
-        string,
-        {
-          id: string;
-          chunks: string[];
-          name: string;
-        }
-      >
-    >,
+    {} as Record<string, Record<string, ImportManifestEntry>>,
```
dai-shi, line 263: Nice, but maybe we should have a separate PR.

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -12,6 +12,16 @@ import { parseFormData } from '../utils/form.js';
 import { streamToString } from '../utils/stream.js';
 import { decodeActionId } from '../renderers/utils.js';
 
+// HACK for react-server-dom-webpack without webpack
```
dai-shi, line 15: It seems like we should finish up #707 before this PR.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: Alright. It's reasonable that it requires the second argument `moduleMap`.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: I still need to understand the issue, but it feels like we want lazy evaluation.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: Yeah... I hope we can improve it.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: Are you sure? I mean waku's approach might be different from nextjs's one.

dai-shi (commented):
This looks like a good improvement.
If possible, we would like to split into multiple PRs and merge gradually.
(And, we definitely need unit tests for transformation. 😓 )

himself65:
> Are you sure? I mean waku's approach might be different from nextjs's one.

It depends on what's the real behavior from webpack, I need double check that

packages/waku/src/lib/renderers/html-renderer.ts
dai-shi: https://github.com/dai-shi/waku/pull/769

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
dai-shi: Hm, I made a test case #771, but it looks like it's transforming well.
Can you give me a failing case?

This is also tried:
```js
const actionsMap = createActions({
  foo: () => {
     "use server"
     // ...
  } 
})
```

dai-shi:
I reverted the src code, but why do all e2e tests fail... should only a new one fail??

himself65:
no it's all basic rsc test, I think something is still wrong

himself65:
For this input:

```tsx
const AI = createAI({
	actions: {
		foo: async function() {
			"use server";
			return 0;
		}
	}
});
export { AI }
```

should emit something like

```tsx
const AI = createAI({actions: {foo: __waku__server__1 });
export { AI };
export const __waku__server__1 = registerServerReference(async function() {
	'use server';
	return 0;
}, 'file_id', '1');
```

himself65:
working on `decodeReply`

himself65:
PR is to complex to rebease/merge, create a new one https://github.com/dai-shi/waku/pull/785

PR #746, opened 2024-06-13
feat: expose hash from `useRouter`

pmelab:
## What was changed?

Expose the current url hash via `UNSTABLE_useRouter`. On the server, its always an empty string, but gets populated on the client after the initial render.

## Why was it changed?

Otherwise inconvenient `useEffect`-gymnastics are necessary to get the current hash reliably into a client component.

## Notes

In the process of writing tests I found another potential issue with `searchParams`.

https://github.com/dai-shi/waku/pull/746/files#diff-a8dab8847d02ba73c6b66e738495556ddda5d2dd01b70322aae006f7669ea237R6-R11

I'm not sure if we should move the double-creation of `URLSearchParams` into the `useRouter` hook. Given some advice, I can try to fix that as well.


packages/waku/src/router/common.ts
```diff
@@ -1,6 +1,7 @@
 export type RouteProps = {
   path: string;
   searchParams: URLSearchParams;
+  hash?: string;
```
dai-shi, line 4: I originally thought of adding `hash`, but wasn't confident if it worked correctly.
Does it work with `router.push('foo#bar')`?

Also, I want to avoid optional:
```suggestion
  hash: string;
```

dai-shi (commented):
Thanks for the suggestion!

e2e/fixtures/use-router/src/TestRouter.tsx
```diff
@@ -0,0 +1,37 @@
+'use client';
+
+import { Link, useRouter_UNSTABLE } from 'waku';
+
+export default function TestRouter() {
+  const router = useRouter_UNSTABLE();
+  // TODO: Why is this necessary? On the client `searchParams`
+  //       is a URLSearchParams object, but on the server it's an array.
```
dai-shi, line 8: I didn't know it's an array on the server. How can it be??

pmelab:
To be transparent, this is the code that we save in the consumer:

```typescript
  const router = useRouter();
  const [hash, setHash] = useState('');
  useEffect(() => {
    setHash(window.location.hash);
  }, [router]);
```

Its not gigantic, but `useEffect` has a bad reputation lately.

e2e/fixtures/use-router/src/TestRouter.tsx
pmelab: Not sure, but without that, it breaks with `get is not a function`. And the debugger showed me an array. Maybe node version differences?

packages/waku/src/router/common.ts
pmelab: > Does it work with router.push('foo#bar')?

Yes, its even covered in the integration test case:
https://github.com/dai-shi/waku/pull/746/files#diff-a8dab8847d02ba73c6b66e738495556ddda5d2dd01b70322aae006f7669ea237R31-R33

e2e/fixtures/use-router/src/TestRouter.tsx
dai-shi: Can you dig into it further? If it's a bug in Waku, we would like to fix it.

packages/waku/src/router/client.ts
```diff
@@ -318,9 +318,22 @@ const RouterSlot = ({
...
+        // so it matches the server-rendered version. Otherwise
+        // there might be hydration errors.
+        hash: '',
+      }) as RouteProps,
   );
+
+  // Update the route post-load to include the current hash.
+  useEffect(() => {
+    setRoute(parseRoute(new URL(window.location.href)));
+  }, [setRoute]);
```
dai-shi, line 335: I think we can omit `setRoute` in deps.
I wonder if we should add deps at all.

The safe bed would be to add `href`.
```suggestion
  }, [window.location.href]);
```

packages/waku/src/router/client.ts
```diff
@@ -318,9 +318,22 @@ const RouterSlot = ({
...
-  const [route, setRoute] = useState(() =>
-    parseRoute(new URL(window.location.href)),
+  const [route, setRoute] = useState(
+    () =>
+      ({
+        ...parseRoute(new URL(window.location.href)),
+        // Make sure there is no hash on the initial route,
+        // so it matches the server-rendered version. Otherwise
+        // there might be hydration errors.
+        hash: '',
```
dai-shi, line 328: It looks a bit hacky. Maybe we make `parseRoute` to accept an option?

dai-shi (commented):
Sorry for the delay.
It looks good overall. Comments below.

e2e/fixtures/use-router/src/TestRouter.tsx
pmelab: Sorry, took me some time to get back to this. I was not able to pinpoint whats the problem. This is what I have found.

At this point, `searchParams` is still an `URLSearchParams` object:
https://github.com/dai-shi/waku/blob/f61b264a466bec88c3c3bfcc5588a83bd39ea4c9/packages/waku/src/router/define-router.ts#L217-L228

But when `ServerRouter` gets rendered, it has become `[]`:
https://github.com/dai-shi/waku/blob/f61b264a466bec88c3c3bfcc5588a83bd39ea4c9/packages/waku/src/router/client.ts#L530-L552

My guess is that the route definition gets serialized by react when passed into the `ServerRouter` client component and the empty `URLSearchParams` map gets turned into `[]`?

With https://github.com/dai-shi/waku/pull/746/commits/ddb40868b4bb75f29218aab781ba5643fb0ec68b I removed the additional construction of `URLSearchParams` to see if it also happens on CI, which it does.

pmelab: @dai-shi : i tried to find out more, but basically thats it. An empty `URLSearchParams` object gets passed as a prop into `createElement` and inside the `ServerRouter` its an empty array. It has to be reacts prop-handling. Maybe it would be safer to stringify `searchParams`?

dai-shi: Yeah, you are absolutely right.

repro:
```bash
$ npm create waku@next
$ cd ...
$ npm i
$ cat foo.cjs
const { renderToPipeableStream } = require('react-server-dom-webpack/server');

const data = new URLSearchParams();

renderToPipeableStream(data).pipe(process.stdout);
$ node --conditions react-server foo.cjs
0:[]
```

I will open another PR to make it a string and see how it goes.

dai-shi: #799

dai-shi:
@pmelab Hi, can you merge `main`?

packages/waku/src/router/client.ts
```diff
@@ -48,15 +48,19 @@ const normalizeRoutePath = (path: string) => {
   return path;
 };
 
-const parseRoute = (url: URL): RouteProps => {
+const parseRoute = (url: URL, ignore_hash: boolean = false): RouteProps => {
```
dai-shi, line 51: Just for preference:
```suggestion
const parseRoute = (url: URL, ignoreHash = false): RouteProps => {
```

packages/waku/src/router/client.ts
```diff
@@ -301,8 +305,15 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
   const refetch = useRefetch();
 
   const [route, setRoute] = useState(() =>
-    parseRoute(new URL(window.location.href)),
+    parseRoute(new URL(window.location.href), true),
```
dai-shi, line 308: So, `ignoreHash` is used only here.

packages/waku/src/router/client.ts
```diff
@@ -301,8 +305,15 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
...
   );
+
+  // Update the route post-load to include the current hash.
+  useEffect(() => {
+    setRoute(parseRoute(new URL(window.location.href)));
```
dai-shi, line 313: Do we not `ignoreHash` here?

dai-shi (commented):
1st review

packages/waku/src/router/client.ts
pmelab: No, thats where we want to get it, as the comment states.

packages/waku/src/router/client.ts
pmelab: Yes, without this, there are hydration errors, because the server has no hash, but the first client render does.

dai-shi: Oh, I see. Can you leave some comments in the code?
I might refactor things around here in the future.

packages/waku/src/router/client.ts
pmelab: I left a note in there.

packages/waku/src/router/client.ts
pmelab: I added some explanations there

packages/waku/src/router/client.ts
```diff
@@ -301,8 +305,19 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
...
+    // to ignore the hash, because on server side there is none.
+    // Otherwise there will be a hydration error.
+    // The client side route, including the hash, will be updated in the effect below.
+    parseRoute(new URL(window.location.href), true),
   );
+
+  // Update the route post-load to include the current hash.
+  useEffect(() => {
+    setRoute(parseRoute(new URL(window.location.href)));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
dai-shi, line 318: Hm, I wonder why we need the exception.

packages/waku/src/router/client.ts
```diff
@@ -301,8 +305,19 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
...
+    // This is the first initialization of the route, and it has
+    // to ignore the hash, because on server side there is none.
+    // Otherwise there will be a hydration error.
+    // The client side route, including the hash, will be updated in the effect below.
+    parseRoute(new URL(window.location.href), true),
   );
+
+  // Update the route post-load to include the current hash.
+  useEffect(() => {
+    setRoute(parseRoute(new URL(window.location.href)));
```
dai-shi, line 317: This actually triggers re-renders every time, even if the hash is `''`. We should probably improve it.

dai-shi (commented):
3rd review.

dai-shi:
working on it.

dai-shi (approved):
Should be good.

PR #748, opened 2024-06-17
load config for start command

hipstersmoothie:
`runStart` wasn't use values from `wake.config.ts`

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -81,7 +81,7 @@ if (values.version) {
       await runBuild();
       break;
     case 'start':
-      await runStart({});
+      await runStart();
       break;
     default:
       if (cmd) {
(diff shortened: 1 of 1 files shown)


dai-shi:
Nice catch. Though I'm not sure if we should make the config available for prd. In the long run, we should probably support config only for dev&config. It's not fixed. Feedback is welcome.

hipstersmoothie:
I'm building a tool on top of waku and really need the ability to set distDir. Maybe it could use the value set in the Vite.config.js?

himself65:
> Though I'm not sure if we should make the config available for prd.

I think we should have, it's very common like CI to generate dist file in another place and different workers host it

dai-shi:
Fair enough. My concern is that not everything is configurable in prd.
So, my plan is to merge this PR in the meantime, and eventually revisit and redesign the config architecture from scratch (will be breaking change anyway).

dai-shi (approved):
The change looks clean. Thanks for your contribution!

hipstersmoothie:
Thank you!

PR #758, opened 2024-06-24
ci: use .nvmrc as default node version

himself65:


.github/workflows/cd.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: pnpm/action-setup@v2
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version-file: '.nvmrc'
```
dai-shi, line 19: I'd simply write `22`. Not everyone uses nvm.

.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       matrix:
         shared: [1, 2, 3, 4]
         os: [ubuntu-latest, windows-latest, macos-latest]
-        version: [18.17.0, 20.8.0]
+        version: [18.17.0, 20.8.0, 22.3.0]
```
dai-shi, line 63: Doesn't it make the CI time 1.5 times longer?

dai-shi (changes requested):

dai-shi:
I don't find the necessity of using .nvmrc.
Adding node 22 in CI is probably good, but not sure if we should run for all pushes.

Closing for now.

PR #765, opened 2024-06-30
fix: hack webpack vite ignore

Aslemammad:


dai-shi:
The commit d6a81d7 has been cherry-picked in #707.

PR #767, opened 2024-07-01
tests: add rscTransform internal function tests

dai-shi:
- [x] server env: static 'use client'
- [x] server env: static 'use server'
- [x] server env: dynamic 'use server'
- [x] client env: static 'use server'


dai-shi:
Maybe I should rename test file name to avoid conflicting with #759.

PR #774, opened 2024-07-03
fix: server action cases

himself65:
Fix some RSC case, will fix https://github.com/dai-shi/waku/pull/745. In the future, I will migrate the logic to https://github.com/himself65/react-server and keep same output

Some issues still happening:

1. `export default async function () {"use server"}` cannot be RSCA
2. `export function a() { "use server" }` should throw error, because it's not async

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
@@ -91,9 +91,17 @@ const createCallExpression = (
   span: { start: 0, end: 0, ctxt: 0 },
 });
 
-const serverActionsInitCode = swc.parseSync(`
+const serverInitCode = swc.parseSync(`
 import { registerServerReference as __waku_registerServerReference } from 'react-server-dom-webpack/server.edge';
 `).body;
+
+const findLastImportIndex = (mod: swc.Module) => {
+  const lastImportIndex = mod.body.findIndex(
+    (node) =>
+      node.type !== 'ExpressionStatement' && node.type !== 'ImportDeclaration',
+  );
+  return lastImportIndex === -1 ? 0 : lastImportIndex;
+};
 
 type FunctionWithBlockBody = (
   | swc.FunctionDeclaration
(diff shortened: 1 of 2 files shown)


packages/waku/tests/vite-plugin-rsc-transform-internals.test.ts
```diff
@@ -68,6 +68,81 @@ export const log = (mesg) => {
...
+    wrappedActions[name] = wrapAction(actions[name]);
+  }
+  return function AI(props) {
+    return jsx(InternalProvider, {
+      actions: wrappedActions,
+      children: props.children,
+    });
+  };
+}`;
+    // fixme: this is wrong, `innerAction.bind` should be `__waku_action1.bind`
```
himself65, line 116: /cc @dai-shi

packages/waku/tests/vite-plugin-rsc-transform-internals.test.ts
```diff
@@ -68,10 +66,117 @@ export const log = (mesg) => {
...
+  });
+
+  test('create server with bind', async () => {
+    const code = `
+'use server';
+import { InternalProvider } from './shared.js';
+import { jsx } from 'react/jsx-runtime';
+
+async function innerAction({ action }, ...args) {
+  'use server';
```
dai-shi, line 120: Okay, you want to support this double `use server` case. I thought about it, but I was not sure if it's valid.

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -417,44 +417,96 @@ export ${name === 'default' ? name : `const ${name} =`} registerClientReference(
...
+      }
+      let useServerPosStart: number;
+      let useServerPosEnd: number;
+      if (hasUseServer === true) {
+        useServerPosStart = 0;
+        useServerPosEnd = 0;
+      } else {
+        useServerPosStart = hasUseServer.span.start - mod.span.start;
+        useServerPosEnd = hasUseServer.span.end - mod.span.start;
+      }
```
dai-shi, line 481: The change above seems a bit ugly.
Let me see if I can do anything with it.

dai-shi (commented):
Thanks for the catch.

packages/waku/tests/vite-plugin-rsc-transform-internals.test.ts
himself65: It's valid in Next.js I think🤔

himself65: I wrote down all cases in sheet
image

dai-shi: yeah, it's reasonable with the object param case.

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
dai-shi: done.

dai-shi:
> 1. `export default async function () {"use server"}` cannot be RSCA

Yeah, I notice that too. Even with top-level "use server", it fails. Gotta fix it, but maybe later.

> 2. `export function a() { "use server" }` should throw error, because it's not async

Originally, I thought sync functions are supported, and I have such an example in 23_actions.
Does nextjs throw an error? I'll leave it to your `react-server`.

himself65:
> Originally, I thought sync functions are supported, and I have such an example in 23_actions.
Does nextjs throw an error? I'll leave it to your react-server.

It must be an async (and next.js forces this) because the server action is function between the client and server, the is no way to be sync

dai-shi:
> > Originally, I thought sync functions are supported, and I have such an example in 23_actions.
> > Does nextjs throw an error? I'll leave it to your react-server.
> 
> It must be an async (and next.js forces this) because the server action is function between the client and server, the is no way to be sync

#778

PR #776, opened 2024-07-04
test: catch an error in client router

t6adev:
It covers #696 and #741.
The test aims at how to implement auth features on the app side.

Considerations:
1. Our client router has to handle 4xx errors that the server responds
2. Middleware can be a validator to respond to 200 or 4xx status on requests 

I'm so glad to send my first PR to Waku, thank you!

### DIFF
--- a/e2e/fixtures/ssr-catch-error/package.json
+++ b/e2e/fixtures/ssr-catch-error/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "ssr-catch-error",
+  "version": "0.1.0",
+  "type": "module",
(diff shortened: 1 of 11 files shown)


dai-shi (commented):
I think you need to add a new entry in tsconfig.e2e.json.
Please run prettier too.

dai-shi (approved):
We may not preserve this capability of using error boundary in client layout, and we may change the capability and this test in the future.
But, meanwhile, adding this test to notice regression would be nice.

Thanks for your contribution.

PR #784, opened 2024-07-09
feat: migrate analyze function to `react-server-action`

himself65:
I finally found a way to release the package. Let me try it in waku

### DIFF
--- a/packages/waku/package.json
+++ b/packages/waku/package.json
@@ -78,7 +78,7 @@
     "@vitejs/plugin-react": "4.3.1",
     "dotenv": "16.4.5",
     "hono": "4.4.11",
-    "react-server-action": "0.2.13",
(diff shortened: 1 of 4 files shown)


packages/waku/package.json
```diff
@@ -78,6 +78,7 @@
     "@vitejs/plugin-react": "4.3.1",
     "dotenv": "16.4.5",
     "hono": "4.4.11",
+    "react-server-action": "0.2.15",
```
dai-shi, line 81: Despite the package name, it supports all RSC transforms, right?

himself65: yes

dai-shi:
I think the goal is to migrate to rolldown and eliminate swc. closing this for now.

PR #785, opened 2024-07-10
fix: edge case when rsa depends on another rsa

himself65:


himself65:
@dai-shi ready to review

packages/waku/src/lib/hono/runner.ts
```diff
@@ -28,6 +28,12 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
         .map(async (middleware) => (await middleware).default(options)),
     ),
   );
+
+  (globalThis as any).__WAKU_HACK_IMPORT__ = async (id: string) => {
```
dai-shi, line 32: `__WAKU_HACK_IMPORT__` is unused, right?

btw, I don't like this hack. there must be a better way.

himself65: yes, React will read buildConfig and use __WAKU_HACK_IMPORT__ to load the react server action

dai-shi: Alright, `__WAKU_HACK_IMPORT__` should only be required for client. Let me see if I can do something.

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -189,9 +189,9 @@ export async function renderRsc(
     ) {
       // XXX This doesn't support streaming unlike busboy
       const formData = parseFormData(bodyStr, contentType);
-      args = await decodeReply(formData);
+      args = await decodeReply(formData, bundlerConfig);
```
dai-shi, line 192: I actually wonder why this works. This is basically resolving client entry. Can it be used for server entry?

himself65: I think it's also resolving the server entry, yes

himself65: check react test case, they have cases to load server modules

https://github.com/facebook/react/blob/29552c7907230222acd3f2c586784d24f9da6200/packages/react-server-dom-webpack/src/__tests__/ReactFlightDOMReply-test.js#L43

https://github.com/facebook/react/blob/29552c7907230222acd3f2c586784d24f9da6200/packages/react-server-dom-webpack/src/__tests__/utils/WebpackMock.js#L130

packages/waku/src/lib/plugins/vite-plugin-rsc-rsdw.ts
```diff
@@ -21,7 +21,7 @@ globalThis.__WAKU_${type}_CHUNK_LOAD__ ||= (
   if (!globalThis.__WAKU_${type}_MODULE_LOADING__.has(id)) {
     globalThis.__WAKU_${type}_MODULE_LOADING__.set(
       id,
-      customImport(id).then((m) => {
+      (customImport ? customImport(id) : import(id)).then((m) => {
```
himself65, line 24: I think this won't work because node condition is not including `react-server`

himself65: that's why you need a vite server to load the module

himself65: and import() doesn't support .ts(x) file in dev side

dai-shi: Okay, but it should work on PRD, right?

himself65: I think so, but ID is not correct; it is now like `/path/to/file` which is presented as HTTP resource

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: In `bundlerConfig`, we use `resolveClientEntry`, so even if it works, it's confusing.

dai-shi:
`__WAKU_HACK_IMPORT__` should be DEV only.
Can you find a solution without using it in PRD?
As I said, it feels strange to reuse bunderConfig for client for server.
Maybe, both `bunderConfigForClient` and `bunderConfigForServer` would be required.

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -104,6 +104,22 @@ export async function renderRsc(
...
 
+  const serverBundlerConfig = new Proxy(
+    {},
+    {
+      get(_target, encodedId: string) {
+        const [fileId, name] = encodedId.split('#') as [string, string];
+        const id = filePathToFileURL(fileId);
+        if (fileId.startsWith('@id/assets/')) {
+          const id = '.' + fileId.slice('@id'.length);
+          return { id, chunks: [id], name, async: true };
```
himself65, line 115: it will be `./assets/rsf0.js#func`

and because rsdw.js is in dist/rsdw.js, it import() will be work

dai-shi (approved):
This looks pretty good! Thanks for your work!

PR #788, opened 2024-07-12
test: add vercel ai fixtures

himself65:
`Error: `useUIState` must be used inside an <AI> provider.`


dai-shi (changes requested):
@himself65 I would like to merge this e2e test to avoid breaking it in the future, but the current fixture is too big. Can you create a minimal fixture and the spec?

himself65:
ok

himself65:
See downstream

PR #793, opened 2024-07-15
feat: migrate cloudflare workers adapter to cloudflare pages

jkhaui:
This PR implements a minimal set of changes so that running waku build with the flag `-- --with-cloudflare` now creates a build specifically for Cloudflare Pages (using their advanced mode strategy). This is the recommended way to deploy web apps on Cloudflare now; worker sites are deprecated.

More discussion and context can be found here: https://github.com/dai-shi/waku/discussions/763

Note: I have tested numerous times that the changes work and can successfully deploy a CF Pages site with no errors. To test this, I used a custom workflow which deploys `examples/10_fs-router` from the monorepo with a compiled version of the waku pnpm workspace. To get this to work I also made the following changes to the example project (these are not committed): 

1. Add `waku.config.ts` with the base path set to `/public/` (explained in the updated docs). 
2. Also needed to add hono as a dependency in the example project, otherwise deploying via Cloudflare would throw an error about a missing hono dependency.

image
image

There is a chance that behaviour could be different when consuming Waku as an external npm package. But I can't think of any good ways to replicate this. So I suggest we just test the end-to-end deployment again after merging and releasing this PR

Just let me know if any changes should be made

### DIFF
--- a/README.md
+++ b/README.md
@@ -850,7 +850,20 @@ Adding the `--with-netlify-static` flag to the build script will produce static 
 }
 ```
 
-### Cloudflare (experimental)
+### Cloudflare Pages (experimental)
+
+**Note:** the current integration with Cloudflare Pages requires a `waku.config.ts` file to be
+added at the root of your project, with the basePath set like so:
+
+```
+export default {
+  basePath: "/public/"
+}
+```
+
+From here, it's recommended to follow Cloudflare Page's workflow of creating a new Pages
+application and connecting it to your GitHub or GitLab repo for automatic deployments.
+See [the docs](https://developers.cloudflare.com/pages/get-started/git-integration/).
 
 ```
 npm run build -- --with-cloudflare
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -752,7 +752,7 @@ export async function build(options: {
       options.deploy.slice('netlify-'.length) as 'static' | 'functions',
     );
   } else if (options.deploy === 'cloudflare') {
-    await emitCloudflareOutput(rootDir, config, DIST_SERVE_JS);
+    await emitCloudflareOutput(rootDir, config);
   } else if (options.deploy === 'partykit') {
     await emitPartyKitOutput(rootDir, config, DIST_SERVE_JS);
   } else if (options.deploy === 'aws-lambda') {
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -298,6 +298,24 @@ const buildServerBundle = async (
   if (!('output' in serverBuildOutput)) {
     throw new Error('Unexpected vite server build output');
   }
+
+  if (serve === 'cloudflare') {
+    // For Cloudflare Pages, create `_routes.json` which prevents calls to
+    // static client assets from invoking the server-side function.
+    // TODO: check for existing `_routes.json` in users' root directory? And
+    // merge this with default staticRoutes provided below
+    const routesJsonFilePath = joinPath(rootDir,
+      config.distDir,
+      '_routes.json');
+    const staticRoutes = {
+      version: 1,
+      include: ['/*'],
+      exclude: ['/public/assets/*']
+    };
+
+    await writeFile(routesJsonFilePath, JSON.stringify(staticRoutes));
+  }
+
   return serverBuildOutput;
 };
 
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -58,6 +58,7 @@ import { emitAwsLambdaOutput } from './output-aws-lambda.js';
 import {
   DIST_ENTRIES_JS,
   DIST_SERVE_JS,
+  DIST_WORKER_JS,
   DIST_PUBLIC,
   DIST_ASSETS,
   DIST_SSR,
@@ -237,7 +238,11 @@ const buildServerBundle = async (
         ? [
             rscServePlugin({
               ...config,
-              distServeJs: DIST_SERVE_JS,
+              distServeJs: serve !== 'cloudflare'
+                ? DIST_SERVE_JS
+                // Cloudflare Pages advanced mode requires the dist entry file
+                // to be named `_worker.js`
+                : DIST_WORKER_JS,
               distPublic: DIST_PUBLIC,
               srcServeFile: decodeFilePathFromAbsolute(
                 joinPath(
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -298,24 +298,6 @@ const buildServerBundle = async (
   if (!('output' in serverBuildOutput)) {
     throw new Error('Unexpected vite server build output');
   }
-
-  if (serve === 'cloudflare') {
-    // For Cloudflare Pages, create `_routes.json` which prevents calls to
-    // static client assets from invoking the server-side function.
-    // TODO: check for existing `_routes.json` in users' root directory? And
-    // merge this with default staticRoutes provided below
-    const routesJsonFilePath = joinPath(rootDir,
-      config.distDir,
-      '_routes.json');
-    const staticRoutes = {
-      version: 1,
-      include: ['/*'],
-      exclude: ['/public/assets/*']
-    };
-
-    await writeFile(routesJsonFilePath, JSON.stringify(staticRoutes));
-  }
-
   return serverBuildOutput;
 };
 
--- a/packages/waku/src/lib/builder/output-cloudflare.ts
+++ b/packages/waku/src/lib/builder/output-cloudflare.ts
@@ -1,14 +1,28 @@
 import path from 'node:path';
 import { existsSync, writeFileSync } from 'node:fs';
+import { joinPath } from '../utils/path.js';
+import { DIST_ASSETS, DIST_PUBLIC } from './constants.js';
 
 import type { ResolvedConfig } from '../config.js';
-import { DIST_PUBLIC } from './constants.js';
 
-// XXX this can be very limited. FIXME if anyone has better knowledge.
 export const emitCloudflareOutput = async (
   rootDir: string,
   config: ResolvedConfig
 ) => {
+  const routesJsonPublicFile = joinPath(rootDir, config.srcDir, DIST_PUBLIC);
+  if (!existsSync(routesJsonPublicFile)) {
+    const routesJsonDistFile = joinPath(rootDir,
+      config.distDir,
+      '_routes.json');
+    const staticRoutes = {
+      version: 1,
+      include: ['/*'],
+      exclude: [`/${DIST_PUBLIC}/${DIST_ASSETS}/*`]
+    };
+
+    writeFileSync(routesJsonDistFile, JSON.stringify(staticRoutes));
+  }
+
   const wranglerTomlFile = path.join(rootDir, 'wrangler.toml');
   if (!existsSync(wranglerTomlFile)) {
     writeFileSync(
--- a/packages/waku/src/lib/builder/output-cloudflare.ts
+++ b/packages/waku/src/lib/builder/output-cloudflare.ts
@@ -7,8 +7,7 @@ import { DIST_PUBLIC } from './constants.js';
 // XXX this can be very limited. FIXME if anyone has better knowledge.
 export const emitCloudflareOutput = async (
   rootDir: string,
-  config: ResolvedConfig,
-  serveJs: string,
+  config: ResolvedConfig
 ) => {
   const wranglerTomlFile = path.join(rootDir, 'wrangler.toml');
   if (!existsSync(wranglerTomlFile)) {
@@ -16,12 +15,10 @@ export const emitCloudflareOutput = async (
       wranglerTomlFile,
       `
 name = "waku-project"
-main = "${config.distDir}/${serveJs}"
-compatibility_date = "2023-12-06"
-compatibility_flags = [ "nodejs_als" ]
+compatibility_date = "2024-04-05"
+compatibility_flags = ["nodejs_compat"]
 
-[site]
-bucket = "./${config.distDir}/${DIST_PUBLIC}"
+pages_build_output_dir = "./${config.distDir}/${DIST_PUBLIC}"
 `,
     );
   }
--- a/packages/waku/src/lib/builder/serve-cloudflare.ts
+++ b/packages/waku/src/lib/builder/serve-cloudflare.ts
@@ -1,28 +1,15 @@
 import { Hono } from 'hono';
-import { serveStatic } from 'hono/cloudflare-workers';
-// @ts-expect-error no types
-// eslint-disable-next-line import/no-unresolved
-import manifest from '__STATIC_CONTENT_MANIFEST';
+import { serveStatic } from 'hono/cloudflare-pages';
 
 import { runner } from '../hono/runner.js';
 
 const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
 let serveWaku: ReturnType<typeof runner> | undefined;
-let staticContent: any;
-
-const parsedManifest: Record<string, string> = JSON.parse(manifest);
 
 const app = new Hono();
-app.use('*', serveStatic({ root: './', manifest }));
+app.use('/assets/*', serveStatic());
 app.use('*', (c, next) => serveWaku!(c, next));
 app.notFound(async (c) => {
-  const path = parsedManifest['404.html'];
-  const content: ArrayBuffer | undefined =
-    path && (await staticContent?.get(path, { type: 'arrayBuffer' }));
-  if (content) {
-    c.header('Content-Type', 'text/html; charset=utf-8');
-    return c.body(content, 404);
-  }
   return c.text('404 Not Found', 404);
 });
 
@@ -34,7 +21,6 @@ export default {
   ) {
     if (!serveWaku) {
       serveWaku = runner({ cmd: 'start', loadEntries, env });
-      staticContent = env.__STATIC_CONTENT;
     }
     return app.fetch(request, env, ctx);
   },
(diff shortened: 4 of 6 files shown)


jkhaui:
I wonder what you think about extending `resolveConfig` with a 2nd arg for the deployment platform, like this:

```
// build.ts

const config = await resolveConfig(options.config, options.deploy);
```

and then

```
// config.ts

export async function resolveConfig(config: Config, deploy?: Deploy) {
  const resolvedConfig: ResolvedConfig = {
    basePath: `${deploy !== 'cloudflare' ? '' : '/public'}/`,
    srcDir: 'src',
    distDir: 'dist',
    preserveModuleDirs: ['pages', 'templates', 'routes', 'components'],
    privateDir: 'private',
    rscPath: 'RSC',
    htmlAttrs: '',
    htmlHead: DEFAULT_HTML_HEAD,
    middleware: DEFAULT_MIDDLEWARE,
    ...config,
  };
  if (!resolvedConfig.htmlHead.includes(ADDITIONAL_HTML_HEAD)) {
    resolvedConfig.htmlHead += ADDITIONAL_HTML_HEAD;
  }
  return resolvedConfig;
}
```

Might not be so robust, but for now it would let users deploy to Cloudflare Pages without having to create a `waku.config.ts` with the base path set to `/public` in prod

dai-shi:
We will be revisiting our config system later, so let's not make changes for now.

packages/waku/src/lib/builder/output-cloudflare.ts
```diff
@@ -7,21 +7,18 @@ import { DIST_PUBLIC } from './constants.js';
...
   if (!existsSync(wranglerTomlFile)) {
     writeFileSync(
       wranglerTomlFile,
       `
 name = "waku-project"
-main = "${config.distDir}/${serveJs}"
-compatibility_date = "2023-12-06"
-compatibility_flags = [ "nodejs_als" ]
+compatibility_date = "2024-04-05"
+compatibility_flags = ["nodejs_compat"]
```
dai-shi, line 19: I was intentionally avoiding nodejs_compat. Do we require it now?

packages/waku/src/lib/builder/output-cloudflare.ts
```diff
@@ -7,21 +7,18 @@ import { DIST_PUBLIC } from './constants.js';
...
 
-[site]
-bucket = "./${config.distDir}/${DIST_PUBLIC}"
+pages_build_output_dir = "./${config.distDir}/${DIST_PUBLIC}"
```
dai-shi, line 21: Hm, I wonder if we can move public files.

```
pages_build_output_dir = "./output"
and
fs.rename("./${config.distDir}/${DIST_PUBLIC}", "./output")
```
something like that.

packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -34,7 +21,6 @@ export default {
   ) {
     if (!serveWaku) {
       serveWaku = runner({ cmd: 'start', loadEntries, env });
-      staticContent = env.__STATIC_CONTENT;
```
dai-shi, line 37: I'm happy with this hack going away.

dai-shi (commented):
first review

jkhaui:
Sure thing

I have a question regarding the default value for `preserveModuleDirs`: just curious if there's a reason for not defaulting to a chunked + minified SSR build?

I'm asking because CF has a 1mb max bundle size for workers on their free tier, so every byte in SSR build counts.

Usually I overcome this by setting `vite.build.minify` to use terser, but this doesn't work with Waku - it breaks the build. 

Setting minify to "esbuild" instead of "terser" sometimes works. I've also tried setting `preserveModuleDirs: []` but it doesn't seem to help

jkhaui:
Thanks for checking out the PR as well - I'll go through your comments tomorrow 👍

dai-shi:
preserveModuleDirs is to make it possible to read files with relative paths, but we might not need to allow it. I don't know why minification fails with or without it.

packages/waku/src/lib/builder/build.ts
```diff
@@ -293,6 +298,24 @@ const buildServerBundle = async (
...
+      config.distDir,
+      '_routes.json');
+    const staticRoutes = {
+      version: 1,
+      include: ['/*'],
+      exclude: ['/public/assets/*']
+    };
+
+    await writeFile(routesJsonFilePath, JSON.stringify(staticRoutes));
+  }
```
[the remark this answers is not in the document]
jkhaui: I think there's a reason I didn't put it there. Let me check.

And thanks for commenting on this, you made me notice that the public and asset paths should probably be variables instead of hardcoded

packages/waku/src/lib/builder/build.ts
```diff
@@ -237,7 +238,11 @@ const buildServerBundle = async (
         ? [
             rscServePlugin({
               ...config,
-              distServeJs: DIST_SERVE_JS,
+              distServeJs: serve !== 'cloudflare'
+                ? DIST_SERVE_JS
+                // Cloudflare Pages advanced mode requires the dist entry file
+                // to be named `_worker.js`
```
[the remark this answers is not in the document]
jkhaui: Yeah for sure. Astro's CF adapter creates a build output that looks like this:

image

I don't know how they are using a custom entry path to `_worker/index.js` unless I browse the source code. Because CFP docs say the entry must be called `_worker.js` and be in the output root directory. 

But the point is it means it is possible to customize the `_worker.js` entry file 👍

packages/waku/src/lib/builder/build.ts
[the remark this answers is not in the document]
jkhaui: I was wrong - we absolutely can (and should) move this to `output-cloudflare.ts`. Good spot!

packages/waku/src/lib/builder/output-cloudflare.ts
jkhaui: I'm not sure tbh, just assumed it wouldn't work unless nodejs_compat is set. Let me try without it and report back.

Any particular reason you think we should avoid it?

packages/waku/src/lib/builder/output-cloudflare.ts
jkhaui: Hmm. This one I'll probably have to leave to you to decide.

I think I had a similar idea. But I decided not to pursue it further, looked like it was getting very complex because the _worker.js fil then needs to be moved.

In other words, if we implement this strategy, then we must also move _worker.js to `/dist/public/output` as CFP requires _worker.js to exist in the root output directory. There's a good chance other import paths will also break after doing this?

README.md
```diff
@@ -850,7 +850,20 @@ Adding the `--with-netlify-static` flag to the build script will produce static
...
 
-### Cloudflare (experimental)
+### Cloudflare Pages (experimental)
+
+**Note:** the current integration with Cloudflare Pages requires a `waku.config.ts` file to be
+added at the root of your project, with the basePath set like so:
+
+```
+export default {
+  basePath: "/public/"
```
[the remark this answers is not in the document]
jkhaui: Yes I have no idea why this happens either. It could be from CFP, Waku, or Vite's build process. Or even a combination, e.g. CFP might not be able to read `import.meta.env` in some cases.

I definitely see it as temporary. My suggestion is, seeing as Waku doesn't currently have lots of users, we accept an initial hacky workaround. I will then use the published package to see if anything behaves differently and try working on a better solution.

packages/waku/src/lib/builder/output-cloudflare.ts
dai-shi: I want to remove nodejs dependency strictly, and cloudflare wrangler gives me errors if some nodejs dependencies remain.

packages/waku/src/lib/builder/output-cloudflare.ts
dai-shi: I'm not sure how much it would break others, but I would like to push it for now.
It eliminates the need of `waku.config.ts`, right?

We should revisit it anyway, when we design our build plugin system in the future (not sure when).

jkhaui: In theory it should, I guess the only way is to try it. Let me do this soon and tell you the results :)

Yeah the way I see this solution, it's just to unblock people like me who want to start working immediately with Waku on a platform outside Vercel. I think even just having experimental support for CFP (& other platforms) will bring in more users instead of waiting for a full rewrite of the plugin system.

And users with existing Waku apps should be able to easily upgrade to the new adapter system when it is ready in future

jkhaui: > Hm, I wonder if we can move public files.
> 
> ```
> pages_build_output_dir = "./output"
> and
> fs.rename("./${config.distDir}/${DIST_PUBLIC}", "./output")
> ```
> 
> something like that.

@dai-shi I need to clarify here, do you mean that `dist/public` should be renamed to `dist/output`? And then `pages_build_output_dir = "./dist/output"` Or am I misunderstanding

rmarscher:
Hi! I've also been working on adjusting the waku cloudflare adapter. I made some progress on the build output. I have it creating the _worker.js as the destination folder, creating a _routes.json and copying the public files to the root of the dist output. I updated the wrangler.toml output to use the [[newer Cloudflare Pages wrangler.toml options](https://developers.cloudflare.com/pages/functions/wrangler-configuration/)](https://developers.cloudflare.com/pages/functions/wrangler-configuration/) (but I’m guessing you already did that here too). I’m just catching up - but I see you found the way some of the other framework adapters handle it.

See https://github.com/dai-shi/waku/compare/main...rmarscher:waku:cloudflare-pages

My changes so far have gotten things working for me when I run `waku build --with-cloudflare` and then run `wrangler pages dev dist`. I haven't added the Cloudflare bindings to the dev server yet. Also, when I try to deploy to cloudflare, I'm getting an error:

```
Failed to publish your Function. Got error: Uncaught Error: No such module "hono".
    imported from "serve.js"
```

I just got that far. I was going to look into why it isn't bundling hono next.

Cloudflare needs a _routes.json file to let it now what are static routes and files that can be cached. I added some code to walk the build public folder and build out the _routes.json. It also can use a generated _headers file to set the cache response headers for certain routes. I didn’t add any custom code for that, but it works if you drop it in your public folder.

I looked at Cloudflare adapters from SvelteKit, Astro and Hono.

- https://github.com/withastro/adapters/blob/main/packages/cloudflare/src/index.ts
- https://github.com/honojs/vite-plugins/blob/main/packages/cloudflare-pages/src/cloudflare-pages.ts
- https://github.com/sveltejs/kit/blob/main/packages/adapter-cloudflare/index.js

They all use the `getPlatformProxy` method from Cloudflare's `wrangler` package to populate the environment bindings in dev mode. https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy

I'm currently trying to figure out a new vite or waku plugin that can handle the wrangler proxy when running `waku dev --with-cloudflare`. Is Vite's configureServer hook the right place to try to inject it? I was going to try that next. https://vitejs.dev/guide/api-plugin#configureserver

The astro adapter seems to be especially thought out. It has deep config options and also does some processing of any redirect files.

jkhaui:
> Hi! I've also been working on adjusting the waku cloudflare adapter. I made some progress on the build output. I have it creating the _worker.js as the destination folder, creating a _routes.json and copying the public files to the root of the dist output. I updated the wrangler.toml output to use the [[newer Cloudflare Pages wrangler.toml options](https://developers.cloudflare.com/pages/functions/wrangler-configuration/)](https://developers.cloudflare.com/pages/functions/wrangler-configuration/) (but I’m guessing you already did that here too). I’m just catching up - but I see you found the way some of the other framework adapters handle it.
> 
> 
> 
> See https://github.com/dai-shi/waku/compare/main...rmarscher:waku:cloudflare-pages
> 
> 
> 
> My changes so far have gotten things working for me when I run `waku build --with-cloudflare` and then run `wrangler pages dev dist`. I haven't added the Cloudflare bindings to the dev server yet. Also, when I try to deploy to cloudflare, I'm getting an error:
> 
> 
> 
> ```
> 
> Failed to publish your Function. Got error: Uncaught Error: No such module "hono".
> 
>     imported from "serve.js"
> 
> ```
> 
> 
> 
> I just got that far. I was going to look into why it isn't bundling hono next.
> 
> 
> 
> Cloudflare needs a _routes.json file to let it now what are static routes and files that can be cached. I added some code to walk the build public folder and build out the _routes.json. It also can use a generated _headers file to set the cache response headers for certain routes. I didn’t add any custom code for that, but it works if you drop it in your public folder.
> 
> 
> 
> I looked at Cloudflare adapters from SvelteKit, Astro and Hono.
> 
> 
> 
> - https://github.com/withastro/adapters/blob/main/packages/cloudflare/src/index.ts
> 
> - https://github.com/honojs/vite-plugins/blob/main/packages/cloudflare-pages/src/cloudflare-pages.ts
> 
> - https://github.com/sveltejs/kit/blob/main/packages/adapter-cloudflare/index.js
> 
> 
> 
> They all use the `getPlatformProxy` method from Cloudflare's `wrangler` package to populate the environment bindings in dev mode. https://developers.cloudflare.com/workers/wrangler/api/#getplatformproxy
> 
> 
> 
> I'm currently trying to figure out a new vite or waku plugin that can handle the wrangler proxy when running `waku dev --with-cloudflare`. Is Vite's configureServer hook the right place to try to inject it? I was going to try that next. https://vitejs.dev/guide/api-plugin#configureserver
> 
> 
> 
> The astro adapter seems to be especially thought out. It has deep config options and also does some processing of any redirect files.

That's awesome, great to see someone else here!

I'll read your message more thoroughly in a sec (having lunch atm), but if you want to catchup on our earlier brainstorming, also check out the discussion I linked in the PR description

rmarscher:
> That's awesome, great to see someone else here!

:wave:

> I'll read your message more thoroughly in a sec (having lunch atm), but if you want to catchup on our earlier brainstorming, also check out the discussion I linked in the PR description

I got past the hono error. It was being added as an external in the rollup config in the vite-plugin-rsc-serve.ts file if opts.serve was "cloudflare" or "partykit".

That allowed my CF Pages project to deploy... but when I try to open it, I get an error:

```
Error: No such module "ssr/assets/events".
  imported from "ssr/assets/rsc1-3ac5d8483.js"
```

That's coming from this line:
```
import EventEmitter$2, { EventEmitter as EventEmitter$1 } from "events";
```

Maybe that needs a `node:` prefix on the import package name.

I tried with nodejs_als and nodejs_compat and both had the same issue (it did work to build and deploy with nodejs_als so far though...). I also tried the `--no-bundle` option of `wrangler pages deploy` and that didn't seem to make a difference either.

I'll keep plugging away...

rmarscher:
Oh actually... it was my client code that was causing the import error. I must have another dependency that has a bundling issue. With just a basic site, it works! https://74bb725c.waku-test.pages.dev/  <- that site loads the message from a json file.

packages/waku/src/lib/builder/output-cloudflare.ts
dai-shi: I'm not sure what's the best, but what I'm suggesting is to restructure (rename) folders in `output-cloudflare.ts` and avoid changing `basePath`.

jkhaui:
> I got past the hono error. It was being added as an external in the rollup config in the vite-plugin-rsc-serve.ts file if opts.serve was "cloudflare" or "partykit".

Yep good stuff - I was just about to tell you to look there before you posted
 
> That allowed my CF Pages project to deploy... but when I try to open it, I get an error:
> 
> ```
> Error: No such module "ssr/assets/events".
>   imported from "ssr/assets/rsc1-3ac5d8483.js"
> ```
> 
> That's coming from this line:
> 
> ```
> import EventEmitter$2, { EventEmitter as EventEmitter$1 } from "events";
> ```
> 
> Maybe that needs a `node:` prefix on the import package name.
> 
> I tried with nodejs_als and nodejs_compat and both had the same issue (it did work to build and deploy with nodejs_als so far though...). I also tried the `--no-bundle` option of `wrangler pages deploy` and that didn't seem to make a difference either.
> 
> I'll keep plugging away...

Yeah, for the first couple of attempts I made, I also tried to follow a similar implementation as the adapters you listed.

And I also faced a tonne of errors from node imports in transitive dependencies that aren't prefixed with `node:`. The challenge (if I understand correctly) is that you can only fix these errors by updating the source code of every incompatible dependency and add the `node:` prefix. Though I'm also unsure why these errors occur in some implementations of Waku + CFP but not others (i.e. like the way I've done this PR). 

How are you deploying to CFP? Are you using a connected git repo? IIRC this also affected the build process behaviour for me

Ultimately, I decided it's not worth the effort atm to figure out all these quirks if the minimal change of renaming `serve.js` to `_worker.js` (along with a couple of other minor changes) appears to be enough to get basic CFP working. I also didn't even look into setting up a platform proxy with Wrangler - as @dai-shi has mentioned, Waku's adapter system will be revamped in future. When this happens it could be a good time to add any dev environment features?

The stuff you've done so far looks great. though. Seeing as our implementations are quite different, I suggest we first aim to merge in this PR with its minimal changes and verify it works when Waku is consumed as an external package. From here, we can then cherry-pick any enhancements from the work you've been doing. E.g. support for CF bindings looks a good candidate

What does everyone think?

packages/waku/src/lib/builder/output-cloudflare.ts
jkhaui: Got it. Unfortunately the renaming from public to output doesn't look to work

image

I also think it should be left as `public` as this is what [Vite expects](https://vitejs.dev/config/shared-options.html#publicdir)

@rmarscher any suggestions here? Did you encounter any challenges with the `public` directory?

rmarscher:
> The stuff you've done so far looks great. though. Seeing as our implementations are quite different, I suggest we first aim to merge in this PR with its minimal changes and verify it works when Waku is consumed as an external package. From here, we can then cherry-pick any enhancements from the work you've been doing. E.g. support for CF bindings looks a good candidate

I submitted a PR to make it a bit easier to review and try the changes I made. I'm heading to bed, but feel free to handle this in any way -- merge this PR and then I can potentially rebase and resubmit or however everyone wants to handle it.

> **edit:** just saw your last message @rmarscher , so you got it working? Nice! Are you able to share what your `dist` directory structure looks like?

Yup. It looks like this. Cheers!

Screenshot 2024-07-17 at 2 39 44 AM

jkhaui:
Closing this so we can consolidate efforts behind [this PR](https://github.com/dai-shi/waku/pull/795). Can always reopen if we change our mind for whatever reason.

PR #796, opened 2024-07-17
fix: Hono runner 404 response

rmarscher:
This fixes the issue for me mentioned in #644 where an empty 200 status response is returned instead of a 404 error page.

I found the RSC middleware doesn't run in that case so the ctx.res is never set. I added a fallback to move the hono res.body to the ctx.body if the ctx.body is empty. I assume that won't be a breaking change but 🤷.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -96,8 +96,9 @@ async function runDev() {
   const config = await loadConfig();
   const app = new Hono();
   app.use('*', runner({ cmd: 'dev', config, env: process.env as any }));
-  // @ts-expect-error bypassing hono default notFound handler
-  app.use('*', (_c, _next) => {})
+  // bypassing hono default notFound handler
+  // @ts-expect-error FIXME there might be a better way to handle this
+  app.use('*', (_c, _next) => {});
   const port = parseInt(values.port || '3000', 10);
   await startServer(app, port);
 }
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -96,6 +96,8 @@ async function runDev() {
   const config = await loadConfig();
   const app = new Hono();
   app.use('*', runner({ cmd: 'dev', config, env: process.env as any }));
+  // @ts-expect-error bypassing hono default notFound handler
+  app.use('*', (_c, _next) => {})
   const port = parseInt(values.port || '3000', 10);
   await startServer(app, port);
 }
--- a/packages/waku/src/lib/hono/runner.ts
+++ b/packages/waku/src/lib/hono/runner.ts
@@ -53,10 +53,12 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
-    return c.body(
-      ctx.res.body || null,
-      (ctx.res.status as any) || 200,
-      ctx.res.headers || {},
-    );
+    if (!c.finalized) {
+      return c.body(
+        ctx.res.body || null,
+        (ctx.res.status as any) || 200,
+        ctx.res.headers || {},
+      );
+    }
   };
 };
--- a/packages/waku/src/lib/hono/runner.ts
+++ b/packages/waku/src/lib/hono/runner.ts
@@ -53,8 +53,7 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
-    /* checking ctx.res.body is a workaround for dev server 404 response */
-    if (!c.finalized/* || ctx.res.body*/) {
+    if (!c.finalized) {
       return c.body(
         ctx.res.body || null,
         (ctx.res.status as any) || 200,
--- a/packages/waku/src/lib/hono/runner.ts
+++ b/packages/waku/src/lib/hono/runner.ts
@@ -53,7 +53,8 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
-    if (!c.finalized) {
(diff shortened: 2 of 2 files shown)


dai-shi (commented):
Thanks for the suggestion!
If it fixes #644, it's great and it's probably a valid solution.
But, I'm not sure how it happens. Do you know what's going on?

packages/waku/src/lib/hono/runner.ts
```diff
@@ -53,6 +53,14 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
+    if (!ctx.res.body) {
+      // RSC didn't run, maybe an error. Try using hono res
+      if (c.res.body) {
+        ctx.res.body = c.res.body;
```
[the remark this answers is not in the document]
dai-shi: wait, i get it. it's already handled in `next()`.

dai-shi (changes requested):
Can you apply this change
https://github.com/dai-shi/waku/pull/644/commits/7a3cb850f178e72458f13a55867f7dfd978f8e3c
and see if it fixes your case?

rmarscher:
> Can you apply this change [7a3cb85](https://github.com/dai-shi/waku/commit/7a3cb850f178e72458f13a55867f7dfd978f8e3c) and see if it fixes your case?

Works for me! I'm glad I was able to help you identify the cause of the issue and good job finding the hono context "finalized" property as a solution. I cherry-picked your commit onto the branch replacing mine. Thanks!

dai-shi:
Thanks for confirming.

Hm, now all e2e tests fail. Can you investigate into it please?

packages/waku/src/lib/hono/runner.ts
```diff
@@ -53,10 +53,12 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
-    return c.body(
-      ctx.res.body || null,
-      (ctx.res.status as any) || 200,
-      ctx.res.headers || {},
-    );
+    if (!c.finalized || ctx.res.body) {
```
rmarscher, line 56: @dai-shi Checking for `!c.finalized || ctx.res.body` instead of just `!c.finalized` gets the e2e tests to pass again.

dai-shi: Thanks for investigating!
Hmm, but is there anything weird in the e2e tests?
I hope to avoid checking `ctx`, but only `c`... 🤔

rmarscher: I see. It looks like the dev server in the e2e tests has an issue. I am seeing it return a 404 error instead of the page response when I use the playwright vscode extension to debug rsc-basic.spec.ts.

I'll comment out the workaround and let the e2e tests fail until the root issue can be found with the e2e tests.

dai-shi: @himself65 do you have any idea?

packages/waku/src/cli.ts
```diff
@@ -96,6 +96,8 @@ async function runDev() {
   const config = await loadConfig();
   const app = new Hono();
   app.use('*', runner({ cmd: 'dev', config, env: process.env as any }));
+  // @ts-expect-error bypassing hono default notFound handler
+  app.use('*', (_c, _next) => {})
```
rmarscher, line 100: @dai-shi I think I figured it out. The default hono notFound handler is not needed for the dev server. The `next()` call is running before the dev server sets the result and that invokes the not found handler here - https://github.com/honojs/hono/blob/main/src/hono-base.ts#L424

I added an additional middleware here to exit the middleware loop without invoking that final next(). I'm not sure if there's a better way to handle this.

rmarscher: I should add that this is only a problem when the dev server is being used without the SSR middleware.

dai-shi: Nice finding! I'm not sure either, but feels like there should be a better way. Maybe we find it in the future.

dai-shi, line 100:
```suggestion
  // bypassing hono default notFound handler
  // FIXME there might be a better way to handle this
  app.use('*', () => {})
```

packages/waku/src/lib/hono/runner.ts
```diff
@@ -53,10 +53,13 @@ export const runner = (options: MiddlewareOptions): MiddlewareHandler => {
       });
     };
     await run(0);
-    return c.body(
-      ctx.res.body || null,
-      (ctx.res.status as any) || 200,
-      ctx.res.headers || {},
-    );
+    /* checking ctx.res.body is a workaround for dev server 404 response */
+    if (!c.finalized/* || ctx.res.body*/) {
```
dai-shi, line 57:
```suggestion
    if (!c.finalized) {
```

dai-shi (commented):
LGTM

final changes:

packages/waku/src/cli.ts
rmarscher: Without ts-ignore, this fails to pass typescript warnings during compilation. It's hard to read the hono middleware type signature to understand if there's another way that will pass type checking.

dai-shi: Oh, I see. Then, just add FIXME comment.

dai-shi (approved):
LGTM! Thanks for your contribution!

dai-shi:
Oops, please run prettier for packages/waku/src/cli.ts.

rmarscher:
> Oops, please run prettier for packages/waku/src/cli.ts.

All set. Sorry about that! I forgot it wasn't set to run on save.

PR #798, opened 2024-07-21
fix: hmr for .env files

Aslemammad:
Resolves #562 

We have to call dotenv on each change so it loads the new variables.

Previously, we used the first iteration of dotenv which happen to be called in the start of application.


packages/waku/src/router/fs-router.ts
```diff
@@ -25,9 +25,9 @@ export function fsRouter(
           { join, dirname, extname, sep },
           { fileURLToPath },
         ] = await Promise.all([
-          import(DO_NOT_BUNDLE + 'node:fs/promises'),
-          import(DO_NOT_BUNDLE + 'node:path'),
-          import(DO_NOT_BUNDLE + 'node:url'),
+          import(/* @vite-ignore */ DO_NOT_BUNDLE + 'node:fs/promises'),
+          import(/* @vite-ignore */ DO_NOT_BUNDLE + 'node:path'),
+          import(/* @vite-ignore */ DO_NOT_BUNDLE + 'node:url'),
```
dai-shi, line 30: I somehow thought it didn't cause issues. https://github.com/dai-shi/waku/pull/792/files#diff-dbb54d801e436acf21fb8afff808d82c614f9f0a7fc5e3ce8a666e437eef4d04
I think I missed something.

packages/waku/src/lib/plugins/vite-plugin-rsc-env.ts
```diff
@@ -11,6 +12,12 @@ export function rscEnvPlugin({
   return {
     name: 'rsc-env-plugin',
     config(viteConfig) {
+      dotenv.config({
+        path: ['.env.local', '.env'],
+        processEnv: (globalThis as any).__WAKU_PRIVATE_ENV__,
+        override: true,
+      });
```
dai-shi, line 19: Hm, I'm not sure how it works, but can we make this explicitly for dev only?

packages/waku/src/router/fs-router.ts
Aslemammad: Yea it does, so many warnings!

packages/waku/src/lib/plugins/vite-plugin-rsc-env.ts
Aslemammad: For sure!

packages/waku/src/lib/plugins/vite-plugin-rsc-env.ts
```diff
@@ -12,11 +14,13 @@ export function rscEnvPlugin({
   return {
     name: 'rsc-env-plugin',
     config(viteConfig) {
-      dotenv.config({
-        path: ['.env.local', '.env'],
-        processEnv: (globalThis as any).__WAKU_PRIVATE_ENV__,
-        override: true,
-      });
+      if (isDev) {
```
dai-shi, line 17: LGTM

packages/waku/src/lib/plugins/vite-plugin-rsc-env.ts
```diff
@@ -11,6 +14,14 @@ export function rscEnvPlugin({
   return {
     name: 'rsc-env-plugin',
     config(viteConfig) {
+      if (isDev) {
+        dotenv.config({
+          path: ['.env.local', '.env'],
+          processEnv: (globalThis as any).__WAKU_PRIVATE_ENV__,
+          override: true,
```
dai-shi, line 21: Oh, I see. This overrides the existing object.
Well, I guess there's a better solution, as now we don't have a worker.

dai-shi: Hm, but, this is fine for now. Let's merge.

dai-shi (approved):

PR #805, opened 2024-07-26
fix: sanitize slugs with `'.'` or `' '`

### LINKED ISSUE #803: Slug with '.' seems to break the build output
To reproduce:

```sh
npm create waku@next
npm install
npm run build
```

Screenshot of build output:

image

Each other slug creates `[slug]/index.html`, so I believe this is likely just an issue of escaping or replacing the `.` in Mr. Mime in this case.

I'd love to help with a PR for this, but wouldn't mind a nudge for where to start :)

tylerlaws0n:
fixes #803

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-react": "^7.34.3",
     "eslint-plugin-react-hooks": "^4.6.2",
-    "prettier": "^3.3.3",
+    "prettier": "^3.3.2",
     "prettier-plugin-tailwindcss": "^0.6.5",
     "terminate": "^2.8.0",
     "typescript": "^5.5.3",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-react": "^7.34.3",
     "eslint-plugin-react-hooks": "^4.6.2",
-    "prettier": "^3.3.2",
+    "prettier": "^3.3.3",
     "prettier-plugin-tailwindcss": "^0.6.5",
     "terminate": "^2.8.0",
     "typescript": "^5.5.3",
--- a/packages/waku/src/lib/builder/constants.ts
+++ b/packages/waku/src/lib/builder/constants.ts
@@ -5,3 +5,4 @@ export const DIST_SERVE_JS = 'serve.js';
 export const DIST_PUBLIC = 'public';
 export const DIST_ASSETS = 'assets';
 export const DIST_SSR = 'ssr';
+export const DOT_MIME = ':mime:';
--- a/packages/waku/src/lib/builder/constants.ts
+++ b/packages/waku/src/lib/builder/constants.ts
@@ -5,4 +5,3 @@ export const DIST_SERVE_JS = 'serve.js';
 export const DIST_PUBLIC = 'public';
 export const DIST_ASSETS = 'assets';
 export const DIST_SSR = 'ssr';
-export const PERIOD_ESCAPE = ':mime:';
--- a/packages/waku/src/lib/builder/constants.ts
+++ b/packages/waku/src/lib/builder/constants.ts
@@ -5,4 +5,4 @@ export const DIST_SERVE_JS = 'serve.js';
 export const DIST_PUBLIC = 'public';
 export const DIST_ASSETS = 'assets';
 export const DIST_SSR = 'ssr';
-export const DOT_MIME = ':mime:';
+export const PERIOD_ESCAPE = ':mime:';
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -175,11 +175,11 @@ export function createPages(
         page as {
           staticPaths: string[] | string[][];
         }
-      ).staticPaths.map((item) =>
-        Array.isArray(item)
-          ? item.map((i) => i.replace(/\./g, ''))
-          : [item.replace(/\./g, '')],
-      );
+      ).staticPaths
+        .map((item) => (Array.isArray(item) ? item : [item]))
+        .map((slug) =>
+          slug.map((i) => i.replace(/\./g, '').replace(/ /g, '-')),
+        );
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -79,7 +79,7 @@
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-react": "^7.34.3",
     "eslint-plugin-react-hooks": "^4.6.2",
-    "prettier": "^3.3.2",
+    "prettier": "^3.3.3",
```
tylerlaws0n, line 82: @dai-shi when I ran `pnpm i` for the first time, it got 3.3.3 because of the ^

I chose to just update here to make that change more obvious.

I can exclude the lockfile and package.json updates in this PR if that is your preference though. Just figured I'd leave it in and let you decide what to do.

dai-shi: I wonder how it happens, but in general I prefer 👇 

> I can exclude the lockfile and package.json updates in this PR if that is your preference though

packages/waku/src/lib/builder/constants.ts
```diff
@@ -5,3 +5,4 @@ export const DIST_SERVE_JS = 'serve.js';
 export const DIST_PUBLIC = 'public';
 export const DIST_ASSETS = 'assets';
 export const DIST_SSR = 'ssr';
+export const PERIOD_ESCAPE = ':mime:';
```
dai-shi, line 8: Is this a Vite thing?

tylerlaws0n: No, this is so that `extname` can be called without matching on `'.'` in the slug folder name. 

As part of the waku build process I am temporarily replacing `.` with this arbitrary constant (code change in `packages/waku/src/router/create-pages.ts`), so that we can call `extname` and do all the other normal logic, then just clean up the temporary symbol at the end (like 598 in `packages/waku/src/lib/builder/build.ts`)

package.json
tylerlaws0n: Updated to exclude those changes! :)

packages/waku/src/lib/builder/constants.ts
tylerlaws0n: Another solution I can suggest is just disallowing `.` in slugs. This could be done with a console warning and removing the `.` altogether instead of replacing it with an escape string to add back in at the end.

tylerlaws0n: Ok, I actually think disallowing `.` in slugs and simply removing them will be the best strategy here. 

This is the build output when doing the escaping I have currently in this PR:

image

And this is what I would expect the output to be if `.` was considered valid in a slug. However, the route `http://localhost:8080/mr. mime` results in a 404 and is unreachable. It could at that point be an issue in `http-server`, but I can't honestly think of a url path that used `.` as part of the path (it is always for the extension).

I'll add the commit to just strip out `.` from slugs, but happy to change if you'd rather another strategy.

tylerlaws0n: Along these same lines: Should `' '` be replaced with `'-'`?

packages/waku/src/router/create-pages.ts
```diff
@@ -175,7 +175,11 @@ export function createPages(
         page as {
           staticPaths: string[] | string[][];
         }
-      ).staticPaths.map((item) => (Array.isArray(item) ? item : [item]));
+      ).staticPaths.map((item) =>
+        Array.isArray(item)
+          ? item.map((i) => i.replace(/\./g, ''))
+          : [item.replace(/\./g, '')],
```
dai-shi, line 181: I would avoid repeating `replace()`.
```js
(Array.isArray(item) ? item : [item]).map(...)
```

packages/waku/src/lib/builder/constants.ts
tylerlaws0n: Tested with Astro (best standard I can think of for handling SSG like this)

for a file named `mr. mime.md` it gives the slug: `mr-mime`

image

I will update to do the following to match this
- remove `.` characters from slugs
- replace ` ` with `-`

dai-shi: > I will update to do the following to match this
> 
> * remove `.` characters from slugs
> * replace ` ` with `-`

Sounds good. Can you also add this current spec to https://github.com/dai-shi/waku/blob/main/docs/create-pages.mdx?

dai-shi:
> I will update to do the following to match this
> 
> * remove `.` characters from slugs
> * replace ` ` with `-`

Sounds good. Can you also add this current spec to https://github.com/dai-shi/waku/blob/main/docs/create-pages.mdx?

packages/waku/src/router/create-pages.ts
```diff
@@ -175,7 +175,11 @@ export function createPages(
         page as {
           staticPaths: string[] | string[][];
         }
-      ).staticPaths.map((item) => (Array.isArray(item) ? item : [item]));
+      ).staticPaths
+        .map((item) => (Array.isArray(item) ? item : [item]))
+        .map((slug) =>
+          slug.map((i) => i.replace(/\./g, '').replace(/ /g, '-')),
+        );
```
dai-shi, line 182: Is this less readable?
```suggestion
        .map((item) => (Array.isArray(item) ? item : [item]).map((i) => i.replace(/\./g, '').replace(/ /g, '-'));
```

packages/waku/src/router/create-pages.ts
```diff
@@ -175,7 +175,11 @@ export function createPages(
         page as {
           staticPaths: string[] | string[][];
         }
-      ).staticPaths.map((item) => (Array.isArray(item) ? item : [item]));
+      ).staticPaths.map((item) =>
+        (Array.isArray(item) ? item : [item]).map((i) =>
+          i.replace(/\./g, '').replace(/ /g, '-'),
+        ),
```
dai-shi, line 181: Looking at the doc change makes me feel like maybe we can define `sanitizeSlug` function out, and do this?
```suggestion
        (Array.isArray(item) ? item : [item]).map(sanitizeSlug),
```

dai-shi (approved):
Thanks so much for your contribution!

PR #807, opened 2024-07-27
docs: inform static users about the getConfig call

### LINKED ISSUE #804: `public/index.html` missing from `dist` on initial build
```
npm create waku
npm install
npm run build -- --with-netlify-static # same happens for vercel static
```

`dist/public/index.html` does not exist, so the output is incomplete for use as a truly static site. Each other page seems to build just fine because they have index.html files, but not for the home route `/`.

Also would love to help with a PR on this one!

This project is awesome btw, thanks! A minimal way to use server components is super cool for the react ecosystem. 🎉

tylerlaws0n:
Warning and fix at build time for static deploy when 'dynamic' is set in the `getConfig`

fixes #804

### DIFF
--- a/README.md
+++ b/README.md
@@ -829,6 +829,8 @@ Adding the `--with-vercel-static` flag to the build script will produce static s
 }
 ```
 
+Note: When rendering in static mode, please be sure to return `render: 'static'` from each of your page files' `getConfig()`.
+
 ### Netlify
 
 Waku projects can be deployed to Netlify with the [Netlify CLI](https://docs.netlify.com/cli/get-started/).
@@ -849,6 +851,8 @@ Adding the `--with-netlify-static` flag to the build script will produce static 
   }
 }
 ```
+
+Note: When rendering in static mode, please be sure to return `render: 'static'` from each of your page files' `getConfig()`.
 
 ### Cloudflare (experimental)
 


packages/waku/src/router/fs-router.ts
```diff
@@ -77,6 +77,19 @@ export function fsRouter(
             render: 'static',
             ...config,
           });
+        } else if (
+          process.env.npm_lifecycle_script?.includes('static') &&
```
dai-shi, line 81: I would like to avoid using `process`.
L18 checks if it's PRD, but in general we make all code (except for dev/build specific files) free from Node.js. This is still confusing and I hope to organize the code structure better.

packages/waku/src/router/fs-router.ts
```diff
@@ -77,6 +77,19 @@ export function fsRouter(
...
+          config?.render === 'dynamic'
+        ) {
+          console.warn(
+            `[WARN] getConfig for path: '${path}' cannot be dynamic for static deploys. Switching to static.`,
```
dai-shi, line 85: I would just warn it and not automatically switch to static.

dai-shi (commented):
Thanks. I generally like the idea of warning this case, but I don't prefer the code.

I think we need to improve our handling of build process (so that, in this case, `fs-router.ts` could know if it's static build.)

packages/waku/src/router/fs-router.ts
tylerlaws0n: Makes sense. 

Should there be some concept of a "build context" where we pass a state object in with all of the state for a given build. 

Here that would mean passing the build mode (static or dynamic) but I imagine other pieces of state would be helpful to pass around to many of the builder functions.

If not this, how else would you like to communicate to fs-router what the build type is?

packages/waku/src/router/fs-router.ts
tylerlaws0n: will make this change when resolving the other comment, thanks!

packages/waku/src/router/fs-router.ts
dai-shi: At this point, I don't think it's possible. It's due to my current design which I find limited, and for now lots of hacks around "build config".
I would like to revamp things around it after finalizing `v0.21.0`. Can consider about "build context" too then.

So, for now let's wait for it and mitigate the issue with readme, please?

dai-shi (approved):
Looks good to me! Can you update the PR title please?

PR #809, opened 2024-07-31
refactor(dev): avoid willBeHandledLater

dai-shi:
This is the first step to hide `willBeHandledLater` from `ctx.devServer`.


packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -126,6 +126,9 @@ const createMainViteServer = (
         external: ['waku'],
         noExternal: ['react-server-dom-webpack'],
       },
+      // FIXME Can we make 'mpa' and let fallback middleware handle DEV?
+      // It will remove `willBeHandled` hack.
```
dai-shi, line 130: I'll work on the second step later, which should remove the HACK depending on `rscPath` below.

PR #811, opened 2024-08-02
chore: add nvmrc and node-version to gitignore

tylerlaws0n:
Feel free to just close this if there's a preference to not have this. 

My motivation for adding this is just that with [fnm](https://github.com/Schniz/fnm) (and other node version managers) - they look for this file or .nvmrc to automatically set the node version. Adding this would make it more convenient to swap between different projects that use different node versions.

I chose 20.8.0 because it is enforced in the `packages/waku/package.json` engines field.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ dist
 .vscode
 *.tsbuildinfo
 .cache
+.nvmrc
+.node-version


dai-shi:
Thanks for your suggestion. I would avoid it for now for some preferences:
- If possible, I would like to avoid creating files and especially dot files. (If it were package.json field, it might be better.)
- There are several node package managers.
- I use several node versions for developing.

tylerlaws0n:
> Thanks for your suggestion. I would avoid it for now for some preferences:
> 
> * If possible, I would like to avoid creating files and especially dot files. (If it were package.json field, it might be better.)
> * There are several node package managers.
> * I use several node versions for developing.

How would you feel about adding `.nvmrc` and/or `.node-version` to the .gitignore file then?

dai-shi:
> How would you feel about adding `.nvmrc` and/or `.node-version` to the .gitignore file then?

I'm fine with it.

dai-shi (approved):

PR #812, opened 2024-08-02
feat: update types for static createPage

tylerlaws0n:


packages/waku/tests/vite-plugin-rsc-analyze.test.ts
```diff
@@ -2,16 +2,12 @@ import { build } from 'vite';
...
 import path from 'node:path';
-import crypto from 'node:crypto';
 import type { LoggingFunction, RollupLog } from 'rollup';
 
 import { rscAnalyzePlugin } from '../src/lib/plugins/vite-plugin-rsc-analyze.js';
 
 const root = fileURLToPath(new URL('./fixtures', import.meta.url));
 
-// FIXME vitest node environment does not have crypto
-globalThis.crypto = crypto as any;
```
tylerlaws0n, line 13: This didn't seem needed anymore. Curious if using node 20 is what fixes this?

I found this related discussion: https://github.com/vitest-dev/vitest/discussions/893

but fwiw this is the output when I `console.log(globalThis.crypto)` when running `vitest --silent=false`

testsvite-plugin-rsc-analyze test ts129

happy to leave this out of the PR, but seems good to remove if it is truly no longer an issue

packages/waku/tests/create-pages.test.ts
```diff
@@ -73,6 +73,14 @@ describe('type tests', () => {
         staticPaths: ['x', 'y', 'z'],
         component: () => 'Hello',
       });
+
+      // also good
+      createPage({
+        render: 'static',
+        path: '/test/[a]/[b]',
```
tylerlaws0n, line 80: @dai-shi This was commented as being an invalid slug, but from my understanding of the docs, nested segment routes are supported in static mode? https://waku.gg/#nested-segment-routes

It seems like what is not supported is the "catch-all" wildcard slugs, so I updated the tests and types here to reflect the docs as best I could.

packages/waku/package.json
```diff
@@ -64,6 +64,7 @@
   "scripts": {
     "dev": "swc src -d dist -w --strip-leading-paths",
     "test": "vitest run",
+    "test:watch": "vitest",
```
tylerlaws0n, line 67: I personally like running the test script in watch mode locally, but happy to leave this out if you'd prefer.

packages/waku/tests/vite-plugin-rsc-analyze.test.ts
tylerlaws0n: So this fails in CI using node 18.17.0!

I was able to replicate this locally as well... so that begs the question, do we need to support node 18, or would upgrading the version used in CI and in the engines field of the package.json to all use node 20 be acceptable?

Or would you prefer we just update the comment here mentioning that this issue is specific to node 18 and no longer an issue in node 20?

dai-shi: Yes, this is for node 18. Feel free to update the comment.
We may drop node 18 in the future and run CI with node 20 and node 22.

packages/waku/package.json
dai-shi: Please leave it as it's called from the monorepo root.
I hope you can use `npx vitest` instead.

packages/waku/tests/create-pages.test.ts
dai-shi: Thanks!!

packages/waku/package.json
dai-shi: I misunderstood. Adding `test:watch` is acceptable. (I confused with changing `test`)

packages/waku/tests/create-pages.test.ts
tylerlaws0n: Is my understanding correct though? There is also code to test that static wildcard support does exist:

https://github.com/dai-shi/waku/blob/main/packages/waku/tests/create-pages.test.ts#L518-L527

https://github.com/dai-shi/waku/blob/main/packages/waku/src/router/create-pages.ts#L196-L201

I can change these in this PR as well of course... I just wanted to be careful and call this code out because it seems to be in conflict with the docs at the moment.

dai-shi: I guess so. But, I'm not super confident with anything. I would not be surprised if there were differences among implementation, tests and docs.
I'm fine if you completely re-implement it from scratch around this path handling (as it's not tightly related with `unstable_defineRouter` which I'll be modifying.)

tylerlaws0n: Sounds good to me.

Here's what I'll plan to start with: 

Breaking up the 4 different types of pages into 4 different functions for handling

They can all be defined inside createPages to share other logic and utilities, but a good starting point to me would be unraveling this code a bit to start making things more clear.

I'll update more here as I make progress and start a discussion if we need to make larger design decisions.

tylerlaws0n:
converted to draft to use this PR as a place to make progress for more refactoring of the `createPages` function.

At a minimum, I want to be confident for what routes are supported in static mode before considering merging this.

tylerlaws0n:
Quick Update: I am focusing on the types first to familiarize myself with the api and what everything is responsible for.

Changes made so far:
- added a utility type for splitting strings which will make parsing routes as string a lot more manageable imo
- added a type for getting the appropriate tuple type relative to the route given for `staticPaths` 

image

Next I am looking into the catch all routes more deeply... I think it might actually be perfectly valid and supported to include these for the static mode, but I want to run the code and test this to see what the limitations would be, then update the types to describe this.

packages/waku/tests/create-pages.test.ts
tylerlaws0n: Following back up here: I updated the tests and disabled wildcard/catch-all routes for  for the static render mode fully.

My guess is that these were added initially in error when adding all types of slug support for each render mode.

But when I stop to think about "what would it mean to be a wildcard route in a static site?" there really is no way to support this.

For a wildcard route to work, you would need to run logic on the request to match the request to the wildcard slug and then inject that as a value into the component for that page. This is simply not possible when you are requesting a static file from a static file server. Your request is taken as a literal representation for the html file you are interested in.

Finally, I am confident this changeset is ready for a look and I will start a discussion with some follow up thoughts about the `createPages` code 

I'll leave this thread open in case there are some further thoughts around this :)

dai-shi:
Is it ready?

tylerlaws0n:
> Is it ready?

I'll need to update the playwright tests as well... forgot about those. 

They're failing because there are tests that expect ssg mode to work for wildcards. @dai-shi do you mind just confirming that what I said above is correct in that wildcard slugs for ssg mode don't make sense and we do not intend to support them?

I will rewrite the tests to expect failure in the `ssg + wildcard slug` scenario if you agree here. I can also update the docs to make this more explicit.

dai-shi:
> confirming that what I said above

sorry, I think I missed it.

> that wildcard slugs for ssg mode don't make sense and we do not intend to support them?

hm? I think slugs should work for ssg, even with wildcard `[...slug]`, as long as we have `staticPaths=`. I think we intentionally support it.

tylerlaws0n:
@dai-shi this should be ready now and updated to support wildcards correctly! including the type matching for slugs :)

packages/waku/tests/create-pages.test.ts
tylerlaws0n: resolved this by appropriately supporting wildcards for static mode... some examples:

```ts
createPage({
  render: 'static',
  path: '/nested/[...ids]',
  // results in paths /nested/foo/bar and /nested/foo/baz and /nested/foo
  staticPaths: [['foo', 'bar'], ['foo','baz'], ['foo']],
  component: ...,
});  
```

dai-shi (approved):
Other than that, it looks good to me. (I didn't read & understand every detail of the type utils.)

packages/waku/src/router/create-pages.ts
```diff
@@ -30,6 +30,27 @@ const sanitizeSlug = (slug: string) =>
...
+ * Type version of `String.prototype.split()`. Splits the first string argument by the second string argument
+ * @example
+ * ```ts
+ * // ['a', 'b', 'c']
+ * type Case1 = Split<'abc', ''>
+ * // ['a', 'b', 'c']
+ * type Case2 = Split<'a,b,c', ','>
+ * ```
+ */
+export type Split<
```
dai-shi, line 43: I would like to not export types until it's necessary.

```suggestion
type Split<
```

dai-shi: > Other than that

I forgot to submit this comment previously...

tylerlaws0n: fixed

dai-shi: Sorry, I wasn't very clear. Can you do it for all new types?

tylerlaws0n: oh, yes! Done!

PR #814, opened 2024-08-05
fix: not found in DEV and refactor fallback handling

dai-shi:
following up #809.

This is still incomplete and need more refactors as noted in `FIXME`.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -96,9 +96,14 @@ async function runDev() {
   const config = await loadConfig();
   const app = new Hono();
   app.use('*', runner({ cmd: 'dev', config, env: process.env as any }));
-  // bypassing hono default notFound handler
-  // @ts-expect-error FIXME there might be a better way to handle this
-  app.use('*', (_c, _next) => {});
+  app.notFound((c) => {
+    // FIXME can we avoid hardcoding the public path?
+    const file = path.join('public', '404.html');
+    if (existsSync(file)) {
+      return c.html(readFileSync(file, 'utf8'), 404);
+    }
+    return c.text('404 Not Found', 404);
+  });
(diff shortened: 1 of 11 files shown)


packages/waku/src/cli.ts
```diff
@@ -96,9 +96,14 @@ async function runDev() {
...
-  // @ts-expect-error FIXME there might be a better way to handle this
-  app.use('*', (_c, _next) => {});
+  app.notFound((c) => {
+    // FIXME can we avoid hardcoding the public path?
+    const file = path.join('public', '404.html');
+    if (existsSync(file)) {
+      return c.html(readFileSync(file, 'utf8'), 404);
+    }
+    return c.text('404 Not Found', 404);
+  });
```
dai-shi, line 106: I wonder if this breaks what we fixed in #796.

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -369,6 +368,7 @@ export const devServer: Middleware = (options) => {
 
     if (
       // HACK depending on `rscPath` is a bad idea
+      // FIXME This hack should be removed as well as `willBeHandled`
       ctx.req.url.pathname.startsWith(config.basePath + config.rscPath + '/') ||
       !(await willBeHandled(ctx.req.url.pathname))
```
dai-shi, line 373: I was hoping to remove this, but I'm not sure how to deal with it with Vite. So, I hope someone will help in the future.

dai-shi:
I hope this doesn't break #644 and others, but not 100% sure. cc @rmarscher 
Let's merge and see if we get any feedback.

rmarscher:
I deployed a test with 0.21.0-beta.5 and it still responds with the expected 404 error and not a 200. 👍

PR #820, opened 2024-08-08
refactor(router): style/performance tweaks to logic and jsdoc added

tylerlaws0n:
What I have focused on so far: 

- Removing unnecessary creation of arrays for loops (using iterability of maps)
- jsdoc for functions that I am reading through to understand in the router
- moved some of the types around
  - Not sure how I feel about this yet... if splitting up the logic for each of these entry points ends up working out, it will be nice to have these types re-used for that. Otherwise, I will revert that part. (CreatePageStaticWithoutSlug & CreatePageStaticWithSlug)
 
Notes:

I think I will be adding some more docs and re-organizing a bit more before calling this PR ready.

I am interested to hear opinions and thoughts on the changes so far though, if any.

### DIFF
--- a/packages/waku/src/lib/utils/path.ts
+++ b/packages/waku/src/lib/utils/path.ts
@@ -105,6 +105,21 @@ export const path2regexp = (path: PathSpec) => {
   return `^/${parts.join('/')}$`;
 };
 
+/**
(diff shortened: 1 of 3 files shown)


dai-shi (approved):
LGTM
Thanks for your work!

PR #822, opened 2024-08-09
chore(deps): update typescript-eslint

dai-shi:


dai-shi:
I don't think this is a correct workaround, but merging for now.
Please someone give us advices.

PR #836, opened 2024-08-19
fix: hot reload rsc

dai-shi:
I forgot to change this in #835. TS doesn't work for strings 😓 .

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -371,9 +371,9 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
   }, [changeRoute]);
 
   useEffect(() => {
-    const callback = (pathname: string, query: string) => {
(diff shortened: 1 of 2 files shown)


packages/waku/src/router/client.ts
```diff
@@ -371,10 +371,10 @@ const InnerRouter = ({ routerData }: { routerData: RouterData }) => {
   }, [changeRoute]);
 
   useEffect(() => {
-    const callback = (pathname: string, searchParamsString: string) => {
+    const callback = (path: string, query: string) => {
       const url = new URL(window.location.href);
-      url.pathname = pathname;
-      url.search = searchParamsString;
+      url.pathname = path;
+      url.search = query;
```
dai-shi, line 377: This is just a refactor unrelated to this PR. (But found it by searching `searchParamsString`.)

PR #838, opened 2024-08-19
test: add vercel ai fixtures

himself65:
Upstream: https://github.com/dai-shi/waku/pull/788


dai-shi (approved):
Thanks for adding this.
My only concern is `ssr-basic` is no longer something minimal, so is `rsc-basic` already.
Other than that, it looks good to me.

PR #850, opened 2024-08-26
[fix] File upload

cjpais:
I have a simple HTML form which submits a file to a server action via multipart form upload. From the client side, the file looks proper, but when the server action receives it, the file is much larger? Is there some intermediate processing of this file or improper handling in the library? 

client sent:
```
File {
  lastModified: 1724514738000
  name: "bara.jpeg"
  size: 2094593
  type: "image/jpeg"
}
```

server received:
```
File {
  size: 3715003,
  type: 'image/jpeg',
  name: 'bara.jpeg',
  lastModified: 1724698582443
}
```

I went through the code, and the bug largely was parsing the body as a string. Instead we have an array buffer so we can manipulate it more directly. It's possible it may still work as a string, but I believe the array buffer leaves more flexibility.

I've only tested directly in the application I've been building


cjpais:
i would guess the right thing to do is probably move to busboy, etc

packages/waku/src/lib/utils/form.ts
```diff
@@ -1,8 +1,16 @@
 // TODO is this correct? better to use a library?
-export const parseFormData = (body: string, contentType: string) => {
+export const parseFormData = (body: ArrayBuffer, contentType: string) => {
```
dai-shi, line 2: Would you be able to create a test for this function?

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -194,13 +194,14 @@ export async function renderRsc(
 
   let decodedBody: unknown | undefined = args.decodedBody;
   if (body) {
-    const bodyStr = await streamToString(body);
+    const bodyBuf = await streamToArrayBuffer(body);
+    const bodyStr = arrayBufferToString(bodyBuf);
```
dai-shi, line 198: We can delay calling this in the else-if block. We just need to check if the buffer has a length.

dai-shi (commented):
Nice catch and this looks great. Thanks for working on it.

dai-shi:
> i would guess the right thing to do is probably move to busboy, etc

We used to use busboy as React lib uses it, but it's Node.js specific and Waku now supports non-Node.js envs.

packages/waku/src/lib/utils/form.ts
cjpais: will try, going on vacation soon so have a lot of other projects to wrap up

packages/waku/src/lib/renderers/rsc-renderer.ts
cjpais: great, latest version should have this

packages/waku/src/lib/utils/form.ts
dai-shi: looking forward to it. take your time.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: Cool, thanks.

cjpais:
tests are a good starting point, but really need more. large files, etc.. the current code does not pass. the tests themselves need some review for accuracy as well

packages/waku/src/lib/utils/stream.ts
```diff
@@ -11,6 +11,31 @@ export const concatUint8Arrays = (arrs: Uint8Array[]): Uint8Array => {
...
+
+  do {
+    ({ done, value } = await reader.read());
+    if (!done && value) {
+      chunks.push(value);
+      totalSize += value.length;
+    }
+  } while (!done);
+
+  const result = new Uint8Array(totalSize);
```
dai-shi, line 29: A little concern for the performance is the two loops, but it's easier to know the `totalSize`.

packages/waku/src/lib/utils/stream.ts
```diff
@@ -31,6 +56,13 @@ export const streamToString = async (
   return outs.join('');
 };
 
+export function arrayBufferToString(buffer: ArrayBuffer): string {
+  const uint8Array = new Uint8Array(buffer);
+  return Array.from(uint8Array)
```
dai-shi, line 61: I'm not familiar but does it create a new array?

dai-shi (commented):
Yeah, those tests are good start. Thanks for working on it. It's hard to tell the accuracy, but it's noticeable.

2nd review 👇 . Maybe we need a test for `arrayBufferToString`.

packages/waku/src/lib/utils/stream.ts
```diff
@@ -31,6 +56,13 @@ export const streamToString = async (
...
+    .map((byte) => String.fromCharCode(byte))
```
dai-shi, line 62: This doesn't seem to work for non-ASCII chars.

Do we have the same issue in `parseFormData`?

cjpais: yep

packages/waku/src/lib/utils/stream.ts
cjpais: array.from does yes, I think we should try to avoid conversions if possible, or convert once and handle from there. this feels like a hack to me

cjpais:
So I initially wrote this code as a hack to solve a specific issue for my project. I think after writing the tests and seeing some of the review comments, it might be nice to review this body of code in total and maybe write a better implementation to handle more cases (like binary, base64, quoted strings, etc). 

Mentioned in discord but I see Hono is doing this. Not sure if applicable given support for non node.js based runtimes, I'm not super familiar with the API's down to that level. But since Hono supports different runtimes, I imagine there are either polyfills or something to get the parsing working.

```
export const bufferToFormData = (
  arrayBuffer: ArrayBuffer,
  contentType: string
): Promise<FormData> => {
  const response = new Response(arrayBuffer, {
    headers: {
      'Content-Type': contentType,
    },
  })
  return response.formData()
}
```

if I get a chance today I may try something as simple as this. (however I am using bun runtime so it probably has support for the node.js apis)

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -9,10 +9,11 @@ import type {
 } from '../../server.js';
 import type { ResolvedConfig } from '../config.js';
 import { filePathToFileURL } from '../utils/path.js';
-import { parseFormData } from '../utils/form.js';
-import { streamToArrayBuffer, arrayBufferToString } from '../utils/stream.js';
+import { streamToArrayBuffer } from '../utils/stream.js';
 import { decodeActionId } from '../renderers/utils.js';
 
+import { bufferToFormData, bufferToString } from 'hono/utils/buffer';
+
```
cjpais, line 16: use Hono directly

packages/waku/src/lib/utils/stream.ts
```diff
@@ -56,13 +56,6 @@ export const streamToString = async (
   return outs.join('');
 };
 
-export function arrayBufferToString(buffer: ArrayBuffer): string {
```
cjpais: Hono provides this now

packages/waku/src/lib/utils/form.ts
cjpais, line 1: deleted because Hono provides it now

packages/waku/src/lib/utils/stream.ts
cjpais: yeah may try to remove the second loop, theres surely a better way

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: This file should not depend on Hono. It can be used without Hono.

dai-shi: This file should not depend on Hono. It can be used without Hono.

dai-shi (changes requested):

packages/waku/src/lib/utils/stream.ts
cjpais: left for now.. according to ai the performance shouldn't be big concern

packages/waku/src/lib/utils/buffer.ts
```diff
@@ -0,0 +1,19 @@
...
+  const response = new Response(buffer, {
+    headers: {
+      'Content-Type': contentType,
+    },
+  });
+  return response.formData();
+};
+
+export const bufferToString = (buffer: ArrayBuffer): string => {
+  if (buffer instanceof ArrayBuffer) {
```
dai-shi, line 14: Do we need this check?

packages/waku/src/lib/utils/stream.ts
dai-shi: I don't trust ai in this case. two loops should be of impact. This is fine for now. Can you leave a FIXME comment in the code?

dai-shi (commented):
Overall looks good. Thanks for working on it.

packages/waku/src/lib/utils/buffer.ts
```diff
@@ -0,0 +1,16 @@
...
+  const response = new Response(buffer, {
+    headers: {
+      'Content-Type': contentType,
+    },
+  });
+  return response.formData();
+};
+
+export const bufferToString = (buffer: ArrayBuffer): string => {
+  const enc = new TextDecoder('utf-8');
```
dai-shi, line 14: Can use the default. https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder

```suggestion
  const enc = new TextDecoder();
```

packages/waku/src/lib/utils/buffer.ts
```diff
@@ -0,0 +1,16 @@
+export const parseFormData = async (
+  buffer: ArrayBuffer,
+  contentType: string,
+): Promise<FormData> => {
+  const response = new Response(buffer, {
+    headers: {
+      'Content-Type': contentType,
```
dai-shi, line 7: Maybe it works with this:
```suggestion
      'content-type': contentType,
```

dai-shi (approved):
This is so great! Thanks for your contribution!

PR #853, opened 2024-08-28
Add support for React Compiler

L-Qun:
Now the users can enable the react compiler by adding a new field `reactCompiler: true` in the `waku.config.ts` file.

Let's take a look at the effects before and after enabling react compiler optimization.

```
const Child = () => {
  console.log("render");
  return <div>Child</div>;
};

export const Counter = () => {
  const [count, setCount] = useState(0);

  const handleIncrement = () => setCount((c) => c + 1);

  return (
      <div>Count: {count}</div>
      <button
        onClick={handleIncrement}
        className="rounded-sm bg-black px-2 py-0.5 text-sm text-white"
      >
        Increment
      </button>
      <Child />
    </section>
  );
};
```

When the Child component re-renders, it will console "render".

before enabling the react compiler:
20240828231420_rec_

after enabling the react compiler:
20240828230748_rec_

### DIFF
--- a/packages/waku/src/config.ts
+++ b/packages/waku/src/config.ts
@@ -2,6 +2,11 @@ import type { Middleware } from './lib/middleware/types.js';
 
 export type { Middleware };
 
+export interface ReactCompilerOptions {
+  compilationMode?: 'infer' | 'annotation' | 'all'
+  panicThreshold?: 'ALL_ERRORS' | 'CRITICAL_ERRORS' | 'NONE'
+}
+
 export interface Config {
   /**
    * The base path for serve HTTP.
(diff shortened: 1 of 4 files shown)


L-Qun:
Accordingly, update the React documentation. https://github.com/reactjs/react.dev/pull/7133

dai-shi:
We don't do this at least for now.
You can create vite.config.ts instead. See: https://x.com/dai_shi/status/1794171384761094385

Now, may I ask you to add `./docs/guides/react-compiler.mdx` for the instruction? That would be helpful.

L-Qun:
https://github.com/dai-shi/waku/pull/857

dai-shi:
Now, I wonder how you made it work without "annotation". 🤔

PR #854, opened 2024-08-28
feat: infer types for waku router

tylerlaws0n:
Currently this changeset gets `createPages` ready to return an inferred type for each response of `createPage`

TODO:

- [x] Create a union type of all possible paths given the pages
- [x] Pass this union to an override-able interface so waku projects can consume the pages type safety (same as tanstack router's approach)
- [x] Use that type in our `<Link />` component
- [x] Update tests and refactor examples to match new api

### DIFF
--- a/docs/create-pages.mdx
+++ b/docs/create-pages.mdx
@@ -19,22 +19,64 @@ import { createPages } from 'waku';
 import { RootLayout } from './templates/root-layout';
 import { HomePage } from './templates/home-page';
 
-export default createPages(async ({ createPage, createLayout }) => {
+const pages = createPages(async ({ createPage, createLayout }) => [
   // Create root layout
   createLayout({
     render: 'static',
     path: '/',
     component: RootLayout,
-  });
+  }),
 
   // Create home page
   createPage({
     render: 'dynamic',
     path: '/',
     component: HomePage,
-  });
-});
+  }),
+]);
+
+export default pages;
 ```
+
+### Router Paths Type Safety
+
+Waku provides inference for the router paths when created pages are returned from the callback passed into `createPages`. The following example shows how a minimal example for how to setup the router paths type safety.
+
+```tsx
+// ./src/entries.tsx
+import { createPages } from 'waku';
+import type { PathsForPages } from 'waku/router';
+
+import { HomePage } from './templates/home-page';
+
+const pages = createPages(async ({ createPage, createLayout }) => [
+  // Create root layout
+  createLayout({
+    render: 'static',
+    path: '/',
+    component: RootLayout,
+  }),
+
+  // Create home page
+  createPage({
+    render: 'dynamic',
+    path: '/',
+    component: HomePage,
+  }),
+]);
+
+declare module 'waku/router' {
+  interface RouteConfig {
+    paths: PathsForPages<typeof pages>;
+  }
+}
+
+export default pages;
+```
+
+Once this is done, any `<Link />` component or hook from `waku/router` that uses paths in your app will use this type. In this case, the one valid use would be `<Link to="/" />`, but as you add more pages to the router, this type will grow to include them.
+
+Note: The file-based router paths will be supported in the future with some form of code-generation to get the types from your local page files.
 
 ### Pages
 
@@ -49,21 +91,21 @@ import { createPages } from 'waku';
 import { AboutPage } from './templates/about-page';
 import { BlogIndexPage } from './templates/blog-index-page';
 
-export default createPages(async ({ createPage }) => {
+export default createPages(async ({ createPage }) => [
   // Create about page
   createPage({
     render: 'static',
     path: '/about',
     component: AboutPage,
-  });
+  }),
 
   // Create blog index page
   createPage({
     render: 'static',
     path: '/blog',
     component: BlogIndexPage,
-  });
-});
+  }),
+]);
 ```
 
 #### Segment routes
@@ -79,7 +121,7 @@ import { createPages } from 'waku';
 import { BlogArticlePage } from './templates/blog-article-page';
 import { ProductCategoryPage } from './templates/product-category-page';
 
-export default createPages(async ({ createPage }) => {
+export default createPages(async ({ createPage }) => [
   // Create blog article pages
   // `<BlogArticlePage>` receives `slug` prop
   createPage({
@@ -87,7 +129,7 @@ export default createPages(async ({ createPage }) => {
     path: '/blog/[slug]',
     staticPaths: ['introducing-waku', 'introducing-create-pages'],
     component: BlogArticlePage,
-  });
+  }),
 
   // Create product category pages
   // `<ProductCategoryPage>` receives `category` prop
@@ -95,8 +137,8 @@ export default createPages(async ({ createPage }) => {
     render: 'dynamic',
     path: '/shop/[category]',
     component: ProductCategoryPage,
-  });
-});
+  }),
+]);
 ```
 
 Static paths (or other values) could also be generated programmatically.
@@ -111,12 +153,14 @@ import { BlogArticlePage } from './templates/blog-article-page';
 export default createPages(async ({ createPage }) => {
   const blogPaths = await getBlogPaths();
 
-  createPage({
-    render: 'static',
-    path: '/blog/[slug]',
-    staticPaths: blogPaths,
-    component: BlogArticlePage,
-  });
+  return [
+    createPage({
+      render: 'static',
+      path: '/blog/[slug]',
+      staticPaths: blogPaths,
+      component: BlogArticlePage,
+    }),
+  ];
 });
 ```
 
@@ -130,15 +174,15 @@ import { createPages } from 'waku';
 
 import { ProductDetailPage } from './templates/product-detail-page';
 
-export default createPages(async ({ createPage }) => {
+export default createPages(async ({ createPage }) => [
   // Create product detail pages
   // `<ProductDetailPage>` receives `category` and `product` props
   createPage({
     render: 'dynamic',
     path: '/shop/[category]/[product]',
     component: ProductDetailPage,
-  });
-});
+  }),
+]);
 ```
 
 For static prerendering of nested segment routes, the `staticPaths` array is instead composed of ordered arrays.
@@ -149,7 +193,7 @@ import { createPages } from 'waku';
 
 import { ProductDetailPage } from './templates/product-detail-page';
 
-export default createPages(async ({ createPage }) => {
+export default createPages(async ({ createPage }) => [
   // Create product detail pages
   // `<ProductDetailPage>` receives `category` and `product` props
   createPage({
@@ -160,8 +204,8 @@ export default createPages(async ({ createPage }) => {
       ['some-category', 'another-product'],
     ],
     component: ProductDetailPage,
-  });
-});
+  }),
+]);
 ```
 
 #### Catch-all routes
@@ -176,7 +220,7 @@ import { createPages } from 'waku';
 
 import { DashboardPage } from './templates/dashboard-page';
 
-export default createPages(async ({ createPage }) => {
+export default createPages(async ({ createPage }) => [
   // Create account dashboard
   // `<DashboardPage>` receives `catchAll` prop (string[])
   createPage({
@@ -184,7 +228,7 @@ export default createPages(async ({ createPage }) => {
     path: '/app/[...catchAll]',
     component: DashboardPage,
   });
-});
+]);
 ```
 
 ### Layouts
@@ -201,14 +245,14 @@ import { createPages } from 'waku';
 
 import { RootLayout } from './templates/root-layout';
 
-export default createPages(async ({ createLayout }) => {
+export default createPages(async ({ createLayout }) => [
   // Add a global header and footer
   createLayout({
     render: 'static',
     path: '/',
     component: RootLayout,
-  });
-});
+  }),
+]);
 ```
 
 ```tsx
@@ -255,14 +299,14 @@ import { createPages } from 'waku';
 
 import { BlogLayout } from './templates/blog-layout';
 
-export default createPages(async ({ createLayout }) => {
+export default createPages(async ({ createLayout }) => [
   // Add a sidebar to the blog index and blog article pages
   createLayout({
     render: 'static',
     path: '/blog',
     component: BlogLayout,
-  });
-});
+  }),
+]);
 ```
 
 ```tsx
--- a/e2e/fixtures/render-type/src/entries.tsx
+++ b/e2e/fixtures/render-type/src/entries.tsx
@@ -3,27 +3,33 @@ import { createPages } from 'waku/router/server';
 import { ServerEcho } from './ServerEcho.js';
 import { ClientEcho } from './ClientEcho.js';
 
-export default createPages(async ({ createPage }) => {
-  createPage({
-    render: 'dynamic',
-    path: '/server/dynamic/[echo]',
-    component: ServerEcho,
-  });
-  createPage({
-    render: 'static',
-    path: '/server/static/[echo]',
-    staticPaths: ['static-echo'],
-    component: ServerEcho,
-  });
-  createPage({
-    render: 'dynamic',
-    path: '/client/dynamic/[echo]',
-    component: ClientEcho,
-  });
-  createPage({
-    render: 'static',
-    path: '/client/static/[echo]',
-    staticPaths: ['static-echo'],
-    component: ClientEcho,
-  });
-});
+// This needs type annotations for the return type of createPages
+// @see https://github.com/microsoft/TypeScript/issues/42873#issuecomment-2065572017
+const router: ReturnType<typeof createPages> = createPages(
+  async ({ createPage }) => [
+    createPage({
+      render: 'dynamic',
+      path: '/server/dynamic/[echo]',
+      component: ServerEcho,
+    }),
+    createPage({
+      render: 'static',
+      path: '/server/static/[echo]',
+      staticPaths: ['static-echo'],
+      component: ServerEcho,
+    }),
+    createPage({
+      render: 'dynamic',
+      path: '/client/dynamic/[echo]',
+      component: ClientEcho,
+    }),
+    createPage({
+      render: 'static',
+      path: '/client/static/[echo]',
+      staticPaths: ['static-echo'],
+      component: ClientEcho,
+    }),
+  ],
+);
+
+export default router;
--- a/examples/07_router/src/entries.tsx
+++ b/examples/07_router/src/entries.tsx
@@ -1,5 +1,6 @@
 import { lazy } from 'react';
 import { createPages } from 'waku';
+import type { PathsForPages } from 'waku/router';
 
 import FooPage from './components/FooPage';
 
@@ -10,51 +11,51 @@ const BarPage = lazy(() => import('./components/BarPage'));
 const NestedBazPage = lazy(() => import('./components/NestedBazPage'));
 const NestedQuxPage = lazy(() => import('./components/NestedQuxPage'));
 
-export default createPages(async ({ createPage, createLayout }) => {
+const pages = createPages(async ({ createPage, createLayout }) => [
   createLayout({
     render: 'static',
     path: '/',
     component: HomeLayout,
-  });
+  }),
 
   createPage({
     render: 'static',
     // render: 'dynamic',
     path: '/',
     component: HomePage,
-  });
+  }),
 
   createPage({
     render: 'static',
     // render: 'dynamic',
     path: '/foo',
     component: FooPage,
-  });
+  }),
 
   createPage({
     render: 'static',
     path: '/bar',
     component: BarPage,
-  });
+  }),
 
   createPage({
     render: 'dynamic',
     path: '/baz',
     // Inline component is also possible.
     component: () => <h2>Dynamic: Baz</h2>,
-  });
+  }),
 
   createPage({
     render: 'static',
     path: '/nested/baz',
     component: NestedBazPage,
-  });
+  }),
 
   createPage({
     render: 'static',
     path: '/nested/qux',
     component: NestedQuxPage,
-  });
+  }),
 
   createPage({
     render: 'static',
@@ -66,7 +67,7 @@ export default createPages(async ({ createPage, createLayout }) => {
         <h3>Static: {id}</h3>
       </>
     ),
-  });
+  }),
 
   createPage({
     render: 'dynamic',
@@ -77,7 +78,7 @@ export default createPages(async ({ createPage, createLayout }) => {
         <h3>Dynamic: {id}</h3>
       </>
     ),
-  });
+  }),
 
   createPage({
     render: 'dynamic',
@@ -85,12 +86,20 @@ export default createPages(async ({ createPage, createLayout }) => {
     component: ({ all }: { all: string[] }) => (
       <h2>Catch-all: {all.join('/')}</h2>
     ),
-  });
+  }),
 
   // Custom Not Found page
   createPage({
     render: 'static',
     path: '/404',
     component: () => <h2>Not Found</h2>,
-  });
-});
+  }),
+]);
+
+declare module 'waku/router' {
+  interface RouteConfig {
+    paths: PathsForPages<typeof pages>;
+  }
+}
+
+export default pages;
--- a/packages/waku/package.json
+++ b/packages/waku/package.json
@@ -42,6 +42,10 @@
       "types": "./dist/server.d.ts",
       "default": "./dist/server.js"
     },
+    "./router": {
+      "types": "./dist/router/base-types.d.ts",
+      "default": "./dist/router/base-types.js"
+    },
     "./router/client": {
       "types": "./dist/router/client.d.ts",
       "default": "./dist/router/client.js"
--- a/packages/waku/src/router/base-types.ts
+++ b/packages/waku/src/router/base-types.ts
@@ -1,0 +1,6 @@
+export type { PathsForPages } from './create-pages-utils/inferred-path-types.js';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface RouteConfig {
+  // routes to be overridden by users
+}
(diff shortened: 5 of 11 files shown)


packages/waku/src/server.ts
```diff
@@ -20,47 +20,49 @@ export type BuildConfig = {
   customData?: unknown; // should be serializable with JSON.stringify
 }[];
 
-export type RenderEntries = (
+export type RenderEntries<Path extends string> = (
   input: string,
   options: {
     params: unknown | undefined;
-    buildConfig: BuildConfig | undefined;
+    buildConfig: BuildConfig<Path> | undefined;
```
dai-shi, line 27: Ahhh, I'm planning to remove buildConfig from here.
Can you try anything (I mean type assertion) without modifying `server.ts` and `define-router.ts` for now?

tylerlaws0n: Since you removed the buildConfig, are these safe to modify now?

dai-shi: I still want to see them unchanged. If some capabilities are missing, can we discuss in advance?

tylerlaws0n: Ok, so I was able to greatly simplify the code that I have so far and added a temporary `test.tsx` that is simply to demo the current behavior of the type inferences that are set up thus far.

So I pulled the callback that is meant to be passed into `createPages` to show that `makeEntries` properly infers the types of each string literal path:

image

This is great, but not quite where I'd like to have the inference yet. This is where I'd like to carry the inference through the defined Router so that the final paths can be inferred through `typeof router` at the end. You can see currently the type of the router is not generic:

image

tylerlaws0n:
@dai-shi & @sandren I just want to mention that while the enforcement of returning the pages here would mean a slight breaking change to `createPages`, it would unlock theoretically typing anything that is statically defined in each call to `createPage`

Right now, I am just inferring `path`, but I intend to follow this up with inferring the full Page object to get things like the `staticPaths` and `render`. If we ever wanted to look into things like typesafe query params, this would be an interesting place to add those features in without any further changes to the external api because we have setup the api for returning the static pages which makes them available for inference.

packages/waku/src/server.ts
tylerlaws0n: I realized something... I don't need to pass the type through `unstable_defineRouter`... I just need to return anything that has the inferred generic. So, I added `ready` to the return of `createPages`

As you can see, now the type of any path is inferred through the response of the router:

image
[file for reference ](https://github.com/dai-shi/waku/pull/854/files#diff-c32c159b88bdd4dc24789fc64dd63421910b3917eec39eb7aa710bc8433d49c7)

I'll plan to extend this to infer the rest of the page type to collect things like the `staticPaths` and render mode in order to generate a final string literal that represents all possible paths for the router to route to.

It's worth acknowledgind that this approach does have the downside of adding to the response body of `createPages` just to get the type of the pages defined, but this seems like a good tradeoff while we do not want to touch `unstable_defineRouter`.

dai-shi: > I don't need to pass the type through unstable_defineRouter

That's what I meant by "I mean type assertion".

tylerlaws0n: Ah! I get it now... I am asserting that type now as `DO_NOT_USE_ready` - I would happily take naming suggestions on this, I just wanted to get the point across that this type is not for actual use in js.

I have the full page types inferred now 🥳 

image

Next steps:
- [ ] Create a union type of all possible paths given the pages
- [ ] Pass this union to an override-able interface so waku projects can consume the pages type safety (same as tanstack router's approach)
- [ ] Use that type in our `<Link />` component
- [ ] Update tests and refactor examples to match new api

tylerlaws0n:
@dai-shi How would you feel about turning [noErrorTruncation](https://www.typescriptlang.org/tsconfig/#noErrorTruncation) on in one of the tsconfigs?

I am currently setting this to true locally while developing then leaving it out of my commits, but here is why it is nice for me:

This is without setting noErrorTruncation:

image

With adding `"noErrorTruncation": true,`:

image

tylerlaws0n:
Another update here: 

image

All possible paths are being inferred from looking at path, render, and staticPaths on each page response 🥳

tylerlaws0n:
> Another update here:
> 
> image
> 
> All possible paths are being inferred from looking at path, render, and staticPaths on each page response 🥳

Just realized that the above screenshot does not handle the catch-all/wildcard static slugs correctly yet, so I fixed that here:

image

tylerlaws0n:
It's working in the router example now... I plan to revert that example to install from a published node version, but am leaving it in there in case @sandren or @dai-shi want to play around with this in that working example package :)

image

examples/07_router/package.json
```diff
@@ -12,7 +12,7 @@
     "react": "19.0.0-rc-7771d3a7-20240827",
     "react-dom": "19.0.0-rc-7771d3a7-20240827",
     "react-server-dom-webpack": "19.0.0-rc-7771d3a7-20240827",
-    "waku": "0.21.1"
+    "waku": "workspace:*"
```
dai-shi, line 15: If it's the same version, it will be the workspace version (unless pnpm changes the behavior).

As this is used as a template the workspace protocol should be avoided.
```suggestion
    "waku": "0.21.1"
```

dai-shi:
> How would you feel about turning [noErrorTruncation](https://www.typescriptlang.org/tsconfig/#noErrorTruncation) on in one of the tsconfigs?

Sounds good. Any drawbacks?

dai-shi:
> It's working in the router example now... I plan to revert that example to install from a published node version

Did it not work without the change?

packages/waku/src/router/client.ts
```diff
@@ -31,13 +31,37 @@ import {
   LOCATION_ID,
 } from './common.js';
 import type { RouteProps, ShouldSkip } from './common.js';
+import type { createPages } from './create-pages.js';
+import type { CollectPaths } from './inferred-router-types.js';
```
dai-shi, line 35: I would like to make this file depends on create-pages.
This file is paired with `define-router.ts` which has no notions about "pages".

packages/waku/src/router/client.ts
```diff
@@ -31,13 +31,37 @@ import {
...
+type RegisteredRouter = Register extends {
+  router: infer UserRouter extends AnyRouter;
+}
+  ? UserRouter
+  : AnyRouter;
+
+export type InferredPaths =
+  CollectPaths<RouterSpec<RegisteredRouter>> extends never
+    ? string
+    : CollectPaths<RouterSpec<RegisteredRouter>>;
```
dai-shi, line 63: Can we just allow overriding the final result?

dai-shi (commented):
Does it make sense? Let me know if my requirement is unclear.

packages/waku/src/router/client.ts
tylerlaws0n: I'd imagine that the most common use case would be to use the type inferred from the router.

That being said, I can add support to loosen the final result. I can think of two ways a user might want to do that:

1. Use inferred router page paths for autocomplete, but allow any string `type Path = '/' | (string & {})`
2. Full override the final result entirely to define whatever the user wants as their paths distinct from what goes through the Router `type Path = '/' | '/about'`

packages/waku/src/router/client.ts
tylerlaws0n: I understand...

I am working in this file because I need to override the type of `to` for the Link component.

Would you like to me to add an export for "./router" and surface the types [there](https://github.com/dai-shi/waku/blob/main/packages/waku/package.json#L18-L53)? 

This client.ts file would then import the `InferredPath` (or whatever it gets renamed to) from that router file.

tylerlaws0n:
> > How would you feel about turning [noErrorTruncation](https://www.typescriptlang.org/tsconfig/#noErrorTruncation) on in one of the tsconfigs?
> 
> Sounds good. Any drawbacks?

None that I have noticed. I've used this setting on my work projects for a few years now and have liked the visibility into the types that it gives.

tylerlaws0n:
> > It's working in the router example now... I plan to revert that example to install from a published node version
> 
> Did it not work without the change?

It won't work without change to get type inference. The pages need to be returned to get that working.

I will loosen the return type of the callback that is passed into `createPages` to have it so that we just fallback to `string` for the paths if they are not yet returning the created pages.

packages/waku/src/router/client.ts
dai-shi: > I need to override the type of `to` for the Link component.

That's fine. A change in this file is acceptable. Just make it agnostic to pages.

packages/waku/src/router/client.ts
dai-shi: I'm not trying to say to loosen the result.

dai-shi:
> > > How would you feel about turning [noErrorTruncation](https://www.typescriptlang.org/tsconfig/#noErrorTruncation) on in one of the tsconfigs?
> > 
> > 
> > Sounds good. Any drawbacks?
> 
> None that I have noticed. I've used this setting on my work projects for a few years now and have liked the visibility into the types that it gives.

Let's open a separate PR for clarification. Please send it.

dai-shi:
> It won't work without change to get type inference. The pages need to be returned to get that working.

I guess you misunderstood my comment. I just mean this change: https://github.com/dai-shi/waku/pull/854#discussion_r1749025630

examples/07_router/package.json
tylerlaws0n: oh, that's cool! I didn't know that.

Just changed it back and you were right, it works just as expected

packages/waku/src/router/client.ts
dai-shi: I'm thinking something like:
```ts
export interface RoutePaths {
  // to be overridden by users
}
```
Or its holder.

packages/waku/src/router/client.ts
tylerlaws0n: What are you hoping to do with overriding the final result then?

packages/waku/src/router/client.ts
tylerlaws0n: Oh I get it! So then would you want to export the generic to pull them from the router and give an example like this?
```ts
import type { PathsForRouter } from 'waku/router';

  interface RouteConfig {
    routes: PathsForRouter<typeof myRouter> | '/some-other-route'; // could also be whatever else they want
  }
}
```

dai-shi: Exactly.

packages/waku/src/router/client.ts
tylerlaws0n: Resolving this... going to change to: https://github.com/dai-shi/waku/pull/854#discussion_r1749041657

packages/waku/src/router/client.ts
tylerlaws0n: Ok, updated! Let me know what you think.

I'll update the docs to reflect the changes in the meantime 👍

examples/07_router/src/entries.tsx
```diff
@@ -10,87 +10,97 @@ const BarPage = lazy(() => import('./components/BarPage'));
...
-    path: '/',
-    component: HomePage,
-  });
+  return [
+    createPage({
+      render: 'static',
+      // render: 'dynamic',
+      path: '/',
+      component: HomePage,
+    }),
```
dai-shi, line 26: Out of curiosity, can we return createLayout too?
```js
const router = createPages(async ({ createPage, createLayout }) => [
  createLayout(...),
  createPages(...),
];
```
Not a strong opinion, but just feels a little bit consistent.

packages/waku/src/router/create-pages.ts
```diff
@@ -180,11 +178,11 @@ export type CreateLayout = <T extends string>(layout: {
   >;
 }) => void;
 
-export function createPages(
+export function createPages<AllPages extends AnyPage[]>(
   fn: (fns: {
     createPage: CreatePage;
     createLayout: CreateLayout;
-  }) => Promise<void>,
+  }) => Promise<AllPages | void>,
```
dai-shi, line 185: Is `void` for backward compatibility? I wonder if it breaks so much.

packages/waku/src/router/index.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+export type PathsForRouter<Router extends AnyRouter> =
+  CollectPaths<RouterSpec<Router>> extends never
+    ? string
+    : CollectPaths<RouterSpec<Router>> & {};
+
+export type InferredPaths = RouteConfig extends { paths: infer UserPaths }
+  ? UserPaths
+  : string;
```
dai-shi, line 22: I don't prefer `index.ts` here. Oh, it's required for `declare module 'waku/router' {` ?
If that's the case, can we only have L10-13 in this file?

Or, I'm fine with `declare module 'waku/router/client' {`. `'waku/router'` seems nicer.

dai-shi (commented):
some comments:

packages/waku/src/router/create-pages.ts
tylerlaws0n: Yes it is explicitly for backwards compatibility. 

If we made a breaking change for this that would simplify our code a bit and encourage the move to type-safe paths. The breaking change is also a very easy fix for users.

I'm happy with whichever you'd prefer though.

packages/waku/src/router/index.ts
tylerlaws0n: I can name the file anything you'd like. `declare module` just looks at the package.json exports and we set the filename there.

examples/07_router/src/entries.tsx
tylerlaws0n: We definitely can! We wouldn't do anything with them so there is no need to, but I also like that example you shared. It has nice visual consistency. 

Would you like me to change to that?

packages/waku/src/router/index.ts
tylerlaws0n: would `src/router/router.ts` be ok?

The module is still `waku/router` externally.

examples/07_router/src/entries.tsx
dai-shi: Yeah, it seems cleaner.

packages/waku/src/router/create-pages.ts
dai-shi: I wouldn't expect many people use `createPages` even for experiments.
This is type only breaking change, so it's reasonable. Let's simplify the code.

packages/waku/src/router/index.ts
dai-shi: Anything other than `index.ts` is okay, but `router/router.ts` isn't nice.

tylerlaws0n: thoughts on `src/router/base.ts` or `src/router/export.ts`?

Or `src/router.ts`? it would seem weird to have a folder named `src/router` and also `src/router.ts` though.

dai-shi: among them, src/router/base.ts sounds good. but then the name accepts more "base" logic in the file.

If this is all about typing, how about src/router/baseTypes.ts?

packages/waku/src/router/create-pages.ts
tylerlaws0n: Removed this ✅

examples/07_router/src/entries.tsx
tylerlaws0n: Updated!

packages/waku/src/router/index.ts
tylerlaws0n: changed to base-types.ts ✅

packages/waku/package.json
```diff
@@ -42,6 +42,10 @@
       "types": "./dist/server.d.ts",
       "default": "./dist/server.js"
     },
+    "./router": {
+      "types": "./dist/router/base-types.d.ts",
+      "default": "./dist/router/base-types.js"
+    },
```
dai-shi, line 48: Ah, now I see it.

packages/waku/src/router/base-types.ts
```diff
@@ -0,0 +1,22 @@
+import type { createPages } from './create-pages.js';
```
dai-shi, line 1: I'm not comfortable having this dependency in this "base" file.

packages/waku/src/router/base-types.ts
```diff
@@ -0,0 +1,22 @@
...
+type AnyRouter = ReturnType<typeof createPages>;
+
+type RouterSpec<Router extends AnyRouter> = Awaited<
+  NonNullable<Router['DO_NOT_USE_ready']>
+>[number];
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface RouteConfig {
+  // routes to be overridden by users
+}
```
dai-shi, line 13: This is fine in this file. Maybe only this.

packages/waku/src/router/base-types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type PathsForRouter<Router extends AnyRouter> =
+  CollectPaths<RouterSpec<Router>> extends never
+    ? string
+    : CollectPaths<RouterSpec<Router>> & {};
```
dai-shi, line 18: This should probably move to `create-pages.ts`

packages/waku/src/router/inferred-router-types.ts
```diff
@@ -0,0 +1,91 @@
...
+export type CollectPaths<EachPage extends AnyPage | void> = // void is returned by createLayout
+  EachPage extends AnyPage
+    ? IsPageWithSlug<EachPage> extends true
+      ? EachPage extends StaticSlugPage
+        ? CollectPathsForStaticSlugPage<EachPage>
+        : EachPage extends DynamicPage
+          ? CollectPathsForDynamicSlugPage<EachPage>
+          : never
+      : EachPage['path']
+    : never;
```
dai-shi, line 91: Hmm, in the source code, I would like to make it clear that:
- "router" is something we define in `unstable_defineRouter`.
- "createPages" is **one instance** of "router".
- Assume there can be another instance of router, like `createFoo` independent from `createPages`.

So, any code specific to `create-pages.ts` should probably move to `src/router/create-pages-utils/*.ts`

dai-shi (commented):
some more comments 😄

packages/waku/src/router/base-types.ts
tylerlaws0n: this generic needs to be exported from one of our modules for users to be able to do

```ts
paths: PathsForRouter<typeof myRouter>
```

I could move this to `createPages`, but then would need to re-export through another file?

dai-shi: yes, it should be exported through `create-pages.ts`.
I was thinking exporting `waku/router/server`, but hmm, yeah `waku/router` should be fine.
Just re-export directly like `export type { ... } from './create-pages.js'`;.

Let me re-consider about `base-types.ts` naming afterwards.

packages/waku/src/router/base-types.ts
tylerlaws0n: removed this dependency with moving RouterSpec and AnyPage to `inferred-router-types.ts`and moved `InferredPaths` to a new file: `types-from-route-config.ts`

I'll move these to `src/router/create-pages-utils/*.ts` per your other thread next, but this change should address the concern of the dep on create-pages

packages/waku/src/router/inferred-router-types.ts
dai-shi: actually `src/router/create-pages/*.ts` is fine, if you prefer. My request is to avoid `index.ts`.

tylerlaws0n: Ok, how's this look?

```
packages/waku/src/router
├── base-types.ts
├── client.ts
├── common.ts
├── create-pages
│   ├── common.ts
│   ├── inferred-router-types.ts
│   └── types-from-route-config.ts
├── define-router.ts
├── fs-router.ts
├── server.ts
└── util-types.ts

2 directories, 10 files
```

packages/waku/src/router/base-types.ts
tylerlaws0n: Moved this to `create-pages/inferred-router-types.ts` and re-exported from `base-types.ts` 👍

packages/waku/src/router/base-types.ts
tylerlaws0n: This and the re-export of PathsForRouter are now the only things in this file

packages/waku/src/router/inferred-router-types.ts
tylerlaws0n: > * "router" is something we define in `unstable_defineRouter`.

For this, would it be good enough to add jsdoc comments to the types in the inferred-router-types.ts to explain where each is called from and what they're used for?

> * "createPages" is **one instance** of "router".
> * Assume there can be another instance of router, like `createFoo` independent from `createPages`.

I was thinking that when we support other router types that these types would still be used to infer the final paths for consistency on our end. For example `AnyRouter` would be widened to support other "router"'s and then from that point on, all the types for path inference would be the same.

packages/waku/src/router/client.ts
```diff
@@ -31,6 +31,7 @@ import {
   LOCATION_ID,
 } from './common.js';
 import type { RouteProps, ShouldSkip } from './common.js';
+import type { InferredPaths } from './create-pages/types-from-route-config.js';
```
dai-shi, line 34: This shouldn't come from `create-pages`. Can you move the definition `InferredPaths` here?

packages/waku/src/router/server.ts
```diff
@@ -1,3 +1,3 @@
 export { unstable_defineRouter, unstable_redirect } from './define-router.js';
-export { createPages } from './create-pages.js';
+export { createPages } from './create-pages/common.js';
```
dai-shi, line 2: I'm sorry for the implicit naming, but "common.ts" has special usage in Waku. We use it when the file is used both from the server and the client.

docs/create-pages.mdx
```diff
@@ -19,23 +19,65 @@ import { createPages } from 'waku';
 import { RootLayout } from './templates/root-layout';
 import { HomePage } from './templates/home-page';
 
-export default createPages(async ({ createPage, createLayout }) => {
+const router = createPages(async ({ createPage, createLayout }) => [
```
dai-shi, line 22: I noticed the confusion starts here. We probably shouldn't call the result of `createPages` just a "router". We should call it pagesRouter or just pages.

examples/07_router/src/entries.tsx
```diff
@@ -1,5 +1,6 @@
 import { lazy } from 'react';
 import { createPages } from 'waku';
+import type { PathsForRouter } from 'waku/router';
```
dai-shi, line 3: Then, this becomes `PathsForPages` or `PathsForPagesRouter`. In the future there could be, `PathsForFooRouter`.

e2e/fixtures/render-type/src/entries.tsx
```diff
@@ -3,27 +3,33 @@ import { createPages } from 'waku/router/server';
...
-  });
-  createPage({
-    render: 'static',
-    path: '/client/static/[echo]',
-    staticPaths: ['static-echo'],
-    component: ClientEcho,
-  });
-});
+// This needs type annotations for the return type of createPages
+// @see https://github.com/microsoft/TypeScript/issues/42873#issuecomment-2065572017
```
dai-shi, line 7: I was curious about this explicit annotation in other repo.
Does the linked issue commend suggest the explicit annotation?

dai-shi (commented):
hope it makes sense.

packages/waku/src/router/server.ts
dai-shi: Re: https://github.com/dai-shi/waku/pull/854#discussion_r1749304119

So, we can't use `common.ts`. Unless you have a better suggestion, we should go back with `create-pages.ts` and `create-pages-utils/*.ts`.

packages/waku/src/router/inferred-router-types.ts
dai-shi: yeah, jsdoc would be helpful.

We don't know the future and how `createFoo` looks like, if it shares some thing, we can restructure. For now, I would like to separate unstable_defineRouter and createPages very clearly. For example, slugs are only for createPages.

docs/create-pages.mdx
tylerlaws0n: i like just `pages` 👍

e2e/fixtures/render-type/src/entries.tsx
tylerlaws0n: Yes... it is apparently a known issue with type inference inside pnpm projects with symlinked packages.

After visiting the discord for [arktype](https://github.com/arktypeio/arktype) I was linked to that issue on the Typescript repo that I left in the comments. Here is the summary that I ended up putting on the arktype discord after a chat with their author:

image

[discord link if you'd like to read the full thread](https://discord.com/channels/957797212103016458/1282402037731950653/1282402037731950653)

[arktype discord if you need a join link](https://arktype.io/discord)

tylerlaws0n:
> hope it makes sense.

Yep! all makes sense and were simple changes I think. 215e6e2

packages/waku/src/router/inferred-router-types.ts
tylerlaws0n: added jsdoc to inferred-path-types ✅

tylerlaws0n:
@dai-shi some idea names for `base-types.ts` now that it only has the empty interface to be overridden externally:

- `external.ts` or `external-types.ts`
- `router-types.ts`
- `public-types.ts`

Also, are there other changes you'd like to see with this PR?

dai-shi:
I'll do a full review later in this week.

packages/waku/src/router/client.ts
```diff
@@ -31,6 +31,11 @@ import {
   LOCATION_ID,
 } from './common.js';
 import type { RouteProps, ShouldSkip } from './common.js';
+import type { RouteConfig } from './base-types.js';
+
+export type InferredPaths = RouteConfig extends { paths: infer UserPaths }
```
dai-shi, line 36: No need to export?
```suggestion
type InferredPaths = RouteConfig extends { paths: infer UserPaths }
```

dai-shi (approved):
Otherwise, LGTM. Note that I didn't read the code in detail.

packages/waku/src/router/client.ts
tylerlaws0n: Good catch! I left this lingering by mistake after moving which file it was in

tylerlaws0n:
> Otherwise, LGTM. Note that I didn't read the code in detail.

Sounds good 👍 - squashed and rebased. Should be good to merge 😄

PR #857, opened 2024-08-29
Instructions on how to enable the React Compiler in a Waku pro…

L-Qun:
Background: https://github.com/dai-shi/waku/pull/853


docs/guides/react-compiler.mdx
```diff
@@ -0,0 +1,31 @@
...
+          ],
+        },
+      }),
+    ],
+    // ...
+  };
+});
+```
+
+For a practical demonstration, visit this [StackBlitz demo](https://stackblitz.com/edit/github-ukt9tm?file=vite.config.ts) which provides a real-world example of the configuration in action.
```
dai-shi, line 31: I've updated the demo.

```suggestion
For a practical demonstration, visit this [StackBlitz demo](https://stackblitz.com/edit/github-h4uxwo?file=vite.config.ts) which provides a real-world example of the configuration in action.
```

As I tried, removing "annotation" will cause errors. Can you please try it? I guess we should add it in the instruction above. What do you think? Do you think we can recommend it even with "annotation"?

docs/guides/react-compiler.mdx
```diff
@@ -0,0 +1,31 @@
+---
+slug: guides/react-compiler
+title: Enabling React Compiler in Waku
+description: Learn how to enable the React Compiler in your Waku project using the Vite Plugin to optimize your React applications.
+---
+
+The React Compiler is a build-time tool designed to optimize React applications automatically. This guide will show you how to enable the React Compiler by integrating it with the `@vitejs/plugin-react`.
```
dai-shi, line 7: I don't think it's build-time only.

L-Qun: image

I copied it from React offical website

docs/guides/react-compiler.mdx
L-Qun: image

I believe we shouldn't add this restriction because if we add an "annotation", users will need add the "use memo" directive manually

I found the error `Error: __vite_ssr_import_1__.c is not a function`, actually I encountered this error when I implement react compiler feature, I can try to fix it issue

dai-shi: > I can try to fix it issue

That will be great.

docs/guides/react-compiler.mdx
dai-shi: Interesting. I wonder if raw Vite app + React Compiler works in DEV mode. I don't think this is Waku specific.

docs/guides/react-compiler.mdx
L-Qun: image

This issue occurs when I pass react compiler config in. So in my [implementation](https://github.com/dai-shi/waku/pull/853) I only pass react compiler config in `viteReact` called by `createMainViteServer`

dai-shi: Do we not need react compiler for server files?

My plan is to allow customizing vite config for each server in v0.22.0. Give me more time.

docs/guides/react-compiler.mdx
himself65: dev mode is also a kind of build-time. built time in this context means it won't haveruntime cost, which will have no exta bundle size & cpu runs when you open browser

dai-shi: I see.

dai-shi:
let's revisit this after v0.22 is finished.

docs/guides/react-compiler.mdx
dai-shi: Actually, I noticed the change isn't breaking much, so I decided to work on it during v0.21.*. 

https://github.com/dai-shi/waku/issues/24#issuecomment-2564364866  @rmarscher

docs/guides/react-compiler.mdx
```diff
@@ -0,0 +1,31 @@
...
+          ],
+        },
+      }),
+    ],
+    // ...
+  };
+});
+```
+
+For a practical demonstration, visit this [StackBlitz demo](https://stackblitz.com/edit/github-h4uxwo?file=vite.config.ts) which provides a real-world example of the configuration in action.
```
dai-shi, line 31:
```suggestion
Check out [a working example](../../examples/05_compiler), or visit the [StackBlitz demo](https://stackblitz.com/github/dai-shi/waku/tree/main/examples/05_compiler) with Waku v0.21.12 and up.
```

dai-shi (commented):
Thanks to #1102.

dai-shi (approved):

PR #859, opened 2024-08-30
fix(client): do not wrap with `startTransition`

dai-shi:
Let users control transition.

### DIFF
--- a/examples/24_nesting/src/components/Counter.tsx
+++ b/examples/24_nesting/src/components/Counter.tsx
@@ -9,6 +9,15 @@ export const Counter = ({ enableInnerApp }: { enableInnerApp?: boolean }) => {
   const refetch = useRefetch();
   const handleClick = () => {
     if (enableInnerApp) {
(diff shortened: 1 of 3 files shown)


examples/24_nesting/src/components/Counter.tsx
```diff
@@ -21,9 +30,12 @@ export const Counter = ({ enableInnerApp }: { enableInnerApp?: boolean }) => {
...
     <div style={{ border: '3px blue dashed', margin: '1em', padding: '1em' }}>
       <p>Count: {count}</p>
-      <button onClick={handleClick} disabled={isPending}>
-        Increment
-      </button>{' '}
+      <button onClick={handleClick}>Increment</button>
+      {enableInnerApp && (
+        <button onClick={handleClickWithTransition} disabled={isPending}>
+          Increment with transition
+        </button>
```
dai-shi, line 37: These two buttons show different behaviors.

Issue #827, opened 2024-08-11
Feature Request: An API for accessing platform specific data/information/APIs (e.g. the Cloudflare ones)


dario-piotrowicz:
Hello :wave:

I'm opening this issue to ask if there is any intention or plan to support platform specific APIs (and I am specifically thinking of the Cloudflare APIs).

I think it might be worth clarifying the above: There is a [Cloudflare deployment option](https://waku.gg/#cloudflare-experimental) to allow Waku applications to be deployed to Cloudflare Pages, this is already very nice, but one of the main reasons/incentives for developers to deploy to Cloudflare Workers/Pages is so that their application can make use of the various resources that the Cloudflare network provides (mainly [bindings](https://developers.cloudflare.com/pages/functions/bindings/) and [`waitUntil`](https://developers.cloudflare.com/workers/runtime-apis/context/#waituntil)). So ideally Waku should provide an API to access such platform specific APIs (I would also imagine that other hosting platforms would have similar APIs but I haven't really looked into that to be honest).

I could not see this mentioned in the [roadmap](https://github.com/dai-shi/waku/issues/24) so I figured I could ask :slightly_smiling_face: 

Additionally I think @rmarscher already started experimenting with this in [PR #795](https://github.com/dai-shi/waku/pull/795) + [gist](https://gist.github.com/rmarscher/9bb6ed54dc9535f4b81bed147204c7e9). But from what I can tell:
 - his dev plugin hasn't been included in his PR, is this something that should not built-in into Waku itself?
 - it seems like he's taking the `executionContext` and simply making it available through `getEnv` ([source](https://gist.github.com/rmarscher/9bb6ed54dc9535f4b81bed147204c7e9#file-waku-cloudflare-dev-server-ts-L33)), is that really the cleanest/best API? (for one, if would override any pre-existing env variable called `executionContext`)
 - either way, he didn't provide a production build solution (e.g. he's exposing `executionContext` only in dev mode)

(@rmarscher please do correct me if I got any of the above wrong :pray:)

So I would imagine that some extra work/thoughts needs to be put into it :thinking: 

Personally I think a new `getPlatform` utility (equivalent to the existing `getEnv`) would make sense, it could expose the platform specific APIs plus any other useful data/information regarding the current platform.

___

<details>

<summary>Other meta-framework examples</summary>

Most meta-frameworks do provide such APIs.

As React meta-frameworks go:

[...]

PR #860, opened 2024-08-30
feat(build): introduce buildData in platformObject

dai-shi:
It's still unstable.

### DIFF
--- a/packages/waku/src/server.ts
+++ b/packages/waku/src/server.ts
@@ -17,14 +17,12 @@ export type BuildConfig = {
   }[];
   context?: Record<string, unknown>;
   customCode?: string; // optional code to inject TODO hope to remove this
-  customData?: unknown; // should be serializable with JSON.stringify
 }[];
 
 export type RenderEntries = (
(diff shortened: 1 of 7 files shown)


packages/waku/src/server.ts
```diff
@@ -149,3 +146,26 @@ export function unstable_getHeaders(): Record<string, string> {
...
+      | 'vercel-serverless'
+      | 'netlify-static'
+      | 'netlify-functions'
+      | 'cloudflare'
+      | 'partykit'
+      | 'deno'
+      | 'aws-lambda'
+      | undefined;
+  };
+} & Record<string, unknown>;
```
tylerlaws0n, line 164: What is the purpose of extending this with `Record<string, unknown>` when `buildData` is also present?

I guess I am just curious what this extra flexibility on the type might be used for. Currently it just looks like `buildData` is all that's being used.

dai-shi: It's for #827 kind of extensions. It's not just for build.

PR #870, opened 2024-09-10
fix: cloudflare routing 404

### LINKED ISSUE #869: Cloudflare pages deploy results in 404 status when index page is "static"
Repro:

```bash
npm create waku@0.9.3
cd waku-project
npm install
npm run build -- --with-cloudflare
npx wrangler pages dev
```

image

It's the same when it's deployed with `npx wrangler pages deploy`.

@jkhaui @rmarscher

rmarscher:
This fixes cloudflare static vs dynamic routing for the home index page. But there are likely more bugs with the _routes.json when trying to nest a dynamic route under a static route path. By using _routes.json, we can route static html files directly off Cloudflare's CDN and not incur any worker function changes. But we need to make sure it only matches the static html routes.

Right now, it is doing a very basic walk of the public output dir to generate a _routes.json. We need to generate a more accurate list of exclude patterns. I am opening this PR as a draft until I can complete this. Fixes #869

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -41,14 +41,30 @@ const srcServeFile = decodeFilePathFromAbsolute(
   ),
 );
 
+const getFiles = (dir: string, files: string[] = []): string[] => {
+  const entries = readdirSync(dir, { withFileTypes: true });
+  for (const entry of entries) {
+    const fullPath = path.join(dir, entry.name);
+    if (entry.isDirectory()) {
+      getFiles(fullPath, files);
+    } else {
+      files.push(fullPath);
+    }
+  }
+  return files;
+};
+
 const WORKER_JS_NAME = '_worker.js';
 const ROUTES_JSON_NAME = '_routes.json';
+const HEADERS_NAME = '_headers';
 
 type StaticRoutes = { version: number; include: string[]; exclude: string[] };
 
 export function deployCloudflarePlugin(opts: {
   srcDir: string;
   distDir: string;
+  rscPath: string;
+  privateDir: string;
 }): Plugin {
   const platformObject = unstable_getPlatformObject();
   let rootDir: string;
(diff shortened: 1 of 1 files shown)


dai-shi:
Let me merge #866 first.

rmarscher:
Wow. #866 is an exciting PR. I will rebase and see if I can finish the rest of the cloudflare routing updates needed here.

rmarscher:
I updated the _routes.json to always start with "/assets/*" in the exclude and then walk the public directory and add the rest of the files, replacing `/index.html` with `/` in the exclude.

I noticed a couple new issues while working on this.

1) The static html routes are put in index.html files and a redirect from the non-trailing slash to the trailing slash. When I changed a route from static to dynamic, my redirect was still cached but I now received an error from https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/renderers/utils.ts#L14 about input should not end in `/`. I cleared the redirect cache and I was able to access it without the trailing slash. I think we might need to open another discussion or issue around trailing slash configuration. Most meta frameworks have a config to allow either. For cloudflare, we could move `/path/index.html` files to `/path.html` which cloudflare will route statically as `/path`.

2) Dynamic routes on cloudflare are failing. I am working to determine which recent update caused it. It is a new issue since 0.21.1. The error is with trying to import `node:fs/promises` in fsRouter.`node:fs` is not available in cloudflare even with nodejs_compat_v2. https://developers.cloudflare.com/workers/runtime-apis/nodejs/#nodejs-api-polyfills

tylerlaws0n:
> For cloudflare, we could move /path/index.html files to /path.html which cloudflare will route statically as /path

Could you please link to the docs for the cloudflare path support?

rmarscher:
> > For cloudflare, we could move /path/index.html files to /path.html which cloudflare will route statically as /path
> 
> Could you please link to the docs for the cloudflare path support?

No problem! https://developers.cloudflare.com/pages/configuration/serving-pages/

rmarscher:
Here are other docs to reference.

We are using "advanced mode" so everything routes through a single `/worker.js/index.js` file right now (the root dir is actually named `worker.js` which is a trick other projects use to encapsulate many files into the advanced mode function). 
https://developers.cloudflare.com/pages/functions/advanced-mode/

There is also additional relevant documentation here. Once advanced mode is enabled, a _routes.json file is required to split dynamic routes that are handler by the worker function and static routes that should be served off the CDN. https://developers.cloudflare.com/pages/functions/routing/

dai-shi (commented):
I didn't test it, but looks fine to me. Can I merge?

dai-shi:
Thanks for confirming.

rmarscher:
Thank you!

PR #872, opened 2024-09-11
feat(deploy): add cloudflare context to platform object

rmarscher:
Example usage from a server component:

```ts
import { unstable_getPlatformObject } from 'waku/server';

const getData = async () => {
  const { env, executionCtx } = unstable_getPlatformObject() as { env: Env; executionCtx: ExecutionContext };
  executionCtx?.waitUntil(
    new Promise<void>((resolve) => {
      console.log("Waiting for 5 seconds")
      setTimeout(() => {
        console.log("OK, done waiting")
        resolve()
      }, 5000)
    }),
  )
  const { results } = await env.DB.prepare(
    "SELECT * FROM users WHERE user_id = ?",
  )
    .bind(userId)
    .all();
  return results;
};
```

Server console output:

```
Waiting for 5 seconds
[wrangler:inf] GET /about 200 OK (30ms)
[wrangler:inf] GET /assets/jsx-runtime-BjG_zV1W.js 304 Not Modified (6ms)
[wrangler:inf] GET /assets/index-CbskofAj.js 304 Not Modified (7ms)
[wrangler:inf] GET /assets/_layout-Shb4QlRw.css 304 Not Modified (9ms)
[wrangler:inf] GET /assets/rsc2-c69df7b3e.js 200 OK (11ms)
[wrangler:inf] GET /assets/client-kczTGGZ_.js 304 Not Modified (16ms)
[wrangler:inf] GET /assets/indexHtml-DCalQi_d.js 304 Not Modified (18ms)
[wrangler:inf] GET /assets/client-CMyJdxTj.js 304 Not Modified (21ms)
[wrangler:inf] GET /assets/rsc0-ba005381c.js 304 Not Modified (4ms)
[wrangler:inf] GET /images/favicon.png 304 Not Modified (3ms)
OK, done waiting
```

The page was already sent to the client and loaded while the promise was still executing.

It is necessary to cast the Cloudflare types. See https://developers.cloudflare.com/workers/languages/typescript/#generate-types-that-match-your-workers-configuration-experimental for info on generating types for your project.

### DIFF
--- a/packages/waku/src/lib/builder/serve-cloudflare.ts
+++ b/packages/waku/src/lib/builder/serve-cloudflare.ts
@@ -1,41 +1,75 @@
 import { Hono } from 'hono';
+import { unstable_getPlatformObject } from 'waku/server';
 import { runner } from '../hono/runner.js';
+import type {
+  ExportedHandler,
+  fetch,
+  Request as CloudflareRequest,
+  Response as CloudflareResponse,
+} from '@cloudflare/workers-types/experimental';
 
 const loadEntries = () => import(import.meta.env.WAKU_ENTRIES_FILE!);
 let serveWaku: ReturnType<typeof runner> | undefined;
 
 export interface CloudflareEnv {
   ASSETS: {
-    fetch: (input: RequestInit | URL, init?: RequestInit) => Promise<Response>;
+    fetch: typeof fetch;
   };
 }
 
 export const app = new Hono<{
   Bindings: CloudflareEnv & { [k: string]: unknown };
 }>();
-app.use('*', (c, next) => serveWaku!(c, next));
+app.use('*', (c, next) => {
+  if (!serveWaku) {
+    throw new Error('serveWaku is not initialized');
+  }
+  const platform = unstable_getPlatformObject();
+  platform.honoContext = c;
+  platform.cf = (c.req.raw as unknown as CloudflareRequest).cf;
+  platform.env = c.env;
+  platform.executionContext = c.executionCtx;
+  return serveWaku(c, next);
+});
 app.notFound(async (c) => {
   const assetsFetcher = c.env.ASSETS;
   const url = new URL(c.req.raw.url);
   const errorHtmlUrl = `${url.origin}/404.html`;
-  const notFoundStaticAssetResponse = await assetsFetcher.fetch(
+  const notFoundStaticAssetResponse = (await assetsFetcher.fetch(
     new URL(errorHtmlUrl),
-  );
+  )) as unknown as Response;
   if (notFoundStaticAssetResponse && notFoundStaticAssetResponse.status < 400) {
     return c.body(notFoundStaticAssetResponse.body, 404);
   }
   return c.text('404 Not Found', 404);
 });
 
-export default {
-  async fetch(
-    request: Request,
-    env: Record<string, string>,
-    ctx: Parameters<typeof app.fetch>[2],
-  ) {
+// Waku getEnv only supports strings
+// Cloudflare injects bindings to env and JSON
+// Use unstable_getPlatformObject() to access cloudflare env and execution context
+// https://developers.cloudflare.com/workers/configuration/environment-variables/#add-environment-variables-via-wrangler
+// https://developers.cloudflare.com/workers/runtime-apis/bindings/
+const extractWakuEnv = (env: Record<string, unknown>): Record<string, string> =>
+  Object.fromEntries(
+    Object.entries(env).filter(([, value]) => typeof value === 'string'),
+  ) as Record<string, string>;
+
+const handler: ExportedHandler<CloudflareEnv & { [k: string]: never }> = {
+  async fetch(request, env, ctx) {
     if (!serveWaku) {
-      serveWaku = runner({ cmd: 'start', loadEntries, env });
+      serveWaku = runner({
+        cmd: 'start',
+        loadEntries,
+        env: extractWakuEnv(env),
+      });
     }
-    return app.fetch(request, env, ctx);
+    return app.fetch(
+      request as unknown as Request,
+      env,
+      ctx,
+    ) as unknown as CloudflareResponse;
   },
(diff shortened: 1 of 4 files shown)


dai-shi:
> Using the getEnv API until #827 is resolved

#860 was my answer. Can you try it? getEnv is for strings.

rmarscher:
I'm getting close. I just need to do testing on this latest implementation. Does https://github.com/dai-shi/waku/pull/872/files#diff-cd6a8e530f81183dc69573add03d2338dfb64976b74eec5eb387e66a713e6522R45-R53 look good? I'm trying to remove non-string vars that Cloudflare puts in env before passing to the runner.

In addition to the [bindings](https://developers.cloudflare.com/workers/runtime-apis/bindings/), I found that cloudflare allows JSON in env vars: https://developers.cloudflare.com/workers/configuration/environment-variables/#add-environment-variables-via-wrangler

rmarscher:
There are some typescript issues due to differences in workerd built-in types like Request/Response but I worked around them with casting.

packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -1,41 +1,75 @@
 import { Hono } from 'hono';
+import { unstable_getPlatformObject } from 'waku/server';
```
dai-shi, line 2: It might not change anything real, but for consistency. (Will be refactored anyway.)
```suggestion
import { unstable_getPlatformObject } from '../../server.js';
```

packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -1,41 +1,75 @@
...
 import { runner } from '../hono/runner.js';
+import type {
+  ExportedHandler,
+  fetch,
```
dai-shi, line 6: This is confusing. So:
```suggestion
  fetch as CloudflareFetch,
```

packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -1,41 +1,75 @@
...
 export const app = new Hono<{
   Bindings: CloudflareEnv & { [k: string]: unknown };
 }>();
-app.use('*', (c, next) => serveWaku!(c, next));
+app.use('*', (c, next) => {
+  if (!serveWaku) {
+    throw new Error('serveWaku is not initialized');
+  }
+  const platform = unstable_getPlatformObject();
+  platform.honoContext = c;
```
dai-shi, line 28: This is not good.
The platform object is something stable all the time. It shouldn't change for each request.
`unstable_getCustomContext` is for each request. And, the hono context is already exposed. #852

packages/waku/src/server.ts
```diff
@@ -167,11 +170,13 @@ type PlatformObject = {
...
-} & Record<string, unknown>;
+} & T;
 
 (globalThis as any).__WAKU_PLATFORM_OBJECT__ ||= {};
 
 // TODO tentative name
-export function unstable_getPlatformObject(): PlatformObject {
+export function unstable_getPlatformObject<
+  T = Record<string, unknown>,
+>(): PlatformObject<T> {
```
dai-shi, line 180: I believe asserting types is more appropriate than annotating types, because it never be type safe.

packages/waku/src/lib/builder/serve-cloudflare.ts
```diff
@@ -1,41 +1,75 @@
...
+      });
     }
-    return app.fetch(request, env, ctx);
+    return app.fetch(
+      request as unknown as Request,
+      env,
+      ctx,
+    ) as unknown as CloudflareResponse;
   },
+  // TODO ability to add other handlers or Durable Objects?
```
dai-shi, line 72: Unless we do this, this PR doesn't provide any more capability than #852.

rmarscher:
@dai-shi Thank you reviewing the PR. You [mentioned the context API](https://github.com/dai-shi/waku/issues/827#issuecomment-2308651317) before and I saw #852 but I hadn't found `unstable_getCustomContext` yet.

Is it correct to access it like this? `unstable_getCustomContext().__hono_context`

```ts
import type { ExecutionContext } from '@cloudflare/workers-types/experimental';
import { unstable_getCustomContext } from 'waku/server';

const getData = async ({ userId }: { userId: string }) => {
  const { env, executionCtx } = unstable_getCustomContext().__hono_context as {
    env: Env
    executionCtx: ExecutionContext
  };
  executionCtx?.waitUntil(
    new Promise<void>((resolve) => {
      console.log("Waiting for 5 seconds");
      setTimeout(() => {
        console.log("OK, done waiting");
        resolve();
      }, 5000);
    }),
  );
  const { results } = await env.DB.prepare("SELECT * FROM users WHERE user_id = ?")
    .bind(userId)
    .all();
  return results;
}
```

I think I will close this PR.

Re:
https://github.com/dai-shi/waku/pull/872#discussion_r1759928284
https://github.com/dai-shi/waku/pull/879#discussion_r1759974584

How could we allow a developer to inject handlers and export Durable Object classes from their worker index file?

The way I have been doing it is to build waku and then copy in a custom index.js file to the dist folder.

We could make that an option to the cloudflare deploy vite plugin - a path to an alternate worker entry file.

Maybe a cloudflare deploy vite plugin option could point to an alternate worker entry file that would be used instead of the default. 

Or could there be another way to implement it using getPlatformObject()?

dai-shi:
> Is it correct to access it like this?

Looks good.
I wasn't aware that the db was in the hono context.
https://developers.cloudflare.com/d1/examples/d1-and-hono/

> How could we allow a developer to inject handlers and export Durable Object classes from their worker index file?

Oh, I think I misunderstood something.
Please note that I have zero experience with Cloudflare.
Would you create a new discussion and describe the goal briefly?
(It sounds somewhat related to the idea of say adding another Vercel function, which is impossible for now.)

dai-shi:
@rmarscher see #884

PR #877, opened 2024-09-14
refactor(build): vercel serve and constants

dai-shi:
following #866


packages/waku/package.json
```diff
@@ -34,6 +34,10 @@
       "types": "./dist/internals.d.ts",
       "default": "./dist/internals.js"
     },
+    "./hono": {
```
tylerlaws0n, line 37: Just curious here, but why do `/hono` and `/internals` need to be in the published package json exports if they are for internal use only?

package.json
```diff
@@ -51,7 +51,7 @@
     "examples:prd:24_nesting": "NAME=24_nesting pnpm run examples:prd",
     "website:dev": "(cd packages/website && pnpm run dev)",
     "website:build": "cd packages/website && pnpm run build",
-    "website:vercel": "pnpm run compile && pnpm run website:build --with-vercel-static && mv packages/website/.vercel/output .vercel/",
+    "website:vercel": "pnpm run compile && pnpm run website:build && mv packages/website/.vercel/output .vercel/",
```
tylerlaws0n: Is this to test the hono server serving the website from vercel?

dai-shi: yes

packages/waku/package.json
dai-shi: I did it at first (locally without commits), but it feels like mixed. `/internals` are "more internal" than `/hono`, so to speak. 😄 
In any case, I might be reconsidering it in the future, before finalizing v1.

dai-shi: Also, I noticed we need to re-export `Hono` to avoid adding in peer dependencies. So, `/internals` are weird for it.

dai-shi: I misunderstood your question probably.
- `/internals` are for someone who might want to experiment something.
- `/hono` is for deploy plugins (actual use case), I wish this could be avoided.

package.json
tylerlaws0n: nice! thanks 👍

packages/waku/package.json
tylerlaws0n: Ah, this makes more sense! Thanks for the explanation

PR #882, opened 2024-09-15
refactor(deploy): aws-lambda serve

dai-shi:
Prior art:
- #877 
- #878


dai-shi:
@aheissenberger Can you test this and see if it works?

rmarscher:
I have been having trouble with the hono import in the built serve-aws-lambda.js file. It looks like it is not building hono/context into the bundle? Here is a section of the built serve-aws-lambda.js file:

```js
const importHono = () => import("./assets/index-BxM3V2xC.js");
const importHonoContextStorage = () => import("hono/context-storage");
const importHonoNodeServerServeStatic = () => import("./assets/serve-static-CLws5J7J.js");
const importHonoAwsLambda = () => import("./assets/index-DM8TMISb.js");
const { Hono } = await importHono();
const { contextStorage } = await importHonoContextStorage();
const { serveStatic } = await importHonoNodeServerServeStatic();
const { handle } = await importHonoAwsLambda();
```

And then I get a runtime error on AWS when it tries to serve a page:
Cannot find package 'hono' imported from /var/task/serve-aws-lambda.js

I'm not sure if this is caused by the way I am trying to build this branch and copy to my AWS project dependencies. I have installed hono 4.6.1 in my AWS project.

rmarscher:
I was able to get it working by commenting out the hono context storage from the serve js content. So that seems to be the only issue. I don't understand how hono/aws-lambda is built correctly but hono/context-storage is not.

rmarscher:
Oh right... probably this... https://github.com/dai-shi/waku/commit/58f7ded82afc5c959c60b269dfb83ec2982fe032#diff-275ea9f25448c4618b0f5d5e886b0f4072dad0167fc254e3804e46be65ec2096R249

rmarscher:
It works for me if I remove `external: ['hono/context-storage']` from the vite config for buildServerBundle and use `import { importHonoContextStorage } from "waku/unstable_hono";` in a server component.

```
import { importHonoContextStorage } from "waku/unstable_hono";

const getData = async () => {
  const { getContext } = await importHonoContextStorage();
  const c = getContext();
  const acceptHeader = c.req.header("Accept");
  return { acceptHeader };
}
```

I had to leave `hono/context-storage` as external in the dev-server-impl.ts though. I tried updating the runner in cli.ts to import from unstable_hono:

```ts
  const contextStorage = (await importHonoContextStorage()).contextStorage;
  app.use(contextStorage());
```

But that results in `Context is not available` error. Leaving cli.ts as is works in dev mode while still using importHonoContextStorage in the server component.

I don't personally have a use case for `waku start` but I have a feeling that might have issues because the hono context in the runner will be different from the one in the build. Maybe the cli could be changed to invoke runners that are part of the server build instead of having a separate runner that imports entries directly.

rmarscher:
I made a related documentation change in https://github.com/dai-shi/waku/pull/897

dai-shi:
@rmarscher Thanks for investigating this.

> Oh right... probably this... [58f7ded#diff-275ea9f25448c4618b0f5d5e886b0f4072dad0167fc254e3804e46be65ec2096R249](https://github.com/dai-shi/waku/commit/58f7ded82afc5c959c60b269dfb83ec2982fe032#diff-275ea9f25448c4618b0f5d5e886b0f4072dad0167fc254e3804e46be65ec2096R249)

If that's the problem, why does cloudflare version work?
Can you confirm if the generated file has `const importHonoContextStorage = () => import("hono/context-storage");` as well?

dai-shi:
> And then I get a runtime error on AWS when it tries to serve a page:
> Cannot find package 'hono' imported from /var/task/serve-aws-lambda.js

Does it work if you install hono manually for your app?

aheissenberger:
no - even after installing hono as a dependency the import of the `hono/context-storage` is not bundled by vite which might be a bug in rollup handling package / exports definitions - all others are included in the `./assets/`directory:
```ts
const importHono = () => import("./assets/index-BxM3V2xC.js");
/*                                            vv no ./assets/..   */
const importHonoContextStorage = () => import("hono/context-storage");

const importHonoNodeServerServeStatic = () => import("./assets/serve-static-CLws5J7J.js");
const importHonoAwsLambda = () => import("./assets/index-DM8TMISb.js");
const { Hono } = await importHono();
const { contextStorage } = await importHonoContextStorage();
const { serveStatic } = await importHonoNodeServerServeStatic();
const { handle } = await importHonoAwsLambda();
```

here is a simple code to debug the problem without deployment to aws:
```js
const requestHandlerPath = "./serve-aws-lambda.js";
const { handler } = await import(requestHandlerPath);
const eventData = {
  version: "2.0",
  routeKey: "$default",
  rawPath: "/",
  rawQueryString: "",
  headers: {
    accept: "*/*",
    "content-length": "0",
    host: "example.com",
    "user-agent": "PostmanRuntime/7.26.8",
    "x-amzn-trace-id": "Root=1-5f84c7a9-0e5b1e1e1e1e1e1e1e1e1e1e",
    "x-forwarded-for": "127.0.0.1",
    "x-forwarded-port": "443",
    "x-forwarded-proto": "https",
  },
  requestContext: {
    accountId: "123456789012",
    apiId: "api-id",
    domainName: "example.com",
    domainPrefix: "example",
    http: {
      method: "GET",
      path: "/",
      protocol: "HTTP/1.1",
      sourceIp: "127.0.0.1",
      userAgent: "PostmanRuntime/7.26.8",
    },
    requestId: "id",
    routeKey: "$default",
    stage: "$default",
    time: "12/Mar/2021:19:03:58 +0000",
    timeEpoch: 1615578238000,
  },
  isBase64Encoded: false,
};
const response = await handler(eventData, {});
console.log(response);
```

I was able to fix the broken vite resolution problem with this `vite.config.ts`:
```ts
import { defineConfig } from 'vite';
import path from "node:path";
export default defineConfig({
    resolve: {
        alias: {
          "hono/context-storage": path.resolve(__dirname, "node_modules/hono/dist/middleware/context-storage/index.js"),
        },
      },
});
```

aheissenberger:
This new aws deployment is missing the option to set the `distDir` with the environment variable `WAKU_BUILD_DIST_DIR`. This is important as the `serverless framework` will deploy the files to `/var/task/dist` with `/var/task/` being the current working directory and there is no option to remove one directory level. This is also the reason why the `./private/` directory does not work when used without joining with the current WAKU_BUILD_DIST_DIR directory.

Deploying with the `cdk framework` will deploy directly to `/var/task/`. 

See old code:
https://github.com/dai-shi/waku/blob/7e296e0e0bea1b173723b561a41225bb1d11b20e/packages/waku/src/lib/builder/serve-aws-lambda.ts#L9-L26

dai-shi:
It should be fixed with #898. Let me merge it.

aheissenberger:
> It should be fixed with #898. Let me merge it.

do I understand correctly that this fix only ignores the rollup path resolution problem by catching and ignoring the error?

aheissenberger:
I suggest to rollback your fix and add this to the plugins e.g. `vite-plugin-deploy-aws-lambda.ts` (last 7 lines):
```ts
export function deployAwsLambdaPlugin(opts: {
  srcDir: string;
  distDir: string;
}): Plugin {
  const platformObject = unstable_getPlatformObject();
  let entriesFile: string;
  return {
    name: 'deploy-aws-lambda-plugin',
    config(viteConfig) {
      const { deploy, unstable_phase } = platformObject.buildOptions || {};
      if (unstable_phase !== 'buildServerBundle' || deploy !== 'aws-lambda') {
        return;
      }
      const { input } = viteConfig.build?.rollupOptions ?? {};
      if (input && !(typeof input === 'string') && !(input instanceof Array)) {
        input[SERVE_JS.replace(/\.js$/, '')] = `${opts.srcDir}/${SERVE_JS}`;
      }

     // fix for rollup resolution problem
      if (viteConfig.resolve===undefined) {
        viteConfig.resolve = {};
      }
      if (viteConfig.resolve.alias===undefined) {
        viteConfig.resolve.alias = {};
      }
      (viteConfig.resolve.alias as Record<string, string>)['hono/context-storage'] = path.resolve('node_modules/hono/dist/middleware/context-storage/index.js');
```

dai-shi:
> > It should be fixed with #898. Let me merge it.
> 
> do I understand correctly that this fix only ignores the rollup path resolution problem by catching and ignoring the error?

No, my expectation is it should work flawlessly if users install `hono` in the app. It ignores if it's not installed.

dai-shi:
> ```js
> (viteConfig.resolve.alias as Record<string, string>)['hono/context-storage'] = path.resolve('node_modules/hono/dist/middleware/context-storage/index.js');
> ```

Are you sure that is required? I can add it if that's the case but I would like to keep #898.

aheissenberger:
> > ```js
> > (viteConfig.resolve.alias as Record<string, string>)['hono/context-storage'] = path.resolve('node_modules/hono/dist/middleware/context-storage/index.js');
> > ```
> 
> Are you sure that is required? I can add it if that's the case but I would like to keep #898.

With the #898 fix and the `hono` package installed, I get no error but the build process will not bundle the import and will still leave the original module import `"hono/context-storage"` :
```ts
const importHono = () => import("./assets/index-BxM3V2xC.js");
/*                                            vv no ./assets/..   */
const importHonoContextStorage = () => import("hono/context-storage");
```
Does it bundle correctly on your system?

You can keep #898 but it will make it harder to understand why the hono context is not available. If hono context should be optional, I suggest to use a documented feature flag to activate but still fail if the hono library is not added as a dependency.

My solution to add the alias to the build configuration is a hack, but I was not able to find the reason why vite/rollup ignores the package.json exports 😞

dai-shi:
> ```ts
> const importHonoContextStorage = () => import("hono/context-storage");
> ```

This shouldn't be so in the latest commit. Can you make sure if you are on the latest commit?

If `hono` isn't installed, then people wouldn't be able to use the hono context, so I think it's fine not to add the middleware.

aheissenberger:
> > ```ts
> > const importHonoContextStorage = () => import("hono/context-storage");
> > ```
> 
> This shouldn't be so in the latest commit. Can you make sure if you are on the latest commit?
> 
> If `hono` isn't installed, then people wouldn't be able to use the hono context, so I think it's fine not to add the middleware.

I use commit 872913670f, have compiled WAKU, have changed the version of WAKU in the `03_demo` to `"waku": "workspace:*"`, have `hono` in the dependencies, and still get  `import("hono/context-storage")` 

A branch of the PR882 with my changes to the `examples/03_demo`:
https://github.com/dai-shi/waku/compare/main...aheissenberger:waku:broken-aws-hono-20240925

dai-shi:
> and still get `import("hono/context-storage")`

If you are talking about "hono/context-storage", it's expected. But, we no longer have `importHonoContextStorage`, and instead this:

```js
let contextStorage;
try {
 ({ contextStorage } = await import('hono/context-storage'));
} catch {}
```



> have `hono` in the dependencies

and still causes the build error? Then we may need config.alias hack... unfortunately.

dai-shi:
Wait, there was something I misunderstood with the plugins. #899 might fix something.

dai-shi:
@aheissenberger A fix is added.



> it is never bundling `hono/context`, which breaks any usage of the `hono/context` after building.

But, I'm not sure why it fails to bundle.
I assume `serve-cloudflare` is working well, but now I'm not so sure. @rmarscher

aheissenberger:
> @aheissenberger A fix has been added.
> 
> > it is never bundling `hono/context`, which breaks any usage of the `hono/context` after building.
> 
> But, I'm not sure why it fails to bundle. I assume `serve-cloudflare` is working well, but now I'm not so sure. @rmarscher

I checked the latest version b34b2a99d83 - no change - still `({ contextStorage } = await import("hono/context-storage"));`. This does not throw any error but will never work as the bundler cannot include the required hono package.

I think I know why it might work with other hosting adapters - the adapters for AWS is compiling external dependencies and does not include a "node_modules" folder for the target system. Your solution expects a node_module folder with a `hono` package at runtime. If using `hono/context-storage` should be optional, provide an other mechanism which is 

Here is  link to the archive which gets transferred to AWS Lambda for the `examples/03_demo`:
https://www.dropbox.com/t/nrlW9Ua35Kg5sWX3

I suggest the following fix:
* remove the package `'hono/context-storage'` from the vite build config defined in `packages/waku/dist/lib/builder/build.js:201`which is the main reason for all the troubles
* now if the package `hono` exists in the project (e.g.`03_demo`) the packages is included. If the package is missing no error on bundling and no error at runtime based on the `try .. catch` construct

```diff
diff --git a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
index 65cebfd2..d30bafc5 100644
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -63,6 +63,13 @@ export function deployAwsLambdaPlugin(opts: {
       if (input && !(typeof input === 'string') && !(input instanceof Array)) {
         input[SERVE_JS.replace(/\.js$/, '')] = `${opts.srcDir}/${SERVE_JS}`;
       }
+      if (Array.isArray(viteConfig?.ssr?.external)) {
+        if (viteConfig.ssr.external.includes('hono/context-storage')) {
+          viteConfig.ssr.external = viteConfig.ssr.external.filter(
+            (ext) => ext !== 'hono/context-storage'
+          );
+        }
+      }
     },
     configResolved(config) {
       entriesFile = `${config.root}/${opts.srcDir}/${SRC_ENTRIES}`;
@@ -71,9 +78,9 @@ export function deployAwsLambdaPlugin(opts: {
       if (source === `${opts.srcDir}/${SERVE_JS}`) {
         return source;
       }
-      if (source === 'hono/context-storage') {
-        return { id: source, external: true };
-      }
+      // if (source === 'hono/context-storage') {
+      //   return { id: source, external: true };
+      // }
     },
     load(id) {
       if (id === `${opts.srcDir}/${SERVE_JS}`) {
```

rmarscher:
I think another fix is to remove `external: ['hono/context-storage'],` from the build config. It works for me if I comment out https://github.com/dai-shi/waku/blob/refactor/deploy/aws-lambda/packages/waku/src/lib/builder/build.ts#L249

That gets the output serve-aws-lambda.js to have 
```js
let contextStorage;
try {
  ({ contextStorage } = await import("./assets/index-ED7G-tsu.js"));
} catch {
}
```

aheissenberger:
What is the reason to set  the `ssr.external: ['hono/context-storage'],` in the first place in build?

> I think another fix is to remove `external: ['hono/context-storage'],` from the build config. It works for me if I comment out https://github.com/dai-shi/waku/blob/refactor/deploy/aws-lambda/packages/waku/src/lib/builder/build.ts#L249
> 
> That gets the output serve-aws-lambda.js to have
> 
> ```js
> let contextStorage;
> try {
>   ({ contextStorage } = await import("./assets/index-ED7G-tsu.js"));
> } catch {
> }
> ```

you are quicker - I was just asking why this exists in the first place. I only tried to fix aws, as I have no knowledge about the other hosting platforms.

rmarscher:
> What is the reason to set the ssr.external: ['hono/context-storage'], in the first place in build?

It was from a previous attempt to get the same module to be imported by the hono runtime and from server components. I'm pretty sure it can be safely removed. I'll give a try with Cloudflare. I have a few Waku apps. One on AWS Lambda and a few deployments on Cloudflare.

I use a fork of this adapter for AWS so I can use 
```
const { streamHandle } = await importHonoAwsLambda();
```
and enable lambda streaming.

rmarscher:
I confirmed it also works great with cloudflare with `external: ['hono/context-storage']` removed from build.ts.

rmarscher:
Here's my fork. It uses streaming and it separates the public assets from the lambda function. That way I don't include any non-needed static assets in the function bundle. I upload the public assets to an S3 bucket and use a fallback origin with cloudfront so to have it look there first before routing to the lambda function. That works very well. https://github.com/dai-shi/waku/compare/refactor/deploy/aws-lambda...rmarscher:waku:slimmer-aws-lambda-sst-patch

aheissenberger:
> I use a fork of this adapter for AWS so I can use
> 
> ```
> const { streamHandle } = await importHonoAwsLambda();
> ```
> 
> and enable lambda streaming.

How about adding AWS Lambda Streaming as an option in the config file? Both variants are useful as AWS Streaming is not supported by the API Gateway.

dai-shi:
> What is the reason to set the `ssr.external: ['hono/context-storage'],` in the first place in build?

It's #894. Without it, it doesn't work with `waku start` (node server).
So, I think it's generally required.

But, I'm fine to remove it for aws-lambda deploy (and cloudflare too?)

dai-shi:
@rmarscher Does d15b11b0 work?

aheissenberger:
> > What is the reason to set the `ssr.external: ['hono/context-storage'],` in the first place in build?
> 
> It's #894. Without it, it doesn't work with `waku start` (node server). So, I think it's generally required.
> 
I have no problems to run `waku start` with `ssr.external: ['hono/context-storage']` removed in `packages/waku/dist/lib/builder/build.js:200`. I tried `03_demo` and `08_cookies` with the node server.

dai-shi:
> I have no problems to run `waku start` with `ssr.external: ['hono/context-storage']` removed in `packages/waku/dist/lib/builder/build.js:200`. I tried `03_demo` and `08_cookies` with the node server.

Hmmm? I just tried `npm run examples:prd:08_cookies` with removing the line, and it shows this error:
```
> waku start

ready: Listening on http://localhost:8080/
waku context [ 'count', '__waku_requestHeaders' ]
Error: Context is not available
```

aheissenberger:
> > What is the reason to set the `ssr.external: ['hono/context-storage'],` in the first place in build?
> 
> It's #894. Without it, it doesn't work with `waku start` (node server). So, I think it's generally required.
> 
I have no problems to run `waku start` with `ssr.external: ['hono/context-storage']` removed in `packages/waku/dist/lib/builder/build.js:200`. I tried `03_demo` and `08_cookies` with the node server.

aheissenberger:
> > I have no problems to run `waku start` with `ssr.external: ['hono/context-storage']` removed in `packages/waku/dist/lib/builder/build.js:200`. I tried `03_demo` and `08_cookies` with the node server.
> 
> Hmmm? I just tried `npm run examples:prd:08_cookies` with removing the line, and it shows this error:
> 
> ```
> > waku start
> 
> ready: Listening on http://localhost:8080/
> waku context [ 'count', '__waku_requestHeaders' ]
> Error: Context is not available
> ```

yes - I get the same error - is it possible to limit `ssr.external: ['hono/context-storage']` to the node server or do we know the need for `ssr.external: ['hono/context-storage']` is also given for all other deployment adapters?

dai-shi:
I'd go the other way around. The node server is the default and deploy adapters are options.

aheissenberger:
@dai-shi What is your suggestion to add an option for a deployment adapter in `waku.config.ts` to allow to activate AWS Lambda streaming?

dai-shi:
> @dai-shi What is your suggestion to add an option for a deployment adapter in `waku.config.ts` to allow to activate AWS Lambda streaming?


Good question. The config stuff is still something I'll be considering. So, I would like to delay the decision. Meanwhile, if you need some configurations in deployment adapters, the easiest is `process.env...` for now.

aheissenberger:
With latest commit both situations, with and without package `hono` installed work for `example/03_demo` and without any changes with `08_cookies` deployed to aws.

aheissenberger:
> Here's my fork. It uses streaming and it separates the public assets from the lambda function. That way I don't include any non-needed static assets in the function bundle. I upload the public assets to an S3 bucket and use a fallback origin with cloudfront so to have it look there first before routing to the lambda function. That works very well. [refactor/deploy/aws-lambda...rmarscher:waku:slimmer-aws-lambda-sst-patch](https://github.com/dai-shi/waku/compare/refactor/deploy/aws-lambda...rmarscher:waku:slimmer-aws-lambda-sst-patch)

@rmarscher when this refactoring is merged we can discuss to add streaming to this adapter. I also deploy the static assets to s3 and use cloudfront to cache and split the requests between waku and the static assent but do not need the output folder to be rearranged.

dai-shi:
> With latest commit both situations, with and without package `hono` installed work for `example/03_demo` and without any changes with `08_cookies` deployed to aws.

Nice. Thanks for comfirning.



> And `resolveId(source)` does not need
> 
> ```ts
>       if (source === 'hono/context-storage') {
>         return { id: source, external: true };
>       }
> ```

Ah, thanks.

dai-shi:
Merging this.

@rmarscher Can you check the cloudflare one too and send a PR?

> @rmarscher when this refactoring is merged we can discuss to add streaming to this adapter.

Nice. Looking forward to a PR based on you guys' discussions.

PR #883, opened 2024-09-15
ci: pnpm cache

himself65:
3mins -> 37s


dai-shi:
🎉

dai-shi:
@himself65 It seems like the cache only works for ubuntu, and for macos and windows, ci installs playwright every time, saying it doesn't find a cache.

https://github.com/dai-shi/waku/actions/runs/11258656376/job/31305801590?pr=958

image

PR #884, opened 2024-09-16
feat: export getHonoContext

dai-shi:
following #852, this should be more explicit. Still unstable, but less experimental.


rmarscher (approved):
💯  Being able to access this from RSC in an actual server environment provides a lot of flexibility and resolves #827 for Cloudflare usage.

PR #886, opened 2024-09-16
feat: entries gen for fs router projects

tylerlaws0n:
This PR will add a vite plugin for running on `waku dev` that will generate `entries.gen.tsx`

Tasks left
- [x] Attempt to find prettier config
- [x] Test with more calls to `getConfig`
- [x] Disable generation when not using fs-router


packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+  await promises.writeFile(OUTPUT_FILE, JSON.stringify(fileInfo), 'utf-8');
+};
+
+export const typegenPlugin = (): Plugin => {
+  return {
+    name: 'vite-file-watcher-plugin',
+
+    configureServer(server) {
```
tylerlaws0n, line 76: I noticed that this runs on startup, but it doesn't seem like it runs when I change a file or add/remove/rename files.

I'll probably read through the vite plugin docs more after getting the rest working unless you know a better way to watch for these changes.

packages/waku/src/lib/utils/vite-loader.ts
```diff
@@ -1,12 +1,31 @@
...
+    },
+  });
+
+  let config: Config = {};
+  try {
+    config = (await viteTempServer.ssrLoadModule(fileURLToFilePath(fileURL)))
+      .default;
+  } finally {
+    await viteTempServer.close();
+  }
```
tylerlaws0n, line 20: I don't particularly like needing to do this temp server thing, but I needed a way to load `config.typegen` before adding the plugin. Is there a better way to do this?

packages/waku/src/config.ts
```diff
@@ -48,6 +48,8 @@ export interface Config {
    * ]
    */
   middleware?: () => Promise<{ default: Middleware }>[];
+
+  typegen?: boolean;
```
dai-shi, line 52: I don't think we need to make it configurable for now.

dai-shi: In the future we may want to configure vite plugins.

packages/waku/src/lib/utils/vite-loader.ts
dai-shi: Does removing the config solve it?

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,102 @@
+import type { Plugin } from 'vite';
+import { promises } from 'fs';
+import path from 'path';
+
+const SRC_DIR = path.resolve(process.cwd(), './src/pages');
```
dai-shi, line 5: seems hard coded. and we don't use `cwd`.

let's revisit when this PR is ready.

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,102 @@
...
+        console.warn('Failed to load page file:', shortPath);
+      }
+    }
+  }
+  await promises.writeFile(OUTPUT_FILE, JSON.stringify(fileInfo), 'utf-8');
+};
+
+export const typegenPlugin = (): Plugin => {
+  return {
+    name: 'vite-file-watcher-plugin',
```
dai-shi, line 74: Let's make it clear that it is for `fs-router`.

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,102 @@
...
+          path: shortPath.replace('index.tsx', ''),
+          src: shortPath, // where to import the component from
+          config,
+        });
+      } catch {
+        console.warn('Failed to load index file:', shortPath);
+      }
+    } else {
+      try {
+        const pageFile = await import(fullPath);
```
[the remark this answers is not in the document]
dai-shi: we shouldn't import probably.
if we need to load, that requires `react-server` condition.
can we do static analysis?

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
dai-shi: https://vitejs.dev/guide/api-plugin#handlehotupdate

dai-shi (commented):
Welcome to the vite world.

packages/waku/src/config.ts
tylerlaws0n: sounds good and yes this allows removing the temp vite server in the vite-loader

tylerlaws0n:
> we shouldn't import probably.
> if we need to load, that requires react-server condition.
> can we do static analysis?

static analysis sounds like it would have a lot of potential gotchas. We could get the whole function, but if it calls something that is imported the function will not be executable. Like [here](https://github.com/dai-shi/waku/blob/main/e2e/fixtures/partial-build/src/pages/page/%5Btitle%5D.tsx)

Going further with static analysis sounds too complicated from what I can think of. 

How could I require the react-server condition you mentioned? I don't actually need to run react for anything but I will need to import from files that have react I think.

packages/waku/src/lib/utils/vite-loader.ts
tylerlaws0n: yes, thanks!

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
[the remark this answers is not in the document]
tylerlaws0n: Do you mean static analysis to get the return result of getConfig directly, or do you mean to get the function then attempt to execute it?

Which leads to the question of what do we do for this scenario: https://github.com/dai-shi/waku/blob/main/e2e/fixtures/partial-build/src/pages/page/%5Btitle%5D.tsx where you need to run the `getConfig` to get the staticPaths result.

I would be fine with defaulting to `string[]` or `string[][]` based on the path if the static analysis.

parsing the static code like this on our own does seem a bit complex though...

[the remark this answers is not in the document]
dai-shi: If we want to run the code, it's probably easier with `waku codegen-fs-router` which is a separate process.

I thought the codegen from pages/* -> entries.tsx is one-to-one source code conversion.
```js
const mod = await import('./pages/[title].js');
createPage({
  path: '/[title]',
  component: mod.default,
  render: dynamic
  ...(await mod.getConfig()),
});
```
Not sure if types are inferred as expected though.

tylerlaws0n:
Commenting my findings from a bit of research so far here

This plugin from the tanstact router is _very_ similar to what we will want. It looks like they just use `watchChange` to trigger their check to generate their routes file:

https://github.com/TanStack/router/blob/main/packages/router-plugin/src/core/router-generator-plugin.ts#L79-L83

I was unable to get watchChange logging in my initial tests with out setup, but this leads me to think that it may be how we configure the vite config for the dev server maybe? Either way, it's a POC that `watchChange` should work for what we want to do.

The downside of their example is that afaict: They do not have a need to import anything like we do for our `getConfig`.

Since at this stage we are really just interested in collecting the routes for some autocomplete on the link paths, I think our best first option is to start with defaulting to `string` for any slug. This makes it so that we can get away with just reading filename and building the paths as a reflection of those filenames. 

I'll put together a draft of what this will look like tomorrow night and in doing so get rid of any notion of importing from the plugin to run getConfig.

tylerlaws0n:
Another thing to add: I no longer think that making a generation of the entries file makes sense. With it just being a plugin that looks at what path you are on, we won't know the `render` mode of any of the routes, so making an `entries.gen.tsx` won't make sense. 

Instead I'll make something like `paths.gen.ts` that has all possible paths in the app for the paths defined in `src/pages` assuming all slugs could be any string

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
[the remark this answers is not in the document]
tylerlaws0n: wow! this works... I didn't even consider generating dynamic imports. This is a great idea.

One thing to note is that it relies on this tsconfig setting:

```json
"allowImportingTsExtensions": true
```

Screenshot 2024-09-18 at 6 54 14 PM

I'm good with adding this if you are though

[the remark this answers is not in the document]
dai-shi: I would like to avoid adding "allowImportingTsExtensions".
Can you change "tsx" to "js" when generating the file?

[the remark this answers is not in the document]
dai-shi: > I didn't even consider generating dynamic imports.

It can be static imports, by the way.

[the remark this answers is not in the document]
tylerlaws0n: ah, yeah that works too. Thanks!

examples/05_nossr/src/pages/about.tsx
```diff
@@ -28,5 +28,5 @@ const getData = async () => {
 export const getConfig = async () => {
   return {
     render: 'static',
-  };
+  } as const;
```
tylerlaws0n, line 31: This is needed to tell typescript that the type is `{ render: 'static' }` and otherwise it will be `{ render: string }` which will not work

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (file.isDirectory()) {
+      results = results.concat(await collectFiles(fullPath));
+    } else if (file.isFile() && fullPath.endsWith('.tsx')) {
+      results.push(fullPath.replace(SRC_DIR, ''));
+    }
+  }
+  return results;
+};
+
+const srcToName = (src: string) => {
```
tylerlaws0n, line 24: Now that I have a working example, I plan to circle back to refactoring some of these variables and functions. I know that I have some duplicative string formatting going on

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,146 @@
...
+    result += `import ${moduleName}${file.hasGetConfig ? `, { getConfig as ${moduleName}_getConfig }` : ''} from './src/pages/${file.src.replace('.tsx', '')}';\n`;
+  }
+
+  result += `\nconst DEFAULT_CONFIG = {render: 'dynamic'} as const;
+  
+  const _pages = createPages(async (pagesFns) => [\n`;
+
+  for (const file of fileInfo) {
+    const moduleName = srcToName(file.src);
+    result += `pagesFns.${file.type === 'layout' ? 'createLayout' : 'createPage'}({ path: '${file.path}', component: ${moduleName}, ...${file.hasGetConfig ? `(await ${moduleName}_getConfig())` : 'DEFAULT_CONFIG'} }),\n`;
```
tylerlaws0n, line 94: @dai-shi Is this line hard to read? If yes, do you have any recommendations for simplifying it?

examples/05_nossr/entries.gen.tsx
```diff
@@ -0,0 +1,36 @@
...
+import type { PathsForPages } from 'waku/router';
+
+import Layout, { getConfig as Layout_getConfig } from './src/pages/_layout';
+import About, { getConfig as About_getConfig } from './src/pages/about';
+import Index, { getConfig as Index_getConfig } from './src/pages/index';
+import NewHi from './src/pages/new/hi';
+
+const DEFAULT_CONFIG = { render: 'dynamic' } as const;
+
+const _pages = createPages(async (pagesFns) => [
```
tylerlaws0n, line 11: all things considered...I am reasonably happy with how this generates

examples/05_nossr/entries.gen.tsx
```diff
@@ -0,0 +1,36 @@
+import { createPages } from 'waku';
+import type { PathsForPages } from 'waku/router';
+
+import Layout, { getConfig as Layout_getConfig } from './src/pages/_layout';
+import About, { getConfig as About_getConfig } from './src/pages/about';
+import Index, { getConfig as Index_getConfig } from './src/pages/index';
+import NewHi from './src/pages/new/hi';
+
+const DEFAULT_CONFIG = { render: 'dynamic' } as const;
```
tylerlaws0n, line 9: @dai-shi is this correct? Do we assume dynamic or static when no getConfig is defined?

tylerlaws0n (commented):
My TODO that I am planning for:

- [ ] Test with more types of `getConfigs`
- [x] clean up the code in the vite plugin file a bit
- [x] run prettier on the generated output programatically
- [ ] fix the vite plugin to hot reload the generation when necessary

examples/05_nossr/entries.gen.tsx
dai-shi: For pages, yes. For layouts, no.

examples/05_nossr/src/pages/about.tsx
dai-shi: This is very unfortunate.

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
dai-shi: there might be room for improvement, but fine for now.

packages/waku/src/lib/utils/vite-loader.ts
```diff
@@ -1,11 +1,13 @@
 import { createServer as createViteServer } from 'vite';
 import { fileURLToFilePath } from '../utils/path.js';
+import { typegenPlugin } from '../plugins/vite-plugin-typegen.js';
 
 export const loadServerFile = async (fileURL: string) => {
   const vite = await createViteServer({
     ssr: {
       external: ['waku'],
     },
+    plugins: [typegenPlugin()],
```
dai-shi, line 10: Can we remove this?

dai-shi (commented):
random replies.

examples/05_nossr/src/pages/about.tsx
tylerlaws0n: I agree... importing from this file to the plugin would be a way to get the literal value into the generation, but I can't think of other ways quite yet.

packages/waku/src/lib/utils/vite-loader.ts
tylerlaws0n: This is needed to run the generation step with `waku dev` - Ideally we'd include this here and I figure out how to run the generation in watch mode if you're ok with that.

dai-shi: I think it should be in dev-server-impl.ts.

This file is an unfortunate hack, and we would like to remove it in the future.

tylerlaws0n: oh! moving this to there fixes the hot reload issues :) thanks for the tip. I'll polish and have something more reviewable in the next couple of days

examples/05_nossr/entries.gen.tsx
tylerlaws0n: fixed this by removing `DEFAULT_CONFIG` and adding inline on the type generation, but matched the behavior of pages as dynamic and layouts as static

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
tylerlaws0n: time to revisit I think haha. Any suggestions for how to make this less hard-coded?

tylerlaws0n: Follow up to this is: How should we detect when a user is using fs-router or not? This also seems like something that might suggest some need to hardcode, but I'm curious if you know a better way.

examples/05_nossr/src/pages/about.tsx
tylerlaws0n: one other way would be to add `satisfies GetConfigResult` instead of `as const` or something like this, but that would probably be less desirable given that it's longer.

dai-shi: `as const` is better.

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
dai-shi: - Use `config.root` instead of `process.cwd`.
- `srcDir` is in config, so get it through plugin options. See other plugins.
- `'pages'` must be hard-coded for now.
- Related, we can't detect if fsRouter is used, but if we don't find `entries.tsx` (aka "managed mode"), then we run the plugin. (see rscManagedPlugin)

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -27,6 +27,7 @@ import { rscManagedPlugin } from '../plugins/vite-plugin-rsc-managed.js';
 import { rscDelegatePlugin } from '../plugins/vite-plugin-rsc-delegate.js';
 import { mergeUserViteConfig } from '../utils/merge-vite-config.js';
 import type { ClonableModuleNode, Middleware } from './types.js';
+import { typegenPlugin } from '../plugins/vite-plugin-typegen.js';
```
dai-shi, line 30: Let's name the file and the function so that it implies a feature for `fs-router` or managed mode (rscManagedPlugin).

examples/05_nossr/entries.gen.tsx
```diff
@@ -0,0 +1,32 @@
+import { createPages } from 'waku';
+import type { PathsForPages } from 'waku/router';
+
+import Layout, { getConfig as Layout_getConfig } from './src/pages/_layout';
```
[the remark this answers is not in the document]
tylerlaws0n: Do we support both `layout.tsx` and `_layout.tsx`? If yes, what happens if a project has both in the same directory?

examples/05_nossr/src/pages/_layout.tsx
```diff
@@ -40,5 +40,5 @@ const getData = async () => {
 export const getConfig = async () => {
   return {
     render: 'static',
-  };
+  } as const;
```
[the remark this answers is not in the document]
tylerlaws0n: Yeah that also works. I chose to put it on the full object as a convention because when `staticPaths` and `render` are both defined this way requires you to write `as const` just once, but `'static' as const` would also require `staticPaths: ['foo'] as const`

examples/05_nossr/entries.gen.tsx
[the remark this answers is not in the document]
dai-shi: layout.tsx works the same as foo.tsx. So it will be `/layout` like `/foo`.

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
tylerlaws0n: all updated 👍

packages/waku/src/lib/plugins/vite-plugin-typegen.ts
```diff
@@ -0,0 +1,180 @@
+import type { Plugin } from 'vite';
+import { promises, readFileSync } from 'fs';
```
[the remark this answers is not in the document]
tylerlaws0n: added node prefix to 'path'  as well

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,194 @@
+import type { Plugin } from 'vite';
+import { promises, readFileSync, existsSync } from 'node:fs';
```
dai-shi, line 2: Can we use `import { ... } from 'node:fs/promises'` ?

tylerlaws0n: Is this ok or did you not want to use sync calls at all? 

```ts
import { readdir, writeFile } from 'node:fs/promises';
import { existsSync, readFileSync } from 'node:fs';`
```

dai-shi: that is ok.

examples/05_nossr/entries.gen.tsx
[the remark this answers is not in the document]
tylerlaws0n: good catch! that would have been in conflict. I fixed this with _Layout for the actual layouts' names and added a page named /layout to the fs-router example

examples/10_fs-router/src/entries.tsx
```diff
@@ -1,11 +0,0 @@
...
-
-declare global {
-  interface ImportMeta {
-    readonly glob: any;
-  }
-}
-
-export default fsRouter(import.meta.url, (file: string) =>
-  import.meta.glob('./pages/**/*.tsx')[`./pages/${file}`]?.(),
-);
```
dai-shi, line 11: Actually, I want to keep this in the example. Can you revert changes in 10_fs-router?

packages/waku/src/lib/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const EXTENSIONS = ['.js', '.ts', '.tsx', '.jsx', '.mjs', '.cjs'];
 export const SRC_MAIN = 'main';
 export const SRC_ENTRIES = 'entries';
+export const SRC_PAGES = 'pages';
```
dai-shi, line 4: Technically, this is customizable, so not a constant like others. Let's move this to the plugin for now.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,195 @@
...
+import path from 'node:path';
+import { SRC_ENTRIES, SRC_PAGES } from '../constants.js';
+import { joinPath } from '../utils/path.js';
+
+const srcToName = (src: string) => {
+  const split = src
+    .split('/')
+    .map((part) => (part[0]!.toUpperCase() + part.slice(1)).replace('-', '_'));
+
+  if (src.endsWith('_layout.tsx')) {
```
dai-shi, line 13: This will confuse with `/pages/foo_layout.tsx`.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,195 @@
...
+    return split.slice(0, -1).join('') + '_Layout';
+  } else if (src.endsWith('index.tsx')) {
```
dai-shi, line 15: same here.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: How about this?
```suggestion
  if (split.at(-1) === '_layout.tsx') {
```

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,195 @@
...
+  let entriesFile: string | undefined;
+  let pagesDir: string | undefined;
+  let outputFile: string | undefined;
+  let formatter = (s: string): Promise<string> => Promise.resolve(s);
+  return {
+    name: 'vite-plugin-fs-router-typegen',
+    apply: 'serve',
+    async configResolved(config) {
+      pagesDir = joinPath(config.root, opts.srcDir, SRC_PAGES);
+      entriesFile = joinPath(config.root, opts.srcDir, `${SRC_ENTRIES}.tsx`);
```
dai-shi, line 61: entriesFile can have different extensions such as `.ts` and `.js`. Please leave a TODO comment unless it's easy to fix.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,195 @@
...
+      } catch {
+        // ignore
+      }
+    },
+    configureServer(server) {
+      if (!entriesFile || !pagesDir || !outputFile || existsSync(entriesFile)) {
+        return;
+      }
+
+      // Recursively collect `.tsx` files in the given directory
```
dai-shi, line 80: readdir has `recursive` option.
https://nodejs.org/api/fs.html#fspromisesreaddirpath-options
image

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,195 @@
...
+      };
+
+      const generateFile = (filePaths: string[]): string => {
+        const fileInfo = [];
+        for (const filePath of filePaths) {
+          // where to import the component from
+          const src = filePath.slice(1);
+          const hasGetConfig = fileExportsGetConfig(filePath);
+
+          if (filePath.endsWith('_layout.tsx')) {
```
dai-shi, line 114: not sure if it has the same issue with `/pages/foo_layout.tsx`.

examples/10_fs-router/src/entries.tsx
tylerlaws0n: reverted 👍

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylerlaws0n: good point, fixed!

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylerlaws0n: updated to check for ts, tsx, js, and jsx

easy change :)

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylerlaws0n: nice! thanks!

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylerlaws0n: yep! fixed, thanks

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -0,0 +1,201 @@
...
+  let entriesFilePossibilities: string[] | undefined;
+  let pagesDir: string | undefined;
+  let outputFile: string | undefined;
+  let formatter = (s: string): Promise<string> => Promise.resolve(s);
+  return {
+    name: 'vite-plugin-fs-router-typegen',
+    apply: 'serve',
+    async configResolved(config) {
+      pagesDir = joinPath(config.root, opts.srcDir, SRC_PAGES);
+      entriesFilePossibilities = ['tsx', 'ts', 'js', 'jsx'].map((ext) =>
```
dai-shi, line 65: Nice. Actually, you want to use `EXTENSIONS` defined in `constants.ts`.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: https://github.com/dai-shi/waku/pull/886#discussion_r1778088965

examples/01_template/src/entries.gen.tsx
```diff
@@ -0,0 +1,32 @@
...
+  }),
+]);
+
+declare module 'waku/router' {
+  interface RouteConfig {
+    paths: PathsForPages<typeof _pages>;
+  }
+}
+
+export default _pages;
```
dai-shi, line 32: While it's nice to test this with `01_template`, this will be the real template for `create-waku`, so I don't want this file to be committed.

examples/01_template/src/pages/_layout.tsx
```diff
@@ -40,5 +40,5 @@ const getData = async () => {
 export const getConfig = async () => {
   return {
     render: 'static',
-  };
+  } as const;
```
dai-shi, line 43: this is good. maybe we need this change in all examples with fs-router (incl. managed) with TypeScript.

tylerlaws0n: I can do that! do you also want me to generate the entries file in all fs router examples?

examples/01_template/src/entries.gen.tsx
tylerlaws0n: deleting and adding to gitignore of 01_template specifically

examples/01_template/src/pages/_layout.tsx
dai-shi: tbh, i don't think we need to commit entries.gen.tsx files at all with this PR.
it was good for the review purpose.

examples/01_template/src/entries.gen.tsx
dai-shi: technically, all examples can be a template. `npm create waku@latest -- --choose`.

examples/01_template/src/pages/_layout.tsx
tylerlaws0n: should I add it to the top level gitignore then?

tylerlaws0n:
Found the issue in the playwritght tests.... the vite plugin is creating bad paths in windows 😅 

dang windows

```
Error: ENOENT: no such file or directory, open 'D:/a/waku/waku/examples/04_cssmodules/src/pagesD:\a\waku\waku\examples\04_cssmodules\src\pages\about.tsx'
```

tylerlaws0n:
@dai-shi I have the fixes for above locally... a few last questions

1. Do you want `entries.gen.tsx` added to each gitignore? (top level of monorepo and all example templates)
2. Do you still want to add `as const` to the examples? I think it is good to have to make it more clear how to get the types working, but up to you.
3. should the generated imports use the `.js` import style that we use internally?

examples/01_template/src/pages/_layout.tsx
dai-shi: Yeah, let's do that.

dai-shi:
> the vite plugin is creating bad paths in windows 😅

Yes, that's a tough one. We spent so many hours and even days fighting with it with @himself65 .

dai-shi:
> Do you want `entries.gen.tsx` added to each gitignore? (top level of monorepo and all example templates)

Only top level. As we discussed, in the template, we expect people to commit `entries.gen.tsx` so that CI can type check, right?

> Do you still want to add `as const` to the examples? I think it is good to have to make it more clear how to get the types working, but up to you.

Yes, for all TS examples.

> should the generated imports use the `.js` import style that we use internally?

No. It can be like `examples/07_router/src/entries.tsx`.

tylerlaws0n:
> No. It can be like examples/07_router/src/entries.tsx.

I think you misunderstood what I was referencing here. I mean `import .. from './_layout.js'` for example

dai-shi:
> > No. It can be like examples/07_router/src/entries.tsx.
> 
> I think you misunderstood what I was referencing here. I mean `import .. from './_layout.js'` for example

Isn't it about the code in `entries.gen.tsx`?

tylerlaws0n:
> > > can be like examples/07_router/src/entries.tsx.
> > 
> > 
> > I think you misunderstood what I was referencing here. I mean `import .. from './_layout.js'` for example
> 
> Isn't it about the code in `entries.gen.tsx`?

this at the top of that file:

```ts
import _Layout, { getConfig as _Layout_getConfig } from './pages/_layout.js';
import About, { getConfig as About_getConfig } from './pages/about.js';
import Index, { getConfig as Index_getConfig } from './pages/index.js';
```

dai-shi:
Yes, we can omit `.js` and should do to be consistent with `examples/07_router/src/entries.tsx`. (Ah, this one is using dynamic import. We can change it to static import.)

tylerlaws0n:
> Yes, we can omit `.js` and should do to be consistent with `examples/07_router/src/entries.tsx`. (Ah, this one is using dynamic import. We can change it to static import.)

Sounds good, but there's no dynamic import? We dropped that idea a while ago

tylerlaws0n:
as I went through each of the examples, I noticed that most use the `entries.tsx` file. Is that something you think should be recommended on waku.gg as the standard? 

My current understanding of the docs is that it only really suggests the fs-router

dai-shi:
> but there's no dynamic import?

never mind. It was referring to `examples/07_router/src/entries.tsx`.

dai-shi:
> as I went through each of the examples, I noticed that most use the `entries.tsx` file. Is that something you think should be recommended on waku.gg as the standard?
> 
> My current understanding of the docs is that it only really suggests the fs-router

So, there are two aspects (or goals) in Waku.

1. Research project: What would be a "minimal" API for an RSC framework.
2. Production-level framework: Let's make React development fun again.

Obviously, I started as the first one, and I'm not thinking it a toy. My vision includes Jotai integration, multiple routers (router-agnostic API), and building ecosystem.
But, many people get attracted to the second one, and it's our current focus. Particularly, `waku/router` plays an important role.

`examples/0[1234]*` are for the second aspect.
`examples/0[57]*` are still for the second aspect, but for advanced users.
`examples/0[89]*` are still for the second aspect, but more like PoC. (Maybe we should convert them to use `waku/router`.)
`examples/[12]*` are for the first aspect, or some experiments for the second.

dai-shi (approved):
Looks great! Outstanding work!

PR #879, opened 2024-09-15
refactor(deploy): cloudflare serve

dai-shi:
Prior art:
- #877 
- #878

Issue #871, opened 2024-09-10
Dynamic routes on Cloudflare fail trying to import node:fs/promises


rmarscher:
The error in the workerd console is

```
Cannot process SSR Error: No such module "node:fs/promises".
    at null.<anonymous> (file:///waku-project/.wrangler/tmp/pages-tQnVbB/entries.js:485:11)
```

That is referencing [these changed lines](https://github.com/dai-shi/waku/pull/860/files#diff-dbb54d801e436acf21fb8afff808d82c614f9f0a7fc5e3ce8a666e437eef4d04L12-L55) in #860. Before that PR, Cloudflare workerd must have been getting routing information from `unstable_buildConfig[0].customData.data`.

To reproduce
npm create waku@0.9.3
cd waku-project
npm install

Then edit src/pages/about.tsx and change getConfig to dynamic:

```ts
export const getConfig = async () => {
  return {
    render: 'dynamic',
  };
};
```

Clone the waku repo.
cd packages/waku
pnpm i
pnpm run compile

Copy or link the built package to the waku project. I just replace the dist folder in node_modules/waku/dist.

In the waku project, run a cloudflare build and run a local workerd server via wrangler to serve the built files locally.

npm run build -- --with-cloudflare
npx wrangler pages dev dist

open http://localhost:8788/about in your browser.

tylerlaws0n:
For this we could definitely add logic into the fs-router and build process for `--with-cloudflare` to polyfill `fs` when dynamic routes are present.

I'll defer to @dai-shi on whether or not that's an ideal approach or not with this though. I haven't looked into the plugins a ton, but maybe this would be considered a responsibility of a cloudflare plugin.

rmarscher:
Thanks for giving it a look. I'm not sure polyfilling `fs` is necessary... it seems the `platformObject.buildData.fsRouterFiles` should be populated. If the build process could pass the fsRouterFiles to the build output vite plugin, it could write to a source file during build that can be imported at runtime and added to the platformObject via the platform server. 

For example, the platformObject here - https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts#L86 - would need buildData.fsRouterFiles to be set. It could serialize that to a json file in the worker code bundle. That json file would be imported and passed to the waku server here https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/builder/serve-cloudflare.ts#L36-L38

[...]

PR #888, opened 2024-09-17
fix(deploy): move saving the build config to deploy plugins

rmarscher:
This (along with the previous #874) resolves the issue in #871. The build config was being written to the wrong location for cloudflare pages This PR exports a function from the build module that appends the build config to the given entries file location. It uses a new unstable build config state variable (`unstable_buildConfigSave`) to track whether or not the build config has be saved to the entries file. This allows the cloudflare deploy plugin to write the build config into the worker function directory instead of the root.

We could merge #879 and then I can edit the base branch to main and continue with this PR. Or however you want to handle it. Thank you!

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -776,8 +776,10 @@ export async function build(options: {
   await buildDeploy(rootDir, config);
   delete platformObject.buildOptions.unstable_phase;
 
-  await appendFile(
-    distEntriesFile,
-    `export const buildData = ${JSON.stringify(platformObject.buildData)};`,
-  );
+  if (existsSync(distEntriesFile)) {
+    await appendFile(
+      distEntriesFile,
+      `export const buildData = ${JSON.stringify(platformObject.buildData)};`,
+    );
+  }
 }
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -1,5 +1,6 @@
 import path from 'node:path';
 import {
+  appendFileSync,
   existsSync,
   mkdirSync,
   readdirSync,
@@ -11,7 +12,7 @@ import type { Plugin } from 'vite';
 
 import { unstable_getPlatformObject } from '../../server.js';
 import { SRC_ENTRIES } from '../constants.js';
-import { DIST_PUBLIC } from '../builder/constants.js';
+import { DIST_ENTRIES_JS, DIST_PUBLIC } from '../builder/constants.js';
 
 const SERVE_JS = 'serve-cloudflare.js';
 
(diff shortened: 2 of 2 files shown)


packages/waku/src/lib/builder/build.ts
```diff
@@ -64,6 +64,7 @@ import { deployCloudflarePlugin } from '../plugins/vite-plugin-deploy-cloudflare
 import { deployDenoPlugin } from '../plugins/vite-plugin-deploy-deno.js';
 import { deployPartykitPlugin } from '../plugins/vite-plugin-deploy-partykit.js';
 import { deployAwsLambdaPlugin } from '../plugins/vite-plugin-deploy-aws-lambda.js';
+import { appendFileSync } from 'node:fs';
```
rmarscher, line 67: The other vite deploy operations are sync which is why this is not using a promise.

dai-shi:
> The build config was being written to the wrong location for cloudflare pages

Yeah, I was also curious about it. Good that you confirm.

packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
```diff
@@ -208,14 +209,18 @@ export default {
         force: true,
       });
 
+      unstable_writeBuildConfig(
+        path.join(outDir, WORKER_JS_NAME, DIST_ENTRIES_JS),
+      );
```
dai-shi, line 214: I don't like the complexity and exposing the function from build.ts (which doesn't work if you own the plugin outside waku repo.)

Let's simply do this here:
```suggestion
      appendFileSync(
        path.join(outDir, WORKER_JS_NAME, DIST_ENTRIES_JS),
        `export const buildData = ${JSON.stringify(platformObject.buildData)};`,
      );
```

We will revisit the build process in the future anyway.

rmarscher: OK. A library of helper functions for deploy authors to use would be nice. The "isNodeCompatible" replacement is another candidate for a helper function:

```ts
      config.ssr.target = 'webworker';
      config.ssr.resolve ||= {};
      config.ssr.resolve.conditions ||= [];
      config.ssr.resolve.conditions.push('worker');
      config.ssr.resolve.externalConditions ||= [];
      config.ssr.resolve.externalConditions.push('worker');
```

rmarscher:
@dai-shi I attempted to move the updating of entries.js with build data to each deploy plugin. I tested with cloudflare and it works great. Implementation in the other deploy plugins needs testing.

packages/waku/src/lib/builder/build.ts
```diff
@@ -774,10 +774,4 @@ export async function build(options: {
 
   platformObject.buildOptions.unstable_phase = 'buildDeploy';
   await buildDeploy(rootDir, config);
-  delete platformObject.buildOptions.unstable_phase;
-
-  await appendFile(
-    distEntriesFile,
-    `export const buildData = ${JSON.stringify(platformObject.buildData)};`,
-  );
```
dai-shi, line 782: Please keep this because this is required even without deploy plugins.
Only change cloudflare deploy plugin. I don't know the future, but my understanding is only cloudflare has special requirement for the dist folder structure.

dai-shi: Wouldn't it be the same previously?
I'm fine to add a check:
```js
if (existsSync(distEntriesFile)) {
  await appendFile(
    distEntriesFile,
    `export const buildData = ${JSON.stringify(platformObject.buildData)};`,
  );
}
```

rmarscher: Ah... good idea.

rmarscher: @dai-shi OK great. I amended this PR.

PR #891, opened 2024-09-18
feat(hono): enable context storage

dai-shi:
Instead of #852 and #884, let's use Hono's `getContext` instead.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -4,6 +4,7 @@ import { pathToFileURL } from 'node:url';
 import { parseArgs } from 'node:util';
 import { createRequire } from 'node:module';
 import { Hono } from 'hono';
+import { contextStorage } from 'hono/context-storage';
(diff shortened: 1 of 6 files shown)


rmarscher:
When testing this, I receive the "Context is not available" error when trying to access the hono context within a server component.

I was able to figure out a way to update the dev server plugin to work with the hono context. That is available. I'm not so sure about this solution... I have to reconstruct the hono app to pass the updated env and ctx to the fetch handler. The hono context is immutable once created.

Hono has an adapter system for their vite plugin. The cloudflare adapter returns the env and executionContext and they mix that in when invoking fetch: https://github.com/honojs/vite-plugins/blob/main/packages/dev-server/src/dev-server.ts#L124-L137

I created a new PR with a new fetch adapter option for the dev server. I think this will be much more user-friendly for cloudflare devs than copy and pasting a custom plugin. #892

dai-shi:
Assuming this PR itself is fine, let's merge.

PR #892, opened 2024-09-18
feat(dev): cloudflare dev server adapter

rmarscher:
Related to #891. This is a better implementation than using a waku plugin.

See https://github.com/honojs/vite-plugins/blob/main/packages/dev-server/src/dev-server.ts#L124-L137 for similar work.

Here is the plugin implementation for comparison - https://gist.github.com/rmarscher/9bb6ed54dc9535f4b81bed147204c7e9


dai-shi:
Oookay, I get that waku middleware system doesn't work for this case (because waku middleware is to hide Hono.)

The solution in this PR looks nice DX-wise, however, it doesn't fit well with the minimal mental model in Waku, and there would be some issue with (peer) dependencies. Let me try another approach.

rmarscher:
Closing for #893.

PR #893, opened 2024-09-19
add: unstable_honoEnhancer in config

dai-shi:
alternate approach to #892.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -94,6 +94,7 @@ if (values.version) {
 
 async function runDev() {
   const config = await loadConfig();
+  const honoEnhancer = config.unstable_honoEnhancer || ((app) => app);
   const app = new Hono();
   app.use('*', runner({ cmd: 'dev', config, env: process.env as any }));
   app.notFound((c) => {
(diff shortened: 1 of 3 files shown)


dai-shi:
@rmarscher Can you try this and see if it works for your case?

packages/waku/src/cli.ts
```diff
@@ -149,7 +152,7 @@ async function runStart() {
     return c.text('404 Not Found', 404);
   });
   const port = parseInt(values.port || '8080', 10);
-  await startServer(app, port);
+  await startServer(honoEnhancer(app), port);
```
dai-shi, line 155: Do we need it for `runStart`?

rmarscher: It could be useful for `runStart`. Maybe the `cmd` and `config` should be additional arguments to `honoEnhancer` too.

dai-shi: `config` feels somewhat recursive. `cmd` seems good, but I actually removed it from `middleware` before, so let's leave it until we see real issues.

dai-shi: There's another concern: loadConfig will be defined in dist/entries.js, so it's dual modules, which may cause something unexpected.

rmarscher (approved):
I confirmed this works for the cloudflare dev server use case. LGTM. :+1: Thank you!

packages/waku/src/cli.ts
rmarscher: OK. No problem. This is not an issue for any of my current use cases.

PR #896, opened 2024-09-21
error in error handling

pmelab:
The error boundary does not render `html` or `body` tags and therefore causes and error itself.


packages/waku/src/router/client.ts
```diff
@@ -551,6 +551,14 @@ export function ServerRouter({
   );
 }
 
+function renderError(message: string) {
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('h1', null, message)),
+  );
+}
```
dai-shi, line 560: This doesn't feel like a final solution, but it looks good to me for now at least.

dai-shi (approved):
Thanks for your contribution!

PR #897, opened 2024-09-22
Update cloudflare guide to use importHonoContextStorage

rmarscher:
Update cloudflare guide to use bundled hono context storage rather than importing directly.

### DIFF
--- a/docs/guides/cloudflare.mdx
+++ b/docs/guides/cloudflare.mdx
@@ -39,10 +39,23 @@ Waku currently uses [Hono](https://hono.dev/) in the [advanced mode worker funct
 Waku uses Hono's [context storage middleware](https://hono.dev/docs/middleware/builtin/context-storage) to make the [Hono context](https://hono.dev/docs/api/context) accessible from anywhere. You can access Cloudflare bindings and execution context from the Hono context:
 
 ```ts
-import { getContext } from 'hono/context-storage';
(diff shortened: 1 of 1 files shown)


docs/guides/cloudflare.mdx
```diff
@@ -39,10 +39,23 @@ Waku currently uses [Hono](https://hono.dev/) in the [advanced mode worker funct
 Waku uses Hono's [context storage middleware](https://hono.dev/docs/middleware/builtin/context-storage) to make the [Hono context](https://hono.dev/docs/api/context) accessible from anywhere. You can access Cloudflare bindings and execution context from the Hono context:
 
 ```ts
-import { getContext } from 'hono/context-storage';
+import { importHonoContextStorage } from 'waku/unstable_hono';
```
dai-shi, line 42: Oh, I see. It doesn't work with cloudflare case either.
Well, as you may guess, we don't like this. It's far better to request users to install `hono`, which should allow the normal one.

rmarscher: Thank you! I'm looking forward to trying #898. It is interesting how you used a dynamic import and an empty virtual module to fix it.

dai-shi: It was actually not very nice. I'm trying a different workaround in #899.

rmarscher:
I confirmed this is resolved with the upcoming deploy plugin and build changes. Closing this PR.

PR #899, opened 2024-09-25
fix(deploy): deno

dai-shi:
following up #880 and others.


dai-shi:
I think I should split the PR for non deno related change.

dai-shi:
> I think I should split the PR for non deno related change.

#902

PR #900, opened 2024-09-25
fix: waku build will bundle browser files in server side

himself65:
Default options for vite includes `browser`, whch might cause some bundler issue

Ref: https://v2.vitejs.dev/config/#resolve-conditions
Downstream: https://github.com/run-llama/LlamaIndexTS/pull/1256


packages/waku/src/lib/builder/build.ts
```diff
@@ -252,6 +252,9 @@ const buildServerBundle = async (
     esbuild: {
       jsx: 'automatic',
     },
+    resolve: {
+      conditions: ['import', 'module', 'default', 'production'],
```
dai-shi, line 256: Shouldn't we remove 'production'? Isn't it automatically added?

packages/waku/src/lib/builder/build.ts
```diff
@@ -374,6 +380,9 @@ const buildClientBundle = async (
       rscTransformPlugin({ isClient: true, isBuild: true, serverEntryFiles }),
       ...deployPlugins(config),
     ],
+    resolve: {
+      conditions: ['import', 'module', 'default', 'browser', 'production'],
```
dai-shi, line 384: Is it intentional to have 'default' before 'browser'?

packages/waku/src/lib/builder/build.ts
himself65: Im not sure to be honest

dai-shi: https://github.com/vitejs/vite/blob/6f030ec15f25a2a1d7d912f1b84d83ebb28a3515/packages/vite/src/node/plugins/resolve.ts#L1123-L1140

seems to work anyway. Let's remove `'production'` from our code.

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -112,6 +112,9 @@ const createMainViteServer = (
         rscHmrPlugin(),
         fsRouterTypegenPlugin(config),
       ],
+      resolve: {
+        conditions: ['import', 'module', 'default'],
```
dai-shi, line 116: MainViteServer is used to bundle for browsers.
It should be in `ssr.resolve.conditions`.
Do we need to change RscViteServer?

Now I wonder if `build.ts` change is correct.

dai-shi (commented):
Can we avoid committing `dist`? It's listed in `.gitignore`.

packages/waku/src/lib/middleware/dev-server-impl.ts
himself65: I tested, `ssrLoadModule` won't use `browser` condition by default

dai-shi: So, what's the original problem?
This MainViteServer it to for browsers, so `browser` condition should be appropriate.

dai-shi:
Closing this as stale for now, but I'm not sure if I understand the problem, so feel free to open a discussion.

Issue #845, opened 2024-08-20
update `@swc/core` package


dai-shi:
If we use v1.7.14, it causes "Error: missing field `ctxt`" with tests.
Can anyone investigate it please?

PR #901, opened 2024-09-25
chore: bump `swc`

himself65:
Fixes: https://github.com/dai-shi/waku/issues/845


packages/waku/package.json
```diff
@@ -88,7 +88,7 @@
   },
   "devDependencies": {
     "@netlify/functions": "^2.8.1",
-    "@swc/cli": "^0.4.0",
+    "@swc/cli": "0.4.1-nightly.20240914",
```
dai-shi, line 91: Does it conflict if we keep using `0.4.0`?

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -71,17 +71,26 @@ export ${name === 'default' ? name : `const ${name} =`} createServerReference('$
...
+    start: 0,
+    end: 0,
+  }) as swc.Span;
+
 const createIdentifier = (value: string): swc.Identifier => ({
   type: 'Identifier',
   value,
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-expect-error
+  ctxt: 0,
```
dai-shi, line 85: Hm, we need it in this position... So, if swc fixes the TS type, the `@ts-expect-error` will be gone?

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -71,17 +71,26 @@ export ${name === 'default' ? name : `const ${name} =`} createServerReference('$
...
   optional: false,
-  span: { start: 0, end: 0, ctxt: 0 },
+  span: createEmptySpan(),
```
dai-shi, line 87: Just to confirm, does it work if this is `span: { start: 0, end: 0 }`? Or, does TS complain?

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
himself65: yes

packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
himself65: span: { start: 0, end: 0 }. swc type is broken

dai-shi: How is it likely that swc will fix this type issue sometime soon-ish?

himself65: /cc @kdy1 @magic-akari

himself65:
well, they recommend write rust

himself65:
not sure If I should put rust version of parser to a higher prority.

dai-shi:
I don't think it's a priority, unless it's the bottleneck of the performance.

dai-shi:
our final goal would be rolldown, maybe?
meanwhile, we can keep using swc from js, or migrate to esbuild??

packages/waku/package.json
dai-shi: oh, it is the `latest`.

PR #902, opened 2024-09-26
fix: nonjs-resolve plugin

dai-shi:
there was a huge misunderstanding.


packages/waku/src/lib/plugins/vite-plugin-nonjs-resolve.ts
```diff
@@ -7,17 +7,16 @@ export function nonjsResolvePlugin(): Plugin {
   return {
     name: 'nonjs-resolve-plugin',
     async resolveId(id, importer, options) {
-      if (!id.endsWith('.js')) {
-        return id;
```
dai-shi, line 11: We shouldn't resolve it because other plugins can't resolve it. We should return `undefined`.

PR #903, opened 2024-09-26
feat: enable compression

sarat1669:
Text compression is not enabled in hono.
When we run a light house score on the sample app generated, we get 95 score.
Screenshot 2024-09-26 at 16 09 15

Enabling compression along with the meta description fix in #904 gives 100/100.


dai-shi:
Hi, can you add the PR description with some details?

dai-shi:
Thanks for the suggestion!

packages/waku/src/cli.ts
```diff
@@ -144,6 +145,7 @@ async function runStart() {
     import(pathToFileURL(path.resolve(distDir, DIST_ENTRIES_JS)).toString());
   const app = new Hono();
   app.use(contextStorage());
+  app.use(compress())
```
dai-shi, line 148: It looks like a good addition, it's not 100% certain if this won't cause any issues for all users.
So, my suggestion is to introduce a new option `--experimental-compress` and use it explicitly.
We can also enable it for the dev server too.

dai-shi (changes requested):
Can you make a change?

packages/waku/src/cli.ts
sarat1669: Done

PR #904, opened 2024-09-26
fix: meta description

sarat1669:
Description meta tags used in the scaffolded apps is wrong as per Google Light House.

`<meta property="description" content="The minimal React framework" />`
Should be 
`<meta name="description" content="The minimal React framework" />`

This along with #903 gives perfect light house score


dai-shi:
Hi, can you add the PR description with some details?

dai-shi (commented):
It looks good to me.

Let's see how @sandren takes it.

sandren (approved):
Nice catch! Thanks. 🙂

PR #910, opened 2024-09-27
feat: deploy aws / add streaming

aheissenberger:
adds streaming by using [`streamHandle`](https://hono.dev/docs/getting-started/aws-lambda#lambda-response-streaming) from Hono.

To switch the handler use the environment variable `DEPLOY_AWS_LAMBDA_STREAMING=true`:
```sh
DEPLOY_AWS_LAMBDA_STREAMING=true pnpm build --with-aws-lambda
```

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -21,7 +21,7 @@ import { runner, importHono, importHonoNodeServerServeStatic, importHonoAwsLambd
 
 const { Hono } = await importHono();
 const { serveStatic } = await importHonoNodeServerServeStatic();
-const { ${lambda_streaming ? 'streamHandle as' : ''} handle } = await importHonoAwsLambda();
+const { ${lambda_streaming ? 'streamHandle:' : ''}handle } = await importHonoAwsLambda();
 let contextStorage;
 try {
  ({ contextStorage } = await import('hono/context-storage'));
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -21,7 +21,7 @@ import { runner, importHono, importHonoNodeServerServeStatic, importHonoAwsLambd
 
 const { Hono } = await importHono();
 const { serveStatic } = await importHonoNodeServerServeStatic();
-const { ${lambda_streaming ? 'streamHandle:' : ''}handle } = await importHonoAwsLambda();
+const { ${lambdaStreaming ? 'streamHandle:' : ''}handle } = await importHonoAwsLambda();
 let contextStorage;
 try {
  ({ contextStorage } = await import('hono/context-storage'));
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -8,7 +8,7 @@ import { DIST_PUBLIC } from '../builder/constants.js';
 
 const SERVE_JS = 'serve-aws-lambda.js';
 
-const lambda_streaming = process.env.DEPLOY_AWS_LAMBDA_STREAMING === 'true';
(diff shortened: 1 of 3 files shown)


aheissenberger:
@rmarscher please help - it returns the wrong result in my tests with a function url and sst

sst.config.ts:
```ts
/// <reference path="./.sst/platform/config.d.ts" />

export default $config({
  app(input) {
    return {
      name: "waku03demo",
      removal: input?.stage === "production" ? "retain" : "remove",
      home: "aws",
    };
  },
  async run() {
    const WakuDemoApp = new sst.aws.Function("WakuDemoApp", {
      url: true,
      streaming: true,
      //timeout: "15 minutes",
      handler: "dist/serve-aws-lambda.handler",
      bundle: "bundle", // disable bundling with esbuild
      copyFiles: [
        {
          from: "dist"
        },
        {
          from: "private"
        }
      ],
      environment: {
        NODE_ENV: "production",
      },
    });
    return {
      api: WakuDemoApp.url,
    };
  },
});
```

The result is this - what suggests that the return is not a stream from WAKU:
```
{"body":"<!DOCTYPE html>
<html>
<head>
..
","headers":{"content-type":"text/html; charset=utf-8"},"statusCode":200,"isBase64Encoded":false}
```

aheissenberger:
@rmarscher it works now

please provide some input why you need to rearrange the dist folder. The sst setup above is only the shortest version to test streaming, other setups should not be a problem.

rmarscher:
Very cool. Thanks for giving SST a try. I rearrange the files to move unnecessary public files out of the function bundle - to reduce the lambda package size.

rmarscher:
Here is a demo I made for how I use SST with waku with an older patch of v0.21.1 - https://github.com/rmarscher/waku-sst-demo 
It uses some core functions they have in a "SsrSite" module that all of their official meta framework components import. I need to update that too. Both SST and waku are moving pretty fast right now. SST's TanstackStart component is pretty similar for reference https://github.com/sst/ion/blob/dev/platform/src/components/aws/tanstack-start.ts.

aheissenberger:
@rmarscher in the serverless framework I only copy the needed folders to the lambda and all static files go to an s3 bucket. In Cloudfront I use behaviors to handle the forwarding of static requests and the default request are handled by the lambda function. I know this is not as fancy as using a Cloudfunction to route all static assets but deployments in non `US East (N. Virginia)` get quit complex if you need to deploy on project to multiple regions as the Cloundfunction have to be deployed to `US East (N. Virginia)`. Yes, this is much simpler with sst ;-)

It should be possible keep the current file structure with the sst framework. Looking at TanstackStart Stack I suggest that @dai-shi asks directly asks for support in creating the stack. Tanstack sst stack was created by [Frank Wang](https://github.com/fwang) the founder of sst.

@rmarscher if you agree, I would finish this PR as creating an sst stack should be in its own PR.

dai-shi:
> Looking at TanstackStart Stack I suggest that @dai-shi asks directly asks for support in creating the stack.

What should I do? Open an issue?

aheissenberger:
> > Looking at TanstackStart Stack I suggest that @dai-shi asks directly asks for support in creating the stack.
> 
> What should I do? Open an issue?

yes, and maybe ping Frank to have a look at WAKU as one of the best alternatives to Next.js

dai-shi:
https://github.com/sst/ion/issues/1159

packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
```diff
@@ -8,6 +8,8 @@ import { DIST_PUBLIC } from '../builder/constants.js';
 
 const SERVE_JS = 'serve-aws-lambda.js';
 
+const lambda_streaming = process.env.DEPLOY_AWS_LAMBDA_STREAMING === 'true';
```
dai-shi, line 11: just naming convention:
```suggestion
const lambdaStreaming = process.env.DEPLOY_AWS_LAMBDA_STREAMING === 'true';
```

dai-shi (commented):
looks good other than that. needs to run prettier?

aheissenberger:
> looks good other than that. needs to run prettier?

Done

dai-shi (approved):
LGTM!

PR #911, opened 2024-09-27
Update cloudflare with hono/context-storage external build fix

rmarscher:
re: https://github.com/dai-shi/waku/pull/882#issuecomment-2378365279
Applies the same fix used for handling hono/context-storage with aws builds.

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -95,6 +95,11 @@ export function deployCloudflarePlugin(opts: {
       rootDir = config.root;
       entriesFile = `${rootDir}/${opts.srcDir}/${SRC_ENTRIES}`;
       const { deploy, unstable_phase } = platformObject.buildOptions || {};
+      if (deploy === 'cloudflare' && Array.isArray(config.ssr.external)) {
+        config.ssr.external = config.ssr.external.filter(
+          (item) => item !== 'hono/context-storage',
+        );
+      }
       if (
         (unstable_phase !== 'buildServerBundle' &&
           unstable_phase !== 'buildSsrBundle') ||
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -100,6 +100,11 @@ export function deployCloudflarePlugin(opts: {
       rootDir = config.root;
       entriesFile = `${rootDir}/${opts.srcDir}/${SRC_ENTRIES}`;
       const { deploy, unstable_phase } = platformObject.buildOptions || {};
+      if (deploy === 'cloudflare' && Array.isArray(config.ssr.external)) {
+        config.ssr.external = config.ssr.external.filter(
+          (item) => item !== 'hono/context-storage',
+        );
+      }
       if (
         (unstable_phase !== 'buildServerBundle' &&
           unstable_phase !== 'buildSsrBundle') ||
@@ -117,9 +122,6 @@ export function deployCloudflarePlugin(opts: {
     resolveId(source) {
       if (source === `${opts.srcDir}/${SERVE_JS}`) {
         return source;
-      }
-      if (source === 'hono/context-storage') {
-        return { id: source, external: true };
       }
     },
     load(id) {
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -17,21 +17,16 @@ import { DIST_ENTRIES_JS, DIST_PUBLIC } from '../builder/constants.js';
 const SERVE_JS = 'serve-cloudflare.js';
 
 const getServeJsContent = (srcEntriesFile: string) => `
-import { runner, importHono } from 'waku/unstable_hono';
+import { runner, importHono, importHonoContextStorage } from 'waku/unstable_hono';
 
 const { Hono } = await importHono();
-let contextStorage;
-try {
- ({ contextStorage } = await import('hono/context-storage'));
-} catch {}
+const { contextStorage } = await importHonoContextStorage();
 
 const loadEntries = () => import('${srcEntriesFile}');
 let serveWaku;
 
 const app = new Hono();
-if (contextStorage) {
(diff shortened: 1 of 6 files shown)


packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
```diff
@@ -118,9 +123,6 @@ export function deployCloudflarePlugin(opts: {
       if (source === `${opts.srcDir}/${SERVE_JS}`) {
         return source;
       }
-      if (source === 'hono/context-storage') {
-        return { id: source, external: true };
-      }
```
dai-shi, line 123: Hmm, actually this is effective for all the case. Let me see if I can fix.

dai-shi:
Actually, @aheissenberger was right. It was my bad.
#894 is good for `waku start`, but #898 was wrong.

dai-shi (commented):
I added more commits, but it should be fine.
If there are still some issues, feel free to open PRs.

dai-shi (approved):

rmarscher:
Oh interesting. Either solution seemed to work from my perspective but I am not using `waku start` for any of my apps. Do I need to go back to importing `importHonoContextStorage` from `waku/unstable_hono` to get the context? I'll test and find out...

dai-shi:
Oh, no... I'm going the circle... Let me check it too.

dai-shi:
As far as I tried `npm run build -- --with-aws-lambda` with `08_cookies`, the `dist` folder looks good to me.

rmarscher:
Yeah, this did not work for me unless I update back to [import from unstable_hone](https://github.com/dai-shi/waku/pull/897/files#diff-002147acc7ea306b86bf950248a9c65c2e80f2a4cd9cc48a08867759c03f60a2R42). Then it works fine. Either is ok with me. It was nice with the version I originally committed here. I was able to bring my own hono (tested with 4.6.3).

rmarscher:
I will test AWS next. I only tested Cloudflare so far.

rmarscher:
Same issue in my aws build. getContext fails if I import from 'hono/context-storage' instead of 'waku/unstable_hono'.

dai-shi:
Maybe it's only `--with-cloudflare` issue. I'm not happy with `importHonoContextStorage` as a user facing api.

dai-shi:
Hmmm? I think I miss something. Can you try `08_cookies` example?

rmarscher:
When I copy 08_cookies App and Counter components into an empty page on my app, I get a fatal error

Error - Context is not available
Stack Trace
getContext (file:///var/task/components/test/App.js:12:11)
InternalAsyncComponent (file:///var/task/components/test/App.js:25:43)

The issue is caused because the entry point ("serve js") for the deploy plugins loads getContext() from waku/unstable_hono with the latest change.

It worked better IMHO when the deploy plugins imported from hono/context-storage and let the deploy server bundle include the app's copy of hono/context-storage rather than waku's. Then my app server components received the same bundled version of hono that I supplied.

dai-shi:
> The issue is caused because the entry point ("serve js") for the deploy plugins loads getContext() from waku/unstable_hono with the latest change.

From my investigation, importHonoContextStorage and hono/context points to the same file after the build.

Let me try it again.
Using waku `main` 49d1ae7, and `examples/08_cookies`:
```bash
pnpm -r --filter='./packages/waku' run compile
cd examples/08_cookies
node ../../packages/waku/dist/cli.js build --with-cloudflare
npx wrangler pages dev
```

Actually I get a different error (unexpected):
```
⎔ Starting local server...
✘ [ERROR] service core:user:waku-project: Uncaught Error: No such module "node:path".
```

But, that brings me a question, how is your setup different?

dai-shi:
> Actually I get a different error (unexpected):

Ah, `08_cookies` have `node:fs`. I just removed node related code in src/entries.tsx and replaced with `const items: unknown[] = [];`

`npx wrangler pages dev` is working fine.
image

rmarscher:
You're right. It does work if I deploy from examples/08_cookies directly but I think it is a side-effect of being in the same monorepo. If you change the hono version in 08_cookies to 4.6.3, you will be able to reproduce the error.

I thought I already tested using 4.6.1 from my other apps and it didn't work, but I could try again. Even if it does work, it's probably not an ideal solution.

This fixes:
```diff
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -17,16 +17,22 @@ import { DIST_ENTRIES_JS, DIST_PUBLIC } from '../builder/constants.js';
 const SERVE_JS = 'serve-cloudflare.js';

 const getServeJsContent = (srcEntriesFile: string) => `
-import { runner, importHono, importHonoContextStorage } from 'waku/unstable_hono';
+import { runner, importHono } from 'waku/unstable_hono';
+
+let contextStorage;
+try {
+ ({ contextStorage } = await import('hono/context-storage'));
+} catch {}

 const { Hono } = await importHono();
-const { contextStorage } = await importHonoContextStorage();

 const loadEntries = () => import('${srcEntriesFile}');
 let serveWaku;

 const app = new Hono();
-app.use(contextStorage());
+if (contextStorage) {
+  app.use(contextStorage());
+}
 app.use('*', (c, next) => serveWaku(c, next));
 app.notFound(async (c) => {
   const assetsFetcher = c.env.ASSETS;
```

rmarscher:
^ same patch works for aws for me too.

dai-shi:
> I think it is a side-effect of being in the same monorepo.

I think I tried it standalone too.
Just double checking; Yes, it works without monorepo.

> If you change the hono version in 08_cookies to 4.6.3, you will be able to reproduce the error.

Ah, that explains. Well, then I change my opinion. If you use a different hono version, `import { getContext } from 'hono/context-storage'` doesn't work, and `from 'waku/unstable_hono'` would be the right solution _for now_. Unless we make hono a peer dependency (which we decided not to), there wouldn't be other solutions. This actually makes me feel like #884 can be a better solution than #891. We should definitely revisit before v1.

dai-shi:
> We should definitely revisit before v1.

Or, I can revert it right away, because this hasn't been released yet.

dai-shi:
#914
sorry for going the full circle. 😄 @rmarscher @aheissenberger

rmarscher:
> #914 sorry for going the full circle. 😄 @rmarscher @aheissenberger

I see. No problem! That works. Thank you so much.

PR #912, opened 2024-09-28
feat: deploy / aws lambda / sst v3 waku stack

aheissenberger:
This is a first draft for a sst stack for WAKU.

Based on the fact, that there can be more than 25 folders (AWS Cloudfront routing limit) on the first level of the public folder only the folder `public/images` and `public/assets` are copied to s3 and forward by CloudFront. 

A possible, but more complex solution could handle this by adding a Cloudfunction to Cloudfront which knows all routes and can forward all existing static root level routes.


aheissenberger:
@rmarscher this is based as suggested on the sst tanstack-start stack

rmarscher:
> @rmarscher this is based as suggested on the sst tanstack-start stack

Thank you for creating this! I hope it helps it get added upstream to SST. 🙏

PR #913, opened 2024-09-28
feat: deploy / aws-lambda / sst waku stack

aheissenberger:
This is a first draft for a sst stack for WAKU.

Based on the fact, that there can be more than 25 folders (AWS Cloudfront routing limit) on the first level of the public folder only the folder public/images and public/assets are copied to s3 and forward by CloudFront.

A possible, but more complex solution could handle this by adding a Cloudfunction to Cloudfront which knows all routes and can forward all existing static root level routes.

### DIFF
--- a/docs/builder/examples/aws-lambda-sst/README.md
+++ b/docs/builder/examples/aws-lambda-sst/README.md
@@ -1,0 +1,31 @@
+# README
+
+## Setup
+
+1. init sst `pnpx sst@latest init`
+2. replace the generated `./sst.config.ts` with `sst.config.ts`from this folder
+3. copy the file `waku.ts` to
+   `.sst/platform/src/components/aws/waku.ts`
+4. change to include the deployment adapter `"build" : "waku build --with-aws-lambda"`
+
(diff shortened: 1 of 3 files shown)


aheissenberger:
@rmarscher this is based as suggested on the sst tanstack-start stack

rmarscher:
> @rmarscher this is based as suggested on the sst tanstack-start stack

It looks great on a quick read. I'll try to test it next week. I wonder if we can get them to accept it upstream...

rmarscher:
I think SST refactored their SsrSite dev server code since this example was created. They might be moving too fast for this to be worthwhile to include here. They need a public SsrSite component that's easier to extend.

I also think it's a little bit better to move the public files out of the lambda bundle. I think we discussed that elsewhere. Waku doesn't have an option to output separate server dist folder from public dist folder. The public folder is inside the dist folder because the server code references some of the public assets. In the Cloudflare adapter, I separate them out. I'd like to make that a more standard build option that our adapters can opt into. We can refactor the build options to have more flexibility. Maybe I can find time to suggest some general build config edits in a PR. https://github.com/dai-shi/waku/discussions/806

dai-shi:
closing as stale. waku's adapter systems has changed quite a lot. feel free to do some experiments.

PR #917, opened 2024-09-29
feat(router): props inferred for page components

tylersayshi:
Happy to move things around as needed :)

This change adds the Page component prop support

image

image


packages/waku/src/router/common.ts
```diff
@@ -1,5 +1,5 @@
-export type RouteProps = {
-  path: string;
+export type RouteProps<Path extends string = string> = {
+  path: Path;
   query: string;
   hash: string;
```
tylerlaws0n, line 4: @dai-shi would `hash: string | undefined` be correct here since these could be used on the server?

Or should undefined only be added to hash for `PageProps`?

tylerlaws0n:
@dai-shi do you have recommendations for which docs to update for this change and which examples are updated to use this?

packages/waku/src/router/common.ts
dai-shi: `RouteProps` defined in common.ts is for both server & client.
`hash` is added afterwards, which only exists on the client.
So, there's already a mismatch, which should be improved.
For now, we should omit `hash` for `PageProps` (not `undefined`).

dai-shi: wait, i think that's my misunderstanding.
`hash` on the server is always `''`, so the type is correct.
we don't need to omit `hash`.

Can you confirm it?

tylerlaws0n: So when running the dev server with examples/10_fs-router the hash is not included on the client or the server


```
// from console.log(JSON.stringify(props))
{"path":"/","query":"","children":{"key":null,"props":{},"_owner":null,"_store":{}}}
```

tylerlaws0n: changing the page to `render: dynamic` had no impact either... Is this a bug?

dai-shi: thanks for confirming. sounds like i need to double check the source code. I'm probably confused with something else (it's probably client code runs on the server in SSR.)

So, I think my original comment is correct at least for now. 👇 


> For now, we should omit `hash` for `PageProps` (not `undefined`).

tylerlaws0n: sounds good, omitted 👍

docs/create-pages.mdx
```diff
@@ -68,6 +68,7 @@ const pages = createPages(async ({ createPage, createLayout }) => [
 declare module 'waku/router' {
   interface RouteConfig {
     paths: PathsForPages<typeof pages>;
+    pages: typeof pages;
```
dai-shi, line 71: So, I'm not sure if you remember our discussion before, but this is going against the abstraction (layers) of waku/router.

#598 "Waku middleware / API stack (v0.20)" may or may not help.

At the `defineRouter` layer, it shouldn't use the notion from `createPages`.
So, `createPages` can depend on `defineRouter`, but not the other way around.

That's why we did `paths: PathsForPages<typeof pages>;` previously.

tylerlaws0n: ah, ya sorry. I remember that discussion now. I had a bit of tunnel vision on writing this out, I'll fix that to be similar to how we did for paths, but for the full page types.

If we do something like 
```ts
pages: GetPages<typeof pages>
```
should I then remove `paths: PathsForPages<typeof pages>` since that can then be inferred from the full pages above internally?

This would get us back down to only one item in RouteConfig and also fall in line with the spirit of that last discussion I think

dai-shi: I don't prefer using the word or notion of "page", which implies `createPages`.
`paths: PathsForPages<typeof pages>` was really good. defineRouter is lower level of createPages. There can exist another implementation instead of createPages that depends on defineRouter. The difficulty is we don't know what it is yet. But, we would try out best. At least defineRouter is lower level and should be minimal.

`getFoo<typeof foo>` doesn't make sense.

tylerlaws0n: I think I might be able to actually just use just the paths that is already there... 🤔

tylerlaws0n: ok, updated to extend the existing use of paths internally, but leave the external `paths: PathsForPages` the same

let me know what you think 😄

dai-shi: Yeah, looks much better.

examples/07_router/src/entries.tsx
```diff
@@ -61,7 +61,7 @@ const pages = createPages(async ({ createPage, createLayout }) => [
     render: 'static',
     path: '/nested/[id]',
     staticPaths: ['foo', 'bar'],
-    component: ({ id }: { id: string }) => (
+    component: ({ id }) => (
```
dai-shi, line 64: Wow, niceeee!

packages/waku/src/router/base-types.ts
```diff
@@ -1,6 +1,13 @@
...
+} from './create-pages-utils/inferred-path-types.js';
+
 export type { PathsForPages } from './create-pages-utils/inferred-path-types.js';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
 export interface RouteConfig {
   // routes to be overridden by users
 }
+
+export type PageProps<Path extends PagePath<RouteConfig>> = PropsForPages<Path>;
```
dai-shi, line 13: This util is specifically for `createPages`, right? Can you add a comment to clarify it?

tylerlaws0n: added the comment. 

I looked at `defineEntries` and that could also use this same type system in theory. Happy to work on that soon if you think that'd be helpful.

docs/create-pages.mdx
dai-shi: So, this line should be removed from the doc?

packages/waku/src/router/base-types.ts
dai-shi: I'm not sure what it's about `defineEntries`. If it's just simply improve type inference, it would be nice. If it's about manual type annotation, it's not very important for now.

tylerlaws0n: I think both might be easy enough to look into. I think most of the hard work was done already for `createPages` and might be easily shareable with other things like `defineEntries`.

dai-shi: > might be easily shareable with other things like `defineEntries`.

idea can be shareable, but not the code. I don't think manual annotation for `defineEntries` is necessary, just type inference would be good.

docs/create-pages.mdx
tylerlaws0n: yep! sorry, forgot about this, thanks

packages/waku/src/router/base-types.ts
```diff
@@ -1,6 +1,14 @@
+import type {
+  PagePath,
+  PropsForPages,
+} from './create-pages-utils/inferred-path-types.js';
+
 export type { PathsForPages } from './create-pages-utils/inferred-path-types.js';
```
dai-shi, line 6: Can we add the similar comment as L13 please?

packages/waku/src/router/client.ts
```diff
@@ -33,7 +33,9 @@ import {
 import type { RouteProps, ShouldSkip } from './common.js';
 import type { RouteConfig } from './base-types.js';
 
-type InferredPaths = RouteConfig extends { paths: infer UserPaths }
+type InferredPaths = RouteConfig extends {
+  paths: { expanded: infer UserPaths };
```
dai-shi, line 37: I don't particularly like this change. `expanded` is anything meaningful here.
As I understand, a change here is necessary to simplify the type overload.
If we can't avoid it, how about this?
 ```suggestion
   paths: { inferred: infer UserPaths };
```

**Review by dai-shi — COMMENTED** [review]
If you have any other solution, it might be considerable.

File: packages/waku/src/router/client.ts (hunk 6 above)
Comment (tylerlaws0n) [reply packages/waku/src/router/client.ts:37]: so there is also this

```ts
export type PagePath<Config> = Config extends {
  paths: {
    initial: infer AllPages;
  };
}
  ? AllPages
  : never;
```

the key thing is that I need to differentiate the inferred routes that are with the slugs and the ones with the slugs replaced with staticPaths or `string`. This is why I went with expanded and initial: 

```ts
export type PathsForPages<PagesResult extends { DO_NOT_USE_pages: AnyPage }> = {
  /** All possible paths with string or static paths in place of slug */
  expanded: CollectPaths<PagesResult['DO_NOT_USE_pages']> extends never
    ? string
    : CollectPaths<PagesResult['DO_NOT_USE_pages']> & {};
  /** Paths with slugs as string literals */
  initial: PagesResult['DO_NOT_USE_pages']['path'];
};
```

I am happy to rename these, but I fear `inferred` is even more vague than the current names since they are both technically inferred types from the base of `RouteConfig`

dai-shi: `PagePath` is for `createPages`, while `InferredPaths` is for `defineRouter` which isn't tied to `ceatePages`.

`PathsForPages` is creating for both and my concern is just naming, because `defineRouter` is designed to be used with something else other than `createPages`, currently.

tylerlaws0n: Yeah, I share that concern... I'll think on new names 🤔, but please feel free to suggest other ideas if you have them :)

dai-shi: > to simplify the type overload.

Maybe, I prefer separating overloads. (Your original idea.)

```ts
    interface RouteConfig {
      paths: PathsForPages<typeof _pages>;
      something: SomethingForPages<typeof _pages>;
    }
```

They are for different purposes and consumed differently. So, explicit separation isn't that bad. Besides, this code will usually auto generated.

dai-shi: Or, maybe this is more correct:
```ts
    interface RouteConfig {
      paths: PathsForPages<typeof _pages>;
    }
    interface PagesConfig {
      pages: typeof _pages;
    }
```

tylerlaws0n: I like this second example as well. Having two interfaces is a downside, but having the flexibility to define other types of router definitions like defineEntries later with this pattern shows the benefit to me.

dai-shi (approved):
LGTM
Ready to merge?

tylersayshi:
Yep! @dai-shi

PR #920, opened 2024-09-30
ci: use hoisted mode for e2e test

himself65:
hoisted mode will make e2e test has same node_modules sturcture with npm


himself65 (commented):
Seems like file id for RSC part is invalid on windows?

himself65:
https://github.com/dai-shi/waku/pull/1087 this PR refactor the install step using npm so hoisted mode will doesn't matter now I think

himself65:
anyway it's better to use hoisted mode in CI so keep away from regression

Issue #885, opened 2024-09-16
(router) improve static pages


dai-shi:
## Background

The two rendering types "static" and "dynamic" are very complicated because there are two aspects RSC and SSR.

This `isStatic` is for RSC:
https://github.com/dai-shi/waku/blob/9ca5161cfba93e402641249a494642a912d25795/packages/waku/src/server.ts#L16

This `isStatic is for SSR:
https://github.com/dai-shi/waku/blob/9ca5161cfba93e402641249a494642a912d25795/packages/waku/src/server.ts#L12

Currently, isStatic for SSR will be true if all of isStatic for RSC are true.

## Expectation

If we mark a "page" or a "layout" to be "static", our expectation is that the rendering only happens just once. (For PRD, only at the build time, and for DEV only the first time and again after hot reload.)

## Actual problem

If we mix "static" and "dynamic" for RSC, the page for SSR will be "dynamic" which is expected, but then RSC's "static" isn't truly static.

In the `01_template`, modify _layout.tsx to be "dynamic", while keeping "static" in pages.tsx. Now, HomePage in page.tsx renders more than expected. In addition to the build time, it renders on production. It only behaves like "static" on client routing thanks to `shouldSkip` mechanism.

## Solution

- As we now have `buildData` #860, we can store RSC text for "static" RSC, even if the SSR is dynamic.
- We use `buildData` in fs-router, so the similar implementation wouldn't be too difficult for PRD.

## Challenges

- In DEV, it would be nice if the behavior is identical to PRD, but we need to delete the RSC text cache (hm, where do we store it?) after hot reload.
- `shouldSkip` should be revisited as it's too complicated. (It's a separate issue and more important than this one.)
- We may have more types than "static" and "dynamic" in the future, but we can reconsider our design at that time. (Trial-and-error is welcome.)

dai-shi:
@tylerlaws0n here's the first write-up.

tylerlaws0n:
@dai-shi Are you suggesting that the static page with a dynamic layout should always become implicitly dynamic to account for the dynamic layout's ability to change?

I think it might help to talk through a scenario where you'd want this mixing of build modes between the layout and the page itself.

Without coming into this problem yet myself, I would think that these two things should simply not be mixed. I feel like if you dynamically update the static page, that is unexpected behavior. And if you statically render a dynamic layout, that is also unexpected.

[...]

PR #922, opened 2024-09-30
docs: add example of dynamic layout with static page

tylersayshi:
you can see the time change as navigating the site which shows that the layout is indeed dynamic

The rest of the home page stays unchanged and can be cached

Example for #885

### DIFF
--- a/examples/25_weave_render/src/entries.tsx
+++ b/examples/25_weave_render/src/entries.tsx
@@ -2,6 +2,8 @@ import { lazy } from 'react';
 import { createPages } from 'waku';
 import type { PathsForPages } from 'waku/router';
 import FooPage from './components/FooPage';
+import BarPage from './components/BarPage';
+import NestedBarLayout from './components/NestedBarLayout';
 
 // The use of `lazy` is optional and you can use import statements too.
 const HomeLayout = lazy(() => import('./components/HomeLayout'));
@@ -24,6 +26,18 @@ const pages = createPages(async ({ createPage, createLayout }) => [
     render: 'static',
     path: '/foo',
     component: FooPage,
+  }),
+
+  createPage({
+    render: 'dynamic',
+    path: '/nested/bar',
+    component: BarPage,
+  }),
+
+  createLayout({
+    render: 'static',
+    path: '/nested',
+    component: NestedBarLayout,
   }),
 ]);
 
--- a/examples/25_weave_render/src/entries.tsx
+++ b/examples/25_weave_render/src/entries.tsx
@@ -3,7 +3,7 @@ import { createPages } from 'waku';
 import type { PathsForPages } from 'waku/router';
 import FooPage from './components/FooPage';
 import BarPage from './components/BarPage';
-import NestedBarLayout from './components/NestedBarLayout';
+import BarLayout from './components/BarLayout';
 import FooLayout from './components/FooLayout';
 
 // The use of `lazy` is optional and you can use import statements too.
@@ -12,7 +12,7 @@ const HomePage = lazy(() => import('./components/HomePage'));
 
 const pages = createPages(async ({ createPage, createLayout }) => [
   createLayout({
-    render: 'dynamic',
+    render: 'static',
     path: '/',
     component: HomeLayout,
   }),
@@ -23,12 +23,6 @@ const pages = createPages(async ({ createPage, createLayout }) => [
     component: HomePage,
   }),
 
-  createPage({
-    render: 'dynamic',
-    path: '/foo',
-    component: FooPage,
-  }),
-
   createLayout({
     render: 'static',
     path: '/foo',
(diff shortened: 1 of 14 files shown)


examples/25_weave_render/src/entries.tsx
```diff
@@ -0,0 +1,36 @@
...
+    path: '/',
+    component: HomePage,
+  }),
+
+  createPage({
+    render: 'static',
+    path: '/foo',
+    component: FooPage,
+  }),
+]);
```
dai-shi, line 28: How about adding another route like `/nested/bar` which has static layout and dynamic page?

tylerlaws0n: added... there is weird behavior though

When I nav to /nested/bar then click the link to go back to home or foo, I get a white screen with no content. Is this a bug? Or did I make a mistake?

dai-shi: Does it happen both in DEV and PRD? Can be a bug.

dai-shi: Wait, `/nested/bar` still has a dynamic root layout?

I think we should make:
- `/foo`: static layout + dynamic page
- `/nested/bar`: static layout + dynamic layout + static page

what do you think?

tylerlaws0n: > Does it happen both in DEV and PRD? Can be a bug.

tylerlaws0n: > Does it happen both in DEV and PRD? Can be a bug.
Yes it happens both

tylerlaws0n: > static layout + dynamic layout + static page

How do I make two layouts on a route?

tylersayshi: ok, I changed it. let me know if that's what you meant though. I'm not super confident with it yet.

tylersayshi:
@dai-shi is your general experience with the _layout pattern positive? Or is it not something you use in practice very much?

examples/25_weave_render/src/entries.tsx
dai-shi: It looks different. Okay, we may not need `nested`. How about this?

```js
  createLayout({
    render: 'static',
    path: '/',
    component: HomeLayout,
  }),

  createPage({
    render: 'static',
    path: '/',
    component: HomePage,
  }),

  createLayout({
    render: 'static',
    path: '/foo',
    component: FooLayout,
  }),

  createPage({
    render: 'dynamic',
    path: '/foo',
    component: FooPage,
  }),

  createLayout({
    render: 'dynamic',
    path: '/bar',
    component: BarLayout,
  }),

  createPage({
    render: 'static',
    path: '/bar',
    component: BarPage,
  }),

  ```

**Comment by dai-shi** [comment]
> @dai-shi is your general experience with the _layout pattern positive? Or is it not something you use in practice very much?

I'm not quite sure if my usage would be similar to everyone's. With fs-router, the root layout would be necessary, but in general, I wouldn't need to use other layouts unless I care performance much.

There are a few things I would like to reconsider, and you can probably help on them:
- separation of unstable_defineRouter and createPages
  - currently, it's a bit confusing. defineRouter can be more primitive, or it's probably just naming.
- improve client cache system with defineRouter
  - Currently it's only layout/layout/page pattern, which is too related with createPages.
  - Components could be freely placed, independent from route path.
- Likewise, shouldSkip mechanism with defineRouter should be revisited.
  - It's hardly undestandable.

File: examples/25_weave_render/src/entries.tsx (hunk 1 above)
Comment (tylersayshi) [reply examples/25_weave_render/src/entries.tsx:49]: yeah, makes sense ! lgtm - changed

**Comment by tylersayshi** [comment]
> > @dai-shi is your general experience with the _layout pattern positive? Or is it not something you use in practice very much?
> 
> I'm not quite sure if my usage would be similar to everyone's. With fs-router, the root layout would be necessary, but in general, I wouldn't need to use other layouts unless I care performance much.
> 
> There are a few things I would like to reconsider, and you can probably help on them:
> 
> * separation of unstable_defineRouter and createPages
>   
>   * currently, it's a bit confusing. defineRouter can be more primitive, or it's probably just naming.
> * improve client cache system with defineRouter
>   
>   * Currently it's only layout/layout/page pattern, which is too related with createPages.
>   * Components could be freely placed, independent from route path.
> * Likewise, shouldSkip mechanism with defineRouter should be revisited.
>   
>   * It's hardly undestandable.

all sounds good to me! I don't have immediate thoughts/ideas for how to approach changing those, but I'll be thinking about what makes sense for them.

File: examples/25_weave_render/src/entries.tsx [hunk 2]
```diff
@@ -0,0 +1,57 @@
+import { lazy } from 'react';
+import { createPages } from 'waku';
+import type { PathsForPages } from 'waku/router';
+import FooPage from './components/FooPage';
+import BarPage from './components/BarPage';
+import BarLayout from './components/BarLayout';
+import FooLayout from './components/FooLayout';
+
+// The use of `lazy` is optional and you can use import statements too.
```
dai-shi, line 9: I think we can remove this lazy usage from the examples.

tylersayshi: I see you merged it, should I make a second PR to remove this?

dai-shi: Yeah, please go ahead. I mean from all examples.

PR #923, opened 2024-09-30
feat: Add hono enhancer to cloudflare and aws lambda

rmarscher:
Update `unstable_honoEnhancer` to accept an async function to support dynamic imports.

Update `unstable_honoEnhancer` to pass in the Hono app and the `serveWaku` Hono middleware handler.

Load waku config and invoke `unstable_honoEnhancer` in cloudflare build. If this looks good, we should all this too all of the deploy options.

With these changes, I am able to update my waku.config.ts to this in my Cloudflare Pages project: 

```ts
import type { Hono, MiddlewareHandler } from "hono";
import type { Config, unstable_HonoEnhancer } from "waku/config";

const addApiRoutes = (app: Hono) => {
  app.get("/api/hello", (c) => {
    return c.json({ hello: "world" });
  });
  return app;
};

// @ts-expect-error TODO figure out the right way to cast types here
const wakuConfig: Config = {
  ...(import.meta.env && !import.meta.env.PROD
    ? {
        unstable_honoEnhancer: async (createApp) => {
          const devServer = (await import("./waku.cloudflare-dev-server")).cloudflareDevServer({
            // Optional config settings for the Cloudflare dev server (wrangler proxy)
            // https://developers.cloudflare.com/workers/wrangler/api/#parameters-1
            persist: {
              path: ".wrangler/state/v3",
            },
          });
          return (app: Hono) => {
            // @ts-ignore ignoring types here for now
            return createApp(addApiRoutes(app));
          }
        },
      }
    : {
      // @ts-ignore ignoring types here for now
      unstable_honoEnhancer: (createApp) => (app) => createApp(addApiRoutes(app)),
    }),
  middleware: () => {
    return [
      import("waku/middleware/dev-server"),
      import("waku/middleware/headers"),
      import("waku/middleware/ssr"),
      import("waku/middleware/rsc"),
    ];
  },
};

export default wakuConfig;
```

I use `waku dev` for local development and my API route is accessible.

I use `NODE_ENV=production npm run build -- --with-cloudflare` to build.

Unfortunately... when I deploy to cloudflare, I am getting a strange error when trying to access the API route:

```
Cannot access 'loadConfig' before initialization
```

for the line with `const configPromise = loadEntries().then((entries) => entries.loadConfig());`

It works without error when I run the build using `npx wrangler pages dev` so that is strange.

### DIFF
--- a/packages/waku/src/config.ts
+++ b/packages/waku/src/config.ts
@@ -52,9 +52,9 @@ export interface Config {
    * Enhancer for Hono
    * Defaults to `undefined`
    */
-  unstable_honoEnhancer?: <Hono>(
-    createApp: (app: Hono) => Hono,
-  ) => Promise<(app: Hono) => Hono> | undefined;
+  unstable_honoEnhancer?:
+    | (<Hono>(createApp: (app: Hono) => Hono) => (app: Hono) => Hono)
+    | undefined;
 }
 
 export function defineConfig(config: Config) {
--- a/packages/waku/src/config.ts
+++ b/packages/waku/src/config.ts
@@ -49,12 +49,12 @@ export interface Config {
    */
   middleware?: () => Promise<{ default: Middleware }>[];
   /**
-   * Enhander for Hono
+   * Enhancer for Hono
    * Defaults to `undefined`
    */
-  unstable_honoEnhancer?:
-    | (<Hono>(createApp: (app: Hono) => Hono) => (app: Hono) => Hono)
-    | undefined;
+  unstable_honoEnhancer?: <Hono>(
+    createApp: (app: Hono) => Hono,
+  ) => Promise<(app: Hono) => Hono> | undefined;
 }
 
 export function defineConfig(config: Config) {
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -15,34 +15,34 @@ const getServeJsContent = (
   distPublic: string,
   srcEntriesFile: string,
 ) => `
-import path from "node:path";
-import { existsSync, readFileSync } from "node:fs";
+import path from 'node:path';
+import { existsSync, readFileSync } from 'node:fs';
 import {
   serverEngine,
   importHono,
   importHonoNodeServerServeStatic,
   importHonoAwsLambda,
-} from "waku/unstable_hono";
+} from 'waku/unstable_hono';
 
 const { Hono } = await importHono();
 const { serveStatic } = await importHonoNodeServerServeStatic();
 const { ${lambdaStreaming ? 'streamHandle:' : ''}handle } = await importHonoAwsLambda();
 
-const distDir = "${distDir}";
-const publicDir = "${distPublic}";
-const loadEntries = () => import("${srcEntriesFile}");
+const distDir = '${distDir}';
+const publicDir = '${distPublic}';
+const loadEntries = () => import('${srcEntriesFile}');
 
 const configPromise = loadEntries().then((entries) => entries.loadConfig());
 
 const createApp = (app) => {
-  app.use(serveStatic({ root: distDir + "/" + publicDir }));
-  app.use(serverEngine({ cmd: "start", loadEntries, env: process.env }));
+  app.use(serveStatic({ root: distDir + '/' + publicDir }));
+  app.use(serverEngine({ cmd: 'start', loadEntries, env: process.env }));
   app.notFound(async (c) => {
-    const file = path.join(distDir, publicDir, "404.html");
+    const file = path.join(distDir, publicDir, '404.html');
     if (existsSync(file)) {
-      return c.html(readFileSync(file, "utf8"), 404);
+      return c.html(readFileSync(file, 'utf8'), 404);
     }
-    return c.text("404 Not Found", 404);
+    return c.text('404 Not Found', 404);
   });
   return app;
 };
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
@@ -25,20 +25,28 @@ const { ${lambdaStreaming ? 'streamHandle:' : ''}handle } = await importHonoAwsL
 
 const distDir = '${distDir}';
 const publicDir = '${distPublic}';
-const loadEntries = () => import('${srcEntriesFile}');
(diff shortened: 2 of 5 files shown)


packages/waku/src/config.ts
```diff
@@ -49,14 +49,16 @@ export interface Config {
    */
   middleware?: () => Promise<{ default: Middleware }>[];
   /**
-   * Enhander for Hono
+   * Enhancer for Hono
```
dai-shi, line 52: Thanks.

packages/waku/src/config.ts
```diff
@@ -49,14 +49,16 @@ export interface Config {
...
    * Defaults to `undefined`
    */
-  unstable_honoEnhancer?:
-    | (<Hono>(createApp: (app: Hono) => Hono) => (app: Hono) => Hono)
-    | undefined;
+  unstable_honoEnhancer?: unstable_HonoEnhancer | undefined;
 }
 
+export type unstable_HonoEnhancer<Hono = never, MiddlewareHandler = never> = (
```
dai-shi, line 58: I would like to avoid exporting type from this file. If you need it in the plugin, simply define it there.

packages/waku/src/config.ts
```diff
@@ -49,14 +49,16 @@ export interface Config {
...
+  createApp: (app: Hono) => Hono,
+) => Promise<(app: Hono, serveWaku: MiddlewareHandler) => Hono>;
```
dai-shi, line 60: `Promise<>` is fine, but exposing `serveWaku` is too much.

dai-shi: Another idea would be:
```ts
(createApp: (app: Hono) => Promise<Hono>) => (app: Hono) => Promise<Hono>
```

dai-shi: > accept an async function to support dynamic imports.

Hm, on second thought, I wonder if promise is really necessary. Do you really need dynamic imports?

Can you try top-level await?

At this point, it's not convincing to change `unstable_honoEnhancer`.

packages/waku/src/config.ts
rmarscher: Yes, that sounds good.

packages/waku/src/config.ts
rmarscher: > Can you try top-level await?

Yes, I can try. It needs to import for the `waku dev` command and not import during build. I'm not sure if there is a better way to do that than using `import.meta.env.PROD`.

>  exposing serveWaku is too much. It doesn't look like a nice enhancer abstraction.

Exposing the built Hono middleware (serveWaku) is useful if the user doesn't want Waku to be the default route. I agree that it doesn't look as clean. Maybe we should leave it out until someone has a more compelling use case? I am fine with transforming the build output for my personal use cases.

I will make adjustments to this PR to remove the API changes.

I will also try to fix the error it throws when deployed to Cloudflare. I tried using top-level await there but the same error occurred.

dai-shi: > useful if the user doesn't want Waku to be the default route.

We expose `runner` for such use cases. I don't think they would use `unstable_honoEnhancer` in such cases. But, maybe I miss some use cases.

> I agree that it doesn't look as clean.

I would like it to be clean. 😄

rmarscher: > We expose `runner` for such use cases. I don't think they would use `unstable_honoEnhancer` in such cases. But, maybe I miss some use cases.

OK great. I will try that for the use case I am thinking of.

rmarscher: > > accept an async function to support dynamic imports.
> 
> Hm, on second thought, I wonder if promise is really necessary. Do you really need dynamic imports?
> 
> Can you try top-level await?

The build fails if wrangler and miniflare are imported. I'm pretty sure there isn't another way around that than conditionally importing with a dynamic import, but I might be missing something.

rmarscher: Yeah... I think there probably is a technique with initializing promises and then awaiting them inside the route handler - similar to how serverEngine is created at the top level and awaited in the handler in some of the platform server code. Thanks. I'll keep working on it.

dai-shi: actually the production mode check can be done outside `app.use` in this case? anyway, hope you can work it around.

rmarscher:
We should probably add `unstable_honoEnhancer` to the other 4 adapters now that we have it working for two of them. I'm not set up to easily test those. I could apply the same update to partykit that I did here. That should work the same...

packages/waku/src/lib/plugins/vite-plugin-deploy-partykit.ts
```diff
@@ -15,30 +15,39 @@ const { Hono } = await importHono();
...
+      return c.body(notFoundStaticAssetResponse.body, 404);
+    }
+    return c.text("404 Not Found", 404);
+  });
+  return app;
+};
 
 export default {
-  onFetch(request, lobby, ctx) {
+  async onFetch(request, lobby, ctx) {
```
rmarscher, line 41: According to https://docs.partykit.io/reference/partyserver-api/#static-onfetch - async is supported for onFetch.

packages/waku/src/lib/plugins/vite-plugin-deploy-aws-lambda.ts
```diff
@@ -15,30 +15,42 @@ const getServeJsContent = (
   distPublic: string,
   srcEntriesFile: string,
 ) => `
-import path from 'node:path';
-import { existsSync, readFileSync } from 'node:fs';
-import { serverEngine, importHono, importHonoNodeServerServeStatic, importHonoAwsLambda } from 'waku/unstable_hono';
+import path from "node:path";
```
dai-shi, line 18: We change prettier config, only for this.
https://github.com/dai-shi/waku/blob/5e183b8ebf693379ce9cd675a971b02ebd562c50/package.json#L58

rmarscher: OK. Sorry about that. I though vscode would use the project settings when I pasted into an unsaved editor buffer and ran the formatter, but I guess not. I copied to the file system and used `pnpx prettier -c --write` instead. Better now. https://github.com/dai-shi/waku/pull/923/commits/fc337de6fa1b2a997d7c5c8d32bd4d8f6690dc3c

packages/waku/src/config.ts
rmarscher: I got this working nicely. https://gist.github.com/rmarscher/9bb6ed54dc9535f4b81bed147204c7e9

dai-shi (approved):
LGTM
I hope eventually you can setup dev environments for other deploy adapters.

PR #925, opened 2024-10-01
fix: router typegen variable naming

### LINKED ISSUE #924: waku dev error with generated fs router types
I have a project that doesn't have a home page / root index.tsx file. I hit an error trying to run the dev server on the latest `main` commit (901594c2195de5a7d601ca6ecda674202e9a0915):

```
Error: ENOENT: no such file or directory, open '/waku-project/src/pages/index.tsx'
    at Object.openSync (node:fs:562:18)
    at readFileSync (node:fs:446:35)
    at fileExportsGetConfig (file:///waku-project/node_modules/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:68:30)
    at generateFile (file:///waku-project/node_modules/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:76:42)
    at updateGeneratedFile (file:///waku-project/node_modules/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:126:51) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/waku-project/src/pages/index.tsx'
}
```

The `filePaths` given to `generateFile` in vite-plugin-fs-router-typegen.ts looks like this for my project:

```
[
  '/_layout.tsx',
  '/_layout.tsx',
  '/_layout.tsx',
  '/index.tsx',
  '/demo.tsx',
  '/index.tsx',
  '/detail.tsx',
  '/index.tsx',
  '/index.tsx',
  '/app.tsx'
]
```

I don't think that's right to not include the entry paths and only list the filenames.

~~Also, I think the `recursive` option for `fs/promises` readdir is newer than some of our targets (https://nodejs.org/api/fs.html#fspromisesreaddirpath-options).~~

rmarscher:
Fixes #924


rmarscher:
I'm getting 404s with this... something must be off with generated `createPage` calls. Updated PR to draft while I look into it.

rmarscher:
This PR works fine on my cloudflare project... looking into the 404s in my aws lambda project... might be unrelated.

rmarscher:
OK. My 404 on AWS was due to test code I had for the `unstable_honoEnhancer` config feature. I just didn't see that before because of this fs router bug. I think this PR is good to merge.

I'm having a lot of trouble testing the latest main due trouble linking the @swc/core dependency. It would be helpful to get a beta build of waku on npm. Thank you!

tylersayshi:
@rmarscher Can you summarize what path issues in specific you were seeing as wrong and what the key changes were? 

From looking at the code changed, it looks like just most of the code for the file path collection was changed. I'm not opposed to changing it at all, but it's hard to see what the precise bug and solution is for what you were seeing.

Also, `filePaths` was a poorly named variable. From looking back at the code, it is supposed to be `/${fileNameWithExtension}` in it's current form. Changing to just `.endsWith` is then a bit of a hack to solve for file names that are potentially incorrect to start.

Referring back to your log of `filePaths` in the reported issue though, those look correct I think, so i think `pagesDir` was actually the variable that was not correctly collected for you. I am open to some of the refactoring you did here, but I think it would be better to fix the `pagesDir` not getting set in response to your issue, then follow up with more changes after if we want.

apologies for this being so long 😅

rmarscher:
OK. I'll create a reproduction.

rmarscher:
There are two main issues that this PR addresses:
1. the paths are not the full relative path from the pagesDir. They are just the file name. I think you have to use path or parentPath depending on the node version - https://nodejs.org/docs/latest-v20.x/api/fs.html#direntparentpath. Since the recursive option doesn't seem very stable, I updated to a recursive function that builds the relative path as it walks the pagesDir.
2. multiple characters in a path that are not valid for javascript variable names are not replaced. you are only doing `.replace` which only replaces the first match. I updated to a global regex that will replace all.

Reproducing is pretty easy. Delete the entries.tsx and main.tsx from `examples/10_fs-route` and run `waku dev`. 

For me, that causes this error:
```
Error: ENOENT: no such file or directory, open '/waku/examples/25_auto-fs-router/src/pages/[name].tsx'
    at Object.openSync (node:fs:562:18)
    at readFileSync (node:fs:446:35)
    at fileExportsGetConfig (file:///waku/packages/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:68:30)
    at generateFile (file:///waku/packages/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:76:42)
    at updateGeneratedFile (file:///waku/packages/waku/dist/lib/plugins/vite-plugin-fs-router-typegen.js:127:51) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/waku/examples/25_auto-fs-router/src/pages/[name].tsx'
}
```

I added some debug statements to `updateGeneratedFile`:

```ts
      const updateGeneratedFile = async () => {
        if (!pagesDir || !outputFile) return;
        const files = await collectFiles(pagesDir);
        console.log('collected files', files);
        const formatted = await formatter(generateFile(files));
        console.log('entries.gen.tsx', formatted);
      };
```

which outputs:

```
collected files [
  '/_layout.tsx',
  '/bar.tsx',
  '/index.tsx',
  '/[name].tsx',
  '/index.tsx'
]
```

before the error.

The error happens when calling `const hasGetConfig = fileExportsGetConfig(filePath);` with a filePath that doesn't exist.

I can temporarily get past the error by changing that line to `const hasGetConfig = false;`

Here is the entries.gen.tsx that it's trying to make:

```ts
import { createPages } from 'waku';
import type { PathsForPages } from 'waku/router';

import _Layout from './pages/_layout';
import Bar from './pages/bar';
import Index from './pages/index';
import SlugName from './pages/[name]';
import Index from './pages/index';

const _pages = createPages(async (pagesFns) => [
  pagesFns.createLayout({ path: '/', component: _Layout, render: 'static' }),
  pagesFns.createPage({ path: '/bar', component: Bar, render: 'dynamic' }),
  pagesFns.createPage({ path: '/', component: Index, render: 'dynamic' }),
  pagesFns.createPage({
    path: '/[name]',
    component: SlugName,
    render: 'dynamic',
  }),
  pagesFns.createPage({ path: '/', component: Index, render: 'dynamic' }),
]);

  interface RouteConfig {
    paths: PathsForPages<typeof _pages>;
  }
}

export default _pages;
```

If you rename `src/pages/bar.tsx` to `src/pages/bar-one-two.tsx`, then you will see the generated variable names can be invalid too. I get an error like this when running `waku dev`:

```
SyntaxError: '=' expected. (5:15)
  3 |
  4 | import _Layout from './pages/_layout';
> 5 | import Bar_one-two from './pages/bar-one-two';
    |               ^
  6 | import Index from './pages/index';
  7 | import SlugName from './pages/[name]';
  8 | import Index from './pages/index';
    at Q4 (file:///Users/robmarscher/Code/open-source/waku/node_modules/.pnpm/prettier@3.3.3/node_modules/prettier/plugins/typescript.mjs:17:76108)
```

I think it hits that error running the formatter before it saves the entries.gen.tsx.

When using the code from this PR, all is good. It looks like this:

```
import Layout from './pages/_layout';
import Baronetwo from './pages/bar-one-two';
import FootestIndex from './pages/foo-test/index';
import Index from './pages/index';
import NestedName from './pages/nested/[name]';
```

I guess technically someone could still do something to mess it up. A file named `route-one.tsx` and another named `routeone.tsx` would collide. If we replaced `-` with `_` instead of empty string, then `route-one.tsx` and `route_one.tsx` would collide. There are probably various other things that could break it that hopefully no one would have a reason to do in practice. And if someone has that use case, then they can just make their own custom entries file.

rmarscher:
Unit tests for some of these vite-plugin-fs-router-typegen.ts functions might be useful.

rmarscher:
> Also, `filePaths` was a poorly named variable. From looking back at the code, it is supposed to be `/${fileNameWithExtension}` in it's current form. Changing to just `.endsWith` is then a bit of a hack to solve for file names that are potentially incorrect to start.
> 
> Referring back to your log of `filePaths` in the reported issue though, those look correct I think

That doesn't make sense to me. I might be missing something. Going off the 10 fs router example, create a `pages/foo/_layout.tsx` file and create a `pages/foo/two-levels/index.tsx` file. For me, that creates an entries.gen.tsx with the following imports:

```
import _Layout from './pages/_layout';
import Bar from './pages/bar';
import Index from './pages/index';
import SlugName from './pages/[name]';
import _Layout from './pages/_layout';
import Index from './pages/index';
import Index from './pages/index';
```

Several of those paths two not resolve and there are duplicate names. After the changes in this PR, those become:

```
import Layout from './pages/_layout';
import Bar from './pages/bar';
import FooLayout from './pages/foo/_layout';
import FooIndex from './pages/foo/index';
import FooTwolevelsIndex from './pages/foo/two-levels/index';
import Index from './pages/index';
import NestedName from './pages/nested/[name]';
```

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -91,20 +71,26 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
...
-        for (const file of files) {
-          if (file.name.endsWith('.tsx')) {
-            results.push('/' + file.name);
+      const collectFiles = async (
+        dir: string,
+        cwd: string = '',
+        files: string[] = [],
+      ): Promise<string[]> => {
+        if (!cwd) {
+          cwd = dir;
```
tylersayshi, line 80: Is cwd needed? each entry from readDir has it's parent path

tylersayshi:
> import Layout from './pages/_layout';
import Bar from './pages/bar';
import FooLayout from './pages/foo/_layout';
import FooIndex from './pages/foo/index';
import FooTwolevelsIndex from './pages/foo/two-levels/index';
import Index from './pages/index';
import NestedName from './pages/nested/[name]';

This is what I initially was aiming for, so thanks for the fix.

tylersayshi:
> Unit tests for some of these vite-plugin-fs-router-typegen.ts functions might be useful.

agreed, would you mind adding some given that you've run into a few of the things we should test for here?

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -91,20 +71,26 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
...
+            if (entry.name.endsWith('.tsx')) {
+              files.push(fullPath.slice(dir.length));
+            }
           }
         }
-        return results;
+        return files;
       };
 
       const fileExportsGetConfig = (filePath: string) => {
```
tylersayshi, line 96: Should filePath be renamed? or is this still right?

tylersayshi (commented):
Thanks for digging into this

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -6,49 +6,29 @@ import { joinPath } from '../utils/path.js';
 
 const SRC_PAGES = 'pages';
 
+const invalidCharRegex = /[^0-9a-zA-Z_$]/g;
```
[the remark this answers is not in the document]
rmarscher: To be correct... it looks like unicode chars might be allowed. And there are reserved words that need to be replaced too if they match exactly. https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-names-and-keywords

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
rmarscher: dir is the starting dir and cwd is the current dir as it recurses. It needs the starting dir to create the relative path. It probably could use `pagesDir` from the outer scope though instead of passing around in the function.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -124,14 +110,14 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
           const src = filePath.slice(1);
           const hasGetConfig = fileExportsGetConfig(filePath);
 
-          if (filePath === '/_layout.tsx') {
+          if (filePath.endsWith('/_layout.tsx')) {
```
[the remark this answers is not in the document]
rmarscher: It's needed for nested index and layout files.

rmarscher:
> > Unit tests for some of these vite-plugin-fs-router-typegen.ts functions might be useful.
> 
> agreed, would you mind adding some given that you've run into a few of the things we should test for here?

We could create a fixture for an e2e test with various pages and layout paths and names to test.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
rmarscher: filePath seems right.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: pagesDir seems good to re-use. and I am seeing now that entry.parentPath is only available in node 20, so nvm about that.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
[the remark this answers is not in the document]
tylersayshi: ahh the filepaths are all relative to pages dir now, that makes sense

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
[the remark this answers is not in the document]
tylersayshi: sure, thanks! I meant a comment in the code with the regex though 😅 - just for anyone looking later without the PR context handy

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
rmarscher: Yeah, they started with `parentPath` and then changed it to just `path`. I tried to use too when generating the _routes.json file for the Cloudflare Pages build and found those issues with it and pivoted to a recursive function - so it was easy for me to apply that here.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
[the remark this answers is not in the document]
rmarscher: Right. It's a good suggestion. I agree we should document it. I just wanted to chat about that first before making revisions.

I found this lib - https://github.com/eemeli/safe-identifier. It references a blog article that doesn't exist anymore that suggests appending hash of the original var name to the sanitized name to ensure that it will be unique. https://web.archive.org/web/20110703045355/https://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/

I think I'll update it to be more correct by using a technique like that.

dai-shi:
> Unit tests for some of these vite-plugin-fs-router-typegen.ts functions might be useful.

I'm not following all the details, but it feels like unit tests are more appropriate than e2e tests. I mean unit tests for more coverage and e2e tests for some typical use cases would be good.

rmarscher:
> > Unit tests for some of these vite-plugin-fs-router-typegen.ts functions might be useful.
> 
> I'm not following all the details, but it feels like unit tests are more appropriate than e2e tests. I mean unit tests for more coverage and e2e tests for some typical use cases would be good.
> 
> General idea: e2e tests are very good and necessary to tests React/RSC framework capabilities, but having more e2e tests slows down our CI, which results in bad DX. So, we prefer unit tests if possible and we are adding more (but not enough) unit tests lately.

Great. That sounds good.

I think there are two issues - needing to use the relative fs route path when creating a list of entries and needing to convert that path into a valid javascript identifier to use as a variable name in the generated file. Tests for those two functions would be helpful. I will add vitest tests.

I think there are two issues - needing to use the relative fs route path when creating a list of entries and needing to convert that path into a valid javascript identifier to use as a variable name in the generated file.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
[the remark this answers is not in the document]
rmarscher: I also found this reference.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#identifiers
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words

We reserved words are not an issue because we uppercase the first char.

rmarscher:
OK @tylersayshi and @dai-shi. I think I have resolved the feedback from the code reviews. Let me know if there is anything else needed to merge it.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -91,20 +71,26 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
...
-        const results: string[] = [];
-        const files = await readdir(dir, {
-          withFileTypes: true,
-          recursive: true,
-        });
-
-        for (const file of files) {
-          if (file.name.endsWith('.tsx')) {
-            results.push('/' + file.name);
+      const collectFiles = async (
```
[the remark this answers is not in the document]
rmarscher: Done.

tylersayshi:
> OK @tylersayshi and @dai-shi. I think I have resolved the feedback from the code reviews. Let me know if there is anything else needed to merge it.

looks like the CI tests broke with the changes and are generating undefined as the "identifier"

image

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -3,53 +3,54 @@ import { readdir, writeFile } from 'node:fs/promises';
...
+export function getImportModuleNames(filePaths: string[]) {
+  const _moduleNames: { [k: string]: true } = {};
+  const moduleNames: { [k: string]: string } = {};
+  for (const filePath of filePaths) {
+    let identifier = toIdentifier(filePath);
+    if (identifier in _moduleNames) {
+      const defaultIdentifier = identifier;
+      const hash = crypto.createHash('md5').update(filePath).digest('hex');
+      let length = 8;
+      while (length < hash.length && identifier in _moduleNames) {
```
tylersayshi, line 44: personal preference would be to just add an incrementing number to the import module names to avoid conflicts instead of importing and adding a full hash (seems like overkill imo) but I won't insist on it.

rmarscher: Yeah, that's true... I created the hash var before I came up with using a map to track collisions. I'll take it out.

rmarscher: @tylersayshi I took it out - https://github.com/dai-shi/waku/pull/925/commits/7f7bd9f2e9029f30d3977aa9cca2ef55b52f4b1a

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -6,50 +6,49 @@ import { joinPath } from '../utils/path.js';
...
+    let identifier = toIdentifier(filePath);
+    if (identifier in _moduleNames) {
+      const defaultIdentifier = identifier;
+      let duplicateSuffix = 1;
+      while (identifier in _moduleNames) {
+        identifier = `${defaultIdentifier}_${duplicateSuffix}`;
+        duplicateSuffix = duplicateSuffix + 1;
+      }
+    }
+    _moduleNames[identifier] = true;
```
tylersayshi, line 47:
```suggestion
  const moduleNameCount: { [k: string]: number } = {};
  const moduleNames: { [k: string]: string } = {};
  for (const filePath of filePaths) {
    let identifier = toIdentifier(filePath);
    if (identifier in moduleNameCount) {
      moduleNameCount[identifier] += 1;
      identifier = `${defaultIdentifier}_${duplicateSuffix}`;
    } else {
      moduleNameCount[identifier] = 0;
    }
```

Thoughts on this? The while loop is unnecessary this way and this feels a bit clearer imo

tylersayshi:
@rmarscher wanna make sure you saw this: https://github.com/dai-shi/waku/pull/925#issuecomment-2387546502

It would probably be good to add a check for undefined in the unit tests for the identifiers as well

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -6,50 +6,54 @@ import { joinPath } from '../utils/path.js';
...
-    );
-  } else {
-    const fileName = split.at(-1)!.replace('-', '_').replace('.tsx', '');
-    return (
-      split.slice(0, -1).join('') +
-      fileName[0]!.toUpperCase() +
-      fileName.slice(1)
-    );
+// from waku/router/common
+export function getInputString(path: string): string {
```
dai-shi, line 10: Can we simply import it from there? We might be changing the implementation in the future.

rmarscher: When I tried adding `import { getInputString } from 'waku/router/common';` to the plugin, I got an error `Missing "./router/common" specifier in "waku" package`.

rmarscher:
> @rmarscher wanna make sure you saw this: [#925 (comment)](https://github.com/dai-shi/waku/pull/925#issuecomment-2387546502)
> 
> It would probably be good to add a check for undefined in the unit tests for the identifiers as well

I'm almost there with a new test that can catch this issue. I pushed a WIP commit but it's still failing.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: Try to use a relative path.

rmarscher: > Try to use a relative path.

Ah right. Thanks.

dai-shi (commented):
LGTM
Can @tylersayshi review?

tylersayshi:
Looks great to me! thanks again :)

rmarscher:
Thanks team!! 🙌  🙌

PR #931, opened 2024-10-03
fix: rename unstable_redirect and add example

dai-shi:
https://github.com/dai-shi/waku/discussions/927#discussioncomment-10825901

### DIFF
--- a/examples/07_router/src/components/Counter.tsx
+++ b/examples/07_router/src/components/Counter.tsx
@@ -1,8 +1,9 @@
 'use client';
 
 import { useState } from 'react';
-
 import { Link, useRouter_UNSTABLE as useRouter } from 'waku/router/client';
+
+import { jump } from './funcs';
 
 export const Counter = () => {
   const { path } = useRouter();
@@ -13,7 +14,12 @@ export const Counter = () => {
       <button onClick={() => setCount((c) => c + 1)}>Increment</button>
       <h3>This is a client component.</h3>
       <span>path: {path}</span>
-      <Link to="/">Go to Home</Link>
+      <p>
+        <Link to="/">Go to Home</Link>
+      </p>
+      <p>
+        <button onClick={() => jump()}>Jump to random page</button>
+      </p>
     </div>
   );
 };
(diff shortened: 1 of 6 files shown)


examples/07_router/src/components/Counter.tsx
```diff
@@ -13,7 +14,12 @@ export const Counter = () => {
       <button onClick={() => setCount((c) => c + 1)}>Increment</button>
       <h3>This is a client component.</h3>
       <span>path: {path}</span>
-      <Link to="/">Go to Home</Link>
+      <p>
+        <Link to="/">Go to Home</Link>
+      </p>
+      <p>
+        <button onClick={() => jump()}>Jump to random page</button>
```
rmarscher, line 21: 😍 server functions are fun

rmarscher (approved):
LGTM. Thanks!

packages/waku/src/router/client.ts
```diff
@@ -522,9 +522,13 @@ export function Router({ routerData = DEFAULT_ROUTER_DATA }) {
...
-                routerData[1]?.forEach((listener) =>
-                  listener(pathname, searchParamsString),
-                );
+                setTimeout(() => {
+                  // HACK we need to wait a tick for React to start rendering.
+                  // FIXME It doesn't feel ideal. Hope to find a better solution.
+                  routerData[1]?.forEach((listener) =>
+                    listener(pathname, searchParamsString),
+                  );
+                });
```
tylersayshi, line 531: Would `startTransition` from react be less of a hack?

dai-shi: Unfortunately, not. I'm not sure if it's React issue or Waku issue. Probably, the latter.

Meanwhile, I used another hack 3fedc84.

dai-shi: Ahhh, unfortunately, this causes a flicker when moving to a new page. Let me revert it and open a new issue with `jump` case.

dai-shi: #936

dai-shi: #937

Issue #930, opened 2024-10-02
development error on windows


himself65:
```ts
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported by the default ESM
loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:227:11)
    at defaultLoad (node:internal/modules/esm/load:109:3)
    at ModuleLoader.load (node:internal/modules/esm/loader:570:13)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:443:56)
    at new ModuleJob (node:internal/modules/esm/module_job:76:27)
    at #createModuleJob (node:internal/modules/esm/loader:456:17)
    at ModuleLoader.getJobFromResolveResult (node:internal/modules/esm/loader:266:34)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:247:17)
    at async onImport.tracePromise.proto (node:internal/modules/esm/loader:482:25) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
Cannot process SSR Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are supported
by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'f:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:227:11)
    at defaultLoad (node:internal/modules/esm/load:109:3)
    at ModuleLoader.load (node:internal/modules/esm/loader:570:13)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:443:56)
    at new ModuleJob (node:internal/modules/esm/module_job:76:27)
    at #createModuleJob (node:internal/modules/esm/loader:456:17)
    at ModuleLoader.getJobFromResolveResult (node:internal/modules/esm/loader:266:34)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:247:17)
    at async onImport.tracePromise.proto (node:internal/modules/esm/loader:482:25) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
```

himself65:
```ts
git clone https://github.com/Fallenhh/hh.sb
pnpm i
pnpm run dev
```

dai-shi:
[...]

PR #932, opened 2024-10-03
fix: dev server on windows

himself65:
Fixes: https://github.com/dai-shi/waku/issues/930


packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -147,7 +148,7 @@ const createMainViteServer = (
       const fileWithAbsolutePath = file.startsWith('/')
         ? file
         : joinPath(vite.config.root, file);
-      return import(/* @vite-ignore */ fileWithAbsolutePath);
+      return import(/* @vite-ignore */ filePathToFileURL(fileWithAbsolutePath));
```
dai-shi, line 151: Ah!

dai-shi (approved):

PR #933, opened 2024-10-03
chore: update deps and react rc

dai-shi:


dai-shi:
ee4e014 fixes the weird error. Would anyone like to create a small reproduction without Waku and report it to React?

PR #914, opened 2024-09-28
prefer getHonoContext to hono/context-storage

dai-shi:
at least for now. this reverts #891 and re-introduce #852 and #884.

https://github.com/dai-shi/waku/pull/911#issuecomment-2380439350

PR #934, opened 2024-10-03
feat: make context available to middleware

rmarscher:
Re: #914  I noticed I had to use `ctx.context.__hono_context` to access the hono context from middleware. While that is ok, I thought it might be useful to try to make the `ctx.context` available to middleware from the `unstable_getCustomContext` API.

### DIFF
--- a/packages/waku/src/lib/hono/engine.ts
+++ b/packages/waku/src/lib/hono/engine.ts
@@ -1,6 +1,9 @@
 import type { Context, Env, MiddlewareHandler } from 'hono';
 
-import { unstable_getCustomContext } from '../../server.js';
+import {
+  runWithRenderStoreInternal,
+  unstable_getCustomContext,
+} from '../../server.js';
 import { resolveConfig } from '../config.js';
 import type { HandlerContext, MiddlewareOptions } from '../middleware/types.js';
 
@@ -59,7 +62,13 @@ export const serverEngine = (options: MiddlewareOptions): MiddlewareHandler => {
         }
       });
     };
-    await run(0);
+    const renderStore = {
+      context: ctx.context,
+      rerender: () => {
+        throw new Error('Cannot rerender');
+      },
+    };
+    await runWithRenderStoreInternal(renderStore, () => run(0));
     if (ctx.res.body || ctx.res.status) {
       return c.body(
         ctx.res.body || null,


packages/waku/src/lib/hono/engine.ts
```diff
@@ -59,7 +62,13 @@ export const serverEngine = (options: MiddlewareOptions): MiddlewareHandler => {
...
       });
     };
-    await run(0);
+    const renderStore = {
+      context: ctx.context,
+      rerender: () => {
+        throw new Error('Cannot rerender');
+      },
+    };
+    await runWithRenderStoreInternal(renderStore, () => run(0));
```
dai-shi, line 71: This doesn't looks nice. This function shouldn't be used outside RSC capabilities.
Consider `unstable_getHonoContext` is for server components.
Middleware can use `__hono_context` directly.
A little concern of mine is there can be two types of middleware, one that depends on hono and the other that doesn't.

dai-shi:
not planned for now.

dai-shi:
(but this brings me back to possibly prefer `hono/context-storage`... 😅 )

rmarscher:
Yes, no problem. As I work on my apps, I will look for ways to replace any specific use of Hono with the Waku context. I actually am seeing less and less of a need for it now that I understand Waku better. We could just make the Cloudflare bindings accessible on the Waku context instead of Hono.

> (but this brings me back to possibly prefer `hono/context-storage`... 😅 )
> or change "render storage" to "request storage" 🤔 . Not anytime soon.

Haha... Not anytime soon sounds good. I'm ready to move onto other things. 😃  But... moving c.env/c.executionCtx to ctx.context.cloudflareEnv / ctx.context.executionCtx for the Cloudflare Pages build might be worthwhile. Then I could avoid mentioning Hono so much in the Cloudflare guide.

dai-shi:
The reason I mark getCustomContext and getHonoContext `unstable` is because I'm not very confident yet. Let's learn more with getting more feedback.

rmarscher:
One thing to note here, if we call the same function (like a data fetcher) that needs something from context, then it is necessary to pass through the context when calling it from waku middleware but use `getCustomContext` when calling it from a server component. I use EffectTS for data fetching on one of my apps and have a "Layer" for my context. When in middleware, I supply the context to the layer and otherwise it fetches it from getCustomContext. I think EffectTS Layers just create an ALS context (https://effect.website/docs/guides/context-management/services). So I'm basically just doing the same thing from this PR in userland now.

PR #936, opened 2024-10-04
revert router/client change in #931

dai-shi:
https://github.com/dai-shi/waku/pull/931#discussion_r1787027823


dai-shi:
Actually, I'm not sure if the flicker was caused by this. Anyway, let's take a deeper look later.

PR #938, opened 2024-10-04
fix(create-waku): downgrade ncc

dai-shi:
`@vercel/ncc@0.38.2` seems to break create-waku's `notifyUpdate`.

Here's the error log:
```
TypeError: undefined is not a function
    at notifyUpdate (file:///.../node_modules/create-waku/dist/index.js:30:110825)
```

@himself65 @ojj1123 do you think it's fixable?

In the meantime, let's downgrade `ncc`.

### DIFF
--- a/packages/create-waku/package.json
+++ b/packages/create-waku/package.json
@@ -29,7 +29,7 @@
     "@types/fs-extra": "^11.0.4",
     "@types/prompts": "^2.4.9",
     "@types/tar": "^6.1.13",
-    "@vercel/ncc": "^0.38.2",
+    "@vercel/ncc": "0.38.1",
     "fs-extra": "^11.2.0",
     "kolorist": "^1.8.0",
     "prompts": "^2.4.2",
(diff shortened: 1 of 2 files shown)


ojj1123:
Is it fixed? or need any investigating? If so, how could I reproduce it?

dai-shi:
It's not fixed. If you update ncc to the latest version, you can reproduce it if you check `dist/index.js`, `createRequire` is gone.

himself65:
we can replace ncc to other bundler like tsup, bunchee or something, they are not using webpack, webpack has too many hakcs on imports

dai-shi:
I hope we can replace it with more lightweight one. What was the reason we need bundling again?

ojj1123:
When merging https://github.com/dai-shi/waku/pull/933, ncc is upgraded in 0.38.2.

> we can replace ncc to other bundler like tsup, bunchee or something, they are not using webpack, webpack has too many hakcs on imports

You mean ncc depends on webpack. Btw why did you use ncc? related PR https://github.com/dai-shi/waku/pull/330

dai-shi:
@ojj1123 @himself65 just opened a new issue #973.

PR #948, opened 2024-10-05
fix: type checking for waku unit tests not working

tylersayshi:
This fixes type checking for the waku unit tests both in CI and in the editor


packages/waku/tests/create-pages.test.ts
```diff
@@ -20,6 +19,7 @@ import { renderToString } from 'react-dom/server';
 import { expectType } from 'ts-expect';
 import type { TypeEqual } from 'ts-expect';
 import type { PathsForPages } from '../src/router/base-types.js';
+import { GetSlugs } from '../src/router/create-pages-utils/inferred-path-types.js';
```
dai-shi, line 22: Isn't it `import type ...` ? I wonder why `"verbatimModuleSyntax": true,` doesn't complain.

tylersayshi: good catch, updated

It isn't complaining for the same reason that the import from the wrong file was not complaining I think... `tests` is not included in the tsconfig & the rootDir is `./src`

https://github.com/dai-shi/waku/blob/main/packages/waku/tsconfig.json#L8

I'll look into fixing this type checking. Basically what we want is for tsc to check the `tests` folder, but not output anything for it.

dai-shi: > `tests` is not included in the tsconfig & the rootDir is `./src`

Ohhhhh, I see. cc @himself65

tylersayshi: Fixed with the latest commit! I'm happy to move things around or change the package json scripts changed as you'd like

packages/waku/package.json
```diff
@@ -67,11 +67,12 @@
   ],
   "scripts": {
     "dev": "swc src -d dist -w --strip-leading-paths",
-    "test": "vitest run",
+    "test": "vitest run && pnpm run test:types",
```
dai-shi, line 70: Hmm, can we simply add a reference in the root tsconfig.json instead?

tylersayshi: If we do that, when we run `compile:types` it would compile all of our `tests` and their types to the `dist` output. This seemed very undesirable, which is why I split them into separate configs, where I can set `noEmit: true` for the tests.

dai-shi: Yeah, we don't want to emit test files for sure.
Let's see if @himself65 has an idea.

himself65: Seems like there's no way to check file type without emitting a file

Related issue: https://github.com/microsoft/TypeScript/issues/40431

dai-shi: thanks for checking out.

packages/waku/tests/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "../",
+    "noEmit": true
+  },
+  "include": ["../src", "."]
```
dai-shi, line 7: `src` is already checked, so this is duplicated, right?
Maybe we should move this file to `packages/waku/tsconfig.test.json`. What do you think?

tylersayshi: as `tests/tsconfig.json` vscode (and any other editor with a typescript server) will find the tsconfig for the tests and use it. (otherwise it reads from `packages/waku/tsconfig.json` which excludes the tests, so type checking does not work. This is why we saw bad imports and no enforcement of the `import type {` specifier before. vscode would need custom configuration to acknowledge `tsconfig.test.json` and so would every other editor. 

`"include": ["../src"]`  is needed so we can import from `../src` which typescript considers a separate "project" unless we include it in the same tsconfig.

image

himself65, line 7: can we add another reference here?

tylersayshi: is this what you mean? 

```suggestion
  "include": ["."],
  "references": [
      {
        "path": "../../../packages/waku"
      },
    ]
``

or do you mean to add a reference to the new tsconfig in the root level tsconfig.json ?

I'm not very familiar with the "references" field just yet honestly

Comment (dai-shi) [reply packages/waku/tests/tsconfig.json:7]: alright.

Comment (himself65) [reply packages/waku/tests/tsconfig.json:7]: I think it will work

File: packages/waku/package.json (hunk 2 above)
Comment (himself65) [reply packages/waku/package.json:70]: there is a workaround, you can set output dir into a single directory, like `../../../dist`?

File: packages/waku/tests/tsconfig.json (hunk 3 above)
Comment (tylersayshi) [reply packages/waku/tests/tsconfig.json]: oh yep! replaced the path with just `".."` but that allows me to remove src from the include. Thanks for the tip!

File: packages/waku/package.json (hunk 2 above)
Comment (tylersayshi) [reply packages/waku/package.json:70]: I don't think I follow... can you explain the workaround a bit more?

Comment (himself65) [inline packages/waku/package.json:70]: oh on, this is weird to me, I thought type check should only in monorepo level

Comment (tylersayshi) [reply packages/waku/package.json:70]: do you mean the `tsc -b` on this line: https://github.com/dai-shi/waku/blob/main/package.json#L11 ?

does this actually type check anything? When I run it, theres no output even when I introduce ts errors in the project for checking its behavior.

Comment (himself65) [reply packages/waku/package.json:70]: THis is what i'm expecting

```diff
Subject: [PATCH] fix: monorepo setup
---
Index: tsconfig.e2e.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tsconfig.e2e.json b/tsconfig.e2e.json
--- a/tsconfig.e2e.json	(revision 1469524a4ec329ce1c87db2a37088d2dbd4cce1d)
+++ b/tsconfig.e2e.json	(date 1728199013977)
@@ -4,7 +4,7 @@
     "outDir": "./dist/e2e",
     "types": ["node"]
   },
-  "include": ["playwright.config.base.ts", "playwright.config.ts", "./e2e"],
+  "include": ["playwright.config.base.ts", "playwright.config.ts"],
   "references": [
     {
       "path": "./e2e/fixtures/rsc-basic/tsconfig.json"
Index: tsconfig.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/tsconfig.json b/tsconfig.json
--- a/tsconfig.json	(revision 1469524a4ec329ce1c87db2a37088d2dbd4cce1d)
+++ b/tsconfig.json	(date 1728199013983)
@@ -39,6 +39,10 @@
     // E2E
     {
       "path": "./tsconfig.e2e.json"
+    },
+    // Test
+    {
+      "path": "./packages/waku/tests"
     }
   ]
 }
Index: packages/waku/tests/tsconfig.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/packages/waku/tests/tsconfig.json b/packages/waku/tests/tsconfig.json
--- a/packages/waku/tests/tsconfig.json	(revision 1469524a4ec329ce1c87db2a37088d2dbd4cce1d)
+++ b/packages/waku/tests/tsconfig.json	(date 1728199060430)
@@ -1,8 +1,10 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "../../../tsconfig.json",
   "compilerOptions": {
-    "rootDir": "../",
-    "noEmit": true
+    "rootDir": "..",
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
+    "lib": ["esnext"]
   },
   "include": ["."],
   "references": [

```

himself65: 1. monorepo `tsc -b` should handle cover the whole type check
2. test package.json should extends the global one, and do not have DOM because it's in vitest(node)

tylersayshi: got it, thanks! Updated to match this.

packages/waku/tests/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "..",
+    "outDir": "./dist",
+    "tsBuildInfoFile": "./dist/.tsbuildinfo",
```
dai-shi, line 6: we are using the default for now.
```suggestion
```

himself65: ok, fine

himself65: I hope it emit to `dist`, so people can remove dist with it

dai-shi: Yeah I understand that that sense, but `dist` also is something to publish, so it doesn't feel very appropriate in another sense.

PR #949, opened 2024-10-05
idea: new defineRouter

dai-shi:
We would like to redesign `defineRouter`. This is why we have `unstable_` prefix.

https://github.com/dai-shi/waku/issues/885#issuecomment-2395208298

> similar snippets in both `defineRouter` and `createPages` are currently assuming 1:1 relationship from path to component id.

That's the exactly one of the reasons. It shouldn't be 1:1. It can be anything, and it can cache each. This will change waku/router/client too including "shouldSkipList".

Starting with a rough idea.


packages/waku/src/router/define-router.ts
```diff
@@ -269,3 +269,33 @@ export function unstable_rerenderRoute(
...
+        }
+      >;
+      noSsr?: boolean;
+    }>
+  >,
+  renderRoute: (
+    path: string, // "/foo/bar"
+    options: {
+      query?: string;
+      skipIds?: string[];
```
tylersayshi, line 296: These are component ids specific to RSC level components, correct? 

Sidenote: I do see not that the server does need to be "told" somehow about which RSCs already exist on the client router, which helps explain the use of `skipIds`

dai-shi: It's the subset of strings specified at L284.

packages/waku/src/router/define-router.ts
```diff
@@ -269,3 +269,33 @@ export function unstable_rerenderRoute(
...
+      components: Record<
+        string, // componentId such as "**/layout" or "**/page" in the case of createPages
+        {
+          isStatic?: boolean;
+        }
+      >;
+      noSsr?: boolean;
+    }>
+  >,
+  renderRoute: (
```
tylersayshi, line 292: This is a small dx idea, but I think it might be helpful to have the params of `defineRouter` to be named in an object param. Something like:

```ts
export default defineEntries({
  renderEntries: async (rscPath) => {
    return {
      App: <App name={rscPath || 'Waku'} />,
    };
  },
  getBuildConfig: async () => [{ pathname: '/', entries: [{ rscPath: '' }] }],
  getSSRConfig: () => {
    throw new Error('SSR is should not be used in this test.');
  },
});
```

just to make it clear what each thing is doing from the consumer perspective without needing to rely on comments

dai-shi: yeah, we can do that. I'll redesign defineEntries too, later.

packages/waku/src/router/client.ts
```diff
@@ -606,3 +607,244 @@ class ErrorBoundary extends Component<
...
+}
+
+/**
+ * ServerRouter for SSR
+ * This is not a public API.
+ */
+export function NewServerRouter({ route }: { route: RouteProps }) {
+  const routeElement = createElement(Slot, { id: getRouteSlotId(route.path) });
+  return createElement(
+    Fragment,
```
tylersayshi, line 836: What is the purpose of this Fragment?

packages/waku/src/router/client.ts
```diff
@@ -606,3 +607,244 @@ class ErrorBoundary extends Component<
...
+    async (responsePromise: Promise<Response>) => {
+      const response = await responsePromise;
+      const has404 = routerData[2];
+      if (response.status === 404 && has404) {
+        // HACK this is still an experimental logic. It's very fragile.
+        // FIXME we should cache it if 404.txt is static.
+        return fetchRsc(encodeRoutePath('/404'));
+      }
+      const data = createData(responsePromise);
+      Promise.resolve(data)
```
tylersayshi, line 780: Does this resolve the responsePromise twice? First on L772 then again here? Or is my understanding off with that?

then if it does resolve twice, shouldn't createData just use `response`?

examples/22_define-router/src/entries.tsx
```diff
@@ -0,0 +1,156 @@
+import { new_defineRouter } from 'waku/router/server';
+import { Slot, Children } from 'waku/client';
```
tylersayshi, line 2: is `waku/client` the best place for these? client to me would mean these are targeted towards the scope of client components, but they're just used here for the router to be able to target where to replace with each level of rendering components (regardless of client vs server)

that is if I understand correctly

tylersayshi (commented):
Overall, I think the new defineRouter implementation reads really well!

This really helps to see what you meant by simplifying the structures in the previous version.

packages/waku/src/router/client.ts
dai-shi: createData requires a promise, so L772 isn't related with L780.

However, yes, `data.then((date) => ` here might work. The reason I added `Promise.resolve()` is because `data` returned by `createData` isn't a real promise. TS types aren't correct here. So, it's just a safe bet. Maybe it's not necessary.

packages/waku/src/router/client.ts
dai-shi: Maybe, it's to match the component tree with NewRouter. If removing it works, we can remove it. TBH, I don't like NewServerRouter hack. I wish just NewRouter worked.

examples/22_define-router/src/entries.tsx
dai-shi: They are client components. We can embed client components in a server component. It's confusing, but it's how it works. createPages API will hide this complexity anyway.

PR #952, opened 2024-10-06
feat: allow optionally controlled root

tylersayshi:


packages/waku/src/router/define-router.tsx
```diff
@@ -46,6 +46,15 @@ const safeJsonParse = (str: unknown) => {
...
 };
 
+export function DefaultRoot({ children }: { children: ReactNode }) {
+  return (
+    <html>
+      <head></head>
+      <body>HELLO WORLD{children}</body>
+    </html>
+  );
+}
```
tylersayshi, line 56: @dai-shi I am pretty certain putting this here is wrong, but at the same time I am unsure where it should go or how it should be written.

Should I be trying to create this with react's `createElement` somehow instead?

dai-shi: I think it should be in `create-pages.ts`. Yes, use `createElement`.

tylersayshi: great! thanks, so the very basic example works. Now I'll work on supporting `_root.tsx`, then updating the examples & tests :)

dai-shi: that's quick!

packages/waku/src/router/create-pages.ts
```diff
@@ -350,6 +390,10 @@ export function createPages<
     },
     async (id, { unstable_setShouldSkip }) => {
       await configure();
+      if (id === 'root') {
+        unstable_setShouldSkip();
```
dai-shi, line 394: Shouldn't it be `unstable_setShouldSkip([])` if it's static?

tylersayshi: like this?

```ts
        if (rootItem?.render === 'dynamic') {
          unstable_setShouldSkip();
        } else unstable_setShouldSkip([]);
```

dai-shi: Yeah.

e2e/fixtures/broken-links/src/pages/_layout.tsx
```diff
@@ -5,12 +5,7 @@ export default async function RootLayout({
 }: {
   children: ReactNode;
 }) {
-  return (
-    <html>
-      <head></head>
-      <body>{children}</body>
-    </html>
-  );
+  return children;
```
dai-shi, line 8: Maybe we can remove the file entirely.

e2e/fixtures/ssg-wildcard/src/pages/_layout.tsx
```diff
@@ -1,14 +1,5 @@
 import type { ReactNode } from 'react';
 
-const Layout = ({ children }: { children: ReactNode }) => (
-  <html>
-    <head>
-      <title>Waku</title>
```
dai-shi, line 6: I think having this title is important.

examples/11_fs-router/src/pages/_root.tsx
```diff
@@ -0,0 +1,22 @@
+import { ReactNode } from 'react';
+
+export default function Root({ children }: { children: ReactNode }) {
+  return (
+    <html>
+      <head>
+        <title>Waku</title>
```
dai-shi, line 7: we don't need this.

packages/waku/src/lib/renderers/rsc-renderer.ts
```diff
@@ -169,11 +169,15 @@ export async function renderRsc(
         elementsPromise = Promise.all([
           elementsPromise,
           renderEntries(rscPath, { rscParams }),
-        ]).then(([oldElements, newElements]) => ({
-          ...oldElements,
-          // FIXME we should actually check if newElements is null and send an error
```
dai-shi, line 174: I don't remember when it becomes `null`. Have you tried anything?

packages/waku/src/router/common.ts
```diff
@@ -12,7 +12,7 @@ export function getComponentIds(path: string): readonly string[] {
     idSet.add(id);
   }
   idSet.add([...pathItems, 'page'].join('/'));
-  return Array.from(idSet);
+  return ['root', ...Array.from(idSet)];
```
dai-shi, line 15: Nice.

tylersayshi:
@dai-shi do you know why `pnpm install` is failing on the CI checks? I can't see anything I might've changed... or maybe I'm misreading the logs, honestly unsure

dai-shi (commented):
1st review

packages/waku/src/lib/renderers/rsc-renderer.ts
tylersayshi: When the path is not found: https://github.com/dai-shi/waku/blob/main/packages/waku/src/router/define-router.ts#L126

Maybe we could throw the error from `renderEntries` when path is not found instead of returning null? That way the path could be included in the error message.

e2e/fixtures/broken-links/src/pages/_layout.tsx
tylersayshi: removed. Do you think the other layouts that just return children should also be removed or do we want to keep some of them?

examples/11_fs-router/src/pages/_root.tsx
tylersayshi: Do you mean the title or the entire file?

I think it's helpful to have one example using the `_root.tsx` file at least

e2e/fixtures/ssg-wildcard/src/pages/_layout.tsx
tylersayshi: added the title back just inside the div which is then lifted up to the head. react does this for us now, right?

e2e/fixtures/broken-links/src/pages/_layout.tsx
dai-shi: Let's remove.

e2e/fixtures/ssg-wildcard/src/pages/_layout.tsx
dai-shi: Yep.

e2e/fixtures/broken-links/src/pages/_layout.tsx
tylersayshi: removed 👍

examples/11_fs-router/src/pages/_root.tsx
dai-shi: I meant the title tag.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: Hmm, I actually don't remember why I wrote this comment, particularly "send an error". Let's make it `console.warn` instead of throwing an error for now.

packages/waku/src/router/create-pages.ts
```diff
@@ -350,6 +390,12 @@ export function createPages<
     },
     async (id, { unstable_setShouldSkip }) => {
       await configure();
+      if (id === 'root') {
+        if (rootItem?.render === 'dynamic') {
+          unstable_setShouldSkip();
+        } else unstable_setShouldSkip([]);
```
dai-shi, line 396:
```suggestion
        } else {
          unstable_setShouldSkip([]);
        }
```

packages/waku/src/lib/renderers/rsc-renderer.ts
tylersayshi: I can change to `console.warn` but I should say: this would throw a runtime error if `null` is returned with the current version on `main` because null cannot be spread `...null` or `...someVariableThatIsNull` both throw runtime errors.

So given that we haven't seen this reported as an issue people run into, it seems like `null` is never actually getting returned which is good.

packages/waku/src/router/create-pages.ts
tylersayshi: fixed... in a separate PR would you like me to turn on this rule: https://eslint.org/docs/latest/rules/curly

It would enforce this preference

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: `{ ...null }` is fine. at least, works on nodejs.

tylersayshi: weird... errors in the browser

image

I'll change to console.warn though!

packages/waku/src/router/create-pages.ts
dai-shi: yeah, please try that.

packages/waku/src/lib/renderers/rsc-renderer.ts
dai-shi: oh... but, yeah.

e2e/fixtures/partial-build/src/pages/_root.tsx
```diff
@@ -0,0 +1,18 @@
...
+        <title>Waku</title>
+      </head>
+      <body>{children}</body>
+    </html>
+  );
+}
+
+export const getConfig = async () => {
+  return {
+    render: 'static',
```
dai-shi, line 16: this is the default, right?

I wonder if we need `_root.tsx` for partial-build fixture. Is there any reason you created this file?

dai-shi (approved):
LGTM

Merging this, but could you add a PR description? It would be nice if it's detailed.

e2e/fixtures/partial-build/src/pages/_root.tsx
tylersayshi: I created this just so that we had a test case that used the _root.tsx file.

If the presence of _root.tsx broke the build for any reason, having this here should catch that.

I'll update the description tho!

Issue #954, opened 2024-10-06
Running `pnpm test` locally results in import/no-unresolved errors


tylersayshi:
When I run `pnpm test` locally, eslint fails with this import rule. In other projects, I often disable this import error since typescript already checks imports and more reliably.

I'm not suggesting to disable the rule here necessarily, but I would like to figure out how to fix it. And also, I'd like to understand how it succeeds in CI. I tried doing a fresh install of all my node_modules and still got the same errors

```sh
...

/Users/tyler/gitspace/waku/packages/create-waku/template/24_nesting/src/components/InnerApp.tsx
  1:26  error  Unable to resolve path to module 'react'  import/no-unresolved

/Users/tyler/gitspace/waku/packages/create-waku/template/24_nesting/src/main.tsx
  1:28  error  Unable to resolve path to module 'react'             import/no-unresolved
  2:41  error  Unable to resolve path to module 'react-dom/client'  import/no-unresolved

✖ 66 problems (66 errors, 0 warnings)
```

@himself65 is this something you'd be familiar with?

himself65:
remove template files, it’s old and copy during build step

tylersayshi:
Awesome, thanks!

PR #955, opened 2024-10-06
chore: update eslint ignore

himself65:
Fixes: https://github.com/dai-shi/waku/issues/954


tylersayshi:
Shouldn't eslint just ignore anything that is gitignored, which would include the templates?

himself65:
> Shouldn't eslint just ignore anything that is gitignored, which would include the templates?

unfortunately, no

https://github.com/eslint/eslint/issues/5848#issuecomment-211434706

dai-shi (approved):
thanks!

PR #956, opened 2024-10-07
chore: lint for curly brackets and fix

tylersayshi:
follow up to: https://github.com/dai-shi/waku/pull/952#discussion_r1789346977

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,7 +33,8 @@
         "destructuredArrayIgnorePattern": "^_"
       }
     ],
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "curly": ["error", "all"]
   },
   "overrides": [
     {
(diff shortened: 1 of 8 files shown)


.eslintrc.json
```diff
@@ -33,7 +33,8 @@
         "destructuredArrayIgnorePattern": "^_"
       }
     ],
-    "react/prop-types": "off"
+    "react/prop-types": "off",
+    "curly": ["error", "all"]
```
dai-shi, line 37: I wish it had a rule for early return. https://eslint.org/docs/latest/rules/curly

tylersayshi: ah, yeah. There are other options, but I don't currently see a way to separate early returns

tylersayshi: multi is sorta similar: https://eslint.org/docs/latest/rules/curly#multi

PR #957, opened 2024-10-08
fix: waku website types for paths

tylersayshi:
cc @sandren 

I don't like the `as const` gotcha that this PR calls attention to.

I'd like to suggest another solution... If we suggest (and maybe enforce) the use of a function to wrap the return of `getConfig` we can infer the type of `render` and `staticPaths` without any need to add `as const`

a playground: https://tsplay.dev/weq8Bw

image

I am open to other ideas of course, but the above seems best for helping with type safety of the return on getConfig and with removing the `as const` gotcha.

I'd address this in a separate PR, but seems good to mention here alongside the issue that came up with this.


packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -142,9 +142,10 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
               hasGetConfig,
             });
           } else if (filePath.endsWith('/index.tsx')) {
+            const path = filePath.replace('index.tsx', '');
             fileInfo.push({
               type: 'page',
-              path: filePath.replace('index.tsx', ''),
+              path: path.length === 1 ? '/' : path.slice(0, -1), // remove trailing slash
```
dai-shi, line 148: Not super readable. How about this?
```js
const path = filePath.slice(0, -'/index.tsx'.length);
  ...
  path: path || '/', // or path === '' ? '/' : path
```

tylersayshi: nice! changed

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -142,9 +142,10 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
               hasGetConfig,
             });
           } else if (filePath.endsWith('/index.tsx')) {
+            const path = filePath.slice(0, -'/index.tsx'.length);
             fileInfo.push({
               type: 'page',
-              path: filePath.replace('index.tsx', ''),
+              path: path.length || '/',
```
dai-shi, line 148: .length seems a bug.
```suggestion
              path: path || '/',
```
or `path: path.length ? path : '/'`.

tylersayshi: whoops, thanks

dai-shi (approved):
LGTM

PR #959, opened 2024-10-09
wip: eliminate server router

dai-shi:


dai-shi:
Ah, it's not possible with `ServerRoot`. Will revisit in the future.

PR #961, opened 2024-10-10
new defineEntries (and context middleware)

dai-shi:
This is to prepare for upcoming api route support.

- [x] context middleware (replacing header middleware)
- [x] new_defineEntries
  - [x] renderRsc
  - [x] renderHtml
  - [x] build
  - [x] server functions


examples/31_minimal/src/entries.tsx
```diff
@@ -1,27 +1,24 @@
...
+export default new_defineEntries({
+  unstable_handleRequest: async (config, ctx) => {
+    const basePrefix = config.basePath + config.rscBase + '/';
+    if (ctx.req.url.pathname.startsWith(basePrefix)) {
+      const rscPath = decodeRscPath(
+        decodeURI(ctx.req.url.pathname.slice(basePrefix.length)),
+      );
+      ctx.res.body = renderRsc(config, ctx, {
+        App: <App name={rscPath || 'Waku'} />,
+      });
```
dai-shi, line 18: This might feel too low-level (it's closer to middleware api), but I can't come up with other ideas (one idea with ALS is dropped).

packages/waku/src/lib/middleware/types.ts
```diff
@@ -4,10 +4,10 @@ import type { EntriesDev, EntriesPrd } from '../../server.js';
...
 export type HandlerReq = {
-  body: ReadableStream | null;
-  url: URL;
-  method: string;
-  headers: Record<string, string>;
+  readonly body: ReadableStream | null;
+  readonly url: URL;
+  readonly method: string;
+  readonly headers: Readonly<Record<string, string>>;
 };
```
tylersayshi: Why not use the `DeepReadonly` type here?

examples/31_minimal/waku.config.ts
```diff
@@ -0,0 +1,9 @@
+/** @type {import('waku/config').Config} */
+export default {
+  middleware: () => [
+    import('waku/middleware/context'),
+    import('waku/middleware/dev-server'),
+    import('waku/middleware/handler'),
+    import('waku/middleware/fallback'),
+  ],
+};
```
tylersayshi, line 9: Is this expected to be a consumer facing api?

it seems a little weird to have all the plugins imported dynamically, but at the same time I can't think of any reasons that it isn't good. It just seems different and I am curious as to why it's like this.

examples/31_minimal/src/entries.tsx
```diff
@@ -1,27 +1,21 @@
...
-  },
-  // getBuildConfig
-  async () => [{ pathname: '/', entries: [{ rscPath: '' }] }],
-  // getSsrConfig
-  async (pathname) => {
-    switch (pathname) {
-      case '/':
-        return {
-          rscPath: '',
-          html: <Slot id="App" />,
```
tylersayshi, line 21: Why is the Slot no longer used?

packages/waku/src/lib/middleware/types.ts
dai-shi: nice call, hm, but this looks fine. just preference?

examples/31_minimal/waku.config.ts
dai-shi: customizing the entire `middleware` allows devs full flexibility. it's not for everyone, though. it's for advanced users.

Ah, I think I get what you mean. This file is temporary. When `new_defineEntries` replaces `defiineEntries`, this file will be removed.

examples/31_minimal/src/entries.tsx
dai-shi: I will work on `html` from now on.

packages/waku/src/lib/middleware/types.ts
tylersayshi: yeah this looks fine to me too, I was just curious.

examples/31_minimal/src/entries.tsx
tylersayshi: what is ALS?

dai-shi: AsyncLocalStorage

dai-shi: The API is improved since this. It now hides ctx and config.

packages/waku/src/lib/renderers/html.ts
```diff
@@ -0,0 +1,269 @@
...
+        headSent = true;
+        data = modifyHeadAndBody(data);
+        controller.enqueue(encoder.encode(data));
+        data = '';
+      }
+    },
+  });
+};
+
+// HACK for now, do we want to use HTML parser?
```
tylersayshi, line 139: html parser sounds nice for this

packages/waku/src/lib/renderers/html.ts
```diff
@@ -0,0 +1,269 @@
...
+                  return { id, chunks: [id], name };
+                }
+                const id = filePathToFileURL(file);
+                (globalThis as any).__WAKU_CLIENT_CHUNK_LOAD__(id);
+                return { id, chunks: [id], name };
+              }
+              // !isDev
+              const id = filePath.slice(config.basePath.length);
+              (globalThis as any).__WAKU_CLIENT_CHUNK_LOAD__(id);
+              return { id, chunks: [id], name };
```
tylersayshi, line 229: Might be more readable if this is put at the top of the function as an early return when `!isDev`

packages/waku/src/minimal/client.ts
```diff
@@ -0,0 +1,321 @@
...
+  renderSlot,
+}: {
+  elementsPromise: Elements;
+  renderSlot: (elements: Record<string, ReactNode>) => ReactNode;
+}) => {
+  const elements = use(elementsPromise);
+  return renderSlot(elements);
+};
+
+export const Slot = ({
```
tylersayshi, line 270: Descriptions, comments, or docs for the slots would be helpful

packages/waku/src/minimal/client.ts
```diff
@@ -0,0 +1,321 @@
...
+      callServer: (funcId: string, args: unknown[]) =>
+        callServerRsc(funcId, args, fetchCache),
+    });
+  const url = BASE_PATH + encodeRscPath(encodeFuncId(funcId));
+  const responsePromise =
+    args.length === 1 && args[0] instanceof URLSearchParams
+      ? fetch(url + '?' + args[0])
+      : encodeReply(args).then((body) => fetch(url, { method: 'POST', body }));
+  const data = enhanceCreateData(createData)(responsePromise);
+  // FIXME this causes rerenders even if data is empty
```
tylersayshi, line 119: Should this just check for if Object.keys(data).length is zero or is it more complicated?

packages/waku/package.json
```diff
@@ -42,6 +42,14 @@
...
+      "types": "./dist/minimal/server.d.ts",
+      "default": "./dist/minimal/server.js"
+    },
+    "./minimal/client": {
+      "types": "./dist/minimal/client.d.ts",
+      "default": "./dist/minimal/client.js"
+    },
     "./server": {
       "types": "./dist/server.d.ts",
       "default": "./dist/server.js"
```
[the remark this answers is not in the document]
dai-shi: bundle size doesn't matter much. it's to be explicit that minimal api is separated and for lib devs.

examples/12_nossr/waku.config.ts
```diff
@@ -1,6 +1,7 @@
 /** @type {import('waku/config').Config} */
 export default {
   middleware: () => [
+    import('waku/middleware/context'),
```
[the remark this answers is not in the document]
dai-shi: kind of. but i wish i had a different name. but hopefully one word.
`storage` (or `async-storage`) is one option, but it may confuse people.
`request-context` is another, but a bit long? what is a "request"?

btw, what i used to call "custom context" becomes "context data".

packages/waku/src/lib/builder/build.ts
```diff
@@ -148,7 +155,7 @@ const analyzeEntries = async (rootDir: string, config: ResolvedConfig) => {
   });
   const clientEntryFiles = Object.fromEntries(
     Array.from(clientFileSet).map((fname, i) => [
-      `${DIST_ASSETS}/rsc${i}-${fileHashMap.get(fname)}`,
+      `${DIST_ASSETS}/rsc${i}-${fileHashMap.get(fname) || 'lib'}`, // FIXME 'lib' is a workaround to avoid `undefined`
```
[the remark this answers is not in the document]
dai-shi: for `waku/client` and `waku/minimal/client`, which are initialized (instead of plugin adding) in L105-106.
It's rscAnalyzePlugin that is responsible for creating the file hash. feels like room for improvement.

packages/waku/src/lib/middleware/handler.ts
```diff
@@ -0,0 +1,125 @@
...
+} as const;
+const CLIENT_PREFIX = 'client/';
+
+export const handler: Middleware = (options) => {
+  const env = options.env || {};
+  setAllEnvInternal(env);
+  const entriesPromise =
+    options.cmd === 'start'
+      ? options.loadEntries()
+      : ('Error: loadEntries are not available' as never);
```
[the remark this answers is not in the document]
dai-shi: no, no. cmd can be 'dev'. i agree this is confusing code. hope there's a good way. (it's currently done like this in all middleware.)

packages/waku/src/lib/middleware/handler.ts
```diff
@@ -0,0 +1,125 @@
...
+      rsdwServer,
+      rdServer,
+      rsdwClient,
+      wakuMinimalClient,
+    };
+    const htmlHead = devServer
+      ? ''
+      : entriesPrd.dynamicHtmlPaths.find(([pathSpec]) =>
+          getPathMapping(pathSpec, ctx.req.url.pathname),
+        )?.[1] || '';
```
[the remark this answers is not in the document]
dai-shi: sure. `!devServer`.

packages/waku/src/lib/middleware/types.ts
```diff
@@ -30,6 +34,12 @@ export type HandlerContext = {
       pathname: string,
     ) => Promise<TransformStream<any, any>>;
   };
+  unstable_modules?: {
+    rsdwServer: unknown;
+    rdServer: unknown;
+    rsdwClient: unknown;
+    wakuMinimalClient: unknown;
+  };
 };
```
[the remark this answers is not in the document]
dai-shi: it's a shorthand for the library names:
https://github.com/dai-shi/waku/pull/961/files#diff-4142accf18d84717732f9aa0ed86505383a968a3369c6049dea5c620454322e0R17-R23

I wonder if we can completely avoid this approach.

packages/waku/src/lib/renderers/html.ts
```diff
@@ -0,0 +1,269 @@
...
+
+const injectHtmlHead = (
+  urlForFakeFetch: string,
+  htmlHead: string,
+  mainJsPath: string, // for DEV only, pass `''` for PRD
+) => {
+  const modifyHeadAndBody = (data: string) => {
+    const closingHeadIndex = data.indexOf(CLOSING_HEAD);
+    let [head, body] =
+      closingHeadIndex === -1
```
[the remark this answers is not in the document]
dai-shi: if you don't write `<head>`?

packages/waku/src/lib/renderers/html.ts
dai-shi: yeah, I thought so and looked for something, but performance may go down with the parser.

packages/waku/src/minimal/client.ts
dai-shi: more complicated. `data` here is a promise.

packages/waku/src/lib/renderers/html.ts
dai-shi: on second thought, it's the problem of `if (isDev)` is too long. Let's leave it for now.

packages/waku/src/lib/renderers/html.ts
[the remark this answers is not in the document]
tylersayshi: oh... haha duh. Idk what I was thinking when I wrote this one, sorry 😅

examples/12_nossr/waku.config.ts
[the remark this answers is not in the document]
tylersayshi: Should this just always be included when middleware is used?

What is the motivation for this being opt-in only?

packages/waku/package.json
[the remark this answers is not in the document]
dai-shi: Maybe maybe not. we call "The minimal API". So, I think it's fine. It's not an advanced api. It feels strange for me that defineRouter would be built on an advanced api. The "core" api can be good, but I think "minimal" is okay.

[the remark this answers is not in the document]
tylersayshi: that makes sense, thanks 👍

examples/12_nossr/waku.config.ts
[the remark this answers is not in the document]
dai-shi: 1. the core shouldn't depend on ALS as much as possible. We can use some basic features even ALS isn't available.
2. this actually allows for other middleware to use the context.

> Should this just always be included when middleware is used?

Did you notice you can even remove `rsc` middleware if you want? So, we give full capability with middleware. There's no "default" or "always" thing.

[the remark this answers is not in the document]
tylersayshi: > the core shouldn't depend on ALS as much as possible. We can use some basic features even ALS isn't available.

This makes sense, thanks! a name idea: `waku/middleware/als` - not sure if it's clear enough, but it does seem most correct once the acronym is understood

> Did you notice you can even remove rsc middleware if you want? So, we give full capability with middleware. There's no "default" or "always" thing.

Yep! this is really cool and lines up with my understanding of how these were working. My confusion was specific to why context might not always be needed, but knowing that the rest of the middleware should work without ALS makes it make sense.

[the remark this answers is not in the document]
dai-shi: > waku/middleware/als

I thought about it, but it's too confusing. It's too short. The alternative is `async-context` https://github.com/tc39/proposal-async-context, but it's not general purpose. It's technically a "request async context/storage".

PR #963, opened 2024-10-12
feat: new create pages to use new defineRouter

tylersayshi:
This adds a new version of `createPages` to use `new_defineRouter` as `new_createPages`

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -819,9 +819,11 @@ export function NewRouter({
             if (has404) {
               routerData[2] = true;
             }
-            Object.keys(rest).forEach((id) => {
-              cachedIdSetRef.current.add(id);
-            });
+            if (isStatic) {
+              Object.keys(rest).forEach((id) => {
+                cachedIdSetRef.current.add(id);
+              });
+            }
           }
         })
         .catch(() => {});
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -819,11 +819,9 @@ export function NewRouter({
             if (has404) {
               routerData[2] = true;
             }
-            if (isStatic) {
-              Object.keys(rest).forEach((id) => {
-                cachedIdSetRef.current.add(id);
-              });
-            }
+            Object.keys(rest).forEach((id) => {
+              cachedIdSetRef.current.add(id);
+            });
           }
         })
         .catch(() => {});
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -483,7 +483,7 @@ export const new_createPages = <
 ) => {
   let configured = false;
 
-  const staticPathSet = new Set<string>();
+  const staticPathMap = new Map<string, PathSpec>();
   const dynamicPagePathMap = new Map<
     string,
     [PathSpec, FunctionComponent<any>]
@@ -536,7 +536,7 @@ export const new_createPages = <
       return { numSlugs, numWildcards };
     })();
     if (page.render === 'static' && numSlugs === 0) {
-      staticPathSet.add(page.path);
+      staticPathMap.set(page.path, pathSpec);
       const id = joinPath(page.path, 'page').replace(/^\//, '');
       registerStaticComponent(id, page.component);
     } else if (
@@ -571,7 +571,10 @@ export const new_createPages = <
               break;
           }
         });
-        staticPathSet.add(page.path);
+        staticPathMap.set(
+          page.path,
+          pathItems.map((name) => ({ type: 'literal', name })),
+        );
         const id = joinPath(...pathItems, 'page');
         const WrappedComponent = (props: Record<string, unknown>) =>
           createElement(page.component as any, { ...props, ...mapping });
@@ -666,8 +669,7 @@ export const new_createPages = <
       }[] = [];
       const rootIsStatic = !rootItem || rootItem.render === 'static';
 
-      for (const path of staticPathSet) {
-        const pathSpec = parsePathWithSlug(path);
+      for (const [path, pathSpec] of staticPathMap) {
         const noSsr = noSsrSet.has(pathSpec);
 
         const pattern = path2regexp(parsePathWithSlug(path));
@@ -685,7 +687,7 @@ export const new_createPages = <
             {},
           ),
           root: { isStatic: rootIsStatic },
-          [`page:${path}`]: { isStatic: staticPathSet.has(path) },
+          [`page:${path}`]: { isStatic: staticPathMap.has(path) },
         };
 
         paths.push({
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -783,7 +783,7 @@ export const new_createPages = <
         }
       }
 
-      // loop over all layouts for route:${path}
+      // loop over all layouts for path
       const routeChildren = [
         ...layoutPaths.map((lPath) => ({
           component: Slot,
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -483,8 +483,7 @@ export const new_createPages = <
 ) => {
   let configured = false;
 
-  // TODO I think there's room for improvement to refactor these structures
-  const fixedPathSet = new Set<[string, PathSpec]>();
+  const staticPathSet = new Set<string>();
   const dynamicPagePathMap = new Map<
     string,
     [PathSpec, FunctionComponent<any>]
@@ -537,7 +536,7 @@ export const new_createPages = <
       return { numSlugs, numWildcards };
     })();
     if (page.render === 'static' && numSlugs === 0) {
-      fixedPathSet.add([page.path, pathSpec]);
+      staticPathSet.add(page.path);
       const id = joinPath(page.path, 'page').replace(/^\//, '');
       registerStaticComponent(id, page.component);
     } else if (
@@ -572,10 +571,7 @@ export const new_createPages = <
               break;
           }
         });
-        fixedPathSet.add([
-          page.path,
-          pathItems.map((name) => ({ type: 'literal', name })),
-        ]);
+        staticPathSet.add(page.path);
         const id = joinPath(...pathItems, 'page');
         const WrappedComponent = (props: Record<string, unknown>) =>
           createElement(page.component as any, { ...props, ...mapping });
@@ -670,16 +666,9 @@ export const new_createPages = <
       }[] = [];
       const rootIsStatic = !rootItem || rootItem.render === 'static';
 
-      for (const [path, pathSpec] of fixedPathSet) {
+      for (const path of staticPathSet) {
+        const pathSpec = parsePathWithSlug(path);
         const noSsr = noSsrSet.has(pathSpec);
-        const isStatic = (() => {
-          for (const [_, [layoutPathSpec]] of dynamicLayoutPathMap) {
-            if (hasPathSpecPrefix(layoutPathSpec, pathSpec)) {
-              return false;
-            }
-          }
-          return true;
-        })();
 
         const pattern = path2regexp(parsePathWithSlug(path));
 
@@ -696,14 +685,14 @@ export const new_createPages = <
             {},
           ),
           root: { isStatic: rootIsStatic },
-          [`page:${path}`]: { isStatic },
+          [`page:${path}`]: { isStatic: staticPathSet.has(path) },
         };
 
         paths.push({
           pattern,
           path: pathSpec,
           routeElement: {
-            isStatic: Object.values(elements).every((x) => x.isStatic),
+            isStatic: true,
           },
           elements,
           noSsr,
@@ -729,7 +718,7 @@ export const new_createPages = <
         paths.push({
           pattern,
           path: pathSpec,
-          routeElement: { isStatic: false },
+          routeElement: { isStatic: true },
           elements,
           noSsr,
         });
@@ -754,7 +743,7 @@ export const new_createPages = <
         paths.push({
           pattern: path2regexp(parsePathWithSlug(path)),
           path: pathSpec,
-          routeElement: { isStatic: false },
+          routeElement: { isStatic: true },
           elements,
           noSsr,
         });
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -1,7 +1,7 @@
 import { createElement } from 'react';
 import type { FunctionComponent, ReactNode } from 'react';
 
-import { unstable_defineRouter } from './define-router.js';
+import { new_defineRouter, unstable_defineRouter } from './define-router.js';
 import type { RouteProps } from './common.js';
 import {
   joinPath,
@@ -15,6 +15,7 @@ import type {
   GetSlugs,
   PropsForPages,
 } from './create-pages-utils/inferred-path-types.js';
+import { Children, Slot } from '../minimal/client.js';
 
 const hasPathSpecPrefix = (prefix: PathSpec, path: PathSpec) => {
   for (let i = 0; i < prefix.length; i++) {
@@ -174,6 +175,19 @@ const DefaultRoot = ({ children }: { children: ReactNode }) =>
     createElement('head', null),
     createElement('body', null, children),
   );
+
+const createNestedElements = (
+  elements: {
+    component: FunctionComponent<any>;
+    props?: Record<string, unknown>;
+  }[],
+) => {
+  return elements.reduceRight<ReactNode>(
+    (result, element) =>
+      createElement(element.component, element.props, result),
+    null,
+  );
+};
 
 export function createPages<
   AllPages extends (AnyPage | ReturnType<CreateLayout>)[],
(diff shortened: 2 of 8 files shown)


tylersayshi:
I'm working on replacing the structures here: https://github.com/dai-shi/waku/blob/main/packages/waku/src/router/create-pages.ts#L190-L205

with a lookup that will be from path to components for that path like is done here: https://github.com/dai-shi/waku/blob/main/examples/22_define-router/src/entries.tsx#L66-L87

Initially I thought this would be good to do after, but as I am seeing the new defineEntries style it seems clear that changing these structures first will be the best approach.

packages/waku/src/router/create-pages.tsx
```diff
@@ -457,3 +458,331 @@ export function createPages<
...
+      [`route:${page.path}`]: (
+        // layout adding is different at each level
+        // TODO loop over the path spec to add layout slots
+        <Slot id="root">
+          <Slot id={`layout:${page.path}`}>
+            <Slot id={`page:${page.path}`} />
+          </Slot>
+        </Slot>
+      ),
+      [`page:${page.path}`]: page.component({ children: <Children /> }),
```
tylersayshi, line 521: @dai-shi when are the page component's props supposed to be passed into it? typescript errors here as missing the page props, but I don't believe this is where they would actually be passed in, is it?

packages/waku/src/router/define-router.ts
```diff
@@ -379,6 +379,9 @@ export function new_defineRouter(fns: {
       pathname,
       pathStatus.isStatic ? {} : { query, skip },
     );
+    if (entries === null) {
+      return null;
+    }
```
dai-shi, line 384: When is this required?

packages/waku/src/router/create-pages.tsx
dai-shi: I think we create an anonymous component, maybe? Can you check how the current `unstable_defineRouter` behaves?

packages/waku/src/router/define-router.ts
tylersayshi: At least for now, this is what is returned when a path is requested that matches no components: https://github.com/dai-shi/waku/pull/963/files#diff-e2495518ca4a32ae11025d1ecca9cb482f9cba4f15c1d8eb8d99bc50cb5f2d13R702

dai-shi: Yeah. I wonder if this would be a hint to revisit new_defineRouter API.
Don't you feel the responsibility between `getPathConfig` and `renderRoute` is a bit vague?

tylersayshi: it does seem like there's overlapping responsibility from my point of view for now.

The weird part to me is that `getPathConfig` returns all the path specs, then `new_defineRouter` uses each of these to then request each path one at a time. 

The simplest solution that would solve this that I can think of is if createPages had a single function for handling the current split responsibilities of `renderRoute` and `getPathConfig`. So it would return a `component` per each path spec for example. 

Perhaps there are problems or limitations on control with this that you can think of maybe? 

A benefit of reducing to a single param for `new_defineRouter` would be that the skip list could be entirely managed from inside `new_defineRouter`.

As I write this out, I am realizing that this idea of consolidation could come at the cost of maybe taking away cache capabilities that `new_createPages` could later implement, so maybe this helps generate a new idea for you?

tylersayshi: @dai-shi Should I continue rewriting createPages for the current version of new_defineRouter or should we resolve this thread before continuing considering the possibility of more breaking new_defineRouter changes?

dai-shi: new_defineEntries will be more low level api, so new_defineRouter should have less responsibility and delegate to createPages.
new_defineRouter should only responsible for waku/router/client capability, and it should be as thin as possible.

I'm not sure what's your latest proposition, but if it follow the above principle, it's considerable.

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +458,277 @@ export function createPages<
...
+        ].filter(Boolean);
+        result['route:/'] = createElement(
+          Slot,
+          { id: 'root' },
+          ...routeChildren,
+        );
+      }
+
+      return processSkip(result);
+    },
```
tylersayshi, line 721: This is the bulk of what's new so far. Does this look on the right track?

I still have a lot of testing and other handling to do, but I think some initial feedback on this draft would be helpful, thanks!

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +458,277 @@ export function createPages<
...
+        return Object.fromEntries(
+          Object.entries(elements).filter(([k]) => !options.skip!.includes(k)),
+        );
+      };
+
+      // TODO it would be good if routeMap can exist as a lookup for all routes in createPages
+
+      const result: Record<string, ReactNode> = {
+        root: rootItem
+          ? rootItem.component({ children: Children() }) // Is this right?
```
dai-shi, line 692: We shouldn't call `Children` function.
```suggestion
          ? rootItem.component({ children: createElement(Children) })
```

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +458,277 @@ export function createPages<
...
+          components: { 'page:/': { isStatic: false } },
+          noSsr,
+        });
+      }
+      for (const [path, [pathSpec]] of wildcardPagePathMap) {
+        const noSsr = noSsrSet.has(pathSpec);
+        paths.push({
+          pattern: path2regexp(parsePathWithSlug(path)),
+          path: pathSpec,
+          components: { 'page:/': { isStatic: false } },
```
dai-shi, line 670: This doesn't seem correct. It should include all components for a "page". It's like all components from the current `getComponentIds`.

packages/waku/src/router/create-pages.ts
dai-shi: It may not work as is, but I think it feels on the right track.

dai-shi (commented):
sorry for the delay.

tylersayshi:
`renderRoute` is in a place where I am ready to test it

I need to work on `getPathConfig` to fix this comment now: https://github.com/dai-shi/waku/pull/963#discussion_r1817881964

I'll follow up here after fixing `getPathConfig` then digging into running and tested everything :)

tylersayshi:
@dai-shi if this code had a layout at `/bar/_layout.tsx` https://github.com/tylersayshi/waku/blob/main/examples/22_define-router/src/entries.tsx#L34-L43

would it change to the following?

```ts
      {
        pattern: '/bar',
        path: [{ type: 'literal', name: 'bar' }],
        components: {
          'route:/bar': { isStatic: true },
          root: { isStatic: true },
          'layout:/': { isStatic: true },
          'layout:/bar': { isStatic: true },
          'page:/bar': { isStatic: true },
        },
      },
```

tylersayshi:
I used the above as my working assumption (many layouts can apply)

I am ready to start testing and debugging this tomorrow. The code needs some polishing after checking things, but I think it's getting close :)

tylersayshi:
@dai-shi are you able to help me out a little with debugging this? 

image

When adding this log:

```ts
      // loop over all layouts for route:${path}
      const routeChildren = [
        ...layoutPaths.map((lPath) => ({
          component: Slot,
          props: { id: `layout:${lPath}` },
        })),
        { component: Slot, props: { id: `page:${path}` } },
      ].filter(Boolean);
      console.log(routeChildren.map(({ props }) => props.id));
      result['route:' + path] = createElement(
        Slot,
        { id: 'root' },
        createNestedElements(routeChildren),
      );

      return processSkip(result);
```

I see this as the output: `['page:/']` so I am confused why it mentions it fails to render slot with id: `page` and not `page:/`

Similar issue happens with `/bar`

image

It feels like I am missing something obvious here, but I don't know what 😅 . Any help would be great! Thanks :)

dai-shi:
> would it change to the following?

Yeah, it looks good.

> ```diff
>       {
>         pattern: '/bar',
>         path: [{ type: 'literal', name: 'bar' }],
>         components: {
>           'route:/bar': { isStatic: true },
>           root: { isStatic: true },
>           'layout:/': { isStatic: true },
> +         'layout:/bar': { isStatic: true },
>           'page:/bar': { isStatic: true },
>         },
>       },
> ```

dai-shi:
> It feels like I am missing something obvious here

It seems it's using the old waku/router/client.

It has to be NewRouter.

https://github.com/dai-shi/waku/blob/82dea924457fba1c60a9dfb071cb69dc44d7bbf2/examples/22_define-router/src/main.tsx#L7

tylersayshi:
> It has to be NewRouter.

Perfect! Seems like it all works now!

Would you like to try it out and leave a new review?

tylersayshi:
looks like the tests are breaking actually... looking into this now 👀

tylersayshi:
hmm does this output for `getPathConfig` look right?

```ts
[
  {
    pattern: '^/$',
    path: [],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/': { isStatic: true },
      'page:/': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/foo$',
    path: [ { type: 'literal', name: 'foo' } ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/foo': { isStatic: true },
      'page:/foo': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/bar$',
    path: [ { type: 'literal', name: 'bar' } ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/bar': { isStatic: true },
      'page:/bar': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/nested/baz$',
    path: [
      { type: 'literal', name: 'nested' },
      { type: 'literal', name: 'baz' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/nested/baz': { isStatic: true },
      'page:/nested/baz': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/nested/qux$',
    path: [
      { type: 'literal', name: 'nested' },
      { type: 'literal', name: 'qux' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/nested/qux': { isStatic: true },
      'page:/nested/qux': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/nested/([^/]+)$',
    path: [
      { type: 'literal', name: 'nested' },
      { type: 'literal', name: 'foo' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/nested/[id]': { isStatic: true },
      'page:/nested/[id]': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/nested/([^/]+)$',
    path: [
      { type: 'literal', name: 'nested' },
      { type: 'literal', name: 'bar' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/nested/[id]': { isStatic: true },
      'page:/nested/[id]': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/404$',
    path: [ { type: 'literal', name: '404' } ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/404': { isStatic: true },
      'page:/404': { isStatic: true }
    },
    noSsr: false
  },
  {
    pattern: '^/baz$',
    path: [ { type: 'literal', name: 'baz' } ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/baz': { isStatic: false },
      'page:/baz': { isStatic: false }
    },
    noSsr: false
  },
  {
    pattern: '^/nested/([^/]+)$',
    path: [
      { type: 'literal', name: 'nested' },
      { type: 'group', name: 'id' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/nested/[id]': { isStatic: false },
      'page:/nested/[id]': { isStatic: false }
    },
    noSsr: false
  },
  {
    pattern: '^/any/(.*)$',
    path: [
      { type: 'literal', name: 'any' },
      { type: 'wildcard', name: 'all' }
    ],
    components: {
      'layout:/': { isStatic: true },
      root: { isStatic: true },
      'route:/any/[...all]': { isStatic: false },
      'page:/any/[...all]': { isStatic: false }
    },
    noSsr: false
  }
]
```

I haven't looked super deeply into the `pattern` but that's so far all that looks a little different so far.

The issue with the tests failing though is that the layout is not being included. My latest commit: [9be57ff](https://github.com/dai-shi/waku/pull/963/commits/9be57ff7e3c84882ed5a9d2bb1c214038500cb73) I thought would bring in the layouts, but still it is missing, which is why I was looking at `getPathConfig`

Nothing inside of createPages stands out as wrong to me just yet though

tylersayshi:
Should the other examples be moved to `new_createPages` as well with this PR?

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+>(
+  fn: (fns: {
+    createPage: CreatePage;
+    createLayout: CreateLayout;
+    createRoot: CreateRoot;
+  }) => Promise<AllPages>,
+) => {
+  let configured = false;
+
+  // TODO I think there's room for improvement to refactor these structures
```
dai-shi, line 486: Do _you_ still think so?

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+      for (const [path, pathSpec] of fixedPathSet) {
+        const noSsr = noSsrSet.has(pathSpec);
+        const isStatic = (() => {
+          for (const [_, [layoutPathSpec]] of dynamicLayoutPathMap) {
+            if (hasPathSpecPrefix(layoutPathSpec, pathSpec)) {
+              return false;
+            }
+          }
+          return true;
+        })();
```
dai-shi, line 681: I personally prefer array.some to iife.

What does this `isStatic` represent? Doesn't it include checking dynamicPagePathMap?

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+              (acc, lPath) => {
+                acc[`layout:${lPath}`] = {
+                  isStatic: !dynamicLayoutPathMap.has(lPath),
+                };
+                return acc;
+              },
+              {},
+            ),
+            root: { isStatic: rootIsStatic },
+            [`route:${path}`]: { isStatic },
```
dai-shi, line 701: The `"route:*"` id can be always `{ isStatic: true }` if what it returns never change.

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+            [`page:${path}`]: { isStatic },
```
dai-shi, line 702: Shouldn't it be `!dynamicPageMap.has(path)`?

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+          },
+          noSsr,
+        });
+      }
+      for (const [path, [pathSpec]] of dynamicPagePathMap) {
```
dai-shi, line 707: Oh, I see. I think I misunderstood the code strucuture.

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,357 @@ export function createPages<
...
+          },
+          noSsr,
+        });
+      }
+      return paths;
+    },
+    renderRoute: async (path, options) => {
+      await configure();
+
+      // Should skip be a Set<string>? https://github.com/dai-shi/waku/blob/main/examples/22_define-router/src/entries.tsx#L63
```
dai-shi, line 760: It's coming from a serialized string, so `Array<string>` should be easier to deal with. (at least, for now)

dai-shi (commented):
So, I find it very difficult to prove correctness, mainly due to the lack of tests (my bad) for static/dynamic component behavior.
Let me play with it and see if I can find anything.

dai-shi:
If we made this change,

```diff
diff --git a/examples/21_create-pages/src/components/NestedBazPage.tsx b/examples/21_create-pages/src/components/NestedBazPage.tsx
index ed7488f7..96f6ddd1 100644
--- a/examples/21_create-pages/src/components/NestedBazPage.tsx
+++ b/examples/21_create-pages/src/components/NestedBazPage.tsx
@@ -2,6 +2,7 @@ const Baz = () => (
   <div>
     <h2>Nested</h2>
     <h3>Baz</h3>
+    <p>{new Date().toISOString()}</p>
   </div>
 );
 
diff --git a/examples/21_create-pages/src/entries.tsx b/examples/21_create-pages/src/entries.tsx
index e214e5a4..9dcdbe81 100644
--- a/examples/21_create-pages/src/entries.tsx
+++ b/examples/21_create-pages/src/entries.tsx
@@ -48,7 +48,7 @@ const pages = new_createPages(
     }),
 
     createPage({
-      render: 'static',
+      render: 'dynamic',
       path: '/nested/baz',
       component: NestedBazPage,
     }),
```

we should see the latest date, but it's not.

I haven't looked into it wether the issue is in `new_defineRouter` and/or `new_createPages`.

packages/waku/src/router/create-pages.ts
tylersayshi: haha 😅 maybe - I think the current structures serve their purpose well and were easy to port over to the new api, so in that way they are nice. additionally I find them well divided and easy to debug. 

I think I'm starting to appreciate them as a nice & expressive way to divide up the pages.

I think one thing we should test later on is just a stress test of adding many pages and layouts and seeing where this function slows down and what might help. I think that would be a better reason to reconsider the structures.

packages/waku/src/router/create-pages.ts
tylersayshi: the problem is that the dynamic layout path map stores the path spec in the values. Maybe this is a sign that this structure in particular could be redone to make the pathspec the key of the map, so that we can just do `dynamicLayoutPathMap.has(pathSpec)`

packages/waku/src/router/create-pages.ts
tylersayshi: to confirm: `isStatic: false` for the `route:*` would only be used by other api's and not createPages?

packages/waku/src/router/create-pages.ts
tylersayshi: yeah.. their split by map so isStatic is assumed inside the block... seems fine to me

packages/waku/src/router/create-pages.ts
tylersayshi: makes sense 👍 - wasn't sure if this crossed the client server boundary or not, so this helps my understanding, thanks!

tylersayshi:
> So, I find it very difficult to prove correctness, mainly due to the lack of tests (my bad) for static/dynamic component behavior. Let me play with it and see if I can find anything.

Changing the other `createPages` examples over to `new_createPages` should help test things a bit, right?

I can add a test for the date bug you mentioned as well though.

tylersayshi:
> we should see the latest date, but it's not.
>
> I haven't looked into it wether the issue is in new_defineRouter and/or new_createPages.

hmmm did you run `pnpm -F waku compile`?

It seems like it's updating for me

https://github.com/user-attachments/assets/983b51d9-c33c-49dd-9fa3-b38584e0dcd1

pretty sure this is the same diff but pasting to be 1000% sure haha

```diff
 !  ~/gitspace/waku   new-create-pages    git diff                                                                                     3m
diff --git a/examples/21_create-pages/src/components/NestedBazPage.tsx b/examples/21_create-pages/src/components/NestedBazPage.tsx
index ed7488f..96f6ddd 100644
--- a/examples/21_create-pages/src/components/NestedBazPage.tsx
+++ b/examples/21_create-pages/src/components/NestedBazPage.tsx
@@ -2,6 +2,7 @@ const Baz = () => (
   <div>
     <h2>Nested</h2>
     <h3>Baz</h3>
+    <p>{new Date().toISOString()}</p>
   </div>
 );
 
diff --git a/examples/21_create-pages/src/entries.tsx b/examples/21_create-pages/src/entries.tsx
index e214e5a..9dcdbe8 100644
--- a/examples/21_create-pages/src/entries.tsx
+++ b/examples/21_create-pages/src/entries.tsx
@@ -48,7 +48,7 @@ const pages = new_createPages(
     }),
 
     createPage({
-      render: 'static',
+      render: 'dynamic',
       path: '/nested/baz',
       component: NestedBazPage,
     }),
```

packages/waku/src/router/create-pages.ts
tylersayshi: actually this is being used: https://github.com/tylersayshi/waku/blob/new-create-pages/packages/waku/src/router/create-pages.ts#L324

I am good with changing to `array.some`, but I am going to do some more thinking about "What does this isStatic represent?" - I'd like to have a better answer and mostly carried this over from the original iteration of `createPages` without thinking much about it

packages/waku/src/router/create-pages.ts
dai-shi: `isStatic` is `new_defineRouter`'s capability.
`route:*`  is how we use `new_defineRouter`, so it's `new_createPages`'s responsibility.
If we return different thing for `route:*`, isStatic should be false, but if we **always** return the same thing for it, isStatic can be true and then new_defineRouter can cache the result.

dai-shi:
> It seems like it's updating for me

I'm sorry, it was about when we do client navigation. Like, clicking "Home" and then clicking "Nested / Baz".

dai-shi:
This is how it works with the current `main`.

https://github.com/user-attachments/assets/5cf872ba-7215-4998-9ee1-2747d388e692

packages/waku/src/router/create-pages.ts
tylersayshi: If I understand correct: should `isStatic` for `route:*` be true when all other components for that path are true?

I guess this also poses the question: Why do we need `route:/path` when we have all the components for the path separately. Is it to support different ways of composing the route from all the individual components?

tylersayshi:
unsure if this fix is the best way to fix the issue, but the issue with nested/baz not re-rendering on client side navigation was because `NewRouter` was cacheing the path

https://github.com/dai-shi/waku/pull/963/commits/5f927b659f66d0fd3bf618afa68b6c538bddd8ad

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,354 @@ export function createPages<
...
+            },
+            {},
+          ),
+          root: { isStatic: rootIsStatic },
+          [`page:${path}`]: { isStatic: false },
+        };
+        paths.push({
+          pattern,
+          path: pathSpec,
+          routeElement: { isStatic: false },
```
dai-shi, line 732: I think routeElement can always be static.

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,354 @@ export function createPages<
...
+            (acc, lPath) => {
+              acc[`layout:${lPath}`] = {
+                isStatic: !dynamicLayoutPathMap.has(lPath),
+              };
+              return acc;
+            },
+            {},
+          ),
+          root: { isStatic: rootIsStatic },
+          [`page:${path}`]: { isStatic },
```
dai-shi, line 699: page element is static if the page is static regardless of other layouts. `isStatic` is from the accumulation, so it seems wrong.

dai-shi (commented):
element.isStatic should only be about the element.

packages/waku/src/router/create-pages.ts
tylersayshi: good call, updated to use `staticPathSet` and simplified that structure while I was working with it

packages/waku/src/router/create-pages.ts
tylersayshi: removed the iife

packages/waku/src/router/client.ts
```diff
@@ -819,9 +819,11 @@ export function NewRouter({
             if (has404) {
               routerData[2] = true;
             }
-            Object.keys(rest).forEach((id) => {
-              cachedIdSetRef.current.add(id);
-            });
+            if (isStatic) {
```
dai-shi, line 822: We should remove this right?

packages/waku/src/router/create-pages.ts
```diff
@@ -457,3 +471,345 @@ export function createPages<
...
+      }
+
+      // loop over all layouts for path
+      const routeChildren = [
+        ...layoutPaths.map((lPath) => ({
+          component: Slot,
+          props: { id: `layout:${lPath}` },
+        })),
+        { component: Slot, props: { id: `page:${path}` } },
+      ].filter(Boolean);
```
dai-shi, line 795: Do you need `.filter(Boolean)`?

PR #964, opened 2024-10-14
chore: start GetConfig types

tylersayshi:
Using `satisfies` to help narrow response of `getConfig` with type checking and autocomplete on what is entered into the config.

### DIFF
--- a/packages/waku/src/router/base-types.ts
+++ b/packages/waku/src/router/base-types.ts
@@ -20,24 +20,7 @@ export interface CreatePagesConfig {
 export type PageProps<Path extends PagePath<CreatePagesConfig>> =
   PropsForPages<Path>;
 
-type StaticConfig = {
-  readonly render: 'static';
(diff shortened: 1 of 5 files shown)


packages/waku/src/router/base-types.ts
```diff
@@ -19,3 +19,25 @@ export interface CreatePagesConfig {
 /** Props for pages when using `createPages` */
 export type PageProps<Path extends PagePath<CreatePagesConfig>> =
   PropsForPages<Path>;
+
+type StaticConfig = {
```
dai-shi: Once again, avoid adding createPage things in base-types.ts. This file is basically for defineRouter with a few exceptions.

We can define somewhere else and re-export it from here so that it allows importing from `waku/router`.

tylersayshi: This is just for example for now. I will clean everything up when we decide on the best strategy for solving the `getConfig` problem.

PR #965, opened 2024-10-14
Get config types w fn

tylersayshi:
This is just to show what I am talking about with using a function.

Hoping to highlight the power this brings both through the step of writing the config with type safety and inferring the result without the need for `as const` `satisfies`

These screenshots show a fake example I made up where I am using the functions inference capabilities to read in the type from a list of slug tuples.

image

image

If we are continuing with generating `createPages` for the fs-router projects I think that something like this might be most ergonomic and type-safe.

If we are indeed 100% firm on not adding the function call still, I think `satisfies` will get us the most type safety, but have the gotcha of easily being forgotten. Because of this ease of forgetting, I think not including the `getConfig` in the generated types file at all would be most responsible.

This just means inferring `string` for any slug instead of having typed out staticPaths. So the final tradeoff to weigh for me is: do we want a function call that properly infers and type checks inside `getConfig` or do we want potentially looser slug types?

Happy to still discuss other options though of course

CC @dai-shi @sandren

### DIFF
--- a/packages/waku/src/router/base-types.ts
+++ b/packages/waku/src/router/base-types.ts
@@ -20,24 +20,7 @@ export interface CreatePagesConfig {
 export type PageProps<Path extends PagePath<CreatePagesConfig>> =
   PropsForPages<Path>;
 
-type StaticConfig = {
-  readonly render: 'static';
-  readonly staticPaths?: never;
+export type GetConfigResult = {
+  readonly render: 'static' | 'dynamic';
+  readonly staticPaths?: readonly string[];
 };
-
-type StaticWithSlugConfig = {
-  readonly render: 'static';
-  readonly staticPaths: readonly string[];
-};
-
-type DynamicConfig = {
-  readonly render: 'dynamic';
-};
-
-export type GetConfig =
-  | (() => Promise<StaticConfig>)
-  | (() => Promise<StaticWithSlugConfig>)
-  | (() => Promise<DynamicConfig>)
-  | (() => StaticWithSlugConfig)
-  | (() => StaticConfig)
-  | (() => DynamicConfig);
--- a/packages/waku/src/router/base-types.ts
+++ b/packages/waku/src/router/base-types.ts
@@ -20,7 +20,11 @@ export interface CreatePagesConfig {
 export type PageProps<Path extends PagePath<CreatePagesConfig>> =
   PropsForPages<Path>;
 
-export type GetConfigResult = {
-  readonly render: 'static' | 'dynamic';
-  readonly staticPaths?: readonly string[];
-};
+export const config = <
+  SP extends string,
+  Fig extends
+    | { render: 'static'; staticPaths?: SP[] | [SP, ...SP[]][] }
+    | { render: 'dynamic' },
+>(
+  c: Fig,
+) => c;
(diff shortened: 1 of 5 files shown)


dai-shi:
My preference:
1. `as const`
2. `GetConfig`
3. `satisfies`
4. 
5. 
6. 
7. function


But, anything is fine if we agree upon something. Or, start with simpler one and move to others based on user feedback (so, pick one for this approach to work well.)

tylersayshi:
Option 1: #964 shows how this would work with `as const` / `satisfies` - the generation is the same for both, but `satisfies` helps with having types colocated with where `getConfig` is written.

Option 2: This PR shows the function approach (I'm happy to avoid this considering the heavy preference to avoid a function call)

Option 3: We export `GetConfig` for help with types when writing `getConfig` then greatly simplify the generation to something like this:

```ts
import type { PathsForPages } from 'waku/router';

type Pages = {
  DO_NOT_USE_pages:
    | {
        render: 'dynamic';
        path: '/';
      }
    | {
        render: 'dynamic';
        path: '/blog/[slug]';
      }
    | {
        render: 'dynamic';
        path: '/blog';
      };
};

  interface RouteConfig {
    paths: PathsForPages<Pages>;
  }
  interface CreatePagesConfig {
    pages: Pages;
  }
}
```

This will avoid any type dependency on the rest of the codebase and only infer types for paths that reflect the fs-router pages defined.

tylersayshi:
I have worries about `as const` and `satisfies` being too much of a gotcha, so since we do not want to require a function call I think simplifying to something like option 3 is best.

tylersayshi:
Another generation option is just:

```ts
  interface RouteConfig {
    paths: ReplaceSlugs<'/' | '/blog/[slug]' | '/blog'>;
  }
}
```

This has the downside of not getting the PageProps, but removes the potentially incorrect specified `render: 'dynamic'`

dai-shi:
I'm not following the generation part, but if we pick `GetConfig` solution, what would be major drawbacks and minor drawbacks?

tylersayshi:
> I'm not following the generation part, but if we were to pick `GetConfig` solution, what would be major drawbacks and minor drawbacks?

It really just comes down to the `staticPaths` which I'd consider a minor drawback.

With the `GetConfig` strategy we cannot get the slugs for a route if they are statically defined, so we infer `string` for every slug.

An example:
```ts
// route: /foo/[slug]
export const getConfig: GetConfig = () => {
  return {
    render: 'static',
    staticPaths: ['slug1', 'slug2']
  }
}

// inferred type for route is then - `/foo/${string}`
```

Using the function call and as const/satisfies strategy results in `'/foo/slug1' | '/foo/slug2'` for this example.

I think the slugs are usually dynamically determined, like `getSlugs` for the waku website which results in `/blog/${string}` regardless of which strategy is chosen because the return type of `getSlugs` is `string[]`. This is what leads me to think this is a minor drawback because it seems rare that slugs would be statically defined in real use cases anyways. And further, `/foo/${string}` is still decent type safety.

dai-shi:
Hmmmm, I don't consider it a minor. In the long run, I think we should move to `satisfies` or function, but that makes me confident that for now `as const` would be the best.

tylersayshi:
> Hmmmm, I don't consider it a minor drawback. In the long run, I think we should move to `satisfies` or function, but that makes me confident that for now `as const` would be the best.

Do you prefer `as const` instead of `satisfies` just because it is shorter/more concise?

dai-shi:
- I believe `satisfies` hasn't be widely adopted yet.
- `satisfies` requires importing a new type (and we need to define a good name.)

PR #966, opened 2024-10-15
feat: regenerate fs-router types

tylersayshi:
This moves the generation from fs-router types off of `createPages`

Happy to move things around or update examples as needed, but I think this solution simplifies our current uses nicely

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
@@ -135,12 +135,7 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
           const hasGetConfig = fileExportsGetConfig(filePath);
 
           if (filePath.endsWith('/_layout.tsx')) {
-            fileInfo.push({
-              type: 'layout',
-              path: filePath.replace('_layout.tsx', ''),
-              src,
-              hasGetConfig,
-            });
+            continue;
           } else if (filePath.endsWith('/index.tsx')) {
             const path = filePath.slice(0, -'/index.tsx'.length);
             fileInfo.push({
--- a/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
@@ -66,7 +66,7 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
       entriesFilePossibilities = EXTENSIONS.map((ext) =>
         joinPath(config.root, opts.srcDir, SRC_ENTRIES + ext),
       );
-      outputFile = joinPath(config.root, opts.srcDir, `${SRC_ENTRIES}.gen.tsx`);
+      outputFile = joinPath(config.root, opts.srcDir, 'pages.gen.ts');
 
       try {
         const prettier = await import('prettier');
@@ -120,13 +120,14 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
         const file = readFileSync(pagesDir + filePath).toString();
 
         return (
-          file.includes('const getConfig') ||
-          file.includes('function getConfig')
+          file.includes('const getConfig =') ||
+          file.includes('function getConfig(')
         );
       };
 
       const generateFile = (filePaths: string[]): string => {
-        const fileInfo = [];
+        const fileInfo: { path: string; src: string; hasGetConfig: boolean }[] =
+          [];
         const moduleNames = getImportModuleNames(filePaths);
 
         for (const filePath of filePaths) {
@@ -139,14 +140,12 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
           } else if (filePath.endsWith('/index.tsx')) {
             const path = filePath.slice(0, -'/index.tsx'.length);
             fileInfo.push({
-              type: 'page',
               path: path || '/',
               src,
               hasGetConfig,
             });
           } else {
             fileInfo.push({
-              type: 'page',
               path: filePath.replace('.tsx', ''),
               src,
               hasGetConfig,
--- a/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
@@ -1,8 +1,9 @@
 import type { Plugin } from 'vite';
 import { readdir, writeFile } from 'node:fs/promises';
-import { existsSync, readFileSync } from 'node:fs';
+import { existsSync } from 'node:fs';
 import { SRC_ENTRIES, EXTENSIONS } from '../constants.js';
 import { joinPath } from '../utils/path.js';
+import { parseFileSync } from '@swc/core';
 
 const SRC_PAGES = 'pages';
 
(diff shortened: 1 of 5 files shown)


tylersayshi:
marking draft to fix the test where this can be generated to a broken file: 

image

tylersayshi:
this: https://github.com/dai-shi/waku/pull/966#issuecomment-2415080194 is fixed ✅

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -120,13 +120,14 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
         const file = readFileSync(pagesDir + filePath).toString();
 
         return (
-          file.includes('const getConfig') ||
-          file.includes('function getConfig')
+          file.includes('const getConfig =') ||
+          file.includes('function getConfig(')
```
dai-shi, line 124: wow, this looks too naive. i wonder what's the best practice with typical vite plugins.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -135,57 +136,54 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
           const hasGetConfig = fileExportsGetConfig(filePath);
 
           if (filePath.endsWith('/_layout.tsx')) {
-            fileInfo.push({
-              type: 'layout',
-              path: filePath.replace('_layout.tsx', ''),
-              src,
-              hasGetConfig,
-            });
+            continue;
```
dai-shi, line 139: Does this mean we no longer check types for `_/layout.tsx`?

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: I completely agree. I was looking into trying to get an AST on demand of each file and I couldn't figure that out. 

It seemed like in vite, the AST is made available on the transform step, but I couldn't find a function to just call and pass text of a file. 

I think this is something we could add babel to achieve, but we also want to avoid adding dependencies, so I hesitated to do that.

But I believe an AST of each file would be the correct solution for checking for named exports.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: Did you mean "/_layout.tsx" btw?

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: Vite does export a function called parseAST which comes from rollup.

I was not able to get this to work and from searching around, it seemed undocumented even in the rollup docs.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: yeah `/_layout.tsx`. What I wonder is, previously, it's generated so if we mistype `render: 'sttatic'` (with `as const`), TS can catch it. But, isn't it no longer the case with this change?

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: Hm, it should be something simpler.
How about evaluating the file content with node:vm?
https://nodejs.org/api/vm.html
I'm totally unsure if it's feasible.

dai-shi: otherwise, we can just use swc for now as we do in other plugins.

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: This would result in `render: 'dynamic'` for the type since that is the fallback for when render does not match static or dynamic.

So this would really just change the types if `staticPaths` was added in addition to the `sttatic` typo. But the behavior would be to just assume string for the slugs, and not error otherwise.

I think if we want to catch typos like that we should move to using `satisfies` probably

dai-shi: I thought with `as const`, this typo can be detected. (If your concern is the TS/JS behavioral mismatch, we can add console.warn for JS, but technically, it's another story.)

tylersayshi: > I thought with as const, this typo can be detected.

How would you expect it to be detected?

> If your concern is the TS/JS behavioral mismatch

this ^ is not a concern here. Js is simply ignored because we are generating types for ts.

dai-shi: because with `as const`, the `render` type is `'sttatic'` literal. Oh, I think you mean how I would notice it. Basically, CI or running tsc command.

tylersayshi: Gotcha! I was able to fix this so that it errors now :)

image

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
tylersayshi: vm doesn't seem to work. It seems like `SourceTextModule` is the closest to what we'd need, but as far as I can tell, it does not have support for working with typescript or jsx and we will need both.

Trying swc now...

tylersayshi: swc works! 🥳 Thanks for the recommendation!!

I'll update with a fix for that check shortly now :)

tylersayshi: fixed!

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
dai-shi: Hmm, I can't reproduce this with 28204476f3d2a119177ad8ce1f005b6eaa68caeb.

tylersayshi: hm, it still seems to work for me. Are you just making the typo in one of the `getConfig`'s in 01_template still?

dai-shi: yes, `/_layout.tsx`.
here's the change:
```diff
diff --git a/examples/01_template/src/pages/_layout.tsx b/examples/01_template/src/pages/_layout.tsx
index 6d227c9f..bc360ef1 100644
--- a/examples/01_template/src/pages/_layout.tsx
+++ b/examples/01_template/src/pages/_layout.tsx
@@ -34,6 +34,6 @@ const getData = async () => {
 
 export const getConfig = async () => {
   return {
-    render: 'static',
+    render: 'sttatic',
   } as const;
 };
```

I get the JS runtime error:
```
Cannot process SSR Error: Invalid layout configuration
```
(I didn't know this is already implemented. 😓 )

tylersayshi: Ah! Yeah this wouldn't currently get checked because the generated types file only reads from pages, not layouts.

If we want to type check for scenarios like this I would recommend the `satisfies` or `GetConfig` approaches we've talked through.

tylersayshi: > I thought this continue was related to this change
Yes, the continue is where I am skipping `_layout.tsx` files for what gets written to pages.gen.ts

> my question was why it needs satisfies, because it seems to me as const can infer to a literal.
So satisfies is not necessary, you are correct that `as const` infers literal. 

So I think what you are leading towards is generating the import and typecheck for the layout files in `pages.gen.ts`. This is possible and I can change this to do that. My hesitance is that it seems like it would be better to check for typos where `getConfig` is defined and not in a separate generated file. 

If this is what you'd prefer though, that is fine, I'll change to add a check.

dai-shi: > it would be better to check for typos where getConfig is defined and not in a separate generated file.

I agree it's better.

My original question was that this leads to capability regression with this PR, but as we already have the validity check in the JS runtime, I notice it's not super important. So, let's leave it. Thanks for the explanation.

dai-shi (approved):
This is a follow-up to #886.

PR #968, opened 2024-10-18
chore: run tsc on examples

dai-shi:
LIMITATION: It depends on `find` and doesn't run on windows.

### DIFF
--- a/examples/43_weave-render/src/components/BarLayout.tsx
+++ b/examples/43_weave-render/src/components/BarLayout.tsx
@@ -41,7 +41,7 @@ const BarLayout = ({ children }: { children: ReactNode }) => {
         </li>
         <li>
           <Link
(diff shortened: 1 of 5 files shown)


examples/43_weave-render/src/components/BarLayout.tsx
```diff
@@ -41,7 +41,7 @@ const BarLayout = ({ children }: { children: ReactNode }) => {
         </li>
         <li>
           <Link
-            to="/nested/bar"
+            to={'/nested/bar' as never}
```
dai-shi, line 44: @tylersayshi does it look okay??

tylersayshi: Sorry, what is the goal of this?

PR #970, opened 2024-10-18
update react rc

dai-shi:


dai-shi:
https://github.com/dai-shi/waku/pull/933#issuecomment-2392649806 seems to be fixed.

PR #974, opened 2024-10-18
refactor: move from ncc to tsup for create-waku

tylersayshi:
It looks like a bundler is needed for helping to minify the published code and transpile from typescript.

I've had a good personal experience with tsup and I saw @himself65 mention it, so figured I'd make this PR to help :)

Closes #973

PR #978, opened 2024-10-19
fix: remove createRequire for package json

tylersayshi:
follow up to this: https://github.com/dai-shi/waku/pull/974#issuecomment-2423476651

### DIFF
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -161,7 +161,9 @@ Options:
 }
 
 async function notifyUpdate() {
-  const packageJson = JSON.parse(readFileSync('../package.json'));
+  const packageJson = JSON.parse(readFileSync('../package.json', {
+    encoding: 'utf8',
+  }));
   const result = await checkForUpdate(packageJson).catch(() => {});
   if (result?.latest) {
     console.log(`A new version of 'create-waku' is available!`);
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -161,11 +161,7 @@ Options:
 }
 
 async function notifyUpdate() {
-  const packageJson = JSON.parse(
-    readFileSync('../package.json', {
-      encoding: 'utf8',
-    }),
-  );
+  const packageJson = JSON.parse(readFileSync('../package.json', 'utf8'));
   const result = await checkForUpdate(packageJson).catch(() => {});
   if (result?.latest) {
     console.log(`A new version of 'create-waku' is available!`);
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -1,4 +1,4 @@
-import { existsSync, readdirSync } from 'node:fs';
+import { existsSync, readdirSync, readFileSync } from 'node:fs';
 import fsPromises from 'node:fs/promises';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
@@ -161,8 +161,7 @@ Options:
 }
 
 async function notifyUpdate() {
-  // eslint-disable-next-line @typescript-eslint/no-require-imports
-  const packageJson = require('../package.json');
+  const packageJson = JSON.parse(readFileSync('../package.json'));
   const result = await checkForUpdate(packageJson).catch(() => {});
   if (result?.latest) {
     console.log(`A new version of 'create-waku' is available!`);
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -7,7 +7,6 @@ import { default as prompts } from 'prompts';
 import { red, green, bold } from 'kolorist';
 import fse from 'fs-extra/esm';
 import checkForUpdate from 'update-check';
-import { createRequire } from 'node:module';
 import {
   getTemplateNames,
   installTemplate,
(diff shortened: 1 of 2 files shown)


packages/create-waku/tsup.config.ts
```diff
@@ -2,8 +2,10 @@ import { defineConfig } from 'tsup';
 
 export default defineConfig((options) => ({
   entry: ['src/index.ts'],
+  external: [/package\.json/],
   format: ['esm'],
   minify: !options.watch,
+  clean: true,
   /** @see https://github.com/egoist/tsup/issues/927#issuecomment-2354939322 */
   banner: {
     js: `import { createRequire } from 'module'; const require = createRequire(import.meta.url);`,
```
ojj1123, line 11: You removed `createRequire`, but Is it needed?

ojj1123 (commented):
👍👍

packages/create-waku/tsup.config.ts
tylersayshi: We could look into moving to inquirer.js from prompts:

https://github.com/SBoudrias/Inquirer.js

It seems like it is more actively maintained, so I'd imagine the esm support is more up to date and we could then remove this banner.

ojj1123: Yeah It feels like prompts is not almost maintained. But both of them have similar features. I wonder @dai-shi thoughts :)

dai-shi: Previously, I rejected the idea #339 by @himself65 because I thought it was too early.
But now that we have more active contributors, I think the time is right.
I don't think prompts is too bad, but if you guys have a better alternative, and feel like maintaining it, I'm fine of changing the dependency library, or completely re-implementing it.

packages/create-waku/src/index.ts
```diff
@@ -162,9 +161,7 @@ Options:
 }
 
 async function notifyUpdate() {
-  // keep original require to avoid
-  //  bundling the whole package.json by `@vercel/ncc`
-  const packageJson = createRequire(import.meta.url)('../package.json');
+  const packageJson = await import('../package.json');
```
dai-shi, line 164: So, is this syntax supported by tsup?

tylersayshi: yep! I think esbuild is technically what supports it under the hood, but still the answer is yes, it's supported.

packages/create-waku/tsup.config.ts
tylersayshi: I tried both enquirer and inquirer and each seems to use dynamic requires of their dependencies, so we'd need to keep this banner even if switching.

So, I don't really see a strong motivator for changing off prompts either yet. 

This issue: https://github.com/terkelg/prompts/issues/393 if it get's worked on should be enough to solve any of the current concerns we have now.

packages/create-waku/src/index.ts
tylersayshi: Actually I was wrong here ^

This errors because of the json import without the attribute, so I just changed it to use `require` which is defined in the banner

the error:

```
TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///.../waku/packages/create-waku/package.json" needs an import attribute of type "json"
```

packages/create-waku/tsup.config.ts
ojj1123: Yeah, That's enough since create-waku doesn't have the logic of complex CLI yet.
So I don't have enough motivation on changing other deps.

packages/create-waku/src/index.ts
```diff
@@ -162,9 +161,8 @@ Options:
 }
 
 async function notifyUpdate() {
-  // keep original require to avoid
-  //  bundling the whole package.json by `@vercel/ncc`
-  const packageJson = createRequire(import.meta.url)('../package.json');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const packageJson = require('../package.json');
```
himself65, line 165: you can use `JSON.parse(fs.readFileSync(..))` if wanna remove require here

tylersayshi: Tried updating this from my phone. I think it's right but I'll need to correct the prettier issues from my laptop later haha

dai-shi: I thought there was a motivation of preferring `require` to `readFileSync`? @himself65

packages/create-waku/src/index.ts
```diff
@@ -162,9 +161,11 @@ Options:
 }
 
 async function notifyUpdate() {
-  // keep original require to avoid
-  //  bundling the whole package.json by `@vercel/ncc`
-  const packageJson = createRequire(import.meta.url)('../package.json');
+  const packageJson = JSON.parse(
+    readFileSync('../package.json', {
+      encoding: 'utf8',
+    }),
```
dai-shi, line 167: We use the shorthand everywhere else.
```suggestion
    readFileSync('../package.json', 'utf8'),
```

packages/create-waku/src/index.ts
himself65: Nope, I just think it’s not a familiar way to use require in ESM

dai-shi (approved):
Thanks guys!!

packages/create-waku/src/index.ts
dai-shi: https://github.com/dai-shi/waku/pull/978#issuecomment-2425461081

PR #980, opened 2024-10-20
fix(plugin/fs-router-typegen): use swc.parseSync instead of parseFileSync

dai-shi:
It fails in StackBlitz as `@swc/was` will be used.


dai-shi:
```
TypeError: swc.parseFileSync is not a function
```
Hm, it doesn't work either.

dai-shi:
Ah, in StackBlitz, `@swc/wasm` is used.

tylersayshi:
> Ah, in StackBlitz, `@swc/wasm` is used.

I'm confused with this. Why was this an issue?

dai-shi:
`@swc/wasm` doesn't export `parseFileSync`.
https://unpkg.com/browse/@swc/wasm@1.7.36/wasm.js

PR #981, opened 2024-10-21
fix: issues with cli reading from package.json

tylersayshi:
fix for https://github.com/dai-shi/waku/pull/978#issuecomment-2425461081

Also fixed an issue where the default folder name was not showing on conflict

### DIFF
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -161,7 +161,7 @@ Options:
 
 async function notifyUpdate() {
   const packageJson = JSON.parse(
-    readFileSync(path.resolve(import.meta.dirname, '../package.json'), 'utf8'),
+    readFileSync(new URL('../package.json', import.meta.url), 'utf8'),
   );
   const result = await checkForUpdate(packageJson).catch(() => {});
   if (result?.latest) {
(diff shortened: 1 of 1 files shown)


packages/create-waku/src/index.ts
```diff
@@ -161,7 +160,9 @@ Options:
 }
 
 async function notifyUpdate() {
-  const packageJson = JSON.parse(readFileSync('../package.json', 'utf8'));
+  const packageJson = JSON.parse(
+    readFileSync(path.resolve(import.meta.dirname, '../package.json'), 'utf8'),
```
dai-shi, line 164: `import.meta.dirname` isn't available in Node 18.

https://nodejs.org/docs/latest-v18.x/api/esm.html#importmetaurl

Can you try this?
```js
import { readFileSync } from 'node:fs';
const buffer = readFileSync(new URL('./data.proto', import.meta.url));
```

dai-shi (commented):
Thanks for the quick fix. See the comment.

packages/create-waku/src/index.ts
tylersayshi: yep, that works, thanks!

dai-shi (approved):
thx!

PR #984, opened 2024-10-29
fix(create-waku): package.json not found

### LINKED ISSUE #983: create-waku Error: ENOENT: no such file or directory, open '../package.json'
The path that create waku is run is inconsistent from when we run it inside the project to when `npm create waku@latest` is run.

This causes the package.json to not be found, so we'll need to adjust for this.

I'm happy to dig into this, but wanted to note it here as being tracked.

tylersayshi:
closes #983

I don't think this can be tested without being published, but I believe this is the fix to the problem.

@dai-shi would you be willing to publish some sort of pre-release version to test this if you agree that it should work?

### DIFF
--- a/packages/create-waku/package.json
+++ b/packages/create-waku/package.json
@@ -16,7 +16,8 @@
   },
   "files": [
     "dist",
-    "template"
+    "template",
+    "package.json"
   ],
   "scripts": {
     "start": "node ./dist/index.js",


tylersayshi:
Oh, I think the issue is just that we need to publish a new version with the latest fix 🤦

dai-shi:
Right, it's not published yet.

PR #987, opened 2024-11-05
feat(router): useRouter with types

tylersayshi:
I'd really like feedback here. I left my notes and thoughts in `docs/use-router.mdx`

### DIFF
--- a/docs/use-router.mdx
+++ b/docs/use-router.mdx
@@ -1,0 +1,96 @@
+---
+slug: use-router
+title: useRouter (experimental)
+description: API for using the router.
+draft: true # Where does the frontmatter get read?
+---
+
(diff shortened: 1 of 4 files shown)


tylersayshi:
If we want to lean into using `nuqs`, I’m thinking of building it in similar to how trpc uses react query as a shim

Something like `nuqs.aboutPage.useQueryState(“age”)`

The parser and key pair would be defined where the route config is defined (entries.tsx or getConfig)

Also, the default useQueryState from `nuqs` should work, but this would be for route level organization.

Still just early thoughts ofc

tylersayshi:
closing this in favor of using external libs for types query param state for now

https://github.com/dai-shi/waku/discussions/1123

PR #945, opened 2024-10-05
feat: enable nodejs 22 support

himself65:

PR #988, opened 2024-11-05
run tests on all node versions

dai-shi:
#945 wasn't enough. cc @himself65

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,13 +12,18 @@ concurrency:
 
 jobs:
   test:
+    name: Test on (Node ${{ matrix.version }})
+    strategy:
+      fail-fast: false
+      matrix:
+        version: [18.17.0, 20.8.0, 22.7.0]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: ${{ matrix.version }}
           cache: 'pnpm'
           cache-dependency-path: '**/pnpm-lock.yaml'
       - run: pnpm install --frozen-lockfile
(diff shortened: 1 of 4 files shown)


dai-shi:
Hey @cjpais, would you be able to help on this?

https://github.com/dai-shi/waku/actions/runs/11680436701/job/32523455047?pr=988

```
 FAIL  tests/buffer.test.ts > parseFormData > should handle base64 encoded content
AssertionError: expected '\u001d\ufffde\ufffdj+\ufffd' to be 'Hello, World!' // Object.is equality

Expected: "Hello, World!"
Received: "�e�j+�"
```

It fails only on Node 22.

himself65 (approved):
LGTM

himself65 (commented):
it's undici, just check the change is from 6.7.0 to 6.7.1

himself65:
They changed the formData implmenetation at some moment 

https://github.com/nodejs/undici/commit/5bb84bafef7421e5b8f6b93cad80c3836796aedb#diff-1ca2b01f3e34a5e04ad9a3d2d337c089326920a9c4f96edaea22ea37abe89745

dai-shi:
Do you mean this? https://github.com/nodejs/undici/releases/tag/v6.7.1

Any idea how we can properly handle it?

himself65:
> Do you mean this? https://github.com/nodejs/undici/releases/tag/v6.7.1
> 
> Any idea how we can properly handle it?

We can do something like

```ts
const needPolyfillPromise = parseFormData(...)
```

himself65:
We will drop nodejs 18 after few months evetually. I think just do some check is ok

tylersayshi:
Should dropping node 18 support happen before v1 is released? Otherwise it'll likely need to wait for a v2 given that that's a fairly significant breaking change?

dai-shi:
node 18 will be dropped at some point, but the issue exists in node 20 too.
I hope @cjpais would take some time and come up with a conclusion.

PR #989, opened 2024-11-06
Switch from Cloudflare Pages to Workers with Static Assets

### LINKED ISSUE #944: Switch back to Cloudflare Workers
Cloudflare recently launched this and it makes sense for us to switch back to it from Cloudflare Pages.
https://developers.cloudflare.com/workers/static-assets/

Other framework implementations are here - https://github.com/cloudflare/workers-sdk/tree/main/packages/create-cloudflare/templates-experimental

I'd also love to see waku included in their repo.

rmarscher:
Fixes #944

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
@@ -5,8 +5,8 @@ import {
   mkdirSync,
   readdirSync,
   renameSync,
-  rmSync,
   writeFileSync,
+  copyFileSync,
 } from 'node:fs';
 import type { Plugin } from 'vite';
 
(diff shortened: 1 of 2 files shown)


packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
```diff
@@ -62,25 +62,20 @@ export default {
...
-      getFiles(fullPath, files);
-    } else {
-      files.push(fullPath);
+const copyFiles = (srcDir: string, destDir: string, extensions: string[]) => {
+  const files = readdirSync(srcDir, { withFileTypes: true });
+  for (const file of files) {
+    const srcPath = path.join(srcDir, file.name);
+    const destPath = path.join(destDir, file.name);
+    if (file.isDirectory()) {
+      mkdirSync(destPath, { recursive: true });
```
tylersayshi, line 71: does this recursive option work on node 18? it might have been with another function call but I think we saw that this got added in node 20 for something else, so it seems good to double check.

packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
```diff
@@ -136,93 +131,36 @@ export function deployCloudflarePlugin(opts: {
...
+      // Traverse assetsDistDir and copy specific files to workerDistDir/public
+      const workerPublicDir = path.join(workerDistDir, DIST_PUBLIC);
+      mkdirSync(workerPublicDir, { recursive: true });
+      copyFiles(assetsDistDir, workerPublicDir, [
+        '.txt',
+        '.html',
+        '.json',
+        '.js',
+        '.css',
+      ]);
```
tylersayshi, line 160: would this break if there were files like a CNAME of favicon?

packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
dai-shi: https://nodejs.org/docs/latest-v18.x/api/fs.html#fsmkdirsyncpath-options

Seems OK with 18.

dai-shi:
@rmarscher Please check the CI error. It fails on Windows.

rmarscher:
> @rmarscher Please check the CI error. It fails on Windows.

I see. Hmm... maybe it needs to get an OS temp dir.

```
EPERM: operation not permitted, rename 'D:\a\waku\waku\examples\31_minimal\dist' -> 'D:\a\waku\waku\examples\31_minimal\_dist'
```

packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
rmarscher: I don't think so. The purpose of this is to copy built public assets that might imported during SSR.

It would be better if we had a manifest of these files to copy rather than making assumptions by file extension.

packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
rmarscher: I think it was readdir's recursive option that isn't consistent across node versions. I think mkdir recursive is more stable.

dai-shi (approved):

PR #969, opened 2024-10-18
chore: run tsc examples 2

dai-shi:
another take on #968.

PR #990, opened 2024-11-06
chore: type checking for examples

tylersayshi:
follow up to #969

putting this up to call out a couple things:

1. I think just adding references to each example in the root level tsconfig is probably the easiest way to do typechecking unless we want to keep something like the current bash script:

```sh
find examples -mindepth 1 -maxdepth 1 -type d ! -name '*_js'
```

2. for "test" it seems to me like `--noEmit` should be true for all cases. emitting types should be the responsibility of our `compile` step, right?

If we like this approach I'll fix the type issue and add the rest of the examples as references (basically the same as #969 haha)

let me know if i'm missing anything ofc though

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -43,6 +43,9 @@
     // Test
     {
       "path": "./packages/waku/tests"
+    },
+    {
+      "path": "./examples/21_create-pages"
     }
   ]
 }
--- a/examples/21_create-pages/src/entries.tsx
+++ b/examples/21_create-pages/src/entries.tsx
@@ -42,7 +42,7 @@ const pages = createPages(async ({ createPage, createLayout, createRoot }) => [
   }),
 
   createPage({
-    render: 'dynamic',
+    render: 'dynamic_badVal',
     path: '/baz',
     // Inline component is also possible.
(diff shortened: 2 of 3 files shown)


tsconfig.json
```diff
@@ -43,6 +43,9 @@
     // Test
     {
       "path": "./packages/waku/tests"
+    },
+    {
+      "path": "./examples/21_create-pages"
```
dai-shi, line 48: I think there was a reason we removed example entries here (we used have them.)
@himself65 Do you remember?

tylersayshi: a friendly reminder on this :) - would be nice to get type checking in for these

dai-shi: Yeah, I don't remember the drawback of this. Maybe just we forget adding the entry when we add a new example. (In that sense, `find` approach isn't that bad.)

himself65: --noEmit will still emit files to `lib` or `build` for a ts monorepo

tylersayshi: That is unfortunate... `find` seems like a good solution to me as well then.

tylersayshi:
closing in favor of current find script

PR #991, opened 2024-11-07
fix: new_defineRouter

dai-shi:


packages/waku/src/router/define-router.ts
```diff
@@ -368,17 +374,41 @@ export function new_defineRouter(fns: {
...
+    }
+    const entries = {
+      ...elements,
+      [ROUTE_SLOT_ID_PREFIX + pathname]: routeElement,
+    };
+    for (const skipId of skip) {
+      if (await canSkip(pathname, skipId)) {
+        delete entries[skipId];
+      }
+    }
```
tylersayshi, line 411:
```suggestion
    const toSkip = await Promise.all(skip.map((skipId) => canSkip(pathname, skipId));
    for (const skipId of toSkip) {
      delete entries[skipId];
    }
```

packages/waku/src/router/define-router.ts
```diff
@@ -368,17 +374,41 @@ export function new_defineRouter(fns: {
...
+    const { routeElement, elements } = await fns.renderRoute(
       pathname,
-      pathStatus.isStatic ? {} : { query, skip },
+      pathStatus.isStatic ? {} : { query },
     );
+    if (
+      Object.keys(elements).some((id) => id.startsWith(ROUTE_SLOT_ID_PREFIX))
+    ) {
+      throw new Error('Element ID cannot start with "route:"');
+    }
```
tylersayshi, line 402: Is this check needed if you're moving to use `routeElement` also?

packages/waku/src/router/define-router.ts
```diff
@@ -368,17 +374,41 @@ export function new_defineRouter(fns: {
           has404,
         };
   };
+  const canSkip = async (
+    pathname: string,
+    slotId: SlotId,
+  ): Promise<boolean> => {
+    const pathConfig = await getMyPathConfig();
```
tylersayshi, line 381: Actually, should `canSkip` take the result of `getMyPathConfig()` as a parameter? It seems unnecessary to await it for every pathname checked.

examples/22_define-router/src/entries.tsx
```diff
@@ -47,109 +47,111 @@ export default new_defineRouter({
...
+          ),
+          'page:/bar': <BarPage />,
+        },
+      };
     }
     if (path === '/nested/baz') {
-      return processSkip({
-        'route:/nested/baz': (
+      return {
+        routeElement: (
```
tylersayshi, line 134: nice! what made you change your mind on this?

tylersayshi (commented):
lgtm 👍

packages/waku/src/router/define-router.ts
dai-shi: This is needed to avoid conflict because we move to `routeElement`.

packages/waku/src/router/define-router.ts
dai-shi: it's possible. await just waits one microtask in this case, so I didn't care much.

packages/waku/src/router/define-router.ts
tylersayshi: Yes, thanks!

packages/waku/src/router/define-router.ts
tylersayshi: Ok!

examples/22_define-router/src/entries.tsx
dai-shi: Do you mean moving skip handling from outside to inside? It's clearer abstraction.
The drawback is that now the user of new_defineRouter doesn't have "skip" information, they can't optimize based on it. (They have to return elements that will be skipped. So, theoretically, it's the waste.)

packages/waku/src/router/define-router.ts
dai-shi: I don't disagree it's not ideal and there is room for improvement. Let's revisit it later as we will be rewriting this anyway.

examples/22_define-router/src/entries.tsx
tylersayshi: thanks, ya that makes sense

packages/waku/src/router/define-router.ts
dai-shi: canSkip also waits only a microtask. If we were to improve it, there must be a better way. So, l'd leave it for now.

PR #1000, opened 2024-11-12
refactor: render-type to use new_createPages and fix layouts for new_createPages

tylersayshi:
moving e2e test example to use `new_createPages`

this also fixes the layouts returned from `renderRoute` to re-use the `pathSpec` of each route to detect nested layouts.

it also fixes the page mapping issue where the props were not yet being injected and now are. This is for the slugs of the route.


packages/waku/src/router/create-pages.ts
```diff
@@ -755,29 +775,37 @@ export const new_createPages = <
...
           rootItem ? rootItem.component : DefaultRoot,
           null,
           createElement(Children),
         ),
-        [`page:${path}`]: createElement(
+        [`page:${routePath}`]: createElement(
           pageComponent,
           null,
           createElement(Children),
         ),
```
tylersayshi, line 798: @dai-shi the page props are not being passed into the page component on the server which causes the test to fail. 

this is something that I was assuming defineRouter would be responsible for. Any thoughts?

dai-shi: Page props are `slug` names, right? It's called `mapping` in the code.
No, it's new_createPages that is responsible for it.

I missed it too, but we need a wrapper component like this:
https://github.com/dai-shi/waku/blob/33bea3d603067c5bc0b64a9db78a41c3a5aff800/packages/waku/src/router/create-pages.ts#L430-L431

tylersayshi:
Just realized this is #1000

🎉🍻

packages/waku/src/router/create-pages.ts
tylersayshi: I partially understand this, but some more explanation might be needed here.

`WrappedComponent` is a `FunctionComponent` and the return type of `renderRoute` is `Record<string, ReactNode>`

If we use WrappedComponent as the return for `page${routePath}`, should the return type be changed to `Record<string, ReactNode | FunctionComponent>` or did you mean to do something different entirely?

----

A separate thing I noticed is that `getPathConfig` has the mapping for each of the static paths. I did a naive attempt to get the mapping here:

https://github.com/tylersayshi/waku/blob/refactor-e2e-test-createPages/packages/waku/src/router/create-pages.ts#L788-L789

this is very clearly wrong when I test it, how would you do this here?

dai-shi: You are right. `renderRoute` returns `Record<string, ReactNode>`, so we need to create the mapping in the `renderRoute` function. I'm not sure if `getPathConfig` works, but if you get the requirement, I'll leave it to you for now. Let me know if you need any further help.

tylersayshi: ah I just had a typo haha - that was way simpler to fix than I expected :)

packages/waku/src/router/create-pages.ts
```diff
@@ -755,29 +771,39 @@ export const new_createPages = <
...
         root: createElement(
           rootItem ? rootItem.component : DefaultRoot,
           null,
           createElement(Children),
         ),
-        [`page:${path}`]: createElement(
+        [`page:${routePath}`]: createElement(
           pageComponent,
-          null,
+          mapping,
```
tylersayshi, line 795: @dai-shi should `query` and `hash` also be sent in here?

dai-shi: If a page is dynamic, I think it should receive `query`.

tylersayshi: thanks, and `hash` only exists on the client, so it is handled inside `NewRouter`? is that right?

tylersayshi: so there's a bug upstream of new_createPages. I think it's upstream of defineRouter too. 

This prevents the use of query here. I'll open a new PR to talk through a fix for that, but this PR seems like a fine change on its own to me

tylersayshi: #1004 will fix the `query`

dai-shi: thanks. yeah, `hash` is client only.

dai-shi (approved):
great to fix it.

PR #1003, opened 2024-11-15
refactor: fs-router to use new_createPages & fix staticPaths

tylersayshi:
Changes:

- pages are built in slices of 2500 to avoid running out of memory when running `Promise.all`
- this moves fs-router to use new_createPages


tylersayshi:
@dai-shi the e2e test is failing because it seems like the 401 error is no longer caught correctly. Is this meant as a responsibility of `NewRouter` or `ErrorBoundary`?

Or do I need to try/catch the `createElement` call on the page component or something else in `new_createPages`?

dai-shi:
> it seems like the 401 error is no longer caught correctly.

I'm not sure what broke it. Can you look into it? It's probably my oversight somewhere.

tylersayshi:
@dai-shi for the error boundary issue here... this screenshot shows in order:

1. a console error for the initial 401 for the route (expected)
2. a log I put in `checkStatus` where we throw the error from: https://github.com/dai-shi/waku/blob/main/packages/waku/src/minimal/client.ts#L31
3. react console errors then hits the default error boundary: https://github.com/dai-shi/waku/blob/main/packages/waku/src/router/client.ts#L612-L635 

image

The test fails because the error is being thrown from here: https://github.com/dai-shi/waku/blob/main/packages/waku/src/minimal/client.ts#L208

The Root is a parent of the `ErrorBoundary` used in the `ssr-catch-error` test, so it makes sense that this would not be caught.

With all of this, I am not sure how we should adjust:
Should the ErrorBoundary for this test be moved to a custom Root component?
Should we be throwing the error from some other component? Like the Slot for the page component?
Something else?

dai-shi:
> The test fails because the error is being thrown from here

Do you mean `fetchRsc` throws?

tylersayshi:
> > The test fails because the error is being thrown from here
> 
> Do you mean `fetchRsc` throws? It should return a promise without throwing.

`fetchRsc` throws because `checkStatus` throws inside `fetchRSC` and the error is not caught inside of `fetchRsc`

packages/waku/src/router/create-pages.ts
```diff
@@ -588,7 +588,7 @@ export const new_createPages = <
           }
         });
         staticPathMap.set(
-          page.path,
+          '/' + pathItems.join('/'),
```
dai-shi, line 591: Would you like to add unit tests for this?

tylersayshi: yea that sounds like a good idea 👀

tylersayshi: added a test for this, but I need to move `createPages.test.ts` to use `new_createPages` next. I'll make a new PR for that if it's ok

tylersayshi:
converted to draft to move over the unit tests and work on the performance failure

packages/waku/src/lib/builder/build.ts
```diff
@@ -559,104 +562,112 @@ const emitHtmlFiles = async (
...
+    } else {
+      await writeFile(destHtmlFile, htmlStr);
+    }
+  };
+
+  const dynamicHtmlPathMap = new Map<PathSpec, string>();
+  for (let start = 0; start * PAGE_SLICE_SIZE < buildConfig.length; start++) {
+    const end = start * PAGE_SLICE_SIZE + PAGE_SLICE_SIZE;
+    await Promise.all(buildConfig.slice(start, end).map(handlePage));
+  }
```
tylersayshi, line 663: this is the critical part of this PR to fix the OOM @dai-shi

packages/waku/src/lib/builder/build.ts
```diff
@@ -559,104 +562,112 @@ const emitHtmlFiles = async (
...
-        });
-        await mkdir(joinPath(destHtmlFile, '..'), { recursive: true });
-        if (htmlReadable) {
-          await pipeline(
-            Readable.fromWeb(htmlReadable as any),
-            createWriteStream(destHtmlFile),
-          );
-        } else {
-          await writeFile(destHtmlFile, htmlStr);
+  const handlePage = async ({
```
dai-shi, line 565: We don't want to use the term "Page" here. Can it be something like `handlePath`?

packages/waku/src/lib/builder/build.ts
dai-shi: 🎉

packages/waku/src/lib/builder/build.ts
tylersayshi: changed to handlePath ✅

tylersayshi:
The bug for ssg-performance is that the layout here: https://github.com/dai-shi/waku/blob/main/e2e/fixtures/ssg-performance/src/pages/_layout.tsx#L4-L11

expects to get the path from props so we will either pass these props to Layouts again or use `path` from the page component which would simplify the test further 

fwiw the `Context` for passing path to the page seems weird when the page component already get's passed `path` on its own

dai-shi:
Okay, this is kind of complicated.
With the current unstable_defineRouter, 
```js
              id.endsWith('/layout')
                ? { path: pathname }
                : { path: pathname, query },
```
we pass `path` for layout and both `path` and `query` for others.
This is an intentional design, and `shouldSkip` checks if `path` is changed.

However, with the new_defineRouter, we no longer have `shouldSkip`, and if slotId is the same, it will be cached. So, if `new_createPages` wants to pass `path` to the layout component, the slotId must be unique to that `path`. Otherwise, it shouldn't pass `path`.

tylersayshi:
> So, if new_createPages wants to pass path to the layout component, the slotId must be unique to that path. Otherwise, it shouldn't pass path.

I dislike the idea of conditionally passing `path` to layout. I think using something like the `useRouter` hook in a client component if the layout really needs the `path` would be clearer. 

I've changed the ssg-performance test to use the path from the page component instead now which seems nice to me

e2e/fixtures/ssg-performance/src/pages/[path].tsx
```diff
@@ -1,8 +1,8 @@
 import { Path } from '../Path.js';
 
-export default async function Test() {
+export default async function Test({ path }: { path: string }) {
   await new Promise((resolve) => setTimeout(resolve, 1000));
-  return <Path />;
+  return <Path path={`/${path}`} />;
```
Comment (tylersayshi) [inline e2e/fixtures/ssg-performance/src/pages/[path].tsx:5]: @dai-shi should the `path` from props include the leading `/`? I think it should. This would make it consistent with `window.location.pathname`

maybe we tackle that in a new PR though if you agree?

dai-shi:
> I dislike the idea of conditionally passing `path` to layout.

I don't mean the idea of conditionally. If we take the other idea, we always pass `path` to layout components.

e2e/fixtures/ssg-performance/src/pages/[path].tsx
Comment (dai-shi) [reply e2e/fixtures/ssg-performance/src/pages/[path].tsx:5]: That's weird. When we say `path` or `pathname` it should start with `'/'`. How come this is happening? If it's a bug already, we would like to fix it before this PR.

Comment (tylersayshi) [reply e2e/fixtures/ssg-performance/src/pages/[path].tsx:5]: ok! I'll get a fix for this up then 👍

tylersayshi:
> > I dislike the idea of conditionally passing `path` to layout.
> 
> I don't mean the idea of conditionally. If we take the other idea, we always pass `path` to layout components.

Isn't it specific to the layout's path though?

like `layout:/` could only get `{ path: '/' }` even if the `layout:/` is used on another page like `page:/foo`?

dai-shi:
No, when I say passing `path`, which is the current behavior, it means the root layout takes `/foo` as the path if the user visits `/foo`. (unless I misunderstand the current behavior.)

tylersayshi:
> No, when I say passing `path`, which is the current behavior, it means the root layout takes `/foo` as the path if the user visits `/foo`. (unless I misunderstand the current behavior.)

oh, yea that makes sense to me. I think I just misunderstood something you said on discord then. I follow this though now, thanks.

Are you ok with not passing the `path` to layouts for now though?

dai-shi:
> Are you ok with not passing the `path` to layouts for now though?

This is technically a breaking change. Can you check README and all examples and see if we describe/use the current capability?

dai-shi:
Even if we are okay with the breaking change, we should change it in a different PR in advance. It means the new_createPages lies its types.

dai-shi:
I think dynamic layout should receive `path` in props. We could get it from context, but it feels like a last resort.

e2e/fixtures/ssg-performance/src/pages/[path].tsx
Comment (tylersayshi) [reply e2e/fixtures/ssg-performance/src/pages/[path].tsx]: this is fixed now... It was actually relying on `[path].tsx` as the filename and it got `path` from the mapping. So I changed the name of the file to `[slug].tsx` and passed `path` into the page props where I had previously missed it.

Follow up question: should we ban `[path].tsx` to avoid name collision for the props?

tylersayshi:
I can add the `path` back to Layout props for this PR to avoid the breaking change and we can discuss that separately. That seems easier.

tylersayshi:
> I think dynamic layout should receive `path` in props. We could get it from context, but it feels like a last resort.

I think static layout works too from testing. Would you double check though?

e2e/fixtures/ssg-performance/src/pages/[path].tsx
Comment (dai-shi) [reply e2e/fixtures/ssg-performance/src/pages/[path].tsx]: > Follow up question: should we ban `[path].tsx` to avoid name collision for the props?

Yeah, using `createPages` allows it, but `fs-router` should warn or error.

dai-shi:
> avoid the breaking change and we can discuss that separately. That seems easier.

Yeah.

> I think static layout works too from testing. Would you double check though?

No, it shouldn't work. We need to use different slotId for different path. Let me try an example.

dai-shi:
Try this:

```diff
diff --git a/examples/21_create-pages/src/components/HomeLayout.tsx b/examples/21_create-pages/src/components/HomeLayout.tsx
index 55e77435..d63ea21a 100644
--- a/examples/21_create-pages/src/components/HomeLayout.tsx
+++ b/examples/21_create-pages/src/components/HomeLayout.tsx
@@ -16,8 +16,9 @@ const Pending = ({ isPending }: { isPending: boolean }) => (
   </span>
 );
 
-const HomeLayout = ({ children }: { children: ReactNode }) => (
+const HomeLayout = ({ path, children }: { path: string, children: ReactNode }) => (
   <div>
+    <h1>Path: {path}</h1>
     <ul>
       <li>
         <Link
```

Aaaand, I found there's a bug in the current `createPages`. So, it's not working as expected anyway. Let's fix it in `new_createPages`.

dai-shi:
> Let's fix it in new_createPages.

Or, change the spec as it's not working correctly anyways.

e2e/fixtures/ssg-performance/src/pages/[path].tsx
Comment (tylersayshi) [reply e2e/fixtures/ssg-performance/src/pages/[path].tsx]: added an error for this scenario in fs-router

tylersayshi:
> > Let's fix it in new_createPages.
> 
> Or, change the spec as it's not working correctly anyways.

what is the bug?

also I see what you mean with it not working for static layouts.

tylersayshi:
converted to draft while we discuss what to do with Layout `path` prop

e2e/fixtures/ssg-performance/src/pages/[slug].tsx
```diff
@@ -1,8 +1,9 @@
+import type { PageProps } from 'waku/router';
 import { Path } from '../Path.js';
 
-export default async function Test() {
+export default async function Test({ path }: PageProps<'/[slug]'>) {
   await new Promise((resolve) => setTimeout(resolve, 1000));
-  return <Path />;
+  return <Path path={path} />;
```
Comment (tylersayshi) [inline e2e/fixtures/ssg-performance/src/pages/[slug].tsx:6]: @dai-shi are you ok with this change to the test? I think for testing ssg-performance it still fits that description, but it is different, so double checking seems good.

Comment (dai-shi) [reply e2e/fixtures/ssg-performance/src/pages/[slug].tsx:6]: Yeah, this is unavoidable because of #1034 change.
It should still be fine for `ssg-performance` spec.
context test should be somewhere else. Besides, using `cache` for server context isn't recommended.

e2e/fixtures/ssg-performance/src/pages/_layout.tsx
```diff
@@ -1,10 +1,5 @@
...
 
-export default function Layout({
-  children,
-  path,
-}: PropsWithChildren<{ path: string }>) {
-  setPath(path);
-  return children;
+export default function Layout({ children }: PropsWithChildren) {
+  return <div>{children}</div>;
 }
```
dai-shi, line 5: I think we can remove this file entirely.

dai-shi (approved):
LGTM. Huge thanks to your effort.

PR #1004, opened 2024-11-16
fix: query string should be passed through URLSearchParams

tylersayshi:
maybe this is not the right way to fix the issue, but this does change the query params to be available on the server which was missing before.

needs to come after #1000

this PR should just be for https://github.com/dai-shi/waku/pull/1004/commits/13a3178dd3f20330ae02e94361d798d9a5437b2e once 1000 is merged


packages/waku/src/router/define-router.ts
```diff
@@ -484,7 +484,7 @@ globalThis.__WAKU_ROUTER_PREFETCH__ = (path) => {
     });
     return {
       rscPath,
-      rscParams: JSON.stringify({ query: searchParams.toString() }),
+      rscParams: new URLSearchParams({ query: searchParams.toString() }),
```
dai-shi, line 487: Nice catch. I missed this.

dai-shi (approved):
LGTM

PR #1008, opened 2024-11-21
fix: Windows path

himself65:


packages/waku/src/lib/plugins/vite-plugin-rsc-transform.ts
```diff
@@ -670,7 +675,10 @@ export function rscTransformPlugin(
     }
     throw new Error('server id not found: ' + id);
   };
-  const wakuDist = joinPath(fileURLToFilePath(import.meta.url), '../../..');
+  const wakuDist = joinPath(
```
dai-shi, line 678: We have `wakuDist` in packages/waku/src/lib/middleware/dev-server-impl.ts too. Should it follow too?

himself65: Tried, but not working, I think all rest path code rely on the starting `/`

himself65: We might need fix `packages/waku/src/lib/middleware/dev-server-impl.ts` in another PR? it might takes longer time

dai-shi: Interesting. It's probably something different. No hurry, but if we could understand the difference better, it might be helpful for the future.

PR #1011, opened 2024-11-23
fix: NewRouter for unstable_defineRouter

dai-shi:
Previously with current Router, this issue wasn't revealed because of other bugs (?), but this change should be reasonable. (I'm not happy with the current overall implementation. Hope to revisit.)

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -651,14 +651,11 @@ const getRouteSlotId = (path: string) => 'route:' + path;
 const NewInnerRouter = ({
   routerData,
   initialRoute,
-  cachedIdSetRef,
-  staticPathSetRef,
 }: {
-  routerData: RouterData;
+  routerData: Required<NewRouterData>;
   initialRoute: RouteProps;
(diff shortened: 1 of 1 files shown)


packages/waku/src/router/client.ts
```diff
@@ -808,7 +808,8 @@ export function NewRouter({
               // FIXME this check here seems ad-hoc (less readable code)
               if (
                 window.location.pathname !== path ||
-                window.location.search.replace(/^\?/, '') !== query
+                (!isStatic &&
+                  window.location.search.replace(/^\?/, '') !== query)
```
tylersayshi, line 812: are we sure the `.replace` is still needed?

tylersayshi: nvm I just doubled checked... it is needed

dai-shi: I think `window.location.search` has `?`, while `query` does not, no?

tylersayshi: you're right. I was mistaken with my initial comment, please disregard.

PR #1013, opened 2024-11-24
fix(router): fallback support in new_defineRouter and new_createPages

dai-shi:
add fallback mechanism for #1003.

### DIFF
--- a/packages/waku/src/minimal/client.ts
+++ b/packages/waku/src/minimal/client.ts
@@ -291,6 +291,7 @@ export const Slot = ({
   children,
   fallback,
   unstable_shouldRenderPrev,
+  unstable_renderPrev,
 }: {
   id: string;
   children?: ReactNode;
@@ -299,6 +300,7 @@ export const Slot = ({
     err: unknown,
     prevElements: Record<string, ReactNode>,
   ) => boolean;
+  unstable_renderPrev?: boolean;
 }) => {
   const elementsPromise = use(ElementsContext);
   if (!elementsPromise) {
@@ -317,6 +319,10 @@ export const Slot = ({
       elements[id],
     );
   };
+  if (unstable_renderPrev && elementsPromise.prev) {
+    const elements = elementsPromise.prev;
+    return renderSlot(elements);
+  }
   return createElement(
     OuterSlot,
     {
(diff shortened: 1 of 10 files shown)


packages/waku/src/router/define-router.ts
```diff
@@ -302,12 +302,14 @@ export function new_defineRouter(fns: {
...
     routeElement: ReactNode;
     elements: Record<SlotId, ReactNode>;
+    fallbackElement?: ReactNode;
   }>;
 }): ReturnType<typeof defineEntries> {
   const platformObject = unstable_getPlatformObject();
   type MyPathConfig = {
     pattern: string;
     pathname: PathSpec;
+    // FIXME Do we need this isStaticRouteElement?
```
tylersayshi, line 312: nope! good catch, safe to remove. This just slipped in as left over

tylersayshi (commented):
This seems fine. I don't yet fully understand what `fallbackElement` is yet though.

packages/waku/src/minimal/client.ts
```diff
@@ -286,19 +295,32 @@ export const Slot = ({
...
+  const renderSlot = (elements: Record<string, ReactNode>, err?: unknown) => {
     if (!(id in elements)) {
       if (fallback) {
+        if (err) {
+          // HACK I'm not sure if this is the right way
+          return createElement(
+            ChildrenContextProvider,
+            { value: createElement(InnerErr, { err }) },
+            fallback,
+          );
```
dai-shi, line 322: I wonder if this hack works for new_createPages fallback. Let check it after merging...

packages/waku/src/minimal/client.ts
```diff
@@ -309,11 +331,17 @@ export const Slot = ({
       elements[id],
     );
   };
+  if (unstable_renderPrev) {
+    if (!elementsPromise.prev) {
+      throw new Error('Missing prev elements');
+    }
+    return renderSlot(elementsPromise.prev);
+  }
```
dai-shi, line 339: I'm not happy with this, and hope to redesign it after finishing our current transition.

packages/waku/src/minimal/client.ts
dai-shi: It doesn't work. so #1015.

PR #1014, opened 2024-11-25
fix: createPages was missing root component for default

tylersayshi:


dai-shi:
Isn't it this?
https://github.com/dai-shi/waku/pull/1014/files#diff-aa5fd10e2174215f8433b37357a7b91805c9a9d4d00c232fd563ce9f1119651dR791

tylersayshi:
oh, yes... I am getting a weird error where it says I am missing a root component in another project, but this would not be the fix

PR #1021, opened 2024-11-28
fix(dev): fix hot-reload with e2e test

### LINKED ISSUE #1016: hot reloading doesn't work even in the reference example
I used `npm create waku@latest` reference example, but hot reloading of client component doesn't work..

This is the repo https://stackblitz.com/~/github.com/dannylin108/waku-starter

https://github.com/user-attachments/assets/a53d3f9c-563d-43fa-8bbe-81e295eb647e

dai-shi:
fix #1016 

It's a regression with various changes. Adding the e2e test will help for the future.


packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -225,7 +225,6 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
         await writeFile(outputFile, formatted, 'utf-8');
       };
 
-      server.watcher.add(opts.srcDir);
```
dai-shi, line 228: Removing this fixes the issue of failing on windows and ubuntu. @tylersayshi

tylersayshi: What if we did `server.watcher.add(pagesDir)` instead? .gen.ts will be outside the pages dir so this should be fine I'd think

dai-shi: We can try, but is removing this a problem? It still generates .gen.ts file.

tylersayshi: The idea for adding it here is that it will updates pages.gen.ts when new pages are added or removed. Without the watcher I don't think it will be able to update while the dev server is running. Just on startup only.

tylersayshi: works on mac... I'll watch to see if the other tests pass but I expect them to 🤞 

https://github.com/dai-shi/waku/pull/1023

tylersayshi: looks like it passed 🎉

dai-shi (commented):
Let's merge.

PR #1023, opened 2024-11-30
fix: watch pages dir for typegen updates

tylersayshi:
Tested this locally and the hot reload still works when changing the components or the pages dir

the key about this is that pages.gen.ts is outside pagesDir so we avoid the issue we were seeing before.

but the upside of still watching this is that when we add/remove pages or update their getConfig, pages.gen.ts will hot update to reflect the changes


tylersayshi:
How was there just one failure on node 22 + windows 😞🤯

We can close this for now and revisit later. It's more important that HMR is reliable.

/close

dai-shi:
> How was there just one failure on node 22 + windows

It's a flaky test. It passes on the second time.

PR #1024, opened 2024-11-30
fix: new_createPages getPathConfig behavior for staticPaths

tylersayshi:
This changes new_createPates getPathMapping to return `/path/[slug]` instead of each staticPath as its own entry.

Sidenote: this also helps with memory usage, but doing the Promise.all with slices of 2500 is still good


tylersayshi:
@dai-shi the 21_create-pages example registers a lot of pages at `/nested/*`

code here: https://github.com/dai-shi/waku/blob/main/examples/21_create-pages/src/entries.tsx#L50-L83

my new code errors on the e2e test for 21_create-pages because it registers the two uses of `/nested/[id]` as duplicate. If I'm honest, there seems to be even more duplicates here.

1. do we need to support `/nested/[id]` for both dynamic and static?
2. should we make a note to consider more of these paths duplicate? It seems potentially confusing that a route can be both a literal path or also a slug or wildcard match. We could communicate some sort of priority, but this seems like a lot of supported overlap to me right now.

packages/waku/tests/create-pages.test.ts
```diff
@@ -685,7 +685,7 @@ describe('createPages', () => {
       {
         elements: {
           root: { isStatic: true },
-          'page:/test/w/x': { isStatic: true },
+          'page:/test/[a]/[b]': { isStatic: true },
```
dai-shi, line 688: I think it's generally a good idea for new_createPages and I believe the current createPages behaves like this.

One note is that this string is a slotId, which is a cache id.
So, technically, the following two are very different behavior-wise:
- Having `/src/pages/foo.tsx` and `/src/pages/bar.tsx`,
- Having `/src/pages/[slug].tsx` with staticPath = `['foo', 'bar']`.

dai-shi:
> Sidenote: this also helps with memory usage, but doing the Promise.all with slices of 2500 is still good

I fully agree with it.

packages/waku/tests/create-pages.test.ts
tylersayshi: If I follow you correctly, We'd have better ability to more granularly cache the page slots if this is: `'page:/test/w/x'`?

Will `defineRouter` handle `'page:/test/[a]/[b]'` and `'page:/test/w/x'` the same right now?

dai-shi: Yeah, it handles the same. `'page:...'` is the createPages thing. defineRouter only knows it's a string id.

dai-shi: > I believe the current createPages behaves like this.

On second thought, that doesn't hold for the new one. It's very differently designed. Not just refactoring.

dai-shi:
I'm not sure abou the status of this PR. Converting to draft for now.

examples/21_create-pages/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "waku-example",
+  "name": "create-pages",
```
dai-shi, line 2: I think we can use the same name for all examples. But, looking at them now, it's not consistent.

tylersayshi: I like giving them unique names to allow for running `pnpm -F create-pages dev` in this case. 

It would also be a nice way to reduce all the example dev scripts in the root package.json if you're ok with this

dai-shi: Oh, I didn't know about `-F`. Any reference?

Examples are standalone apps. What you get with `npm create waku@latest` is the example.

Also, this should work too: https://stackblitz.com/github/dai-shi/waku/tree/v0.21.7/examples/01_template

tylersayshi: Nice!

And here's a reference, it's the same as --filter

https://pnpm.io/filtering

dai-shi: Ah, well I often use `--filter`.

tylersayshi: Would you want to use it for the examples?

dai-shi: Yeah, I don't like having too many `"example:*"` entries.

PR #1025, opened 2024-12-01
fix: watch pages dir

tylersayshi:
This will watch the pages die for updates to generate types for pages


tylersayshi:
Reopening since #1023 was just failing due to flakey test

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -225,6 +225,8 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
         await writeFile(outputFile, formatted, 'utf-8');
       };
 
+      server.watcher.add(pagesDir);
```
dai-shi, line 228: Can you help me understand what will fail without this line?

tylersayshi: oh I thought this was needed to watch the pages directory. I remember reading something that suggested doing this when initially creating the plugin.

I just tested with and without this line though and it seems like it actually has no affect and the plugin reruns on changes without this line. Closing this MR and won't re-open since it is not needed after testing.

PR #1030, opened 2024-12-03
refactor: rely on pnpm filter to run specific examples

tylersayshi:
could probably do similar with the website, but scoping this PR to just the examples seems like an appropriate start


CONTRIBUTING.md
```diff
@@ -20,14 +20,7 @@ pnpm run compile
...
 ```shell
-pnpm run examples:dev:01_template
-```
-
-or if you want to run a specific example:
-
-```shell
-cd examples/01_template
-pnpm run dev
+pnpm -F 01_template dev # to run examples/01_template in dev mode
```
dai-shi, line 23: This is so good!
Can you add an instruction for the `pnpm run examples:prd:01_template` equivalent?

tylersayshi: is this good?

```shell
pnpm -F 01_template dev # to run examples/01_template in dev mode
```

`build` and `start` can be run with the same pattern:

```shell
pnpm -F 01_template build
pnpm -F 01_template start
```

---

Or do you want to explicitly call out:

`pnpm -F 01_template build &&  pnpm -F 01_template start`

dai-shi: Hm, whichever is fine. I'd probably use some hack personally. So, just consider a better way for docs.

tylersayshi: updated to the above without the `&&` 

happy hacking haha

dai-shi (approved):
LGTM

I thought a drawback for me was autocompletion, but it seem like pnpm has the support: https://pnpm.io/8.x/completion

PR #1031, opened 2024-12-03
docs: add root element API

sandren:


packages/website/private/contents/post-005.mdx
```diff
@@ -0,0 +1,74 @@
...
+export const getConfig = async () => {
+  return {
+    render: 'static',
+  };
+};
+```
+
+
+
+### Low-level API
```
tylersayshi, line 39: Can refer to `createPages` as something like "config based routing"? Not necessarily that exactly, but "low-level api" seems more descriptive of something like `defineRouter` to me where the target user is library authors.

packages/website/private/contents/post-005.mdx
```diff
@@ -0,0 +1,74 @@
...
+  createPage({
+    render: 'dynamic',
+    path: '/',
+    component: HomePage,
+  });
+});
+```
+
+## Stay cozy...
+Over the next few weeks, we’ll share more updates about other recently released features in past Waku 0.21.x versions. Happy holidays!
```
tylersayshi, line 74: ☃️

packages/website/src/pages/_layout.tsx
```diff
@@ -9,16 +9,11 @@ type RootLayoutProps = { children: ReactNode };
...
-        <Providers>
-          <Meta />
-          <div id="__waku">{children}</div>
-          <Analytics />
-        </Providers>
-      </body>
-    </html>
+    <Providers>
+      <Meta />
+      <div id="__waku">{children}</div>
```
tylersayshi, line 14: this is just used for the website, right? 

It reminds me of `<div id="__next" />` which is the root element nextjs uses for their react app. 

Would we be open to maybe renaming this? sorry for the nit

tylersayshi (commented):
thanks for adding these! Nice to see the apis progressing :)

dai-shi (commented):
LGTM

dai-shi:
Can you run prettier on packages/website/private/contents/post-005.mdx?

packages/website/private/contents/post-005.mdx
dai-shi: Yeah, we kind of have two lower level apis. When we say low-level api in README, it's `createPages`. Let's call `defineRouter` minimal api or api for lib authors. I like the idea of calling config-based routing, though.

packages/website/src/pages/_layout.tsx
dai-shi: > It reminds me of `<div id="__next" />`

This is exactly the one for Waku, as I understand.

tylersayshi: is this the only place we reference this id though?

 https://github.com/dai-shi/waku/blob/main/packages/website/src/styles.css#L12-L19

packages/website/private/contents/post-005.mdx
dai-shi: > Let's call defineRouter minimal api or api

I actually confused with defineEntries. Anyways, createPages is a low-level api for app developers and define* is for lib authors.

packages/website/src/pages/_layout.tsx
dai-shi: yeah, not forcing it at the moment. never mind. though, actually, i'm curious what was your suggestion about renaming.

tylersayshi: I guess my main question is just for why we're using "__waku" here. I shouldn't suggest changing it before understanding why it's used.

For now it seems like it is only used for styling in which case, I would think a class name would make more sense. But I am probably just missing something.

dai-shi: I'm not sure either. Can also be a historical reason.

tylersayshi: not a big deal either way, thanks 👍

PR #1034, opened 2024-12-06
fix(new_createPages): path prop for dynamic layouts only and fix type

tylersayshi:
This change marks the type of static layouts to not receive `path` as a prop. For layouts to receive `path` as a prop, currently this will need to be made dynamic.

this also includes a fix to new_createPages where we consider the pattern of the staticPaths with slug paths to be the same pattern across each use of a staticPath as it is return from `getPathConfig`

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -144,13 +144,21 @@ export type CreatePage = <
   'unstable_disableSSR'
 >;
 
-export type CreateLayout = <Path extends string>(layout: {
-  render: 'static' | 'dynamic';
-  path: PathWithoutSlug<Path>;
-  component: FunctionComponent<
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
LGTM

Technically, the type change affects to the current createPages, but let's assume this is temporarily.

dai-shi:
@tylersayshi Can you please summarize changes in the PR description?

pmelab:
I just tried to update our app to the latest version, and i tripped over this one. It relied on the current page path being available in the layout. What is the reason for this change?

@dai-shi Is there another way to get the current page path within a server-component layout?

dai-shi:
Is your layout static? static layouts are eagerly cached and `path` isn't reliable.

pmelab:
Yes it is. Thanks, then I'll have to find another way around that.

PR #1039, opened 2024-12-07
chore(create-pages): replace createPages with new_createPages

tylersayshi:
Removes old version of createPages in favor of new_createPages


dai-shi (approved):
LGTM

PR #1042, opened 2024-12-08
fix(hmr): file add and delete runtime error with fs-router

tylersayshi:
Resolves runtime error caused by removing a page file when running the dev server with fs-router projects


packages/waku/src/router/fs-router.ts
```diff
@@ -27,6 +26,7 @@ export function fsRouter(
         encoding: 'utf8',
         recursive: true,
       });
+      // dev and build only
```
dai-shi, line 29: It this commend moved intentionally?

tylersayshi: typo, thanks, fixed

dai-shi (approved):
LGTM

PR #1044, opened 2024-12-08
react 19 stable

dai-shi:
Finally!

It's unfortunate that Waku isn't v1 yet.

### DIFF
--- a/e2e/fixtures/broken-links/package.json
+++ b/e2e/fixtures/broken-links/package.json
@@ -9,9 +9,9 @@
     "start": "waku start"
   },
   "dependencies": {
-    "react": "19.0.0-rc-7670501b-20241124",
-    "react-dom": "19.0.0-rc-7670501b-20241124",
-    "react-server-dom-webpack": "19.0.0-rc-7670501b-20241124",
+    "react": "19.0.0",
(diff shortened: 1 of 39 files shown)


tylersayshi:
Should we keep any of our tests on canary?

dai-shi:
> Should we keep any of our tests on canary?

I thought about it too, but we don't want to increase CI time for each PR.
I will probably create a separate PR so that we can run tests on canary manually, on demand.

tylersayshi:
> > Should we keep any of our tests on canary?
> 
> I thought about it too, but we don't want to increase CI time for each PR. I will probably create a separate PR so that we can run tests on canary manually, on demand.

we could run a github action exclusively on main to test canary versions. 

This would mean that PRs would have the same pipeline time, but we run an additional check after it is merged to main, so we should know of issues on canary before we go go do a release. 

If it is its own CI process we can re-use the same tests too and just run the pnpm update to canary on the CI job.

I'm happy to work on a PR for this if it sounds good to you

dai-shi:
> we could run a github action exclusively on main to test canary versions.

good idea.

> I'm happy to work on a PR for this if it sounds good to you

Yeah, please go ahead!

PR #1052, opened 2024-12-10
fix(typegen): abstract DO_NOT_USE_pages away from pages.gen.ts

tylersayshi:
Simplified generated type output for waku website as example:

image


dai-shi (approved):
LGTM

PR #1058, opened 2024-12-11
docs: add type-safe routing

sandren:


packages/website/private/contents/post-006.mdx
```diff
@@ -0,0 +1,60 @@
...
+export default async function SomePage() { ... }
+
+export const getConfig = async () => {
+  const staticPaths = await getStaticPaths();
+
+  return {
+    render: 'static',
+    staticPaths,
+  } as const; // NEW!
+};
```
tylersayshi, line 29:
```suggestion
export const getConfig = async () => {
  return {
    render: 'static',
    staticPaths: ['foo', 'bar'],
  } as const; // NEW!
};
```
could we do the example like this? It's less "real world", but since we don't fill `getStaticPaths` here, it will show a more type-safe usage as the example.

packages/website/private/contents/post-006.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+const getStaticPaths = async () => { ... }
+```
+
+### Developer experience
+
+Afterwards enjoy type checks and IDE autocompletion for the `<Link>` component’s `to` prop as well as parameters of `useRouter()` methods like `router.push()`.
```
tylersayshi, line 36: We also export `PageProps` which can be used as `export const MyPage = (props: PageProps<'/link/[slug]'>) => {}` which would be nice to mention I think.

packages/website/private/contents/post-006.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+import { Link, useRouter } from 'waku';
+
+const Component = () => {
+  const router = useRouter();
+
+  return (
+    <>
+      <Link to="/invalid/link">Wrong link</Link>
+      <button onClick={() => router.push('/invalid/path')}>Wrong path</button>
```
tylersayshi, line 49: I forget, does this have type safety too?

tylersayshi (approved):
lgtm overall, thanks! just a couple small things

dai-shi:
This is work in progress, right? Please run prettier when ready.

PR #1067, opened 2024-12-14
fix(use-router): router.push bug

### LINKED ISSUE #1060: useRouter push does not load the given page
### What
When pushing to a page with the hook `useRouter_UNSTABLE`, the page does not load on first try. An explicit `router.reloaded()` is needed to make it work.
It worked fine in waku version `0.21.2`
### How to reproduce
Run the project found [here](https://github.com/pfongkye/waku-use-router) and click on the button "Use router to change page"

### Misc
waku version where issue was detected: `0.21.8`
The code can be found [here](https://github.com/pfongkye/waku-use-router/blob/main/src/pages/index.tsx#L16).

tylersayshi:
fix #1060

### DIFF
--- a/e2e/fixtures/use-router/src/pages/dynamic.tsx
+++ b/e2e/fixtures/use-router/src/pages/dynamic.tsx
@@ -1,14 +1,22 @@
-import { Link } from 'waku';
+'use client';
+
+import { Link, useRouter_UNSTABLE } from 'waku';
 import TestRouter from '../TestRouter.js';
 
-const Page = () => (
-  <>
-    <h1>Dynamic</h1>
-    <p>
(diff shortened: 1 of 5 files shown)


e2e/fixtures/use-router/src/pages/dynamic.tsx
```diff
@@ -1,14 +1,22 @@
-import { Link } from 'waku';
+'use client';
```
dai-shi, line 1: This `/dynamic` page is used for other specs, so it feels safe to avoid changing it to a client component.

dai-shi (approved):
Let's merge this. If it doesn't solve the issue, we should to the second iteration.

PR #1074, opened 2024-12-17
chore(types): obscure types for page props

tylersayshi:
This is to hide the implementation details of `PropsForPages` in the exported `PageProps` type

re #1058


tylersayshi:
closing... this actually doesn't work as I expected 🫠

the type still shows through once it is selected and hovered.

image

PR #1077, opened 2024-12-18
fix(create-pages): layouts missing with more nesting

### LINKED ISSUE #1055: Non-root layouts not loaded in v0.21.8
I think there's a bug in v0.21.8. The list of found path segments with _layout.tsx files is incorrect. It includes `undefined` and `[object]` strings. I think the fix is to use `segment.name` instead of `segment` [edit: and use `index` instead of `index - 1`].

```
--- a/dist/router/create-pages.js
+++ b/dist/router/create-pages.js
@@ -185,10 +185,10 @@ export const createPages = (fn)=>{
     };
     const getLayouts = (spec)=>{
         const pathSegments = spec.reduce((acc, segment, index)=>{
-            if (acc[index - 1] === '/') {
-                acc.push('/' + segment);
+            if (!acc[index] || acc[index] === '/') {
+                acc.push('/' + segment.name);
             } else {
-                acc.push(acc[index - 1] + '/' + segment);
+                acc.push(acc[index] + '/' + segment.name);
             }
             return acc;
         }, [
```

There should probably be a test case to catch this...

tylersayshi:
Adds tests for layouts with nesting and slugs

fixes #1055

### DIFF
--- a/e2e/21_create-pages_standalone.spec.ts
+++ b/e2e/21_create-pages_standalone.spec.ts
@@ -32,6 +32,12 @@ async function testRouterExample(page: Page, port: number) {
   await page.goto(`http://localhost:${port}/foo`);
   await expect(page.getByRole('heading', { name: 'Foo' })).toBeVisible();
 
+  // /nested/foo is defined as a staticPath of /nested/[id] which matches this layout
+  await page.goto(`http://localhost:${port}/nested/foo`);
+  await expect(
+    page.getByRole('heading', { name: 'Deeply Nested Layout' }),
+  ).toBeVisible();
+
   await page.goto(`http://localhost:${port}/nested/baz`);
   await expect(
     page.getByRole('heading', { name: 'Nested Layout' }),
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -8,6 +8,7 @@ import {
   parsePathWithSlug,
   getPathMapping,
   path2regexp,
+  pathSpecAsString,
 } from '../lib/utils/path.js';
 import type { PathSpec } from '../lib/utils/path.js';
 import type {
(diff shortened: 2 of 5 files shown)


e2e/21_create-pages_standalone.spec.ts
```diff
@@ -32,6 +32,12 @@ async function testRouterExample(page: Page, port: number) {
   await page.goto(`http://localhost:${port}/foo`);
   await expect(page.getByRole('heading', { name: 'Foo' })).toBeVisible();
 
+  // /nested/foo is defined as a staticPath of /nested/[id] which matches this layout
+  await page.goto(`http://localhost:${port}/nested/foo`);
+  await expect(
+    page.getByRole('heading', { name: 'Deeply Nested Layout' }),
+  ).toBeVisible();
+
```
dai-shi, line 40: That was my bad too, but it's actually not a good idea to rely on examples. This is kind of historical reason, but we now have some standalone spec with fixtures.

Let's revisit it with a different PR later.

packages/waku/src/router/create-pages.ts
```diff
@@ -135,14 +136,14 @@ export type CreateLayout = <Path extends string>(
   layout:
     | {
         render: 'dynamic';
-        path: PathWithoutSlug<Path>;
+        path: Path;
```
dai-shi, line 139: I'm not sure about this change, but count on you.

dai-shi (approved):
LGTM

packages/waku/src/router/create-pages.ts
tylersayshi: without this change there is a type error if we attempt to define a layout for `/foo/[id]` which I think would be valid.

the fs router equivalent would be `pages/foo/[id]/_layout.tsx`

PR #1078, opened 2024-12-18
fix(handler): wait for first chunk in stream

dai-shi:
It gives a chance for non async components to write to context or something.
It's pretty limited and I'm not sure if we keep this feature in the future.

### DIFF
--- a/packages/waku/src/lib/utils/stream.ts
+++ b/packages/waku/src/lib/utils/stream.ts
@@ -87,15 +87,11 @@ export const streamFromPromise = (promise: Promise<ReadableStream>) =>
     },
   });
 
-export const waitForFirstByte = (
+export const waitForFirstChunk = async (
   stream: ReadableStream,
 ): Promise<ReadableStream> => {
   const [stream1, stream2] = stream.tee();
   const reader = stream1.getReader();
-  return new Promise<ReadableStream>((resolve, reject) =>
-    reader.read().then(
-      () => resolve(stream2),
-      (error) => reject(error),
-    ),
-  );
+  await reader.read();
+  return stream2;
 };
--- a/packages/waku/src/lib/utils/stream.ts
+++ b/packages/waku/src/lib/utils/stream.ts
@@ -86,3 +86,16 @@ export const streamFromPromise = (promise: Promise<ReadableStream>) =>
       }
     },
   });
+
+export const waitForFirstByte = (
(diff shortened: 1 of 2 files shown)


rmarscher:
This seems to be working well for me. Thank you.

packages/waku/src/lib/utils/stream.ts
```diff
@@ -86,3 +86,16 @@ export const streamFromPromise = (promise: Promise<ReadableStream>) =>
...
+): Promise<ReadableStream> => {
+  const [stream1, stream2] = stream.tee();
+  const reader = stream1.getReader();
+  return new Promise<ReadableStream>((resolve, reject) =>
+    reader.read().then(
+      () => resolve(stream2),
+      (error) => reject(error),
+    ),
+  );
+};
```
rmarscher, line 101: I was thinking it might be good to configure the reader to only read one byte and then release the stream. I asked github copilot for some help there and this is what it came up with:

```
export const waitForFirstByte = (
  stream: ReadableStream,
): Promise<ReadableStream> => {
  const [stream1, stream2] = stream.tee();
  const reader = stream1.getReader();
  
  return new Promise<ReadableStream>((resolve, reject) => {
    reader.read({ min: 1, max: 1 }).then(
      ({ done }) => {
        reader.releaseLock();
        if (done) {
          reject(new Error('Stream ended before first byte'));
        } else {
          resolve(stream2);
        }
      },
      (error) => {
        reader.releaseLock();
        reject(error);
      }
    );
  });
};
```

I tested that it works.

dai-shi: - We should handle empty stream, so `if (done)` should just resolve.
- I'm not sure if `{ max: 1 }` helps performance. It feels like an extra cost.
- releaseLock is technically unnecessary as we don't reuse stream1. Not sure if releasing lock improves performance

So, I think what I mean was `waitForFirstChunkOrClose`.

dai-shi: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read
`read` takes no options.

rmarscher: :+1: Yeah, that all makes sense. 🤦 I should have known to check web api instead of node js to verify too.

PR #1079, opened 2024-12-18
chore: create monorepo e2e spec

rmarscher:
This needs more work to figure out how to install dependencies in the test monorepo and link in the waku package in a way that closely matches the way it works with published npm packages.


dai-shi:
@himself65 Can you help this?

rmarscher:
Thank you for the help. Basically, I am trying to do these steps:

```
cd /tmp
cp -r ~/waku/e2e/fixtures/monorepo .
cd monorepo
npm i
cd packages
npm i --save waku@latest
npm run dev
```

^^ and then verify it loads the home page without errors

It should install waku at /tmp/monorepo/node_modules which has then causes some different behavior with vite paths in the past because it is outside the /tmp/monorepo/packages/waku-project root dir.

When I run the above commands and it installs 0.21.9, it is working.

If I instead `npm i --save waku@0.21.8`, I get this error:

`Cannot process SSR Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './router/client.js?v=1ef4d55f' is not defined by "exports" in /tmp/monorepo/node_modules/waku/package.json imported from /tmp/monorepo/node_modules/waku/dist/lib/middleware/dev-server-impl.js`

himself65 (changes requested):
I think this structure is not working as expected with pnpm monorepo?
pnpm will resolve all deps in the root node_modules, which I think the current e2e cannot really test the prod case?

dai-shi:
@himself65 Can we do something like the standalone setup? It uses `npm` though.

himself65:
> @himself65 Can we do something like the standalone setup? It uses `npm` though.

Im trying to do that

himself65:
I was trying to do that in https://github.com/dai-shi/waku/pull/920

rmarscher:
I got the e2e test to work by passing a package path to set the cwd for the commands. https://github.com/dai-shi/waku/pull/1079/files#diff-810f0b71160f0ddae45c40c1b3b4534a9a48171823eebf14d7d67ba557f6f4ad

I like how #1101 has the ability to test a few different package managers. We can close this PR and track there instead if it is better.

I see even though e2e tests are working, something is broken in the regular test suite checking imports in the components I copied from the standard template to the monorepo example. I think it would be good to test something similar to the create-waku template that uses fs router with the monorepo.

rmarscher:
I fixed the tests with `// eslint-disable-line import/no-unresolved`. :confused: 🤷

dai-shi:
> I like how #1101 has the ability to test a few different package managers. We can close this PR and track there instead if it is better.

@himself65 seems busy, so making this pr incorporate #1101 sounds good.

rmarscher:
Great. I did that. All seems to be passing.

e2e/utils.ts
```diff
@@ -164,16 +174,22 @@ export const prepareStandaloneSetup = (fixtureName: string) => {
...
     if (mode !== 'DEV' && !built) {
-      rmSync(`${standaloneDir}/dist`, { recursive: true, force: true });
+      rmSync(`${standaloneDir}${packageDir}/dist`, {
+        recursive: true,
+        force: true,
+      });
       execSync(
         `node ${join(standaloneDir, './node_modules/waku/dist/cli.js')} build`,
-        { cwd: standaloneDir },
+        { cwd: join(standaloneDir, packageDir) },
```
dai-shi, line 192: Make it consistent with L186. Either use `join` for both, or use string concatenation for both. L209 too.

rmarscher: Yeah, thanks. Updated L186 to `join`.

dai-shi:
I hope @himself65 has a chance to review this, but I'm not sure.

rmarscher:
OK. It looks safe to merge to me. I personally use bun package manager, but the issues I saw with it in the past were the same as an npm monorepo (related to hoisted dependencies). I don't think we need another e2e test for bun.

dai-shi (approved):
LGTM

PR #1080, opened 2024-12-20
fix(create-pages): dynamic wildcard pages runtime errored

tylersayshi:
needed to lookup the wildcard page component

e2e test added to check


dai-shi (approved):
LGTM

Let's keep it mind that we should refactor 21_create-pages_standalone.spec.ts to avoid using the example, and use a new fixture instead.

PR #1012, opened 2024-11-24
feat: migrate to vite v6

himself65:
Blocking issues:

- [x] `polyfillModulePreload is deprecated. Use modulePreload.polyfill instead. (x6)`
- [x] rscRsdwPlugin parse code is somehow incorrect, I think the code it receives is already a bundlered code, that's why `const index = code.indexOf('function requireAsyncModule(id)');` will get `-1`
- [x] vite issue. https://github.com/vitejs/vite/pull/18361 
- [x] no styling in dev. 21_create-pages_standalone spec fails.
- [x] react-error-boundary. ssr-catch-error spec fails.

Other issues:

- [x] proper fix for `resolve.conditions` hack
- [x] proper fix for `vite-loader.ts` hack
- [ ] drop `viteServer.hot`. https://github.com/dai-shi/waku/pull/1012#issuecomment-2505383317

PR #1081, opened 2024-12-20
fix(style-dev): style css for dev with vite js module

tylersayshi:
This will fix the missing styles with vite 6 for #1012

It works by running js from the browser to create the styles with CSR in dev. Before we were extracting the source for plain css and passing those from the dev server in a <style> tag, but since that breaks in vite 6, this is an easy solution that will handle styles more consistently and without errors.


dai-shi (approved):
It just looks okay to me.

PR #1083, opened 2024-12-22
fix(handler): return promise of stream with allReady

dai-shi:
Fixes #1082 
Reverts #1078

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -431,15 +431,16 @@ const willEmitPublicIndexHtml = async (
     return false;
   }
   const utils = {
-    renderRsc: () => {
+    renderRsc: async () => {
       throw new Error('Cannot render RSC in HTML build');
     },
-    renderHtml: () => {
+    renderHtml: async () => {
+      const body: ReadableStream & { allReady: Promise<void> } =
+        // HACK this might not work in an edge case
+        stringToStream('DUMMY') as never;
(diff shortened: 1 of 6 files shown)


dai-shi:
I think we can revert #1078 with this change. cc @rmarscher

rmarscher:
> I think we can revert #1078 with this change. cc @rmarscher

allReady waits for all of the suspense boundaries to render. #1078 just waits for the first chunk to start sending. They are not the same. I think it's best to keep the waitForFirstByte to allow more flexibility with running server-side code before the first suspense boundary.

dai-shi:
> > I think we can revert #1078 with this change. cc @rmarscher
> 
> allReady waits for all of the suspense boundaries to render. #1078 just waits for the first chunk to start sending. They are not the same. I think it's best to keep the waitForFirstByte to allow more flexibility with running server-side code before the first suspense boundary.

I understand it's not the same, but `waitForFirstChunk` is very limited either.
Isn't it correct that `handler` now behaves like how `rsc`/`ssr` did previously?

---

If `waitForFirstChunk` is truly what peopole want, I think it can be implemented with a separate middleware.

rmarscher:
> Isn't it correct that `handler` now behaves like how `rsc`/`ssr` did previously?

Ah, ok. Yes. I just tested with v0.21.10. This is great. The handleRequest promise now waits on the 
initial entry layout and page render functions. If I add a suspense boundary to a layout, it will send the outer shell when it is ready. Perfect.

PR #1086, opened 2024-12-22
Refactor remove hack

tylersayshi:
related to #1012


dai-shi:
Thanks for your challenge!!

PR #1093, opened 2024-12-26
breaking(minimal): handleBuild API for build

dai-shi:
also changes define-router

### DIFF
--- a/examples/35_nesting/src/entries.tsx
+++ b/examples/35_nesting/src/entries.tsx
@@ -47,31 +47,27 @@ export default defineEntries({
         )}</script>`;
       const tasks = [
         async () => ({
-          type: 'htmlHead' as const,
-          pathSpec: [],
-          head: generateHtmlHead(),
-        }),
-        async () => ({
           type: 'file' as const,
-          pathname: rscPath2pathname('AppWithoutSsr'),
+          pathname: rscPath2pathname(''),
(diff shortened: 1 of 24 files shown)


tylersayshi:
this change is significant. I'm curious, what was the motivation for this refactor?

dai-shi:
It's to support api routes. Long journey.

tylersayshi:
> It's to support api routes. Long journey.

wow! I did not expect this to be so disruptive to the current code 😲

dai-shi:
API routes are technically already possible with 39_api example.
But, I would like to build static files like RSS. It's completely a new capability.

examples/35_nesting/src/entries.tsx
```diff
@@ -32,22 +33,43 @@ export default defineEntries({
...
+        })),
+        async () => ({
+          type: 'defaultHtml' as const,
+          pathname: '/',
+          htemlHead: generateHtmlHead(),
+        }),
+        async () => ({
+          type: 'defaultHtml' as const,
+          pathname: '/no-ssr',
+          htemlHead: generateHtmlHead(),
```
rmarscher, line 70: @dai-shi I think htemlHead might be a typo?

dai-shi: Oh, no. Yeah, thanks for pointing it out.

PR #1094, opened 2024-12-27
chore(e2e): add website to smoke tests

dai-shi:
to catch the error we missed.


tylersayshi:
the bug is here: https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/renderers/html.ts#L211

we call __WAKU_CLIENT_CHUNK_LOAD__ when on the server where only __WAKU_SERVER_CHUNK_LOAD__ exists

dai-shi:
closing in favor of #1095.

PR #1096, opened 2024-12-27
fix(vite-6): requireAsyncModule is defined twice in same file in vite 6

tylersayshi:
the patch needs to replace all instances of this function definition so that requireAsyncModule can be guaranteed to have the waku global loading methods

for #1012 

link to commit specific to this PR: [`50cb7a8` (#1096)](https://github.com/dai-shi/waku/pull/1096/commits/50cb7a8db25516ee7e2c941b3dda2500a3d0fc64)


tylersayshi:
closing this since it was merged into #1012

PR #1098, opened 2024-12-28
feat(use-router): change default scroll restoration behavior

### LINKED ISSUE #791: Add scroll-related options to useRouter()
Hello,

I've noticed a bug-like behavior when handling state with query parameters. Every time the query changes, the scroll position resets to the top of the page. This can be disruptive to the user experience, especially on pages with a lot of content.
To address this, I suggest adding scroll-related options to the useRouter() hook. This would allow developers to control scroll behavior when navigating or updating query parameters.

If you're open to it, I'd be willing to work on this feature. Would you be interested in a pull request that implements this functionality?

Thank you for your consideration.

tylersayshi:
fixes #791

### Will always scroll

- router.reload
- router.back
- `<Link to={foo} />`

### Scrolls when path changes

- router.replace 
- router.push

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -233,7 +233,7 @@ export function Link({
           '',
           url,
         );
-        changeRoute(route, { shouldScroll: !!route.hash });
+        changeRoute(route, { shouldScroll: true });
       });
     }
     props.onClick?.(event);
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -352,6 +352,20 @@ const InnerRouter = ({
     });
   }, [initialRoute]);
 
+  const handleScroll = useCallback((shouldScroll = true) => {
+    if (!shouldScroll) {
+      return;
+    }
+    const { hash } = window.location;
+    const { state } = window.history;
+    const element = hash && document.getElementById(hash.slice(1));
+    window.scrollTo({
+      left: 0,
+      top: element ? element.getBoundingClientRect().top + window.scrollY : 0,
+      behavior: state?.waku_new_path ? 'instant' : 'auto',
+    });
+  }, []);
+
   const changeRoute: NewChangeRoute = useCallback(
     (route, options) => {
       const { skipRefetch } = options || {};
@@ -365,7 +379,7 @@ const InnerRouter = ({
         setRoute(route);
       });
     },
-    [refetch, staticPathSet],
+    [refetch, staticPathSet, handleScroll],
   );
 
   const prefetchRoute: PrefetchRoute = useCallback(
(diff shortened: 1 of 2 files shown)


tylersayshi:
I am skeptical that `preserveScroll: true` is the best default for all scenarios, so I'd like to propose a couple scenarios:

1. going back to pages previously visited should restore scroll
2. clicking hash links should do a smooth scroll on that page to that link
3. updating query params `?foo=bar` probably should not trigger a scroll action.

Idea: 

maybe we consider different defaults in different scenarios:
- router.back & router.reload `preserveScroll: true`
- - router.replace & router.push `preserveScroll: false`
- `<Link to={path} />` scrolls only when hash is present in path by default (else does not scroll)

I think if we follow a more declarative ruleset along these lines we can remove the `waku_new_path` and also avoid devs being surprised by the scroll restoration in scenarios where they normally would not expect it to be restored.

packages/waku/src/router/client.ts
```diff
@@ -299,18 +301,6 @@ class ErrorBoundary extends Component<
...
-// Eventually replaces Router and so on
-// -----------------------------------------------------
-
-type NewChangeRoute = (
-  route: RouteProps,
-  options?: {
-    skipRefetch?: boolean;
-  },
-) => void;
-
```
dai-shi, line 313: Oops. Thanks.

packages/waku/src/router/client.ts
```diff
@@ -344,7 +334,18 @@ const InnerRouter = ({
...
+  const handleScroll = useCallback(() => {
+    const { hash } = window.location;
+    const { state } = window.history;
+    const element = hash && document.getElementById(hash.slice(1));
+    window.scrollTo({
+      left: 0,
+      top: element ? element.getBoundingClientRect().top + window.scrollY : 0,
+      behavior: state?.waku_new_path ? 'instant' : 'auto',
+    });
+  }, []);
```
dai-shi, line 346: Can we define this outside `useRouter`?

dai-shi (commented):
looks good to me otherwise.

packages/waku/src/router/client.ts
tylersayshi: moved ✅

dai-shi (approved):
Let's see how this goes.

wesbos:
Is there a way to use <Link> and not have the page scroll? NExt.js has `scroll={false}` which is nice to prevent scrolling on things like pagination.

tylersayshi:
> Is there a way to use <Link> and not have the page scroll? NExt.js has `scroll={false}` which is nice to prevent scrolling on things like pagination. 

Not yet, I'll add a way! Thanks for the feedback :)

dai-shi:
@wesbos Hi, thanks for the feedback.

We could consider adding such option.
Can you elaborate the use case to help us understand the limitation?

(I would like to keep `<Link>` as simple as `<a>`, and add flexibility to `useRouter`, because I believe `<a>` doesn't have `scroll={false}` property. But, the idea might change with your explanation.)

wesbos:
Thanks for the reply - Here is a use case for where I need it.  Most frameworks provide this, and I fear pushing people to use a hook will result in them using buttons over links too often. 

https://github.com/user-attachments/assets/69d6ad57-5815-4fbe-8b15-320670ff3e7d

tylersayshi:
> and I fear pushing people to use a hook will result in them using buttons over links too often.

completely agreed - thank you!

dai-shi:
Yeah, that's convincing. Thanks for the explanation with the demo, Wes!

@tylersayshi Let's do that.

PR #1099, opened 2024-12-28
feat(define-router): handleApi

dai-shi:


packages/waku/src/router/define-router.ts
```diff
@@ -91,15 +92,37 @@ export function unstable_defineRouter(fns: {
...
     pattern: string;
-    staticElementIds: SlotId[];
-    isStatic?: boolean | undefined;
-    specs: { noSsr?: boolean; is404: boolean };
+    specs: {
+      isStatic?: true;
+      staticElementIds?: SlotId[];
+      noSsr?: true;
+      is404?: true;
+      isApi?: true;
```
tylersayshi, line 124: Do we still plan to have `getApiConfig` separate from `getRouteConfig`? 

Including `isApi` here makes it seem like we'll use this type for both kinds of configs, api and route.

Or is this just temporary and getApiConfig will come later?

tylersayshi (approved):
the rest lgtm

packages/waku/src/router/define-router.ts
dai-shi: It is an internal implementation detail. The define-router users won't see it. Yeah, I think separating them would be good. They have different configs and I would like to pair getFooConfig and handleFoo.

PR #1102, opened 2024-12-28
feat(config): granular vite configs and React Compiler support

dai-shi:
- [x] refactor
- [x] treeshake
- [x] 'common'
- [x] 41_type-alias
- 37_css 👉 FIXME
- [x] ssr-basic (tsconfig)
- [x] ssr-catch-error
- [x] 05_compiler


examples/41_path-alias/waku.config.ts
```diff
@@ -0,0 +1,23 @@
...
+export default defineConfig({
+  unstable_viteConfigs: {
+    'dev-main': getConfig,
+    'dev-rsc': getConfig,
+    'build-analyze': getConfig,
+    'build-server': getConfig,
+    'build-ssr': getConfig,
+    'build-client': getConfig,
+  },
+});
```
tylersayshi, line 23: would it make sense to have a "common" field that applies to the viteConfig in all environments/steps? 

This seems like a good example for when that would be useful

dai-shi: Yeah, it's possible.

dai-shi:
> 37_css

@tylersayshi I need your help along with https://github.com/dai-shi/waku/pull/1105#issuecomment-2564586772.
Currently, it's broken in PRD too.

tylersayshi:
> > 37_css
> 
> @tylersayshi I need your help along with [#1105 (comment)](https://github.com/dai-shi/waku/pull/1105#issuecomment-2564586772). Currently, it's broken in PRD too.

I'll have time to look tomorrow 👍

tylersayshi:
If the change from earlier today is blocking for 37_css could we try reverting to the other vote styled plugin to unblock for now?

it seems like both plugins have issues, so we probably shouldn’t let this block us.

dai-shi:
> If the change from earlier today is blocking for 37_css

I'm not sure. Maybe it's not related. I asked you because you should be more familiar. If another one works better, I'm fine about reverting too.

dai-shi:
Oh, this was much more challenging than I first expected.

PR #1103, opened 2024-12-28
chore: add git hook for lint check

himself65:


tylersayshi:
Isn't this made redundant with the CI check? What's the reason for adding a commit hook?

himself65:
> Isn't this made redundant with the CI check? What's the reason for adding a commit hook?

I have to open PR and see I forget run prettier, which CI runs double times

tylersayshi:
> I have to open PR and see I forget run prettier, which CI runs double times

Could we create separate package.json "scripts" for link and prettier checking?

Something like:

```json
"scripts": {
  "format": "prettier -c .",
  "lint": "eslint .",
```

Then you could just run `pnpm format && pnpm lint` if you'd like to check them through the CLI on demand.

Another option would be to optionally set your machine to run these scripts with `.git/hooks` [docs link](https://git-scm.com/book/ms/v2/Customizing-Git-Git-Hooks).

himself65:
Yeah, I know how to add that, but I think we should add this auto fix in monorepo as there're more poeple jump in

tylersayshi:
It would be nice if we could configure the editor configs for our project to autofix correctly instead of adding a step to each commit.

For example `.vscode/settings.json`:

```json
{
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "editor.formatOnSave": true,
}
```

tylersayshi:
there is also this for eslint

```json
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": "explicit"
  },
```

himself65:
I don't use vscode :(

But I think add lint-staged in git hooks with husky (which is under node ecosystem) will save up more time for DX.

tylersayshi:
> I don't use vscode :(

What do you use? Every editor should have a concept of formatting on save and as long as this is configured to use prettier by default, everything should autofix nicely.

himself65:
> > I don't use vscode :(
> 
> What do you use? Every editor should have a concept of formatting on save and as long as this is configured to use prettier by default, everything should autofix nicely.

I use WebStorm, and yes that's why I opened this PR, I have three PCs and many different projects. I don't want setup stuff (like auto fix here) individually, I just usually do git commit.

But it looks out of context, this PR is for adding a auto lint fix script when you pnpm install and git commit. Of course I know how to fix the lint or add a lint script on my editor.

tylersayshi:
> But it looks out of context, this PR is for adding a auto lint fix script when you pnpm install and git commit. Of course I know how to fix the lint or add a lint script on my editor.

I think it's relevant because fixing lint and prettier (for me) are responsibilities of my editor. Then if I miss something (we have CI to catch that as the fallback).

CI is also a great catch all for managing the fact that devs could have any number of editors/setups.

husky also acts as a catch all (it runs for everyone), but for devs who have done the setup for autofixing (and fixing lint errors where they're highlighted), husky is strictly slowing them down.

---

I think editor configs in the waku repo (for prettier) like `vscode/settings.json` or `.idea/workspace.xml` (for webstorm) are a good way to have it so that you and other devs don't need to think about how it is setup (it is just done once from the repo).

The package.json scripts also help to run these as needed.

To be honest about husky specifically though, it would strictly slow me down, so I would really like to solve this another way.

dai-shi:
> Isn't this made redundant with the CI check? What's the reason for adding a commit hook?

Yes, I don't prefer having shared git hook config, unless it's like a company repository where I can force all employees with a single workflow.
In my workflow, very sometimes, I _would_ like to commit without formatting for dev history, before pushing it to the remote origin.

> I have to open PR and see I forget run prettier, which CI runs double times

I can understand that. Just have your git hook configured locally. Feel free to add the instruction to `CONTRIBUTING.md`.

PR #1104, opened 2024-12-28
ci: split e2e test to a single file

himself65:
only run E2E test when necessary to reduce unused CI check and cost

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,34 +1,26 @@
-name: CI
+name: E2E
 
 on:
   push:
     branches: [main]
+    paths:
+      - 'packages/**'
+      - 'examples/**'
+      - 'e2e/**'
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -1,34 +1,24 @@
...
+    paths:
+      - 'packages/**'
+      - 'examples/**'
+      - '.github/workflows/e2e.yml'
   pull_request:
     types: [opened, synchronize]
+    paths:
+      - 'packages/**'
+      - 'examples/**'
+      - '.github/workflows/e2e.yml'
```
dai-shi, line 15: Nice idea!
Do we need to add `'e2e/**'` too?

himself65: yes

dai-shi (approved):
LGTM. Thanks!

PR #1107, opened 2024-12-30
chore(.github/workflows): update action version used in 'uses'

sukvvon:
* update `actions/setup-node` and `actions/upload-artifact`


dai-shi (approved):
Thanks.

PR #1108, opened 2024-12-30
feat(create-pages): api handling

tylersayshi:
This adds api route handling in createPages with tests

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -624,7 +624,15 @@ export const createPages = <
       }
       const { handler } = apiPathMap.get(routePath)!;
 
-      const req = new Request(path, request);
+      const req = new Request(
+        new URL(
+          path,
+          typeof window !== 'undefined'
+            ? window.location.href
+            : 'http://localhost', // FIXME what should this be?
+        ),
+        request,
+      );
       const res = await handler(req);
 
       return {
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -625,12 +625,7 @@ export const createPages = <
       const { handler } = apiPathMap.get(routePath)!;
 
       const req = new Request(
-        new URL(
-          path,
-          typeof window !== 'undefined'
-            ? window.location.href
-            : 'http://localhost', // FIXME what should this be?
-        ),
+        new URL(path, options.headers['host'] ?? 'http://localhost'),
         options,
       );
       const res = await handler(req);
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -625,7 +625,12 @@ export const createPages = <
       const { handler } = apiPathMap.get(routePath)!;
 
       const req = new Request(
-        new URL(path, options.headers['host'] ?? 'http://localhost'),
+        new URL(
+          path,
+          options.headers['host']
+            ? `https://${options.headers['host']}`
+            : 'http://localhost',
+        ),
         options,
       );
       const res = await handler(req);
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -153,11 +153,7 @@ export type CreateApi = <Path extends string>(params: {
   path: Path;
   mode: 'static' | 'dynamic';
   method: Method;
-  handler: (req: {
-    body: ReadableStream | null;
-    headers: Readonly<Record<string, string>>;
-    method: string;
-  }) => Promise<Response>;
+  handler: (req: Request) => Promise<Response>;
 }) => void;
 
 type RootItem = {
@@ -195,20 +191,6 @@ const createNestedElements = (
       createElement(element.component, element.props, result),
     null,
   );
-};
-
-const convertResponse = (
-  res: Response,
-): {
-  body?: ReadableStream;
-  headers?: Record<string, string | string[]>;
-  status?: number;
-} => {
-  return {
-    ...(res.body ? { body: res.body } : {}),
-    headers: Object.fromEntries(res.headers.entries()),
-    status: res.status,
-  };
 };
 
 export const createPages = <
@@ -642,7 +624,14 @@ export const createPages = <
       }
       const { handler } = apiPathMap.get(routePath)!;
 
-      return convertResponse(await handler(request));
+      const req = new Request(path, request);
+      const res = await handler(req);
+
+      return {
+        ...(res.body ? { body: res.body } : {}),
+        headers: Object.fromEntries(res.headers.entries()),
+        status: res.status,
+      };
     },
   });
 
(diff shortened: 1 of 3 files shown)


packages/waku/src/router/create-pages.ts
```diff
@@ -147,6 +147,18 @@ export type CreateLayout = <Path extends string>(
       },
 ) => void;
 
+type Method = 'GET' | 'POST' | 'PUT' | 'DELETE';
+
+export type CreateApi = <Path extends string>(params: {
+  path: Path;
+  render: 'static' | 'dynamic';
```
dai-shi, line 154: I'm not sure if we call `render` for api.

packages/waku/src/router/create-pages.ts
```diff
@@ -147,6 +147,18 @@ export type CreateLayout = <Path extends string>(
...
+  method: Method;
+  handler: (params: {
+    body: ReadableStream | null;
+    headers: Readonly<Record<string, string>>;
+  }) => Promise<unknown>;
```
dai-shi, line 159: As we discussed internally, this should be:
```suggestion
  handler: (req: Request) => Promise<Response>;
```

dai-shi (commented):
review on the types.

packages/waku/src/router/create-pages.ts
tylersayshi: changed to `mode` for now... let me know what you think

packages/waku/src/router/create-pages.ts
tylersayshi: There's a type error if I use the full `Request`, so I re-used the type from defineRouter for req

dai-shi: Please create `Request` object.
```js
const convertToRequest = ({ body, headers, method }) => {
  return new Request(...);
}

Comment (tylersayshi) [reply packages/waku/src/router/create-pages.ts]: updated

File: packages/waku/src/router/create-pages.ts (hunk 1 above)
Comment (dai-shi) [reply packages/waku/src/router/create-pages.ts]: I'm neutral. @sandren may have an idea.

File: packages/waku/src/router/create-pages.ts [hunk 3]
```diff
@@ -568,6 +606,41 @@ export const createPages = <
...
+        throw new Error('Route not found: ' + path);
+      }
+      const { handler } = apiPathMap.get(routePath)!;
+
+      const req = new Request(
+        new URL(
+          path,
+          typeof window !== 'undefined'
+            ? window.location.href
+            : 'http://localhost', // FIXME what should this be?
```
tylersayshi, line 632: @dai-shi do you know how this should be handled? `Request` needs a valid url, which seems to require a host. When we run the tests `localhost` is the "host".

dai-shi: I think having always `localhost` is okay. When we learn more use cases, we can reconsider. As a workaround, we could use `unstable_getHeaders` for "host".

```js
new URL(path, 'http://localhost')
```

packages/waku/src/router/create-pages.ts
```diff
@@ -568,6 +606,41 @@ export const createPages = <
...
+      await configure();
+
+      return Array.from(apiPathMap.values()).map(({ pathSpec, mode }) => {
+        return {
+          path: pathSpec,
+          isStatic: mode === 'static',
+        };
+      });
+    },
+    handleApi: async (path, request) => {
```
dai-shi, line 619: let's not use `request` as it's confusing. `options` or `params`?

packages/waku/src/router/create-pages.ts
tylersayshi: hm I get this when calling `unstable_getHeaders` from `createPages.handleApi`

image

I'm ok with leaving it as localhost though.

dai-shi: Hmm, I'm not sure why it errors. let's look into it separately.

Actually, you can probably read "host" header from params.headers, but if it's not available, we can fallback to `localhost`.

tylersayshi: like this?

```ts
      const req = new Request(
        new URL(path, options.headers['host'] ?? 'http://localhost'),
        options,
      );
```

dai-shi: I'm not sure but should we need the protocol part too?

tylersayshi: image

I think it's needed? This is the output when run through my browser console

dai-shi: Yes, I mean the same and `options.headers['host']` doesn't include the protocol part, right?

tylersayshi: Good point about the protocol. 

However, handApi will run on our server, right?

So will we always want to request to http://localhost actually? I think it makes sense to remove the use of the host header since we expect this to request from the server this code executes on.

Correct me if I'm off here though.

dai-shi: It runs on our server (dynamic) or the build time (static).

I could imagine some use cases would check "host" and change the behavior based on it. So, at least they should be able to read the "host" header.
At that point, creating `URL` with the "host" header makes sense, but we can defer the decision until we see clear requirements.

packages/waku/src/router/create-pages.ts
```diff
@@ -147,6 +147,15 @@ export type CreateLayout = <Path extends string>(
       },
 ) => void;
 
+type Method = 'GET' | 'POST' | 'PUT' | 'DELETE';
```
dai-shi, line 150: I don't mind, but any reason for this choice?

packages/waku/src/router/define-router.ts
```diff
@@ -101,7 +101,7 @@ export function unstable_defineRouter(fns: {
   handleApi?: (
     path: string,
     options: {
-      body: ReadableStream | null;
+      body?: ReadableStream | null;
```
dai-shi, line 104: We always pass body, so this doesn't have to be optional.
```suggestion
      body: ReadableStream | null;
```

dai-shi (approved):
LGTM

packages/waku/src/router/create-pages.ts
```diff
@@ -568,6 +606,41 @@ export const createPages = <
...
+        throw new Error('Route not found: ' + path);
+      }
+      const { handler } = apiPathMap.get(routePath)!;
+
+      const req = new Request(
+        new URL(
+          path,
+          options.headers['host']
+            ? `https://${options.headers['host']}`
+            : 'http://localhost',
```
dai-shi, line 632:
```suggestion
          // TODO consider if we should apply `Forwarded` header here
          'http://localhost',
```

packages/waku/src/router/define-router.ts
dai-shi: Hmm, this was for tests. Can you fix the test somehow?

packages/waku/src/router/create-pages.ts
tylersayshi: I was hoping just to make this typesafe, but happy to add more methods or remove it if it get's in the way

PR #1110, opened 2024-12-31
chore(eslint): remove comment about '.wrangler', simplify the array inside 'includes' in 'tsconfig.eslint.json'

sukvvon:
* remove comment about `.wrangler`
* Simplify the array inside `includes` in `tsconfig.eslint.json`.

### DIFF
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -5,10 +5,11 @@
     "allowJs": true
   },
   "include": [
-    "./eslint.config.js",
-    "packages/**/*",
-    "examples/**/*",
-    "e2e/**/*",
-    "./playwright.config.ts"
+    "eslint.config.js",
+    "packages/",
+    "examples/",
+    "examples/44_cloudflare/.wrangler/**/*",
+    "e2e/",
+    "playwright.config.ts"
   ]
 }
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -8,7 +8,6 @@
     "eslint.config.js",
     "packages/",
     "examples/",
-    "examples/44_cloudflare/.wrangler/**/*",
     "e2e/",
     "playwright.config.ts"
   ]
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -46,4 +46,5 @@ export default tseslint.config(
       'import/no-unresolved': 'off',
     },
   },
+  { ignores: ['examples/44_cloudflare/.wrangler/'] },
 );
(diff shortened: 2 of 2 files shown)


tsconfig.eslint.json
```diff
@@ -5,10 +5,11 @@
...
-    "packages/**/*",
-    "examples/**/*",
-    "e2e/**/*",
-    "./playwright.config.ts"
+    "eslint.config.js",
+    "packages/",
+    "examples/",
+    "examples/44_cloudflare/.wrangler/**/*",
+    "e2e/",
+    "playwright.config.ts"
```
dai-shi, line 13: Can you explain how it works and reference we prefer these syntax?

sukvvon: @dai-shi 

https://www.typescriptlang.org/tsconfig/#include

> These filenames are resolved relative to the directory containing the tsconfig.json file.

By this, omitting `./` will act based on the current directory and will not affect the file path.

> If the last path segment in a pattern does not contain a file extension or wildcard character, then it is treated as a directory, and files with supported extensions inside that directory are included (e.g. .ts, .tsx, and .d.ts by default, with .js and .jsx if [allowJs](https://www.typescriptlang.org/tsconfig/#allowJs) is set to true).

According to this, `packages/**/*` can be simplified to `packages/` because the last path segment is `packages`.
Similarly, `examples/**/*` can be written simply as `examples/` and `e2e/**/*` as `e2e/`.

Both `e2e` and `e2e/` are considered directories because they do not contain extensions or wildcards. `e2e` and `e2e/` are technically identical, but I thought adding `/` could more clearly indicate that they are directory paths.

Because `.wrangler` is a file extension that TypeScript does not recognize, `.wrangler` files inside the directory are automatically ignored, even if we include a directory like `exemples/`. Therefore I added `examples/44_cloudflare/.wrangler/**/*` to let Typescript recognize it separately.

rmarscher: Interesting. The Cloudflare dev server generates a `examples/44_cloudflare/.wrangler/types/runtime.d.ts` file and I guess that happens to pass Waku's eslint rules when it is included in the typescript project.

I think it still might be better to just ignore eslint from the contents of any `.wrangler` directory. It is outside of our control. It doesn't exist until the 44_cloudflare example dev server is run so it doesn't cause an eslint failure on a fresh checkout of the code, but it is convenient when developing to not have to remove before running `pnpm test`.

dai-shi: @sukvvon Thanks for the explanation. Let's do @rmarscher suggestion: Remove L11 in tsconfig.json and revert eslint.config.js change and remove FIXME comment there.

dai-shi (approved):
LGTM

PR #1111, opened 2024-12-31
chore(.github/workflows): update 'node-version' to 20, simplify 'pnpm install', 'actions/setup-node'

sukvvon:
* In CI, `pnpm install` means `pnpm install --frozen-lockfile`
* Update `node-version` `18` to `20`
* We don't need to specify `cache-dependency-path` in `actions/setup-node`, monorepo only use `./pnpm-lock.yaml`, also we only use `./pnpm-lock.yaml`

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,7 +30,6 @@ jobs:
         with:
           node-version: 20
           cache: 'pnpm'
-          cache-dependency-path: '**/pnpm-lock.yaml'
       - run: pnpm install
       - run: pnpm run compile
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e.yml
```diff
@@ -28,10 +28,9 @@ jobs:
       - uses: pnpm/action-setup@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: 18
+          node-version: 20
```
dai-shi, line 31: We still support 18, so it was intentional. I don't mind changing to 20 though for e2e.

PR #1116, opened 2025-01-02
fix: handleApi and related issues (with example defineRouter with api call)

tylersayshi:
@dai-shi when I curl `http://localhost:3000/api/hi` I see the log, but the response is still html. 

Am I setting this up wrong, or is there more work to do so that we response with 'Hello World' here?


examples/22_define-router/src/entries.tsx
```diff
@@ -193,4 +193,24 @@ export default defineRouter({
...
+    {
+      path: [
+        { type: 'literal', name: 'api' },
+        { type: 'literal', name: 'hi' },
+      ],
+      mode: 'static',
+      method: 'GET',
+      handler: async () => {
+        return new Response('Hello World');
+      },
```
dai-shi, line 206: getApiConfig doesn't take these things. It's handleApi's job.

examples/22_define-router/src/entries.tsx
```diff
@@ -193,4 +193,24 @@ export default defineRouter({
...
+    },
+  ],
+  handleApi: async (path) => {
+    console.log('api called', path);
+    return {
+      headers: {},
+      status: 200,
+    };
```
dai-shi, line 214: Though, this looks valid and I expect it returns an empty content. Not sure where it returns an html.

examples/22_define-router/src/entries.tsx
tylersayshi: Ah, right. Should the return type of body be anything that is json serializable then?

Right now it's `Record<string, unknown> | null` right?

dai-shi: It's `ReadableStream | null`.

examples/22_define-router/src/entries.tsx
tylersayshi: it's just when body is undefined or null 👀 - I can look into it

tylersayshi: I found a place where we can set the body in the middleware, but not sure if that is the right place to fix this or not. Let me know what you think of the change and where may be better to fix it if that's the case

dai-shi: `body` is optional, so you don't need to set it. Or, I misunderstand your question.

tylersayshi: This change in the middleware handler: https://github.com/dai-shi/waku/pull/1116/files#diff-4142accf18d84717732f9aa0ed86505383a968a3369c6049dea5c620454322e0L157-L160

is to fix the issue in the description: https://github.com/dai-shi/waku/pull/1116#issue-2766388289

the e2e tests confirm that it is the wrong fix, but I think there is a case where we should return an empty body from the server and instead we send html still.

dai-shi: Yeah, that's a wrong fix. It's not the handler job to create the body. Maybe, I misunderstand the hono behavior.

dai-shi: Can you check if `c.body(null)` is causing the problem?

https://github.com/dai-shi/waku/blob/52f7ad61ad456df65db3221407cca810c95d7646/packages/waku/src/lib/hono/engine.ts#L64

tylersayshi: It isn't this^. 

Somehow the middleware handler is setting the `body` to a readablestream (it should just not be defined on the ctx object I think), but I can't see where it is making the change yet.

dai-shi: looking into it. it looks like it's vite behavior.

dai-shi: so, there were various bugs, most of which are my bad.

examples/22_define-router/src/entries.tsx
```diff
@@ -193,4 +193,17 @@ export default defineRouter({
     }
     throw new Error('renderRoute: No such path:' + path);
   },
+  getApiConfig: async () => [
+    {
+      path: [
+        { type: 'literal', name: 'api' },
+        { type: 'literal', name: 'hi.txt' },
```
dai-shi, line 200: The extension is necessary, if `isStatic=true`. Otherwise, unnecessary, I think.

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -203,10 +203,13 @@ const createMainViteServer = (
...
   // TODO We might be able to elminate this function
   const willBeHandled = async (pathname: string) => {
+    if (pathname.startsWith('/@')) {
+      return true;
+    }
     const vite = await vitePromise;
     try {
       const result = await vite.transformRequest(pathname);
-      return !!result;
+      return !!result?.map;
```
dai-shi, line 212: This feels too hacky. If anyone has any idea, it would be appreciated.

dai-shi:
As it turns out, there were several issues and fixed!

PR #1117, opened 2025-01-02
feat(fs-router): api file support

tylersayshi:
add initial api file router support with simple example


packages/waku/src/lib/types.ts
```diff
@@ -100,3 +101,5 @@ export type HandlerRes = {
   headers?: Record<string, string | string[]>;
   status?: number;
 };
+
+export type Method = (typeof METHODS)[number];
```
dai-shi, line 105: Can you move this and `METHODS` into create-pages.ts for now?
In the future, this might be changed, but for now, lower level api below create-pages just use `string` for method.

examples/11_fs-router/src/pages/api/hello.ts
```diff
@@ -0,0 +1,7 @@
+export async function GET() {
+  return new Response('Hello from API!');
+}
+
+export async function POST(req: Request) {
+  return new Response(`Hello from API! ${new URL(req.url).pathname}`);
```
dai-shi, line 6: My bad. That was my misunderstanding. I thought `req.url` is a URL object.

You can do this if you want. (Though, it shows "localhost", right?)

```suggestion
  return new Response(`Hello from API! ${req.url}`);
```

dai-shi: Not very important in any case. Let's merge. You can change it afterwards if you want.

tylersayshi: I'll probably change this to show the body or something when adding other examples anyways

PR #1057, opened 2024-12-10
fix(workflow): canary-ci again

dai-shi:
following #1051 and #1054, which are reverted in #1062.

- [ ] make CI with canary work
- [ ] consider supporting any react version

PR #1118, opened 2025-01-02
ci: manual canary ci job

tylersayshi:
🐤

Cleanup todo:

- [x] remove wip file
- [x] add synchronous back to normal e2e file
- [x] change reference to dai-shi/waku ... `@main` for uses field for reusable workflow file


.github/workflows/canary-e2e.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  canary-ci:
+    if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/canary-ci run') }}
+    uses: dai-shi/waku/.github/workflows/e2e.yml@main
+    with:
+      react_version: 'canary'
```
dai-shi, line 12: I would like to make `react_version` configurable with the comment.

```
/canary-ci run 19.1.0-canary-3b009b4c-20250102
```

At that point, it's not canary only, so we might want to change the command.

.github/workflows/canary-e2e.yml
```diff
@@ -0,0 +1,12 @@
+name: Canary CI
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  canary-ci:
+    if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/canary-ci run') }}
+    uses: dai-shi/waku/.github/workflows/e2e.yml@main
```
dai-shi, line 10: I think `@main` is the default and we can omit it. Oh, it's different syntax. So:

```suggestion
    uses: ./.github/workflows/ci.yml
```

.github/workflows/e2e.yml
```diff
@@ -15,6 +15,11 @@ on:
       - 'examples/**'
       - 'e2e/**'
       - '.github/workflows/e2e.yml'
+  workflow_call:
+    inputs:
+      react_version:
+        required: false
```
dai-shi, line 21: Isn't it required?

packages/waku/package.json
```diff
@@ -97,6 +97,9 @@
   "devDependencies": {
     "@netlify/functions": "^3.0.0",
     "@swc/cli": "^0.5.2",
+    "react": "19.0.0",
+    "react-dom": "19.0.0",
+    "react-server-dom-webpack": "19.0.0",
```
dai-shi, line 102: Can't we remove this?

dai-shi (commented):
1st review.

.github/workflows/canary-e2e.yml
dai-shi: IIRC, this should be in a step for visibility. See:

https://github.com/dai-shi/waku/pull/1057/files#diff-742d29bcd0bfa0c762922f6d92f9c0d0818927e44846e67a8575e289560f89b7R10-R11

dai-shi: Oh, we need to run both e2e.yml and test.yml. (previously it was combined.)

packages/waku/package.json
tylersayshi: this is necessary as a devDep. `pnpm update` does not apply to peerDeps, so without this the build ends up with two versions of react: react canary & whatever resolves to `~19.0.0` at time of the test run.

.github/workflows/canary-e2e.yml
tylersayshi: > IIRC, this should be in a step for visibility. See:

this is an invalid github action. I checked it by using the "edit file" button

image

So I changed it to not be a step to be valid. 

What is the visibility you're referring to though?

dai-shi: Oh, okay. it was about the visibility in the PR page. Let's revisit if we actually see a problem.

packages/waku/package.json
dai-shi: got it.

.github/workflows/canary-e2e.yml
dai-shi: Maybe just `canary-ci.yml`?

tylersayshi: updated the filename and added the support for the configurable with a comment run

.github/workflows/canary-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+    needs: parse-version
+    uses: ./.github/workflows/e2e.yml
+    with:
+      react_version: ${{ needs.parse-version.outputs.react_version }}
+
+  canary-ci-test:
+    needs: parse-version
+    uses: ./.github/workflows/test.yml
+    with:
+      react_version: ${{ needs.parse-version.outputs.react_version }}
```
dai-shi, line 35: Please add a new line at the end of the file.

.github/workflows/canary-ci.yml
```diff
@@ -0,0 +1,35 @@
...
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  parse-version:
+    if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/canary-ci run') }}
+    runs-on: ubuntu-latest
+    outputs:
+      react_version: ${{ steps.get-version.outputs.react_version }}
```
dai-shi, line 12: Do we need this?

.github/workflows/test.yml
```diff
@@ -24,6 +29,9 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.version }}
-          cache: 'pnpm'
-      - run: pnpm install
+          cache: ${{ !inputs.react_version && 'pnpm' || '' }}
+      - if: ${{ inputs.react_version }}
+        run: |
+          pnpm -r update react@${{ inputs.react_version }} react-dom@${{ inputs.react_version }} react-server-dom-webpack@${{ inputs.react_version }}
+      - run: pnpm install --frozen-lockfile
```
dai-shi, line 36: We decided to remove it as it's the default behavior.
```suggestion
      - run: pnpm install
```

.github/workflows/canary-ci.yml
tylersayshi: my understanding is that this is how github is able to know to store this value so it's available in the other jobs to be referenced as `react_version: ${{ needs.parse-version.outputs.react_version }}`

.github/workflows/canary-ci.yml
tylersayshi: weird, I thought prettier always did this. added though ✅

tylersayshi:
> Otherwise, it looks good to me. Is it ready?

Yeah, I think so!

.github/workflows/canary-ci.yml
dai-shi: https://github.com/pmndrs/jotai/blob/93f2e16a4b4976eaf060273ea293d07c9d6e6d61/.github/workflows/ecosystem-ci.yml is my work. I didn't use `outputs:` and it seems working. Can you try removing it?

tylersayshi:
with this failing with rc.1 - I think there is an issue with how react is being updated and installed

.github/workflows/canary-ci-wip.yml
```diff
@@ -0,0 +1,11 @@
+name: E2E WIP - Test Canary
+
+on:
+  pull_request:
+    types: [synchronize]
+
+jobs:
+  canary-ci:
+    uses: tylersayshi/waku/.github/workflows/e2e.yml@canary-ci-again
```
dai-shi, line 9: By the way, this is a very nice way. I hadn't thought about it.

dai-shi:
e2e test failing with the latest canary seems like a real issue for us (related with react-compiler). We should fix it but not in this PR. canary-ci will help confirming the fix.

PR #1126, opened 2025-01-05
fix: netlify relative fs path

### LINKED ISSUE #909: (deploy/netlify) `included_files` in netlify.toml doesn't seem effective
How to reproduce: deploy `08_cookies` example to Netlify

The error:
```
Error: ENOENT: no such file or directory, open '/var/task/netlify/private/items.json'
```

rmarscher:
Changes the netlify functions output dir from netlify/functions to netlify-functions.

This is a breaking change for existing netlify projects using dynamic rendering. It requires updating the functions directory for existing projects from the default "netlify/functions" to "netlify-functions".

Edit the netlify.toml `[functions]` section and set

```
  directory = "netlify-functions"
```

In a monorepo, the directory value needs to also include the base path from the root of the monorepo.

Fixes #909


dai-shi (approved):
LGTM!

PR #1132, opened 2025-01-07
docs: examples for using api handling defineRouter and createPages

tylersayshi:
hello world api route handlers for defineRouter and createPages

both seem to work nicely!


examples/22_define-router/src/entries.tsx
```diff
@@ -207,10 +208,44 @@ export default defineRouter({
...
+      ],
+    },
   ],
-  handleApi: async () => {
-    return {
-      status: 200,
-    };
+  handleApi: async (path) => {
+    if (path === '/api/hi.txt') {
+      const hiTxt = await readFile('./public/hi.txt');
```
dai-shi, line 220: Reading file from this place is more confusing, as we can't see if `waku build` emits public/hi.txt.

Either, just embed string like below, or place the file somewhere else.

tylersayshi: I'll place the file elsewhere so we can have both string example and file examples

Where should it go? src/files/hi.txt?

dai-shi: actually, this file should be in `dist`. so, `./private/hi.txt` would be the only solution.

tylersayshi: So it's good as is then?

dai-shi (approved):

PR #1136, opened 2025-01-08
fix(dev): proper not found handling

### LINKED ISSUE #1135: Missing route / 404 causes "Connection closed." error with dev server
```
npm create waku@latest
cd waku-project
npm run dev
open http://localhost:3000/asdf
```

It hits the root error boundary (which we are discussing removing in #1131) and displays "Connection closed."

If you then 

```
npm run build
npm run start
open http://localhost:8080/asdf
```

you will see `404 Not Found` from our default Hono `notFound` handler.

That is the expected behavior.

dai-shi:
close #1135

### DIFF
--- a/e2e/ssr-basic.spec.ts
+++ b/e2e/ssr-basic.spec.ts
@@ -25,6 +25,11 @@ for (const mode of ['DEV', 'PRD'] as const) {
       await expect(page.getByTestId('count')).toHaveText('3');
     });
 
+    test('not found', async ({ page }) => {
+      await page.goto(`http://localhost:${port}/does-not-exist.txt`);
+      await expect(page.getByText('Not Found')).toBeVisible();
+    });
+
     test('vercel ai', async ({ page }) => {
       await page.goto(`http://localhost:${port}/`);
       const aiLocator = page.getByTestId('vercel-ai');


packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -460,6 +460,8 @@ export const devServer: Middleware = (options) => {
     const body = await readablePromise;
     if (body) {
       ctx.res.body = body;
+    } else if (ctx.res.status === 404) {
+      delete ctx.res.status;
```
rmarscher, line 464: I'm guessing this was the main fix? When the body and status are set, it finalizes the hono request, but by deleting the status here, you let it pass through to the Hono error handler.

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -128,7 +128,7 @@ const createMainViteServer = (
               include: ['react-server-dom-webpack/client.edge'],
             },
           },
-          appType: 'custom',
+          appType: 'mpa',
```
rmarscher, line 131: "mpa" is a server-rendered multi-page app. and "custom" is a spa with a server for rsc?

packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
```diff
@@ -44,19 +44,23 @@ export function rscIndexPlugin(opts: {
...
-              res.setHeader('content-type', 'text/html; charset=utf-8');
-              res.end(content);
-            })
-            .catch((err) => {
-              console.error('Error transforming index.html', err);
-              res.statusCode = 500;
-              res.end('Internal Server Error');
-            });
+        server.middlewares.use((req, res, next) => {
+          if (req.url === opts.basePath) {
```
rmarscher, line 48: the html transform is only needed on the root path to support the "custom" (spa + rsc) mode?

rmarscher (approved):
:+1:

packages/waku/src/lib/middleware/dev-server-impl.ts
dai-shi: This actually doesn't matter, but feels more consistent. (I recently changed from 'custom' to 'mpa', hoping to fix something.)

packages/waku/src/lib/plugins/vite-plugin-rsc-index.ts
dai-shi: This was the major fix. This overrides (continues) Vite's default behavior.

PR #1137, opened 2025-01-08
fix(path-types): allow # and ? links for to prop

### LINKED ISSUE #1134: types for routing are too strict
The route types that are generated don't allow for search parameters like `/doc?id=test`.
I would suggest instead of generating `'/doc'` as a type it should generate 
```ts
`/doc${ `?${string}` | '' }`
```

tylersayshi:
fixes #1134

I'd like to have more robust testing for these path types but that seems lower priority to setup for now. Maybe a v1 alpha task?


dai-shi (approved):
LGTM

PR #1140, opened 2025-01-09
fix(create-pages-eg): show example with simple response

tylersayshi:
example can respond with simple response object that can accept anything the normal `Response` object would handle

https://developer.mozilla.org/en-US/docs/Web/API/Response/Response


dai-shi (approved):
Very nice.

PR #1141, opened 2025-01-09
fix(create-pages): e2e test createApi and fix method+path combos

tylersayshi:
e2e tests for createApi usage in createPages

chore: remove default main.tsx from spec (happy to revert this if we want it there for some reason)


e2e/fixtures/create-pages/src/entries.tsx
```diff
@@ -105,6 +106,34 @@ const pages: ReturnType<typeof createPages> = createPages(
...
+    }),
+
+    createApi({
+      path: '/api/hi',
+      mode: 'static',
+      method: 'GET',
+      handler: async () => {
+        return new Response('hello world!');
+      },
+    }),
```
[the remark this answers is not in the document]
tylersayshi: fixed!

packages/waku/src/router/create-pages.ts
```diff
@@ -256,12 +254,28 @@ export const createPages = <
     }
   };
 
-  const pathExists = (path: string) => {
+  const getApiRoutePath: (
+    path: string,
+    method: string,
+  ) => string | undefined = (path, method) => {
+    const allPaths = [...apiPathMap.keys()];
+    for (const pathKey of allPaths) {
```
dai-shi, line 262: Does this work?
```suggestion
    for (const pathKey of apiPathMap.keys()) {
```

packages/waku/src/router/create-pages.ts
```diff
@@ -256,12 +254,28 @@ export const createPages = <
...
+      if (m === method && getPathMapping(parsePathWithSlug(p!), path)) {
+        return p;
+      }
+    }
+  };
+
+  const pagePathExists = (path: string) => {
+    const apiRoutes = new Set(
+      [...apiPathMap.keys()].map((pathKey) => pathKey.split(' ')[1]),
+    );
```
dai-shi, line 273: Something like this:
```js
const hasApiRoute = (path) => {
  for (const pathKey of apiPathMap.keys()) {
    if (pathKey.split(' ')[1] === path) {
      return true;
    }
  }
  return false;
};
```

e2e/create-pages.spec.ts
```diff
@@ -116,5 +116,32 @@ for (const mode of ['DEV', 'PRD'] as const) {
       ).toBeVisible();
       ({ port, stopApp } = await startApp(mode));
     });
+
+    test('api hi.txt', async () => {
+      const res = await fetch(`http://localhost:${port}/api/hi.txt`);
```
dai-shi, line 121: If this is better, you might want to change define-route.spec.ts. Another PR is fine.

e2e/fixtures/create-pages/src/entries.tsx
```diff
@@ -99,6 +100,44 @@ const pages: ReturnType<typeof createPages> = createPages(
...
+      method: 'POST',
+      handler: async (req) => {
+        const body = await req.text();
+        return new Response(`POST to hello world! ${body}`);
+      },
+    }),
+
+    createApi({
+      path: '/api/empty',
+      mode: 'static',
```
dai-shi, line 135: It's probably a doc issue, but if we use static mode, we recommend using file extension like `/api/empty.txt`. Otherwise, the static build will be `/api/empty/index.html`.

packages/waku/src/router/create-pages.ts
```diff
@@ -227,27 +236,26 @@ export const createPages = <
     [PathSpec, FunctionComponent<any>]
   >();
   const apiPathMap = new Map<
-    string,
+    string, // `${method} ${path}`
```
dai-shi, line 239: Seems room for improvement, but just a nit.

packages/waku/src/router/create-pages.ts
```diff
@@ -388,12 +413,16 @@ export const createPages = <
     if (configured) {
       throw new Error('createApi no longer available');
     }
-    if (apiPathMap.has(path)) {
-      throw new Error(`Duplicated api path: ${path}`);
+    if (apiPathMap.has(`${method} ${path}`)) {
+      throw new Error(`Duplicated api path+method: ${path} ${method}`);
+    } else if (mode === 'static' && staticApiPaths.has(path)) {
+      throw new Error('Static API Routes cannot share paths: ' + path);
```
dai-shi, line 419: Oh, I see what you mean. It's a valid point.

dai-shi (approved):
LGTM

e2e/create-pages.spec.ts
tylersayshi: yep - this is definitely better. I'll change that in another PR. 

It's better because:

- it's closer to the intended use of the api routes
- it runs WAY faster in the e2e test suite

e2e/fixtures/create-pages/src/entries.tsx
tylersayshi: We *can* enforce that with types, but I think docs are the right way to encourage that to start.

packages/waku/src/router/create-pages.ts
```diff
@@ -150,12 +150,21 @@ export type CreateLayout = <Path extends string>(
...
+        mode: 'static';
+        method: 'GET';
+        handler: (req: Request) => Promise<Response>;
+      }
+    | {
+        path: Path;
+        mode: 'dynamic';
+        method: Method;
+        handler: (req: Request) => Promise<Response>;
+      },
```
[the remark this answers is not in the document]
tylersayshi: yep! If we want

packages/waku/src/router/create-pages.ts
tylersayshi: do you have any suggestions? this composite key approach feels a tiny bit tricky to use, but it solves the problem really well

dai-shi: Not right away. I'll let you know if something arises.

PR #1142, opened 2025-01-09
test(define-router): handleApi and getApiConfig e2e tests

tylersayshi:
add api handling support to define-router spec and test the result


dai-shi (approved):
LGTM

PR #1144, opened 2025-01-11
chore: example with post method for createApi

tylersayshi:
updates create pages example to use two methods on the same path and read the body for POST


examples/21_create-pages/src/entries.tsx
```diff
@@ -140,6 +140,16 @@ const pages = createPages(
       },
     }),
 
+    createApi({
+      path: '/api/hi/[name]',
+      method: 'POST',
```
dai-shi, line 145: I think with `POST` we want to have an example with reading the HTTP request body.

tylersayshi: changed!

PR #1147, opened 2025-01-13
refactor(create-pages): dynamic handlers and static handler

tylersayshi:
refactor createApi to support multiple handlers on dynamic route with single createApi call

this also enforces with types static paths only have a single handler

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -150,19 +150,20 @@ export type CreateLayout = <Path extends string>(
 
 type Method = 'GET' | 'POST' | 'PUT' | 'DELETE';
 
+type ApiHandler = (req: Request) => Promise<Response>;
+
 export type CreateApi = <Path extends string>(
   params:
     | {
-        path: Path;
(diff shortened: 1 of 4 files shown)


packages/waku/src/router/create-pages.ts
```diff
@@ -150,19 +150,20 @@ export type CreateLayout = <Path extends string>(
...
-        handler: (req: Request) => Promise<Response>;
+        handler: ApiHandler;
       }
     | {
-        path: Path;
         mode: 'dynamic';
-        method: Method;
-        handler: (req: Request) => Promise<Response>;
+        path: Path;
+        handlers: Partial<Record<Method, ApiHandler>>;
```
dai-shi, line 166: Why do you want `Partial`?

tylersayshi: I think this example explains it better than I can with words

https://tsplay.dev/wOgPpw

dai-shi: Oh, yeah. I miss that...

PR #1153, opened 2025-01-14
fix: github workflow for canary-ci

dai-shi:
#1118 seems to have some issues for comment triggers.


dai-shi:
We can't test it on the branch, so let's just merge.

PR #1164, opened 2025-01-18
fix(typegen): skip parsing errors and remove formatting

### LINKED ISSUE #1148: Remove prettier format step from generation and add formatter ignore comments
remove prettier formatting from: https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts#L73-L75

Ignore recommendations for:

- [deno fmt](https://docs.deno.com/runtime/reference/cli/fmt)
- [Prettier](https://prettier.io/)
- ~[Biome](https://biomejs.dev/formatter)~
- ~[dprint](https://dprint.dev/)~

plan for fixing: https://github.com/dai-shi/waku/issues/1148#issuecomment-2599867558

tylersayshi:
fixes: [discord comment](https://discord.com/channels/627656437971288081/1083774812129857566/1330093730354561034)

fixes #1148

related to #1163

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
@@ -172,37 +172,42 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
         }
 
         let result = `// deno-fmt-ignore-file
+// biome-ignore format: generated types do not need formatting
 /* eslint-disable */
+// prettier-ignore
 import type { PathsForPages, GetConfigResponse } from 'waku/router';\n\n`;
 
         for (const file of fileInfo) {
           const moduleName = moduleNames[file.src];
           if (file.hasGetConfig) {
-            result += `import type { getConfig as ${moduleName}_getConfig } from './${SRC_PAGES}/${file.src.replace('.tsx', '')}';\n`;
+            result += `// prettier-ignore\nimport type { getConfig as ${moduleName}_getConfig } from './${SRC_PAGES}/${file.src.replace('.tsx', '')}';\n`;
           }
         }
 
-        result += `\ntype Page =\n`;
+        result += `\n// prettier-ignore\ntype Page =\n`;
 
         for (const file of fileInfo) {
           const moduleName = moduleNames[file.src];
           if (file.hasGetConfig) {
-            result += `| ({path: '${file.path}'} & GetConfigResponse<typeof ${moduleName}_getConfig>)\n`;
+            result += `| ({ path: '${file.path}' } & GetConfigResponse<typeof ${moduleName}_getConfig>)\n`;
           } else {
-            result += `| {path: '${file.path}'; render: 'dynamic'}\n`;
+            result += `| { path: '${file.path}'; render: 'dynamic' }\n`;
           }
         }
 
-        result += `;
+        result =
+          result.slice(0, -1) +
+          `;
 
-  declare module 'waku/router' {
-    interface RouteConfig {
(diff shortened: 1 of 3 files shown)


packages/website/private/contents/post-006.mdx
```diff
@@ -69,7 +69,7 @@ export const getConfig = async () => {
 
 ### Type generation
 
-Note that a new `pages.gen.ts` file is present in your project. Check this file into your code repository for compile-time type checking in your CI/CD pipeline.
+Note that a new `pages.gen.ts` file is present in your project. Check this file into your code repository for compile-time type checking in your CI/CD pipeline. It should also be ignored by your formatter and linter (e.g. adding `pages.gen.ts` to `.prettierignore`).
```
tylersayshi, line 72: @dai-shi & @sandren is it ok to add this edit? If not, where should I document this recommendation?

maybe we should add notes about type safety to the readme?

packages/waku/src/lib/plugins/vite-plugin-fs-router-typegen.ts
```diff
@@ -177,7 +171,9 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
           }
         }
 
-        let result = `import type { PathsForPages, GetConfigResponse } from 'waku/router';\n\n`;
+        let result = `// deno-fmt-ignore-file
+/* eslint-disable */
```
dai-shi, line 175: Can we use `prettier-ignore-start`?
https://prettier.io/docs/en/ignore.html#range-ignore

tylersayshi: image

when I test it here, it does not work

dai-shi: Maybe it doesn't work for js files. Hmm.
Well, I think supporting prettier out of the box is important.
Can we add prettier-ignore for _all_ lines?

tylersayshi: > Well, I think supporting prettier out of the box is important.

agreed

image

it looks like that does technically work when used on each block

I wish there was a better way though

tylersayshi: this is the generation for 01_template with the latest changes

```ts
// deno-fmt-ignore-file
// biome-ignore format: generated types do not need formatting
/* eslint-disable */
// prettier-ignore
import type { PathsForPages, GetConfigResponse } from 'waku/router';

// prettier-ignore
import type { getConfig as About_getConfig } from './pages/about';
// prettier-ignore
import type { getConfig as Index_getConfig } from './pages/index';

// prettier-ignore
type Page =
| ({ path: '/about' } & GetConfigResponse<typeof About_getConfig>)
| ({ path: '/' } & GetConfigResponse<typeof Index_getConfig>);

// prettier-ignore
  interface RouteConfig {
    paths: PathsForPages<Page>;
  }
  interface CreatePagesConfig {
    pages: Page;
  }
}
```

packages/website/private/contents/post-006.mdx
dai-shi: I'm not sure, but I'd avoid the edit and add notes at the end instead.

Anyway, we don't need `.prettierignore` for this anymore, right? I think we can remove the addition entirely.

tylersayshi: agreed - removed ✅

dai-shi (approved):
Looks good to me. Let me merge #1163 here.

PR #1167, opened 2025-01-19
docs: PageProps for readme and add as const to getConfig return

tylersayshi:
updated readme to include the changes from: https://waku.gg/blog/type-safe-routing

added some missing code snippet langs too


sandren (approved):
LGTM! 🙂

PR #1184, opened 2025-01-23
docs: Form example improvement and removed unused type references

tylersayshi:
fixing the FIXME for the server form with a more robust form example

also removed the type references since they didn't seem to do anything


dai-shi (approved):
LGTM

PR #1188, opened 2025-01-24
docs: Form example for demo

tylersayshi:
adding a more user ready version of a progressive forms example

image


tylersayshi:
ok it's ready for review now. Sorry for all the changes.

examples/06_form-demo/src/components/Form.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const Form = ({
+  message,
+  greet,
+}: {
+  message: string;
+  greet: (formData: FormData) => Promise<void>;
+}) => (
+  <div style={{ border: '3px blue dashed', margin: '1em', padding: '1em' }}>
+    <p>{message ?? ' '}</p>
```
dai-shi, line 28: `message` is a string, so `?? ' '` would never happen.

examples/06_form-demo/src/components/funcs.ts
```diff
@@ -0,0 +1,17 @@
...
+// module state on server
+let message = '';
+
+export const getMessage = async () => message;
+
+export const greet = async (formData: FormData) => {
+  // simulate a slow server response
+  await new Promise((resolve) => setTimeout(resolve, 1000));
+  message = `Hello ${formData.get('name') || 'Anonymous'} from server!`;
+  unstable_rerenderRoute('/');
```
dai-shi, line 14: It's nice to have this in this example, but it makes me wonder if this example is more than just "form demo".

examples/06_form-demo/src/components/funcs.ts
```diff
@@ -0,0 +1,17 @@
+'use server';
+
+import { unstable_rerenderRoute } from 'waku/router/server';
+
+// module state on server
+let message = '';
```
dai-shi, line 6: Maybe as a "demo", we may want to write it a file, instead of a module variable?

examples/36_form/src/components/App.tsx
```diff
@@ -16,7 +16,7 @@ const App = ({ name }: { name: string }) => {
           <h1>Hello {name}!!</h1>
           <h3>This is a server component.</h3>
           <Counter increment={increment} />
-          <Form message={getMessage()} greet={greet} />
+          <Form message={await getMessage()} greet={greet} />
```
dai-shi, line 19: Wait, this change is very different. I think it's intentional to not have `await`.

dai-shi (commented):
While we can improve things later, the last comment blocks me to merge immediately.

examples/06_form-demo/src/components/funcs.ts
tylersayshi: What would you rename it to? 🤔

examples/36_form/src/components/App.tsx
tylersayshi: I was seeing errors in the e2e tests without this because it was attempting to render the unawaited promise which threw an error.

could you checkout the first commit of this PR and see if there's a different fix I should do?

tylersayshi: it didn't cause issue with 36_form I was just matching the behavior after seeing issues with it in the demo, but I am happy to revert just 36_form if you think it is still correct after checking

examples/06_form-demo/src/components/funcs.ts
tylersayshi: good suggestion! I added a file with some text. Let me know what you think of it :)

examples/06_form-demo/src/components/funcs.ts
dai-shi: No immediate ideas. 😅

examples/36_form/src/components/App.tsx
dai-shi: Let me check.

examples/06_form-demo/src/components/funcs.ts
tylersayshi: "form demo" seems fitting to me I think. `rerenderRoute` can be used for other things, but progressive forms will be a very normal use of it.

examples/36_form/src/components/App.tsx
dai-shi: `
Error: Objects are not valid as a React child (found: object with keys {page:/, layout:/, root, route:/, ROUTE, IS_STATIC}). If you meant to render a collection of children, use an array instead.
`

It feels like a bug somewhere. But, I can't spot it yet.

tylersayshi: it looks like the object that we use here: https://github.com/dai-shi/waku/blob/main/packages/waku/src/router/client.ts#L477-L481

So maybe we are crossing paths with the promise for the Router and the getMessage promise? idk how such a thing would happen though

tylersayshi: Also if it's helpful. It's reproducible locally on the first load of the / route after starting the dev server. Then you need to restart the server to see the issue after that.

dai-shi: If we disable SSR, the error won't happen. So, I assume it's SSR related.

dai-shi: Spent the whole day and just found out how to reproduce the error without waku/router.

```diff
diff --git a/examples/36_form/src/components/App.tsx b/examples/36_form/src/components/App.tsx
index afb38d4f..4091d6ba 100644
--- a/examples/36_form/src/components/App.tsx
+++ b/examples/36_form/src/components/App.tsx
@@ -15,7 +15,9 @@ const App = ({ name }: { name: string }) => {
         >
           <h1>Hello {name}!!</h1>
           <h3>This is a server component.</h3>
+          {/*
           <Counter increment={increment} />
+          */}
           <Form message={getMessage()} greet={greet} />
           <ServerForm />
         </div>
```

That reproduces the error with 36_form.

dai-shi (approved):
Thanks for working on this.

PR #1185, opened 2025-01-23
chore: update tailwind 4

dai-shi:
split from #1183

PR #1189, opened 2025-01-26
refactor(tailwind): bump all projects to v4 with vite tailwind plugin

tylersayshi:
docs: https://tailwindcss.com/blog/tailwindcss-v4

swapped to vite plugin to improve dev perf as well per the rec from the blog post.


e2e/fixtures/monorepo/packages/waku-project/src/styles.css
```diff
@@ -1,4 +1,21 @@
...
+*/
+@layer base {
+  *,
+  ::after,
+  ::before,
+  ::backdrop,
+  ::file-selector-button {
+    border-color: var(--color-gray-200, currentColor);
+  }
+}
```
tylersayshi, line 21: @dai-shi this snippet came from the auto-updater. I am tempted to remove it everywhere from our styles.css files since I do not believe we rely on it at all. 

Does that sound ok with you?

dai-shi: Sure, you know, I like things minimal.

tylersayshi:
@dai-shi it seems like the examples work with vite.config.ts but not with the `common` member of `unstable_viteConfig` in waku.config.ts

The website seems to be having a different bug that I am investigating - weird that that wasn't there before

dai-shi:
> but not with the `common` member of `unstable_viteConfig` in waku.config.ts

Is it for DEV or PRD? Or both?

tylersayshi:
> > but not with the `common` member of `unstable_viteConfig` in waku.config.ts
> 
> Is it for DEV or PRD? Or both?

both

dai-shi:
Now that you have tried and learned it, what do you think about #1185 approach without vite plugin? Do you think using the vite plugin is less confusing or more confusing?

tylersayshi:
> Now that you have tried and learned it, what do you think about #1185 approach without vite plugin? Do you think using the vite plugin is less confusing or more confusing?

from a user perspective, if the vite solution works I think the complexity is equivalent

postcss.config.ts or adding a vite plugin seem like the same sort of setup step to me.

---

I think it is worth following the tailwind recommendation of using the vite plugin for "performance and the best developer experience".

Maybe we could give an easier way to add vite plugins to waku though? this seems a bit verbose for just "add a plugin"

```ts
export default defineConfig({
  unstable_viteConfigs: {
    common: () => ({
      plugins: [tailwindcss()],
    }),
  },
});
```

dai-shi:
> this seems a bit verbose for just "add a plugin"

It's still "unstable".

> I think it is worth following the tailwind recommendation of using the vite plugin

I don't disagree, but we might not be ready yet.

> postcss.config.js or adding a vite plugin seem like the same sort of setup step to me.

So, I'm fine with postcss.config.js for an intermediate/temporary solution. But, it's up to you.

tylersayshi:
@dai-shi i agree with all the above. I'll set it back to the postcss setup in the morning :)

tylersayshi:
@dai-shi I was able to get this working with vite.config.ts + the hack to skip css files in processModule. Everything should be ready now I think :)

packages/website/src/styles.css
```diff
@@ -1,58 +1,64 @@
-@import url('https://fonts.googleapis.com/css2?family=Alegreya:ital,wght@0,400;0,500;0,600;0,700;0,800;0,900;1,400;1,500;1,600;1,700;1,800;1,900&display=block');
-@import url('https://fonts.googleapis.com/css2?family=Nunito:ital,wght@0,400;0,500;0,600;0,700;0,800;0,900;1,400;1,500;1,600;1,700;1,800;1,900&display=block');
-@import url('https://fonts.googleapis.com/css2?family=Inter&display=block');
-@import url('https://fonts.googleapis.com/css2?family=Fira+Code&display=block');
-@tailwind base;
-:root {
-  @apply bg-black;
+@import url('https://fonts.googleapis.com/css2?family=Alegreya:ital,wght@0,400;0,500;0,600;0,700;0,800;0,900;1,400;1,500;1,600;1,700;1,800;1,900&display=block')
```
dai-shi, line 1: Missing semicolons?

tylersayshi: hmm it's supposed to look like this: https://tailwindcss.com/docs/preflight#overview

prettier is reformatting it to this: (line added)

```css
@import url('https://fonts.googleapis.com/css2?family=Inter&display=block')
layer(base);
```

maybe we need a change in the prettier config or add a plugin? I'll look

tylersayshi: So I don't think the layer on a newline is technically a syntax error (the styles still work when it is like this)

but we can add `/* prettier-ignore */` before each of these `@import` calls if you'd prefer to see them each on their own line. The rule that it's hitting is the line width, so it is wrapping the line. You can double check this by running prettier on `@import url('https://fonts') layer(base);` (it will not line wrap)

dai-shi: I was just curious about the diff. If it's intentional, it's fine. I'm not suggesting to ignore prettier.

tylersayshi: Ok then, I'll leave it 👍

dai-shi (approved):
LGTM

We are still interested in learning why `unstable_viteConfigs.common` doesn't work for the tailwindcss vite plugin. For now, having vite.config.ts is fine, but we should be able to configure everything through waku.config.ts. Having two config files doesn't feel nice.

PR #1192, opened 2025-01-27
refactor: move skip logic from define-router to create-pages

dai-shi:
I was skeptical about it. Actually, we tried kind of both in our iteration.
Having `skip` logic in defineRouter can't support actually skipping rendering component. So, this moves the logic and the responsibility from define-router to create-pages.

cc @tylersayshi


dai-shi:
> Having `skip` logic in defineRouter can't support actually skipping rendering component.

That's my misunderstanding as create-pages doesn't render component.
Let me revert it.

PR #1194, opened 2025-01-27
fix(router): static route handling

dai-shi:
There was a bug that `skip` didn't work for rootElement and routeElement. This also does some refactors.


tylersayshi (approved):
lgtm 🚀

PR #1196, opened 2025-01-28
docs: Example with hono rpc route

tylersayshi:
This probably needs a good bit of feedback before merge, but I wanted to get the example proved out and in a place where I could share :).

I got the thought to put this together from [our discord](https://discord.com/channels/627656437971288081/1083774812129857566/1333117498723274833)

This is essentially the honorpc corollary setup to the trpc setup in create t3 turbo:
https://github.com/t3-oss/create-t3-turbo/blob/main/apps/nextjs/src/app/api/trpc/%5Btrpc%5D/route.ts


examples/07_extra-hono-router/src/components/Form.tsx
```diff
@@ -0,0 +1,127 @@
...
+    try {
+      const response = await client.api.hono.hello.$get();
+      const data = await response.json();
+      setName(data.message);
+    } catch {
+      setName('Failed to get name');
+    } finally {
+      setIsGetting(false);
+    }
+  }, []);
```
tylersayshi, line 81: I should probably refactor this to use tanstack query if we're ok with adding the library to this example

dai-shi:
I'm not sure how I feel about this. It doesn't seem like Waku's best practice. At least, it shouldn't be examples/0*. I'm not even sure hono in hono is a good practice.
trpc would fit better, if the goal is type safety. But, we should prefer server functions... So, while it's valid, adding this example feels misguiding.

examples/07_extra-hono-router/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "07_extra-hono-router",
+  "version": "0.1.0",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "dev": "VITE_EXPERIMENTAL_WAKU_ROUTER=true waku dev",
+    "build": "VITE_EXPERIMENTAL_WAKU_ROUTER=true waku build",
+    "start": "VITE_EXPERIMENTAL_WAKU_ROUTER=true waku start"
```
dai-shi, line 9: Don't add the flag in package.json.

tylersayshi:
> I'm not sure how I feel about this. It doesn't seem like Waku's best practice. At least, it shouldn't be examples/0*. I'm not even sure hono in hono is a good practice. trpc would fit better, if the goal is type safety. But, we should prefer server functions... So, while it's valid, adding this example feels misguiding.

I’ll close this then. If we want a trip example or something I’ll open a new PR for that later.

I wasn’t sure about best practice or not on this and don’t have anything against closing it.

tylersayshi:
Hmm from the hobo docs it looks like calling new Hono more than once is indeed a recommended path

https://hono.dev/docs/guides/best-practices#building-a-larger-application

dai-shi:
> https://hono.dev/docs/guides/best-practices#building-a-larger-application

I'm not sure what they do

https://github.com/honojs/hono/blob/main/src/hono-base.ts#L372-L378

is the same as what we do. Feels more indirection, and less efficient.

> I think it's a useful example

I think it's "valid" but not sure if it's recommended. Can you explain when we need this over server functions?

tylersayshi:
> I think it's "valid" but not sure if it's recommended. 

I agree

> Can you explain when we need this over server functions?

I can't really think of a reason for now honestly. Server functions probably lead to a better DX anyways.

I find this to be a really interesting experiment, but I think I agree with your hesitance to share as an example. I'll close this for good now.

PR #1199, opened 2025-01-29
feat(core): generalize elements

dai-shi:
elements are actually used beside react nodes. For now, we hack with an array. But, this should support it better, even though it loses some type checks.


tylersayshi:
What's an example of an `Element` which isn't a `ReactNode`?

dai-shi:
> What's an example of an `Element` which isn't a `ReactNode`?

We already use it in waku/router like `IS_STATIC` boolean. I'll also be using this new capability for 54_jotai.

PR #1203, opened 2025-01-30
fix(core): add hash to rsf{n}

### LINKED ISSUE #1202: Is there any way to guarantee rsf{n} endpoints are stable across builds?
I noticed that in development, calling a server action resulted in a call to an endpoint like `/RSC/F/_/{filepath}/{functionName}.txt`; however, when running the production build, those endpoints are more like `/RSC/F/assets/rsf0/{functionName}.txt`.

That `rsf0` bit is generated here: https://github.com/dai-shi/waku/blob/c0e1f3f8dedcf118790fac419f0284249e8d9971/packages/waku/src/lib/builder/build.ts#L198

The `0` appears to be based on whatever vite chunk the function ends up in, and I can see multiple endpoints in my test build along the lines of `1`, `2`, etc.

Is there any way to guarantee the stability of these endpoints across builds? What happens if I produce a new build and a server function I was calling ends up in a different vite entry, but someone's client hasn't been refreshed yet and so the client ends up calling the wrong endpoint?

dai-shi:
close #1202

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -101,12 +101,12 @@ const analyzeEntries = async (rootDir: string, config: ResolvedConfig) => {
   const wakuMinimalClientDist = decodeFilePathFromAbsolute(
     joinPath(fileURLToFilePath(import.meta.url), '../../../minimal/client.js'),
   );
-  const clientFileSet = new Set<string>([
-    wakuClientDist,
-    wakuMinimalClientDist,
+  const clientFileMap = new Map<string, string>([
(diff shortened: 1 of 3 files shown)


knpwrs:
Stale clients will still have failed requests. Can we have some sort of event on the frontend to recover from that?

dai-shi:
> Stale clients will still have failed requests. Can we have some sort of event on the frontend to recover from that?

The issue is not only with `rsf` but with `rsc` too.
I think it's for #1037 which is still under discussion. I have a vague idea, but not concrete yet.

PR #1207, opened 2025-02-03
feat: support view transitions

pmelab:
Adjust the client side router to use `startViewTransition` if available and provide and example.

- [x] Link navigation
- [x] Programmatic navigation
- [x] Browser prev/next buttons

Does not work in Firefox, since it does not support that API yet.
Also there is no automated test since I really have no idea how to test that.

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -364,7 +364,7 @@ const InnerRouter = ({
       };
 
       if ('startViewTransition' in document) {
-        performChange();
+        document.startViewTransition(performChange);
       } else {
         performChange();
       }
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -349,25 +349,17 @@ const InnerRouter = ({
   const changeRoute: ChangeRoute = useCallback(
     (route, options) => {
       const { skipRefetch } = options || {};
-      const performChange = () => {
-        startTransition(() => {
-          if (!staticPathSet.has(route.path) && !skipRefetch) {
-            const rscPath = encodeRoutePath(route.path);
-            const rscParams = createRscParams(route.query);
-            refetch(rscPath, rscParams);
-          }
-          if (options.shouldScroll) {
-            handleScroll();
-          }
-          setRoute(route);
-        });
-      };
-
-      if ('startViewTransition' in document) {
-        document.startViewTransition(performChange);
-      } else {
-        performChange();
-      }
+      startTransition(() => {
+        if (!staticPathSet.has(route.path) && !skipRefetch) {
+          const rscPath = encodeRoutePath(route.path);
+          const rscParams = createRscParams(route.query);
+          refetch(rscPath, rscParams);
+        }
+        if (options.shouldScroll) {
+          handleScroll();
+        }
+        setRoute(route);
+      });
     },
     [refetch, staticPathSet],
   );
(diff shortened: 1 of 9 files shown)


packages/waku/src/router/client.ts
```diff
@@ -349,17 +349,25 @@ const InnerRouter = ({
...
+            refetch(rscPath, rscParams);
+          }
+          if (options.shouldScroll) {
+            handleScroll();
+          }
+          setRoute(route);
+        });
+      };
+
+      if ('startViewTransition' in document && !skipRefetch) {
```
dai-shi, line 366: Why do we need `skipRefetch` here?

pmelab: I took over conditions to not start the transition if it wouldn't change anything, but that's completely wrong here.

pmelab:
I'm struggling to resolve the one remaining issue:

When clicking a broken link, we fall back to loading the RSC payload for 404 pages here:

https://github.com/dai-shi/waku/blob/b9031fed2c0cf3654af86e9e07d7e46e7ca6a2cd/packages/waku/src/router/client.ts#L470

I stepped through the debugger. The 404 data is there, and for the remaining process it should be the same as a regular route change. But for some reason we eventually end up here:

https://github.com/dai-shi/waku/blob/b9031fed2c0cf3654af86e9e07d7e46e7ca6a2cd/packages/waku/src/minimal/client.ts#L280

pmelab:
More findings:

It seems like react rendering continues while `startViewTransition` is in progress and renders the non-existent page a second time. And thats where the error happens.

pmelab:
Okay, so it turns out the solution for 404 and redirect cases is to simply don't do view transitions in these cases. But now back-navigation on error-pages does not work any more. I suppose we should skip the transitions there too, but I don't know how to detect that from within `changeRoute`. Does anybody have an idea?

dai-shi:
> 404 and redirect cases

tbh, I'm not super happy with 404 handling. What's the redirect case?

> back-navigation on error-pages

Is it only the back navigation on the error pages? Sounds tough.

pmelab:
> > 404 and redirect cases
> 
> tbh, I'm not super happy with 404 handling. What's the redirect case?

It's about these two test cases:
https://github.com/dai-shi/waku/blob/46a5e1ff34a75e0b13f89b380517fd3b71a53738/e2e/broken-links.spec.ts#L92-L116

I was not able to 100% understand what happens. In both cases the page gets the RSC payload that belongs to a different route. Without view transitions this is fine, because the route is adjusted before the second render. When view transitions are active, the second render starts again with the initial route, which has no matching file in `/RSC/R`, and that's what failed. I _think_ this happens because `startViewTransition` (maybe?) runs the state updates concurrently.

This was resolved by simply not doing a view transition for those additional route changes that happen in 404 and redirect cases..

> > back-navigation on error-pages
> 
> Is it only the back navigation on the error pages? Sounds tough.

Yeah, that's the last bit that does not work. Still need to investigate the details.

pmelab:
I just learned about `ViewTransition` coming to react (https://github.com/facebook/react/pull/31975), and in the light of this, I don't think a deep framework integration makes sense.

As a more light-weight intermediate solution I added #1217, to have optional transitions for links until then.

What do you think, @dai-shi ?

dai-shi:
Oh, I thought you have learned something from `ViewTranstion` from React. My bad. Yeah, we should follow React's capabilities.

dai-shi:
Closing in favor of #1217.

PR #1215, opened 2025-02-06
feat(create-pages): support for exactPath

tylersayshi:
escape hatch for allowing any routing conventions at all


packages/waku/src/router/create-pages.ts
```diff
@@ -107,19 +108,19 @@ export type CreatePage = <
   WildSlugKey extends string,
   Render extends 'static' | 'dynamic',
   StaticPaths extends StaticSlugRoutePaths<Path>,
+  EP extends boolean | undefined = undefined,
```
dai-shi, line 111: Can we make it explicit?
```suggestion
  ExactPath extends boolean | undefined = undefined,
```

dai-shi (approved):
LGTM

PR #1217, opened 2025-02-07
feat: link view transitions

pmelab:
Alternative approach to #1207. Instead of integrating view transitions for each route change, they are just optionally added to the `Link` click event.

* Less intrusive
* Other cases (e.g. programmatic navigation) can be handled on the application level.
* Does not support browser prev/next buttons.


dai-shi:
Is this a temporary solution until React provides `ViewTransitioin`?

pmelab:
> Is this a temporary solution until React provides ViewTransitioin?

Yes. It's not perfect, but maybe good enough until then.

dai-shi:
I'm not super familiar with view transitions, but when we use it, do we use `isPending` from `useTransition` as well? Looking at the code, it feels a little bit overlapping.

dai-shi:
Let me add some commits.

dai-shi (approved):
My change makes it less handy, but I would like it as a temporary solution.

pmelab:
Sorry, that was too fast for me. I still had it on my list to investigate if this works for all cases.

> I'm not super familiar with view transitions, but when we use it, do we use isPending from useTransition as well? Looking at the code, it feels a little bit overlapping.

`startViewTransition` takes a snapshot, executes the transition callback, waits for any promise that was returned and then transitions to the new DOM state. I think in our case it would wait for `changeRoute` to  only, but this is not actually the full 
transition if the route takes longer to resolve, right?

I'll create some new tests for that when I have some time.

dai-shi:
I thought the example was working well.
Isn't it working?

> I still had it on my list to investigate if this works for all cases.

Oops, sorry about it.

PR #1219, opened 2025-02-08
feat(minimal): unstable_suspendWithPrev

dai-shi:
for #1212


dai-shi:
https://github.com/dai-shi/waku/discussions/1212#discussioncomment-12171263

let's revisit it in the future.

PR #1222, opened 2025-02-10
feat(fs-router): allow _components & _hooks in the pages folder

### LINKED ISSUE #1220: feat(fs-router): allow components in the pages folder
Currently we can put components in `./src/components` or anywhere and use them in pages.
For example, `./src/components/counter.tsx` and `./src/pages/foo/index.tsx`
However, sometimes people want to put component files in the pages folder, especially if the components are not reusable, but they are only separated to get organized.
So, for example, `./src/pages/foo/_components/counter.tsx`.
It's currently not possible because it becomes a new page.
I wonder if we can treat `_components` a special name for avoiding generating pages.
cc @tylersayshi

tylersayshi:
ignores any files inside `_components` and `_hooks` folders inside `src/pages`.

Previously, these would be considered to be pages but they will now be ignored. This allows easier co-location of components with their pages.

resolves #1220


dai-shi (approved):
LGTM
We need docs for fs-router...

PR #1224, opened 2025-02-11
Fix import of platform data file for cloudflare builds when using dynamic routes

### LINKED ISSUE #1223: Bug: `pathConfig.find is not a function` cloudflare deployment target error since `0.21.17`
I'm attempting to get a waku-based app working for Cloudflare Workers, however **since version `0.21.17`** (_`0.21.16`_ seems to work fine) I get the following error:

```
pathConfig.find is not a function
```

I haven't been able to fully identify why this error is happening, but with a bit of local debugging it seems to point to this code: https://github.com/dai-shi/waku/blob/cf58f612527df9a5b708f225320c5760612664cb/packages/waku/src/router/define-router.ts#L185-L186

`pathConfig` [here](https://github.com/dai-shi/waku/blob/cf58f612527df9a5b708f225320c5760612664cb/packages/waku/src/router/define-router.ts#L185) seems to have a `.default` that is a getter that is the array of paths.

Patching the logic to access `.default` after the await - runs into yet another error so I don't think that's the full solution.

Here's a relatively minimal reproduction of the issue: https://github.com/hamlim/waku-cloudflare-assets

To reproduce:
- Clone the repo
- Run `bun i` to install deps
- Run `bun run worker:dev`
  - This command does a cloudflare build and then runs `wrangler dev`

If you revert [this commit](https://github.com/hamlim/waku-cloudflare-assets/commit/f6f8eb33dc9b5c52c7f77f4ed098a0de2234796f), and re-run `bun i` and then `bun run worker:dev` - the app seems to load fine!

hamlim:
Additionally, it then applies a fix to the generated `entries.js` file from cloudflare builds to access the `default` export from the `platform-data/` asset that is imported.

I used a new example app to test this change (which I can remove if we don't want to add it), before applying the fix I was able to reproduce the `pathConfig.find` bug from #1223, and then after the fix the issue no longer happened!

As noted above, and example app probably isn't the best use for this manual test fixture, ideally it'd be an automated test fixture like those within the `e2e` directory, however I'm not sure if the existing e2e tests run against the cloudflare deployment target or not 🤔 

Fixes #1223


packages/waku/src/lib/plugins/vite-plugin-deploy-cloudflare.ts
```diff
@@ -231,7 +231,8 @@ export function loadPlatformData(key) {
   switch (key) {
     ${Array.from(keys)
       .map(
-        (k) => `case '${k}': return import('./${DIST_PLATFORM_DATA}/${k}.js');`,
+        (k) =>
+          `case '${k}': return import('./${DIST_PLATFORM_DATA}/${k}.js').then(m => m.default);`,
```
hamlim, line 235: This is the core change - other changes in this file were from running prettier.

Also to note - this now mirrors the structure here within the `build.ts` file:

https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/builder/build.ts#L791

dai-shi (commented):
Isn't is possible to modify 44_cloudflare example instead of adding a new one?

hamlim:
> Isn't is possible to modify 44_cloudflare example instead of adding a new one?

Sure! I'll apply the changes to that existing example!

examples/44_cloudflare/wrangler.toml
```diff
@@ -2,7 +2,7 @@
 # See https://developers.cloudflare.com/workers/wrangler/configuration/
 name = "waku-project"
 compatibility_date = "2024-09-23"
-compatibility_flags = [ "nodejs_als" ]
+compatibility_flags = [ "nodejs_compat" ]
```
dai-shi, line 5: I would like to make sure it works with `nodejs_als` only. Can you avoid `readFile` and read just test from js file instead?

dai-shi (commented):
Please preserve the format with prettier.

examples/44_cloudflare/wrangler.toml
hamlim: Updated!

hamlim:
Formatting should be fixed, and I was able to change back to the `nodejs_als` compat flag as well!

examples/44_cloudflare/src/entries.tsx
```diff
@@ -1,56 +1,159 @@
-import { unstable_defineEntries as defineEntries } from 'waku/minimal/server';
```
dai-shi, line 1: I'm not sure what's the original intention of using waku/minimal/server over createPages, but maybe it's fine. cc @rmarscher

dai-shi (approved):
Thanks for your contribution!

examples/44_cloudflare/src/entries.tsx
rmarscher: I think I was just copying other examples that seemed to limit the surface area of the test by using the minimal API.

I was actually working on a new example that could be used a template for Cloudflare projects when using create-waku or c3. It's similar to the default template but with a few extra Cloudflare-specific examples and the dev server middleware.

PR #1227, opened 2025-02-13
feat(create-pages): pathless routes or grouped routes

tylersayshi:
This will add support for grouped paths in createPages. resolves #977 

TODO

- [x] pages: Add failing test
- [x] pages: Implement support to pass the tests
- [x] layouts: Add failing test
- [x] layouts: Implement support to pass the tests
- [x] path typegen should remove the groups
- [x] root level group layouts do not yet work (e.g. `/(home)`)


tylersayshi:
@rmarscher I'd appreciate help stress testing this. The initial implementation here feels almost too simple (I'm expecting edge cases and whatnot)

packages/waku/tests/create-pages.test.ts
```diff
@@ -1581,3 +1581,126 @@ describe('createPages - exactPath', () => {
...
+        render: 'static',
+        path: '/(group)/test',
+        component: TestPage,
+      }),
+    ]);
+    const { getRouteConfig, handleRoute } = injectedFunctions();
+    expect(await getRouteConfig()).toEqual([
+      {
+        elements: {
+          'layout:/(group)': { isStatic: true },
```
tylersayshi, line 1686: @dai-shi is this correct or does it need to be `layout:/`?

rmarscher:
> The initial implementation here feels almost too simple

Yeah, it seems correct to just strip the path out in the right places though. Did you test adjacent groups and routes with multiple groups?

/(home)/_layout.tsx
/(home)/index.tsx
/(about)/_layout.tsx
/(about)/about.tsx
/(about)/about/(me)/_layout.tsx
/(about)/about/(me)/me.tsx
/_layout.tsx

I saw you are already testing that it throws an error for conflicting routes. Looks good.

tylersayshi:
> Yeah, it seems correct to just strip the path out in the right places though. Did you test adjacent groups and routes with multiple groups?

thanks for the suggestions! I added a couple todos from testing. The rest seemed good

packages/waku/tests/create-pages.test.ts
dai-shi: Yes, because client caches elements based on ids.

You want to add a test case for:
- createLayout `path: '/'` and
- createLayout `path: '/(group)'`

dai-shi: Also, add a case for
- createPage `path: '/(group)'` which should fail?

tylersayshi: > createPage path: '/(group)' which should fail?

This would likely be a duplicate of `/` since that page is usually present. It shouldn't fail on its own though because in theory you could have 
```
/(home)/index
/(home)/other
/foo
```
since there is no /index, there is no duplicate error here.

dai-shi (approved):
I'm not 100% sure if this covers all edge cases, but at least there shouldn't be any regressions.

PR #1231, opened 2025-02-15
feat(router): configuration of scroll behavior for Link and useRouter

tylersayshi:
follow up for: https://github.com/dai-shi/waku/pull/1098#issuecomment-2660096203

cc @wesbos thanks for this call out


packages/waku/src/router/client.ts
```diff
@@ -114,7 +114,16 @@ export function useRouter_UNSTABLE() {
   }
   const { route, changeRoute, prefetchRoute } = router;
   const push = useCallback(
-    (to: InferredPaths) => {
+    (
+      to: InferredPaths,
+      options?: {
+        /**
+         * indicates if the link should scroll or not on navigation
+         * @default true
```
dai-shi, line 122: I don't like options being default true.

Looking at the code, it's a doc issue. It's not "default true".

packages/waku/src/router/client.ts
```diff
@@ -125,16 +134,27 @@ export function useRouter_UNSTABLE() {
...
+         */
+        scroll?: boolean;
+      },
+    ) => {
       const url = new URL(to, window.location.href);
       const newPath = url.pathname !== window.location.pathname;
       window.history.replaceState(window.history.state, '', url);
-      changeRoute(parseRoute(url), { shouldScroll: newPath });
+      const shouldScroll = options?.scroll ?? newPath;
+      changeRoute(parseRoute(url), { shouldScroll: shouldScroll });
```
dai-shi, line 157:
```suggestion
      const shouldScroll = options?.scroll ?? newPath;
      changeRoute(parseRoute(url), { shouldScroll });
```

OR

```suggestion
      changeRoute(parseRoute(url), { shouldScroll: options?.scroll ?? newPath });
```

packages/waku/src/router/client.ts
```diff
@@ -186,6 +211,7 @@ export function Link({
   unstable_prefetchOnEnter,
   unstable_prefetchOnView,
   unstable_startTransition,
+  scroll = true,
```
dai-shi, line 214: Hmm, I don't prefer "default true". Can it be the same default behavior with `newPath` like useRouter?

tylersayshi:
@dai-shi all very true. I didn't give the api enough though. How about something like this:
```ts
/**
 * Indicates that the link should not scroll when clicked. By default, waku will 
 * scroll when the path changes. This is a way to explicitly not change the scroll 
 * position on navigation.
 */
 dontScroll?: boolean;
```
alternate names: `disableScroll` or `noScroll`

Let me know what you think and I'll change the code accordingly

dai-shi:
@tylersayshi I'm afraid my comment wasn't very clear.

I'm actually fine with `scroll` naming, but the options are:

- `undefined` (default): scroll based on `newPath` check (= current Link behavior)
- `false`: no scroll
-  `true`: always scroll

packages/waku/src/router/client.ts
```diff
@@ -125,16 +136,30 @@ export function useRouter_UNSTABLE() {
...
+         */
+        scroll?: boolean;
+      },
+    ) => {
       const url = new URL(to, window.location.href);
       const newPath = url.pathname !== window.location.pathname;
       window.history.replaceState(window.history.state, '', url);
-      changeRoute(parseRoute(url), { shouldScroll: newPath });
+      const shouldScroll = options?.scroll ?? newPath;
+      changeRoute(parseRoute(url), { shouldScroll });
```
dai-shi, line 162: The coding style is not very consistent with `push` L140.

dai-shi:
```suggestion
      changeRoute(parseRoute(url), {
        shouldScroll: options?.scroll ?? newPath,
      });
```

tylersayshi: ah, good catch, fixed

packages/waku/src/router/client.ts
```diff
@@ -125,16 +136,29 @@ export function useRouter_UNSTABLE() {
         '',
         url,
       );
-      changeRoute(parseRoute(url), { shouldScroll: newPath });
+      changeRoute(parseRoute(url), {
+        shouldScroll: options?.scroll ?? newPath,
+      });
```
dai-shi, line 141: Maybe this too?
```suggestion
      changeRoute(parseRoute(url), { shouldScroll: options?.scroll ?? newPath });
```

tylersayshi: the multi line one is what prettier changed both to... I need to fix my local prettier it seems

dai-shi: oh, I see.

dai-shi (approved):
LGTM

PR #1232, opened 2025-02-16
feat(create-waku): --project-name option

### LINKED ISSUE #1208: create-waku non-interactive
I'm working on a PR to add create-waku to Cloudflare's C3 CLI tool. 

I'm trying to figure out if there is a way to run create-waku so that it doesn't require any interactive prompts.

I'm not sure if there is a way to pre-fill the project name. I think it would be expected to use the first positional argument for that. Or a `--project-name` option would be fine too. It was looking at the code and at the `prompts` npm package and I'm not seeing it.

In the create-cloudflare implementation for Remix - https://github.com/cloudflare/workers-sdk/blob/main/packages/create-cloudflare/templates-experimental/remix/c3.ts#L13 - the project name is passed as the first positional argument to the `npm create remix` command.

The create-remix cli has a block of code that attempts to detect if it should run in a non-interactive mode. They also look for a custom environment variable (CREATE_REMIX_FORCE_INTERACTIVE).

```ts
  // Adapted from https://github.com/sindresorhus/is-interactive
  return Boolean(
    process.stdout.isTTY &&
      process.env.TERM !== "dumb" &&
      !("CI" in process.env)
  );
```

I could potentially help and create a PR for create-waku. Thanks!

rmarscher:
Fixes #1208


packages/create-waku/src/index.ts
```diff
@@ -98,17 +101,20 @@ async function doPrompts() {
...
         {
           name: 'projectName',
-          type: 'text',
+          type: values['project-name'] ? null : 'text',
           message: 'Project Name',
           initial: defaultProjectName,
-          onState: (state: any) => (targetDir = String(state.value).trim()),
+          onState: (state: any) =>
+            (targetDir =
+              String(values['project-name']).trim() ||
```
rmarscher, line 116: This might be unnecessary... checking...

rmarscher: Yeah, removed the change from there.

dai-shi: Do we still need this?
```js
let targetDir = values['project-name'] || '';
```
Or,
```js
let targetDir = '';
```
works?

rmarscher: Yes, we still need that. Otherwise, it prompts for package name because `targetDir` is empty.

dai-shi (commented):
Thanks for confirming. Then, I hope this works.

packages/create-waku/src/index.ts
```diff
@@ -98,14 +101,14 @@ async function doPrompts() {
   const templateNames = await getTemplateNames(templateRoot);
 
   const defaultProjectName = 'waku-project';
-  let targetDir = '';
+  let targetDir = values['project-name'] || '';
```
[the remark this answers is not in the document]
rmarscher: Yes. I made this update and tested it. Looks good.

dai-shi (approved):
LGTM

Issue #1228, opened 2025-02-13
Feature Request: Support for `bun` package manager when running `create-waku`


hamlim:
Would you be open to also supporting `bun` as a package manager when creating a new waku app?

To be clear, running `bun create waku@latest` works today, but it falls back on `npm` to install dependencies after creating the app, which results in a `package-lock.json`. Usually what I then do is cd into the new app and `rm package-lock.json` and then run `bun i` manually.

It'd be nice to automatically infer `bun` like what happens for `pnpm`, `yarn`, and `npm` already today!

dai-shi:
If you are suggesting a support in https://github.com/dai-shi/waku/blob/main/packages/create-waku/src/index.ts, feel free to send a PR!

hamlim:
Yea, I'll open a PR soon!

PR #1233, opened 2025-02-16
Refactor 44_cloudflare as a template for Cloudflare C3

rmarscher:
I am working on a PR to add waku to create-cloudflare and I'd like to turn 44_cloudflare into the template that we use. 

Running

```sh
npm create cloudflare@latest my-waku-app -- --framework=waku --experimental`
```

 would invoke

```sh
npm create waku@latest -- --project-name my-waku-app --example https://github.com/dai-shi/waku/tree/main/examples/44_cloudflare
```

if they accept the PR.

If I understand correctly, #1228 was reproducible with an "out of the box" `npm create waku@latest` with the managed entries and filesystem router. I am pretty sure this does not revert any of the important parts of #1224 and 
this template would still catch that bug. But I'd appreciate your review @hamlim. Thanks! I replaced the src/components with the defaults from examples/01_template. I added an example for accessing the Hono context (and Cloudflare execution context) at runtime but how to avoid errors at build time.


examples/01_template/src/pages/index.tsx
```diff
@@ -6,7 +6,7 @@ export default async function HomePage() {
   const data = await getData();
 
   return (
-    <div>
+    <div className="container">
```
dai-shi, line 9: Is this by mistake?

dai-shi:
> I'd like to turn 44_cloudflare into the template that we use.

If that's the purpose, let's rename 44_cloudflare to 07_cloudflare.

rmarscher:
> I think you mean #1223 not #1228, and I think that was true if I remember correctly.
> tl;dr: Yea I think your changes still will test/validate that flow
Whoops, yes. I copy/pasted the wrong issue number. Thank you!

examples/01_template/src/pages/index.tsx
rmarscher: Yes.

rmarscher:
I'm seeing an issue with Suspense not working when I run the built code. The server seems to be waiting for the promise to resolve. It is working in dev mode.

rmarscher:
> I'm seeing an issue with Suspense not working when I run the built code. The server seems to be waiting for the promise to resolve. It is working in dev mode.

It looks like this isn't a new issue. I went back to waku v0.21.9 and was able to reproduce. It's only an issue with the wrangler dev server. It doesn't happen when you actually deploy to Cloudflare. Maybe their dev server doesn't support streaming responses -- or maybe they have a flag to enable streaming responses. I'll investigate and open another issue for that. I think this PR is ready for review / merge. Thanks.

dai-shi (approved):
Thanks for working on it.

rmarscher:
> It looks like this isn't a new issue. I went back to waku v0.21.9 and was able to reproduce. It's only an issue with the wrangler dev server. It doesn't happen when you actually deploy to Cloudflare. Maybe their dev server doesn't support streaming responses -- or maybe they have a flag to enable streaming responses. I'll investigate and open another issue for that.

#1237

PR #1234, opened 2025-02-17
fix(tailwind): swap @tailwindcss/vite for @tailwindcss/postcss

tylersayshi:
some nested components not working with the tailwind vite plugin. the hmr seemed broken when using the tailwind plugin as well.

related issue:
https://github.com/tailwindlabs/tailwindcss/issues/16585


dai-shi (approved):
LGTM

dai-shi:
@tylersayshi ah, we missed `07_cloudflare`. please open a follow-up pr.

rmarscher:
@tylersayshi Do you think Waku will switch back to the tailwind plugin instead of postcss since they are addressing the original issue? Does the tailwind vite plugin handle vendor prefixes based on browserslist? I couldn't find any mention of it.

tylersayshi:
> Do you think Waku will switch back to the tailwind plugin instead of postcss since they are addressing the original issue?
yes! tailwind released a change that is expected to fix the issues we were seeing, but we do need to still test those changes before recommending it again. If you have time to try out the tailwind vite plugin on your project(s) that would be helpful!
> Does the tailwind vite plugin handle vendor prefixes based on browserslist?
no idea... sounds like a good question for the tailwind discussions board.

dai-shi:
Please be aware we had two issues in the previous plugin. Another one was hot reload, right?

tylersayshi:
> Please be aware we had two issues in the previous plugin. Another one was hot reload, right?

right! i'd suspect we may need to do some work to integrate the plugin through the waku config `viteConfigs` field to get it working as expected all the time.

dai-shi:
@tylersayshi should we revisit this again?

tylersayshi:
> @tylersayshi should we revisit this again?

yep! If I have time I'll try to get to it, otherwise I made the issue and tagged it as a good first issue. It'd be really good for any new contributors.

dai-shi:
I'm not sure if it's good as a first issue, but maybe so if @tylersayshi can review the PR.

PR #1238, opened 2025-02-17
feat(create-waku): --template option

rmarscher:
Also related to #1208. I didn't realize template was different from example.

### DIFF
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -70,6 +70,9 @@ const { values } = parseArgs({
     example: {
       type: 'string',
     },
+    template: {
+      type: 'string',
+    },
     help: {
       type: 'boolean',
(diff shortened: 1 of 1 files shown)


dai-shi:
My intention was to use `--example` for template too. You can specify the example with the tag.

dai-shi:
Hm, but yeah, this makes sense for your use case.

packages/create-waku/src/index.ts
```diff
@@ -70,6 +70,9 @@ const { values } = parseArgs({
     example: {
       type: 'string',
     },
+    template: {
+      type: 'string',
+    },
```
dai-shi, line 75: Please update the usage `displayUsage`.

dai-shi (approved):
Should be good.

PR #1242, opened 2025-02-17
Fix response streaming with wrangler dev server

### LINKED ISSUE #1237: Streaming dynamic server responses are buffered when testing with `wrangler dev`
It doesn't happen in production with a Cloudflare Worker or with `waku dev`. I found a [related issue](https://github.com/cloudflare/workers-sdk/issues/8004) in the workers-sdk repo. It seems to be due to automatic compression of html and txt responses. It looks like there is a [workaround](https://github.com/cloudflare/workers-sdk/issues/6577#issuecomment-2387260894) - but I think we can wait for an upstream fix to [#8004](https://github.com/cloudflare/workers-sdk/issues/8004).

rmarscher:
Fixes #1237.

### DIFF
--- a/examples/07_cloudflare/src/lib/hono.ts
+++ b/examples/07_cloudflare/src/lib/hono.ts
@@ -1,4 +1,4 @@
-import { getHonoContext as _getHonoContext } from 'waku/unstable_hono';
+import { getHonoContext as getHonoContextFromWaku } from 'waku/unstable_hono';
 import { isBuild } from './waku';
 
 export function getHonoContext() {
@@ -6,7 +6,7 @@ export function getHonoContext() {
     return undefined;
   }
   try {
-    return _getHonoContext<{ Bindings: Env }>();
+    return getHonoContextFromWaku<{ Bindings: Env }>();
   } catch (e) {
     if (isHonoContextUnavailableError(e)) {
       return undefined;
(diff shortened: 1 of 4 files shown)


examples/07_cloudflare/waku.cloudflare-middleware.ts
```diff
@@ -0,0 +1,74 @@
...
+export type HandlerRes = {
+  body?: ReadableStream;
+  headers?: Record<string, string | string[]>;
+  status?: number;
+};
+export type HandlerContext = {
+  readonly req: HandlerReq;
+  readonly res: HandlerRes;
+  readonly context: Record<string, unknown>;
+};
```
rmarscher, line 20: Can we export this from waku? I think only the Middleware type is exported from `'waku/config'` and I don't see a way to extract HandlerContext.

examples/07_cloudflare/src/lib/hono.ts
```diff
@@ -1,12 +1,12 @@
-import { getHonoContext as getHonoContextFromWaku } from 'waku/unstable_hono';
+import { getHonoContext as _getHonoContext } from 'waku/unstable_hono';
```
dai-shi, line 1: It's our coding style to use `_` prefix for "unused values", so we want to avoid using `_` for other meanings.

examples/07_cloudflare/waku.cloudflare-middleware.ts
dai-shi: I'll consider it in the future, but not soon.

Meanwhile:
```ts
type HandlerContext = Parameters<ReturnType<Middleware>>[0];
```

rmarscher: Thanks. I don't think I've used `<Parameters>` before. Good to know.

rmarscher:
Thanks for the review @dai-shi. I resolved the comments.

dai-shi (approved):
It looks pretty hacky, but I assume it's unavoidable.

PR #1246, opened 2025-02-19
fix(router): _components and _hooks should be ignored by path typegen

### LINKED ISSUE #1244: Placing .tsx files in `_components` and `_hooks` directory generates paths in `pages.gen.ts`
https://github.com/user-attachments/assets/24ebffad-fd0c-4810-9388-32ab4bcb370d

<br/>

If the extension is changed to .ts, paths are not generated.
Accessing the generated paths in the browser results in '404 Not Found'.

tylersayshi:
fixes #1244

### DIFF
--- a/packages/waku/src/lib/utils/path.ts
+++ b/packages/waku/src/lib/utils/path.ts
@@ -205,9 +205,3 @@ export const getPathMapping = (
   }
   return mapping;
 };
-
-const IGNORED_PATH_PARTS = new Set(['_components', '_hooks']);
-
-/** Ignore paths like `_components` and `_hooks` in pages dir */
-export const isIgnoredPath = (paths: string[]) =>
-  paths.some((p) => IGNORED_PATH_PARTS.has(p));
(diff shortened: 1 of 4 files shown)


packages/waku/src/lib/utils/path.ts
```diff
@@ -205,3 +205,9 @@ export const getPathMapping = (
   }
   return mapping;
 };
+
+const IGNORED_PATH_PARTS = new Set(['_components', '_hooks']);
```
dai-shi, line 209: I'm not very comfortable with having this in this file.
This is very specific to `fs-router`.
If you want to have it outside `src/router/fs-router.ts`, consider `src/lib/utils/fs-router.ts`.

tylersayshi: moved to `src/lib/utils/fs-router.ts` ✅ 
it seemed wrong to import from `src/router/fs-router.ts` to the typegen plugin, so a utils file makes sense to me.

dai-shi (approved):
LGTM

PR #1249, opened 2025-02-19
feat: bypass Link click handler if alt click or defaultPrevented

### LINKED ISSUE #1248: It should be possible to open a Link in a new tab using browser default behavior
We always preventDefault() on our Link component. I think we should have an option to allow a meta-click to follow the default browser behavior.

rmarscher:
Fixes #1248.

### DIFF
--- a/e2e/fs-router.spec.ts
+++ b/e2e/fs-router.spec.ts
@@ -81,5 +81,18 @@ for (const mode of ['DEV', 'PRD'] as const) {
       await page.goto(`http://localhost:${port}/_components/Counter`);
       await expect(page.getByText('404 Not Found')).toBeVisible();
     });
+
+    test('alt click', async ({ page }) => {
+      await page.goto(`http://localhost:${port}`);
+      await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
+      await page.click("a[href='/foo']", {
+        button: "right",
+      });
+      await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
+      await page.click("a[href='/foo']", {
+        modifiers: ['ControlOrMeta'],
+      });
+      await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
+    });
   });
 }
--- a/e2e/fs-router.spec.ts
+++ b/e2e/fs-router.spec.ts
@@ -86,7 +86,7 @@ for (const mode of ['DEV', 'PRD'] as const) {
       await page.goto(`http://localhost:${port}`);
       await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
       await page.click("a[href='/foo']", {
-        button: "right",
+        button: 'right',
       });
       await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
       await page.click("a[href='/foo']", {
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -81,12 +81,12 @@ const parseRouteFromLocation = (): RouteProps => {
   return parseRoute(new URL(window.location.href));
 };
 
-function isAltClick(event: MouseEvent<HTMLAnchorElement>) {
+const isAltClick = (event: MouseEvent<HTMLAnchorElement>) => {
   return (
     event.button !== 0 ||
     !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey)
   );
-}
+};
 
 let savedRscParams: [query: string, rscParams: URLSearchParams] | undefined;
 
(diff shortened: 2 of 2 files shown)


rmarscher:
I think this is a draft until we create an e2e test.

packages/waku/src/router/client.ts
```diff
@@ -261,9 +263,16 @@ export function Link({
...
+    const isAltClick =
+      unstable_defaultAltClick &&
+      Boolean(
+        event.metaKey || event.ctrlKey || event.shiftKey || event.button !== 0,
+      );
     const url = new URL(to, window.location.href);
-    if (url.href !== window.location.href) {
+    if (!isAltClick) {
+      event.preventDefault();
+    }
```
dai-shi, line 274: Why can it be this?
```js
if (isAltClick) {
  return;
}
```

rmarscher: That skips the onClick prop. What do you think about firing the onClick prop first? React Router does that and then doesn't run the internal handler if the onClick calls preventDefault().
https://github.com/remix-run/react-router/blob/react-router%407.2.0/packages/react-router/lib/dom/lib.tsx#L628-L635

dai-shi: Oh, that looks great. Any drawbacks?

rmarscher: Not that I can think of. It looks good to me. I'm traveling for the next few days so I probably won't push another update until the beginning of next week.

dai-shi: ping

packages/waku/src/router/client.ts
```diff
@@ -81,6 +81,13 @@ const parseRouteFromLocation = (): RouteProps => {
   return parseRoute(new URL(window.location.href));
 };
 
+function isAltClick(event: MouseEvent<HTMLAnchorElement>) {
```
dai-shi, line 84: Can you make it an arrow function? (Just a preference)

packages/waku/src/router/client.ts
```diff
@@ -265,8 +272,7 @@ export function Link({
       };
     }
   }, [unstable_prefetchOnView, router, to]);
-  const onClick = (event: MouseEvent<HTMLAnchorElement>) => {
-    event.preventDefault();
+  const internalOnClick = (_event: MouseEvent<HTMLAnchorElement>) => {
```
dai-shi, line 275: It's not confusing without it.
```suggestion
  const internalOnClick = () => {
```

packages/waku/src/router/client.ts
```diff
@@ -265,8 +272,7 @@ export function Link({
       };
     }
   }, [unstable_prefetchOnView, router, to]);
-  const onClick = (event: MouseEvent<HTMLAnchorElement>) => {
-    event.preventDefault();
```
dai-shi, line 269: Removing this causes e2e tests to fail, doesn't it?

packages/waku/src/router/client.ts
rmarscher: All set

packages/waku/src/router/client.ts
rmarscher: Yes, that was a mistake. I thought I moved it, not deleted.

packages/waku/src/router/client.ts
```diff
@@ -81,6 +81,13 @@ const parseRouteFromLocation = (): RouteProps => {
   return parseRoute(new URL(window.location.href));
 };
 
+const isAltClick = (event: MouseEvent<HTMLAnchorElement>) => {
+  return (
```
dai-shi, line 85: oops, I mean to avoid the return statement too.

dai-shi (approved):
Looks nice.

e2e/fs-router.spec.ts
```diff
@@ -81,5 +81,18 @@ for (const mode of ['DEV', 'PRD'] as const) {
       await page.goto(`http://localhost:${port}/_components/Counter`);
       await expect(page.getByText('404 Not Found')).toBeVisible();
     });
+
+    test('alt click', async ({ page }) => {
```
dai-shi, line 85: @rmarscher This test seems a little bit flaky.
https://github.com/wakujs/waku/actions/runs/14210540846/job/39816803332

dai-shi: https://github.com/wakujs/waku/actions/runs/14210920213/job/39817854884
It happens often on macos.

dai-shi: https://github.com/wakujs/waku/actions/runs/14218743873/job/39841399899
Happening more than expected.

dai-shi: https://github.com/wakujs/waku/pull/1319#issuecomment-2777395732

@tylersayshi

tylersayshi: This seems worth a try: https://github.com/wakujs/waku/pull/1351

dai-shi: I can retry "E2E on macos-latest (Node 22.7.0) - (2/4)" several times there.

PR #1256, opened 2025-02-21
fix(create-pages): Ensure dynamic layouts receive path prop

hamlim:
The logic seemed to be reversed when checking if the current layout segment is dynamic or static (see `!` in create-pages.ts), so the `path` prop was never provided.

Didn't run the E2E tests locally but verified by adding a temp test to `create-pages.test.ts` locally.


tylersayshi:
Yep this is exactly right, thanks for the fix… my bad here 🙈

tylersayshi (approved):

PR #1257, opened 2025-02-22
fix(router): avoid default startTransition

### LINKED ISSUE #1255: Dynamic Layouts with Suspense boundaries do not stream when changing pages
I've modified 01_template to show the issue: https://github.com/wesbos/waku-dynanic-suspense/tree/main/examples/01_template 

When a layout is dynamic and there is a suspense boundary in that layout, Clicking a <Link> to changing pages waits for the entire output to be resolved. 

It should change pages as soon as the unsuspended bits of the components are rendered. 

My use case: I have a dynamic component in the footer of my layout. I actually don't want it to re-run every page change, but that might be another issue.

Here is a video showing the issue: https://share.cleanshot.com/7L0FXm9N

dai-shi:
close #1255

- [x] add failing test
- [x] fix it

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -3,7 +3,6 @@
 import {
   createContext,
   createElement,
-  startTransition,
   useCallback,
   useContext,
   useEffect,
@@ -196,8 +195,7 @@ export function useRouter_UNSTABLE() {
 
 export type LinkProps = {
   to: InferredPaths;
-  pending?: ReactNode;
-  notPending?: ReactNode;
+  children: ReactNode;
   /**
    * indicates if the link should scroll or not on navigation
    * - `true`: always scroll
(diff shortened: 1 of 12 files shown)


packages/waku/src/router/client.ts
```diff
@@ -196,16 +195,16 @@ export function useRouter_UNSTABLE() {
...
   /**
    * indicates if the link should scroll or not on navigation
    * - `true`: always scroll
    * - `false`: never scroll
    * - `undefined`: scroll on path change (not on searchParams change)
    */
   scroll?: boolean;
-  children: ReactNode;
+  unstable_pending?: ReactNode;
+  unstable_notPending?: ReactNode;
```
dai-shi, line 207: This change is technically breaking change. I wasn't confident with this, so marking as unstable.

PR #1261, opened 2025-02-24
fix(router): experimental custom error handler

dai-shi:
follow up #1247

experimental router api for #1239

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -37,6 +37,10 @@ import {
 } from './common.js';
 import type { RouteProps } from './common.js';
 import type { RouteConfig } from './base-types.js';
+import {
+  hasStatusCode,
+  hasLocationHeader,
+} from '../lib/utils/custom-errors.js';
 
 type AllowPathDecorators<Path extends string> = Path extends unknown
   ? Path | `${Path}?${string}` | `${Path}#${string}`
@@ -342,6 +346,58 @@ export class ErrorBoundary extends Component<
   }
 }
 
+const NotFound = () => {
+  // FIXME totally unsure if this is a desired behavior
+  useEffect(() => {
+    window.location.replace('/404');
+  });
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('h1', null, 'Not Found')),
+  );
+};
+
+const Redirect = ({ loc }: { loc: string }) => {
+  // FIXME unsure if this is a desired behavior
+  useEffect(() => {
+    window.location.replace(loc);
+  });
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('a', { href: loc }, 'Redirect')),
+  );
+};
+
+class CustomErrorHandler extends Component<
+  { children: ReactNode },
+  { error?: unknown }
+> {
+  constructor(props: { children: ReactNode }) {
+    super(props);
+    this.state = {};
+  }
+  static getDerivedStateFromError(error: unknown) {
+    return { error };
+  }
+  render() {
+    if ('error' in this.state) {
+      const error = this.state.error;
+      if (hasStatusCode(error)) {
+        if (error.statusCode === 404) {
+          return createElement(NotFound);
+        }
+        if (hasLocationHeader(error)) {
+          return createElement(Redirect, { loc: error.locationHeader });
+        }
+      }
+      throw error;
+    }
+    return this.props.children;
+  }
+}
+
 const getRouteSlotId = (path: string) => 'route:' + path;
 
 const handleScroll = () => {
@@ -543,7 +599,7 @@ export function Router({
       return data;
     };
   const initialRscParams = createRscParams(initialRoute.query);
-  return createElement(
+  const root = createElement(
     Root as FunctionComponent<Omit<ComponentProps<typeof Root>, 'children'>>,
     {
       initialRscPath,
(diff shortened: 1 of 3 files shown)


packages/waku/src/router/client.ts
```diff
@@ -342,6 +346,58 @@ export class ErrorBoundary extends Component<
...
+    window.location.replace(loc);
+  });
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('a', { href: loc }, 'Redirect')),
+  );
+};
+
+class CustomErrorHandler extends Component<
```
rmarscher, line 373: Can we add a way to pass in components to use instead of the defaults? I want to be able to define my own NotFound component, for example. Maybe we can pass it through as a `<Router />` prop or another method?

https://github.com/vercel/next.js/blob/v15.1.0/packages/next/src/client/components/http-access-fallback/error-boundary.tsx#L169-L177

rmarscher (commented):
Thank you for working on this.

Do we need to also make updates here on the server-side? https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/middleware/handler.ts#L162-L192

I think it would ideally render a new RSC stream. We have Hono fallback 404 html that can be configured somewhere. Without a new RSC render, I think new NotFound component will only happen when navigating on the client.

To test, I added `unstable_notFound()` to the pages/about.tsx in the 01_template. I updated `pages/index.tsx` and `pages/about.tsx` be dynamic while leaving `_layout.tsx` as static.

I may have done something wrong, but I'm not seeing the Not Found custom error getting caught by the custom error boundary even with client-side routing.

packages/waku/src/router/client.ts
```diff
@@ -342,6 +346,58 @@ export class ErrorBoundary extends Component<
...
+  // FIXME totally unsure if this is a desired behavior
+  useEffect(() => {
+    window.location.replace('/404');
+  });
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('h1', null, 'Not Found')),
+  );
+};
```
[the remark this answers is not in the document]
dai-shi: Without the stream, we show `src/pages/404.txt` if available, so I would like to keep it consistent. However, there's also `public/public.html`. So, I guess nothing works ideally as of now. We probably need to define the expected behavior from scratch.

It feels impossible to behaves the same between with stream and without it.

packages/waku/src/router/client.ts
```diff
@@ -342,6 +346,58 @@ export class ErrorBoundary extends Component<
...
+const Redirect = ({ loc }: { loc: string }) => {
+  // FIXME unsure if this is a desired behavior
+  useEffect(() => {
+    window.location.replace(loc);
+  });
+  return createElement(
+    'html',
+    null,
+    createElement('body', null, createElement('a', { href: loc }, 'Redirect')),
+  );
```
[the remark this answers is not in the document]
dai-shi: Hm, I think we need to put `CustomErrorHandler` in `Router` then.

packages/waku/src/router/client.ts
dai-shi: I think you would like to use an error that isn't captured by CustomErrorHandler. Let's revisit it as our design doesn't seem very solid.

dai-shi:
> Do we need to also make updates here on the server-side? https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/middleware/handler.ts#L162-L192

I don't believe so, but we might be having a different assumption. And, I don't believe my assumption is correct.

> I think it would ideally render a new RSC stream. We have Hono fallback 404 html that can be configured somewhere. Without a new RSC render, I think new NotFound component will only happen when navigating on the client.

I think it should show `public/404.html`, but not sure.

> To test, I added `unstable_notFound()` to the pages/about.tsx in the 01_template. I updated `pages/index.tsx` and `pages/about.tsx` be dynamic while leaving `_layout.tsx` as static.
> 
> I may have done something wrong, but I'm not seeing the Not Found custom error getting caught by the custom error boundary even with client-side routing.

The custom error boundary would work only when you throw an error in server-side suspense. Did you add Suspense on the server?

Oh, this seems pretty complicated. Thanks for giving me feedback. I would like to make Waku as thin as possible, but also with a good default. Hope we find a balanced solution.

dai-shi:
> Do we need to also make updates here on the server-side? https://github.com/dai-shi/waku/blob/main/packages/waku/src/lib/middleware/handler.ts#L162-L192

Actually, I think I misunderstood something. Will fix it.



> Maybe `error.digest` is the only way to pass custom error information?

Oh... That's my oversight. I'll look into it.

dai-shi:
```js
> const e = new Error('err')
undefined
> e.foo=1
1
> const e2 = structuredClone(e)
undefined
> e2.foo
undefined
```
I see.

dai-shi:
serializeErrorValue
https://github.com/facebook/react/blob/25677265038b89c1ee3000e0669339ed160d9d75/packages/react-server/src/ReactFlightServer.js#L3096-L3126

It doesn't have any `.digest` handling. It looks like some React feature.

packages/waku/src/router/client.ts
rmarscher: This comment was about being able to create different element output than `<html><body><h1>Not Found</h1></body></html>` component when there is a stream. If it can render a [404 page component if available](https://github.com/dai-shi/waku/pull/1261/files#r1967194656), that works.

PR #1268, opened 2025-02-25
fix(create-waku): Fix detection of bun using npm_config_user_agent env var

hamlim:
I noticed that running `bun create waku@latest` and `bunx create-waku@latest` didn't seem to detect `bun` as the package manager as expected based on my changes in #1229.

However - I was able to verify that both formats should work when I tested with a new dummy package: `create-debug-create`

```sh
bun create debug-create@latest
# and
bunx create-debug-create@latest
```

will print some things to the console (e.g. `process.env`, `process`, etc).

Based on running both of those - it seems like the `npm_config_user_agent` env var _does_ get populated with `bun` as expected (not sure how I was testing before where it didn't work - maybe just with running the CLI directly via `bun ./cli.mjs`?)

### DIFF
--- a/packages/create-waku/src/index.ts
+++ b/packages/create-waku/src/index.ts
@@ -19,12 +19,7 @@ import {
 
 const userAgent = process.env.npm_config_user_agent || '';
 
-// Bun doesn't update `npm_config_user_agent`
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
LGTM

PR #1269, opened 2025-02-26
feat(create-pages): Support HEAD and OPTIONS api handlers

hamlim:
This change adds support for `createApi` handlers to handle both `HEAD` and `OPTIONS` request methods.

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -21,7 +21,17 @@ import { Children, Slot } from '../minimal/client.js';
 import { ErrorBoundary } from '../router/client.js';
 
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods partial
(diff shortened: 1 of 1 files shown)


tylersayshi:
The more I think about this... I wonder if we should just allow all methods. Our API route handlers do give the Request as is, so realistically the implementation of handling TRACE and CONNECT would be deferred to the author of the API route.

I suspect there to not be a use case for TRACE or CONNECT, but since it is on the users to handle requests anyways, I don't see a reason to disallow the methods explicitly.

(apologies for my wavering opinion on this. Just aiming to be as thoughtful as possible for it.)

dai-shi:
Let me merge this. @tylersayshi Feel free to send a follow-up PR.

hamlim:
> (apologies for my wavering opinion on this. Just aiming to be as thoughtful as possible for it.)

No worries at all, I totally understand! I was going back and forth a bit on it as well - happy to open a followup to add those in unless someone beats me to it!

PR #1270, opened 2025-02-26
fix: customizable notFound

dai-shi:
following up:
- #1267 

related: 
- #1247 
- #1261 
- #1239 

Tasks:
- [x] add failing test
- [x] fix it

----

~This might not support @rmarscher 's use case. I'll address separately if it doesn't.~


rmarscher:
This looks good to me. This is testing that src/pages/404.tsx will be rendered with client-side routing with and without Suspense?

dai-shi:
> This is testing that src/pages/404.tsx will be rendered with client-side routing with and without Suspense?

Exactly.

dai-shi:
I would like to avoid 404 hack in router/client. However, it wasn't very easy. Let's tackle it separately.

packages/waku/src/router/define-router.ts
```diff
@@ -312,30 +312,39 @@ export function unstable_defineRouter(fns: {
...
+        if (info?.status !== 404) {
+          throw e;
         }
       }
-      const rscPath = encodeRoutePath(pathname);
-      const rscParams = new URLSearchParams({ query });
-      const entries = await getEntries(rscPath, rscParams, input.req.headers);
-      if (!entries) {
+      if (await has404()) {
+        return { ...(await renderIt('/404', '')), status: 404 };
```
rmarscher, line 344: :+1:

PR #1271, opened 2025-02-26
feat(create-pages): Allow for all request methods within createApi

hamlim:
Follow up to #1269, allowing for support for all request methods to be handled by handlers within `createApi`


tylersayshi:
LGTM, thanks!

PR #1274, opened 2025-02-27
@tanstack/react-query example with RSC and SSR

rmarscher:


rmarscher:
I think this works???

rmarscher:
I'm not exactly sure if this example is a good idea. It seems to get stuck during build time SSR for some reason. I tried to make a version with static pages and it hangs during build.

This is basically using server functions as RPC calls and relying on @tanstack/react-query to solve the concerns that the React docs have mentioned in the past with using server functions to load data.

It looks like `react-query-next-experimental` only uses `import { useServerInsertedHTML } from 'next/navigation'` - so maybe it's not that hard to [port it to Waku](https://github.com/TanStack/query/blob/main/packages/react-query-next-experimental/src/HydrationStreamProvider.tsx).

dai-shi:
RSC's server functions are technically tied to React (or Waku's `Root` component), so I'm not sure if it's a good idea to use them as general RPC functions. It might be interesting to explore the possibility, but it doesn't seem a mainstream conceptually.

rmarscher:
I think it is conceptually similar to this Tanstack Start loader example: https://github.com/TanStack/router/blob/main/examples/react/start-basic-react-query/app/routes/posts.%24postId.tsx#L8-L16

I will close this. I agree that it is not React's intent for server functions to be used for just data loading and RPC... but... we have this statement in the Waku readme:

>  Once you get the hang of it, you’ll appreciate how awesome it is to flexibly move server-client boundaries with a single line of code as your full-stack React codebase evolves over time. It’s way simpler than maintaining separate codebases for your backend and frontend.

I have found that it is a really nice developer experience to not have to add JSON or or RPC API endpoints. We can do it all with server components and server functions like this if we are careful about it.

dai-shi:
Yes, it's possible as some of examples have done it. Data fetching with server functions isn't very recommended, but I think it's allowed.

I wasn't sure if there's any implication on the RQ side, especially for caching.

rmarscher:
What if the server functions in this example returned React elements instead of data objects? And the client component just outputs the returned server components with the react-query wrapper to handle caching and refetching? If we remove the server-side fetch and streaming from this example, I think it would work with 51_spa.

dai-shi:
Yeah, I think so too. I didn't mean to be too opposed to the original idea.

By the way, I was hoping to add SSR to 52_tanstack-router. Feel free to take the challenge.

PR #1276, opened 2025-03-01
refactor: config prd instead of pure config

dai-shi:


dai-shi (commented):
There are some undesired type assertions. I hope to fix them in the future refactors.

PR #1280, opened 2025-03-02
experimental: error callback for middleware

dai-shi:
(I'm still not sure how `onError` works. Isn't it documented?)

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -104,7 +104,14 @@ async function runDev() {
     if (values['experimental-compress']) {
       app.use(compress());
     }
-    app.use(serverEngine({ cmd: 'dev', config, env: process.env as any }));
+    app.use(
+      serverEngine({
+        cmd: 'dev',
+        config,
+        env: process.env as any,
+        unstable_onError: new Set(),
+      }),
+    );
     app.notFound((c) => {
       // FIXME can we avoid hardcoding the public path?
       const file = path.join('public', '404.html');
@@ -157,7 +164,12 @@ async function runStart() {
     }
     app.use(serveStatic({ root: path.join(distDir, DIST_PUBLIC) }));
     app.use(
-      serverEngine({ cmd: 'start', loadEntries, env: process.env as any }),
+      serverEngine({
+        cmd: 'start',
+        loadEntries,
+        env: process.env as any,
+        unstable_onError: new Set(),
+      }),
     );
     app.notFound((c) => {
       // FIXME better implementation using node stream?
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -566,6 +566,7 @@ const emitStaticFiles = async (
         config,
         { unstable_modules },
         elements,
+        new Set(),
         options?.moduleIdCallback,
       ),
     renderHtml: async (
@@ -578,6 +579,7 @@ const emitStaticFiles = async (
         { unstable_modules },
         defaultHtmlHead + (options.htmlHead || ''),
         elements,
+        new Set(),
         html,
         options.rscPath,
       );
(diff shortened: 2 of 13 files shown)


rmarscher:
There is documentation for `onError` in the parameters section of [renderToReadableStream](https://react.dev/reference/react-dom/server/renderToReadableStream#rendertoreadablestream)

> onError: A callback that fires whenever there is a server error, whether [recoverable](https://react.dev/reference/react-dom/server/renderToReadableStream#recovering-from-errors-outside-the-shell) or [not.](https://react.dev/reference/react-dom/server/renderToReadableStream#recovering-from-errors-inside-the-shell) By default, this only calls console.error. If you override it to [log crash reports,](https://react.dev/reference/react-dom/server/renderToReadableStream#logging-crashes-on-the-server) make sure that you still call console.error. You can also use it to [adjust the status code](https://react.dev/reference/react-dom/server/renderToReadableStream#setting-the-status-code) before the shell is emitted.

We should add `console.error(err);` to `onError`. I see that is missing from Waku's implementation.

Each of these usage sections discuss how onError can be used:

- [Logging crashes on the server](https://react.dev/reference/react-dom/server/renderToReadableStream#logging-crashes-on-the-server)
- [Recovering from errors inside the shell ](https://react.dev/reference/react-dom/server/renderToReadableStream#recovering-from-errors-inside-the-shell)
    > If there is an error while generating the shell, both onError and your catch block will fire. Use onError for error reporting and use the catch block to send the fallback HTML document
- [Recovering from errors outside the shell ](https://react.dev/reference/react-dom/server/renderToReadableStream#recovering-from-errors-outside-the-shell)
- [Setting the status code](https://react.dev/reference/react-dom/server/renderToReadableStream#setting-the-status-code)
- [Handling different errors in different ways](https://react.dev/reference/react-dom/server/renderToReadableStream#handling-different-errors-in-different-ways)

Logging crashes on the server is the primary use case that I have for onError. I have server-side console.error hooked up to my error tracking - so really I won't have to add anything else if we add that. But... for people that want to do their own instrumentation and maybe ignore certain errors, it would be really great to register a callback to do that.

In those React usage docs, the methods for setting status code or making the onError error accessible to the request handler are interesting... but as soon as there is a suspense boundary, they won't work as expected because the onError will happen after the response starts streaming.

If I want to add error reporting for everything collected by `ctx.unstable_errs` - how can I do that? Create Waku middleware and pipe the ctx.body through my own transform that reports new ctx.unstable_errs as they become available and then one final check when the stream is complete?

rmarscher:
> > Isn't it documented?

> There is documentation for `onError` in the parameters section of [renderToReadableStream](https://react.dev/reference/react-dom/server/renderToReadableStream#rendertoreadablestream)

Sorry... I realize I was confusing react-dom/server renderToReadableStream with [react-server-dom-webpack renderToReadableStream](https://github.com/facebook/react/blob/main/packages/react-server-dom-webpack/src/server/ReactFlightDOMServerEdge.js#L62-L76). I wonder if they are intended to work in a similar way?

dai-shi:
Yeah, I've read the react-dom docs once before.

> We should add `console.error(err);` to `onError`. I see that is missing from Waku's implementation.

Nice catch.

> If I want to add error reporting for everything collected by `ctx.unstable_errs` - how can I do that? Create Waku middleware and pipe the ctx.body through my own transform that reports new ctx.unstable_errs as they become available and then one final check when the stream is complete?

Yes, my expectation is write your own middleware. I haven't tried anything and not sure if it works as you expect. Feedback is welcome.

rmarscher (approved):
LGTM. I think it is ready to merge unless you aren't sure that you like how the callbacks are being passed through.

dai-shi:
Thanks for the review. I'm not 100% confident, but as it's marked as `unstable_`, I'm fine to merge.

PR #1281, opened 2025-03-02
breaking: string only waku.config.ts

dai-shi:
This is for v0.22.0.

Previously, waku.config.ts allows to put functions directly. Instead of that, this one is to take strings only.

Here's the patch for 38_cookies examnple:
```diff
diff --git a/examples/38_cookies/waku.config.ts b/examples/38_cookies/waku.config.ts
index d64ef237..60a72ce8 100644
--- a/examples/38_cookies/waku.config.ts
+++ b/examples/38_cookies/waku.config.ts
@@ -1,10 +1,10 @@
 import { defineConfig } from 'waku/config';
 
 export default defineConfig({
-  middleware: () => [
-    import('waku/middleware/context'),
-    import('./src/middleware/cookie.js'),
-    import('waku/middleware/dev-server'),
-    import('waku/middleware/handler'),
+  middleware: [
+    'waku/middleware/context',
+    './src/middleware/cookie.js',
+    'waku/middleware/dev-server',
+    'waku/middleware/handler',
   ],
 });
```

Tasks:
- [x] DEV
- [x] PRD
- [x] `--with-cloudflare`
- [x] other deploy plugins
- [x] docs


dai-shi:
@rmarscher I get a weird error with `pnpm -F 07_cloudflare build --with-cloudflare`. Can you have a look when you get a chance?

fengzilong:
It seems we will lose type checking for middleware in waku config file. Would you consider supporting both `Promise<{ default: Middleware }>[]` and `string[]`?

rmarscher:
> @rmarscher I get a weird error with `pnpm -F 07_cloudflare build --with-cloudflare`. Can you have a look when you get a chance?

It looks like `import.meta.env && !import.meta.env.PROD` is true during build. The intent is to only set `unstable_honoEnhancer` during dev mode.

rmarscher:
I'm curious what is the reason for this change. Does it enable hot reload for middleware code? Or some other use case for the middleware to be imported within the vite context?

packages/waku/src/lib/plugins/vite-plugin-rsc-entries.ts
```diff
@@ -17,6 +17,8 @@ const CONFIG_FILE = 'waku.config.ts'; // XXX only ts extension
 export function rscEntriesPlugin(opts: {
   basePath: string;
   rscBase: string;
+  middleware: string[];
```
tylersayshi, line 20: I can follow up this PR with type suggestions for the waku supported plugins if you're open to it. It would still allow any string, but suggest the built in ones.

dai-shi: Can you highlight your idea with a few lines of code?

dai-shi:
> The intent is to only set `unstable_honoEnhancer` during dev mode.

It's much easier in that case. Let me see.

dai-shi:
It still doesn't work. It fails with building.

packages/waku/src/lib/plugins/vite-plugin-rsc-entries.ts
tylersayshi: playground: https://tsplay.dev/we3Xem

image

dai-shi: Thanks. I see this doesn't break anything. Not sure how much it helps because it's very unlikely to people to type it. (Besides, this isn't a fix for the regression.)

I don't think we need it at least in this PR.

dai-shi:
> Hmm, but it's still "build" from the waku perspective.

It's my misunderstanding. Now, it builds fine with `import.meta.PRD` condition.

@rmarscher Can you look into how e2e is failing?

rmarscher:
```sh
cd examples/07_cloudflare
pnpm run build
pnpm run start
```

open http://localhost:8787/

I am getting an error:

```
TypeError: Cannot read properties of undefined (reading 'find')
```

It is coming from here:

```ts
const htmlHead = !devServer && entriesPrd.dynamicHtmlPaths.find(([pathSpec]) => getPathMapping(pathSpec, ctx.req.url.pathname))?.[1] || "";
```

`entriesPrd.dynamicHtmlPaths` must be undefined. I'm not sure what that is about.

Are you seeing the same error?

dai-shi:
> Are you seeing the same error?

Oh, yeah, I didn't notice it when I run it previously. This should be fixable. Thanks.

brrock:
If this hasn't already been added if the user upgrades to this version instead of output an error message log something nicer like a way to migrate, so loads of issues don't flood in

dai-shi:
I believe Waku users are kind enough not to flood.

Issue #698, opened 2024-05-05
[BUG]: basePath is not working except default value


PerfectPan:
Reproduce Repo: https://github.com/PerfectPan/waku-issue-691

Setting basePath manualy in waku.config.ts will cause error except `/`.
```ts
/** @type {import('waku/config').Config} */
export default {
  basePath: '/test/'
};
```

build mode: load js and css file error(`pnpm build && pnpm start`)
![CleanShot 2024-05-05 at 14 01 46@2x](https://github.com/dai-shi/waku/assets/24316656/b68ec33e-cc78-4b85-9153-3a010cc1fc4f)

dev mode: unknown error(`pnpm dev`)
![CleanShot 2024-05-05 at 13 57 04@2x](https://github.com/dai-shi/waku/assets/24316656/7ee2cc14-a238-4031-9988-6a8d4e9bc128)

Some previous reserach can be seen in https://github.com/dai-shi/waku/pull/697#issuecomment-2094563046

Would like to solve it later and also welcome other one to work on this.

dai-shi:
Thanks for opening an issue.
Let's fix it step by step:
1. SSG
2. build
3. dev

By SSG, I mean to build static files and serve from a different folder:
```bash
pnpm builld
mkdir ssg
cp -r dist/public ssg/test
npx serve -s ssg
```

yuusheng:
Hi! I'm trying to solve this. But I encounter some problems:

1. It seems that we should not add basePath on `entriesFile` in devServerImpl.ts, as it will be a true file path when vite handle this path(It should be `.../waku_project/src/entries` instead of `.../waku_project/test/src/entries`)
2. But when I use `const entriesFile = ${vite.config.root}/${config.srcDir}/${SRC_ENTRIES}\;`, `globalThis.__WAKU_PREFETCHED__` whill no longer inject by the ssr middleware. So it will fetch `http://localhost:3000/RSC/test.txt`. I can't figure out how to solve this. Do you have some recommends @dai-shi ?

dai-shi:
> 1. It seems that we should not add basePath to `entriesFile` in devServerImpl.ts

Do we do that? Maybe, it's for SSG, but not sure about DEV.

I haven't looked into this yet, so my question might be appropriate, but can you describe the issue you are tackling? Aren't there several issues for `basePath` to work as expected?

yuusheng:
> Do we do that? Maybe, it's for SSG, but not sure about DEV.
[...]

PR #1282, opened 2025-03-02
fix: basePath

yuusheng:
fix: https://github.com/dai-shi/waku/issues/698


yuusheng:
Still working in progress. Currently ssr is working in dev mode. But src/main is 404.

image

yuusheng:
Some other things need fixing:
1. Redirect to basePath if user opens http://localhost:3000
2. Allow using both `http://localhost:3000/test/` and `http://localhost:3000/test`. Only `http://localhost:3000/test/` works for now

dai-shi:
Hi, thanks for working on it.
Can you please create e2e/fixtures/base-url and the spec, instead of  
examples/55_base-url ? We need it to maintain the capability.

dai-shi:
> 1. Redirect to basePath if user opens http://localhost:3000

I think it's out of the scope.

yuusheng:
> Can you please create e2e/fixtures/base-url and the spec, instead of
examples/55_base-url ?

OK, I will remove that when the PR is ready to merge. It's quite hard to debug without that 😭.

> I think it's out of the scope.

As this is the default behavior of Vite, I think this is more friendly to developer. Or we can talk about that in the future instead of in this PR

dai-shi:
> OK, I will remove that when the PR is ready to merge. It's quite hard to debug without that 😭.

Alright, then create both `e2e/fixtures/base-url` and the spec, as well as `examples/55_base-url` and remove `examples/55_base-url` before merging.
The e2e spec is required to review the PR. You can take time. I'm looking forward to it.

> As this is the default behavior of Vite,

Hmm, I'm not sure why it's the default behavior. How does it behave after the build?

yuusheng:
> Hmm, I'm not sure why it's the default behavior. How does it behave after the build?

Vite will redirect only in dev mode as it'll be all static assets after build.

dai-shi:
> Vite will redirect only in dev mode

That's what I expected. So, I thought it would be better to keep the behavior the same in both DEV and PRD modes in Waku, but I don't mind following Vite's convention if that helps.

yuusheng:
Our project structure will change a lot after https://github.com/wakujs/waku/pull/1493 gets merged. This will mostly fix the basePath problem, but we'll still have some issues. For example, `<Link to="/foo" />` should route to /{basePath}/foo. I'll work on this after the pull request is merged.

Issue #1033, opened 2024-12-04
revalidatePath / single Server request with updated rsc data


aheissenberger:
When using Next.js it is possible to call a server action and update the interface with **one network request**.
The request to the server will contain the payload and the response will return the result of the server function **and the update serialized data** of the components.

The `router.refresh()`, which is similar to Waku `router.reload()` is disabled and still the interface ist updated based on the use of [`revalidatePath('/')`](https://nextjs.org/docs/app/api-reference/functions/revalidatePath) in the server action `addTodo()`.

How can I achieve this with the Waku framework? Currently I use `router.reload()` which leads to an extra request to the server to update the interface.

Here is a link to the example code: https://v0.dev/chat/IRiWT7TzftT?b=b_QaETSgGhtSy

```tsx
'use client'

import { useState } from 'react'
import { useRouter } from 'next/navigation'
import { addTodo } from '../../lib/actions'

export function AddTodoForm() {
  const [text, setText] = useState('')
  const router = useRouter()

  const handleSubmit = async (e: React.FormEvent) => {
    e.preventDefault()
    if (text.trim()) {
      await addTodo(text)
      setText('')
      //router.refresh()
    }
  }

  return (
    <form onSubmit={handleSubmit} className="mb-4">
      <input
        type="text"
        value={text}
        onChange={(e) => setText(e.target.value)}
        placeholder="Add a new todo"
        className="border p-2 mr-2"
      />
      <button type="submit" className="bg-blue-500 text-white p-2 rounded">
        Add Todo
      </button>
    </form>
  )
}
```

```ts
'use server'

import { revalidatePath } from 'next/cache'
import { addTodoToStore, toggleTodoInStore, deleteTodoFromStore } from './todos'

export async function addTodo(text: string) {
  await addTodoToStore(text)
  revalidatePath('/')
}
```

Next.js 15 todo app:
https://github.com/aheissenberger/demo-nextjs-todo

Request Payload Next.js 15 todo app:
![SCR-20241204-svfr](https://github.com/user-attachments/assets/7aa7c6d2-8a3d-4386-8205-6e357b9fecfc)
Request Response:
![SCR-20241204-svja](https://github.com/user-attachments/assets/1ad17c24-5528-4378-b19a-8f33b06b41c1)

dai-shi:
> it is possible to call a server action and update the interface with **one network request**.

[...]

PR #1284, opened 2025-03-03
proposal: one possible way to implement revalidate API in server action

himself65:
Here is just my idea with issue https://github.com/dai-shi/waku/issues/1033 and some other updates

Basically, this will solve:

- `rerender` can only trigger once before, but now it could triggerd multiple times (I think?), in the future we could support like `revalidateTag` similarlly using this trick.
- allow throw error and call `rerender` simultaneously

---

Obviously the code is messup, I think it can only works locally dev only

I used streamableValue in `ai/rsc`, I cannot find a better alnertive API in exsiting npm, but it already very simple concept (a Suspense + Iterator wrapper)

### DIFF
--- a/packages/waku/src/lib/utils/create-streamable-value.ts
+++ b/packages/waku/src/lib/utils/create-streamable-value.ts
@@ -1,0 +1,348 @@
+export const HANGING_STREAM_WARNING_TIME_MS = 15 * 1000;
+export const STREAMABLE_VALUE_TYPE = Symbol.for('ui.streamable.value');
+
+export type StreamablePatch = undefined | [0, string]; // Append string.
+
+declare const __internal_curr: unique symbol;
+declare const __internal_error: unique symbol;
+
+/**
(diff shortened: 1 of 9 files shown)


packages/waku/src/lib/utils/create-streamable-value.ts
himself65, line 1: https://github.com/vercel/ai/blob/main/packages/ai/rsc/streamable-value/read-streamable-value.tsx

dai-shi:
While I understand the goal, the current implementation doesn't feel very comfortable. I feel like there should be a better / more minimal solution.
Thanks for the suggestion anyway.

dai-shi:
I would like to understand the use case and problem better. Let's discuss it separately. Open a new discussion?

This PR doesn't seem to match with our design:
- streamable data should be user code
- It seems like you have a misconception about "rerender" / "revalidate" and the blame is on me...

Closing this.

PR #1285, opened 2025-03-03
chore: rename `_value` to a unique symbol

himself65:
`_value` is always confusing me, I hope it could be a meaningful name.

### DIFF
--- a/packages/waku/src/lib/constants.ts
+++ b/packages/waku/src/lib/constants.ts
@@ -1,3 +1,4 @@
 export const EXTENSIONS = ['.js', '.ts', '.tsx', '.jsx', '.mjs', '.cjs'];
 export const SRC_MAIN = 'main';
 export const SRC_ENTRIES = 'entries';
+export const FUNCTION_RESULT = Symbol.for('waku.function.result');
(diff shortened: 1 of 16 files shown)


dai-shi:
To be honest, I'm not a big fan of this. Especially, the use of Symbol.for doesn't feel comfortable.

packages/waku/src/lib/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const EXTENSIONS = ['.js', '.ts', '.tsx', '.jsx', '.mjs', '.cjs'];
 export const SRC_MAIN = 'main';
 export const SRC_ENTRIES = 'entries';
+export const FUNCTION_RESULT = Symbol.for('waku.function.result');
```
rmarscher, line 4: It seems that

```
export const FUNCTION_RESULT = '_value';
```

would accomplish a similar goal while not introducing `Symbol.for`?

dai-shi: I'm fine with that, if that helps a lot.


btw, consider `_` prefix means "reserved" in this case.

We used to have this check:
https://github.com/dai-shi/waku/blob/8a54eab1cb3bb04296f57c77aae999fda328184f/packages/waku/src/lib/renderers/rsc-renderer.ts#L139-L141

dai-shi:
On second thought, this `_value` only appears in Minimal API. So, it's not something we should expose from `waku/config`.
We may reconsider exporting something from Minimal API, or prefering `_result` over `_value` in the future.

PR #1287, opened 2025-03-04
experimental: update e2e fixture for errors in ctx

rmarscher:
```sh
cd e2e/fixtures/ssr-catch-error
pnpm run dev
```

Open http://localhost:3000/

In the dev server logs, you should see the syncronously thrown error available to middleware in ctx.unstable_errs.

Click on "Delayed page" link. 

You should see in the log that ctx.unstable_errs is `undefined` in the middleware for that request. This is because the error is thrown after delay and there is a suspense boundary.

The unstable_onError does capture the error though. I do think it is correct to additionally add console.error into the rsc server onError so it is always printed to the console.

### DIFF
--- a/packages/waku/src/lib/middleware/types.ts
+++ b/packages/waku/src/lib/middleware/types.ts
@@ -16,6 +16,7 @@ export type HandlerContext = {
   readonly context: Record<string, unknown>;
   readonly data: Record<string, unknown>;
   unstable_errs?: unknown[];
+  unstable_onError?: (err: unknown) => void;
   unstable_devServer?: {
     rootDir: string;
     resolveClientEntry: (id: string) => string;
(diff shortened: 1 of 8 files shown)


packages/waku/src/lib/middleware/types.ts
```diff
@@ -16,6 +16,7 @@ export type HandlerContext = {
   readonly context: Record<string, unknown>;
   readonly data: Record<string, unknown>;
   unstable_errs?: unknown[];
+  unstable_onError?: (err: unknown) => void;
```
dai-shi, line 19: Hmmm, I don't prefer this. At least, we don't need both. But still, `ctx` doesn't feel the right place for "callbacks".

rmarscher: Yes. I wasn't really sure where to put it. I mainly just wanted to provide an example for you. Feel free to close and take it from here. Thank you!

dai-shi: I'll have a closer look.

dai-shi: I just updated #1280. Feel free to send another draft PR if you like.

PR #1290, opened 2025-03-05
fix(create-waku): fix cli and swap to clack + pico

### LINKED ISSUE #1289: create-waku confuses the target dir
The issue happens when a project is created named waku-project and then any other project is attempted to be made in the same directory. 

you can test this by:
1. make a project with the default name in a directory
2. make a new project with a new name in that same directory as a sibling/ this will register as a duplicate of waku-project but then go in the correct new name folder

Also, if a project exists named `foo` and another is attempted to be made also named `foo` the duplicate will not be detected.

It would be nice if we could break the prompts into steps instead of the single list api which is done by `prompts`. We've discussed changing before, but I think this issue with the side effect for targetDir which does not work is motivation for considering a different CLI library.

tylersayshi:
prompts => @clack/prompts
kolorist => picocolors

these libraries are copying the pattern set by create-vite

fixes #1289


tylersayshi:
draft to fix e2e test setup

tylersayshi:
@dai-shi would you be ok with testing the cli with execa like is done with create-vite

https://github.com/vitejs/vite/blob/main/packages/create-vite/__tests__/cli.spec.ts

I am having issues handling the stdin and stdout with clack and vite's test seems like a clean way to solve for this part.

tylersayshi:
> I'm not very familiar but sounds ok. Any drawbacks?

the drawback with unit testing that I thought initially was that it would just not test interactive mode. 

to test interactive mode though, we should mimic the @clack/prompts tests. They mock writable and readable to run the tests through vitest.

https://github.com/bombshell-dev/clack/blob/main/packages/core/test/prompts/prompt.test.ts

So I think two tests would be good

1. test create-waku with args
2. test create-waku in interactive mode by mocking writable and readable.

Do you think both are needed for this PR or would a follow up be on for one of them?

dai-shi:
> Do you think both are needed for this PR or would a follow up be on for one of them?

I'll leave it to you. Our requirement is not break the current behavior. (btw, I'm planning another patch release, so let me know if you go with separate PRs.)

tylersayshi:
I'll plan for just this one PR and both tests mentioned above added. I want to be as careful as I can to not cause regression with this.

 It will need to be tomorrow though. I'm out for trivia tonight. ✌️

tylersayshi:
@dai-shi it looks like node 18 is not supported by one of the new dependencies

https://github.com/dai-shi/waku/actions/runs/13734124124/job/38415692370?pr=1290

any chance you'd want to drop node 18 support sooner than originally planned? 😅

tylersayshi:
planning to wait for when we drop node 18 to mark this ready

PR #1295, opened 2025-03-07
feat(router): fs-router support apiDir configurability

### LINKED ISSUE #1293: File system router API root path customization
It would be great to allow the root `'api'` path to be configurable. One of my apps already has a legacy API proxied at `/api`.

https://github.com/dai-shi/waku/blob/v0.21.22/packages/waku/src/router/fs-router.ts#L85

Thanks.

tylersayshi:
resolves #1293

### DIFF
--- a/packages/waku/src/lib/plugins/vite-plugin-rsc-managed.ts
+++ b/packages/waku/src/lib/plugins/vite-plugin-rsc-managed.ts
@@ -20,7 +20,7 @@ export default fsRouter(
   (file) => import.meta.glob('/${srcDir}/pages/**/*.{${EXTENSIONS.map((ext) =>
     ext.replace(/^\./, ''),
   ).join(',')}}')[\`/${srcDir}/pages/\${file}\`]?.(),
-  {pagesDir:'${options.pagesDir}', apiDir:'${options.apiDir}'},
+  { pagesDir: '${options.pagesDir}', apiDir: '${options.apiDir}' },
 );
 `;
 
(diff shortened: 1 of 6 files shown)


packages/waku/src/lib/plugins/vite-plugin-rsc-managed.ts
```diff
@@ -42,6 +42,7 @@ export function rscManagedPlugin(opts: {
   basePath: string;
   srcDir: string;
   pagesDir: string;
+  apiDir: string;
```
dai-shi, line 45: Ah, if we'd like to add this, we need to add `apiDir` in config.

tylersayshi: added ✅ 
it seemed right to match pagesDir everywhere rather than choose a point to hardcode apiDir again

rmarscher:
Thank you! :+1:

packages/waku/src/lib/plugins/vite-plugin-rsc-managed.ts
```diff
@@ -20,7 +20,7 @@ export default fsRouter(
   (file) => import.meta.glob('/${srcDir}/pages/**/*.{${EXTENSIONS.map((ext) =>
     ext.replace(/^\./, ''),
   ).join(',')}}')[\`/${srcDir}/pages/\${file}\`]?.(),
-  '${pagesDir}',
+  {pagesDir:'${options.pagesDir}', apiDir:'${options.apiDir}'},
```
dai-shi, line 23:
```suggestion
  { pagesDir: '${options.pagesDir}', apiDir: '${options.apiDir}' },
```

Issue #1296, opened 2025-03-07
Query strings of the url of the request object inside an API handler disappear


arianrhodsandlot:
Take this API as an example
```ts
createApi({
  path: '/api',
  render: 'dynamic',
  handlers: {
    async GET(req) {
      console.log(req.url)
      return new Response('hi')
    }
  }
})
```

When visiting "/api?id=123" , we expect the `req.url` inside the handler to be something like "http://localhost:3000/api?id=123" while we get "http://localhost:3000/api" acctually.

Reproduction: https://stackblitz.com/edit/stackblitz-starters-vcjzzkxf

Not sure whether the root cause of this issue is the same as #1291 .

dai-shi:
Nice catch! Thanks.

I'll fix define-router. @tylersayshi Would you take care of the rest?

PR #1297, opened 2025-03-07
fix(define-router): pass through url for api

dai-shi:
for #1296 (not the entire fix).

### DIFF
--- a/packages/waku/src/router/define-router.ts
+++ b/packages/waku/src/router/define-router.ts
@@ -115,6 +115,7 @@ export function unstable_defineRouter(fns: {
   handleApi?: (
     path: string,
     options: {
+      url: URL;
       body: ReadableStream | null;
       headers: Readonly<Record<string, string>>;
       method: string;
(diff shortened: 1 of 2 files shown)


packages/waku/src/router/define-router.ts
```diff
@@ -367,6 +369,7 @@ export function unstable_defineRouter(fns: {
             type: 'file',
             pathname,
             body: handleApi(pathname, {
+              url: new URL(pathname, 'http://localhost:3000'),
```
tylersayshi, line 372: Should this always be localhost:3000?

Is the port configurable?

dai-shi: It shouldn't matter. They shouldn't rely on it. It can be without the port.
```suggestion
              url: new URL(pathname, 'http://localhost'),
```

dai-shi: Or, we could make `url` optional. Do you prefer that?

tylersayshi: hmm i'll look as I implement the remaining parts

PR #1298, opened 2025-03-07
fix(handler): tee body for decode post action

### LINKED ISSUE #1291: APIs created by `createApi` crash when receiving a `FormData`
This is an API endpoint declared in the `entries.ts` file:
```ts
export default createPages(async ({ createPage, createApi }) => [
  createPage(/* omitted... */),

  createApi({
    path: '/api',
    render: 'dynamic',
    handlers: {
      POST: async () => {
        return new Response('api')
      },
    },
  })
])
```

When we make a request with `FormData` assigned to the request body, the API will crash with a message "Response body object should not be disturbed or locked".

I created a demo to reproduce the issue: https://stackblitz.com/edit/stackblitz-starters-vsnmntgt

Click the "test api" button on the preview page then the error message will reveal.

dai-shi:
fix #1291

### DIFF
--- a/packages/waku/src/lib/renderers/rsc.ts
+++ b/packages/waku/src/lib/renderers/rsc.ts
@@ -207,9 +207,17 @@ export async function decodePostAction(
       typeof contentType === 'string' &&
       contentType.startsWith('multipart/form-data')
     ) {
-      const bodyBuf = await streamToArrayBuffer(ctx.req.body);
+      const [stream1, stream2] = ctx.req.body.tee();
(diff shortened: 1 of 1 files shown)


dai-shi:
While trying this, I found 36_form example doesn't work as expected. I'm not sure if it's a regression, or it has never worked before...

dai-shi:
> While trying this, I found 36_form example doesn't work as expected. I'm not sure if it's a regression, or it has never worked before...

Oops, no. This PR breaks it. I need to change something.

PR #1304, opened 2025-03-10
fix(router): decode path for storing elements and slug as prop

### LINKED ISSUE #1302: Single quotes in route params will cause the page crash
Let's say given a route "/book/[name]" in a waku app, when we visit a path containing a single quote `'`, for example, "/book/Webster's%20Dictionary", the page will crash with a message "Invalid element: route:/book/Webster's%20Dictionary".

Reproduction: https://stackblitz.com/edit/stackblitz-starters-yjfjlhke

---

Additionally, the route params recieved in the page entry component are not decoded with `decodeURIComponent`.
```ts
export default function Book({ name }) {
  // We got `Webster's%20Dictionary` rather than `Webster's Dictionary` here
  return <div>{name}</div>
}
```
Not sure is this behaviour by design or not? As I tested, Next.js' app router doesn't decode it too, but its legacy page router will decode it when calling `useRouter` to retrieve the route params.

tylersayshi:
Let me know if this approach is ok. The change makes it so that id's internally for routes are always decoded. This makes it so the prop is decoded when it is passed. It also fixes the runtime error. The runtime error happened when the id as being looked up in elements as decoded, but internally stored as encoded, so our invalid element id error was thrown.

fixes #1302

### DIFF
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -67,6 +67,13 @@ for (const mode of ['DEV', 'PRD'] as const) {
       ).toBeVisible();
     });
 
+    test("nested/cat's pajamas", async ({ page }) => {
+      await page.goto(`http://localhost:${port}/nested/cat's%20pajamas`);
+      await expect(
+        page.getByRole('heading', { name: "Dynamic: cat's pajamas" }),
+      ).toBeVisible();
+    });
+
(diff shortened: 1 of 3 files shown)


dai-shi:
I might overlook something, but it looks good to me. Would it be possible to add an e2e test case?

tylersayshi:
added!

PR #1311, opened 2025-03-17
docs: Waku router under the hood and add guides to website

tylersayshi:
This is an idea for a router post. I am happy to reword, restructure or whatever.

It was helpful for me to organize my thoughts to look a bit closer at what is loaded when on each route, so I think a post to talk through some of this has some value.

Excited to hear what y'all think :)

### DIFF
--- a/docs/guides/router-fetch-strategy.mdx
+++ b/docs/guides/router-fetch-strategy.mdx
@@ -23,7 +23,7 @@ For dynamic pages, the expectation is that they should be re-requested from the 
 Fear not! We can take advantage of a very common trick for speeding the perceived load time of dynamic pages.
 
 ```tsx
-<Link unstable_prefetchOnEnter to='/dynamic'>
+<Link unstable_prefetchOnEnter to="/dynamic">
   Dynamic
 </Link>
 ```
--- a/docs/guides/router-fetch-strategy.mdx
+++ b/docs/guides/router-fetch-strategy.mdx
@@ -1,9 +1,7 @@
 ---
-slug: waku-router-under-the-hood
+slug: router-fetch-strategy
 title: Waku Router Under the Hood
 description: A brief look into the mechanics of the Waku Router
-author: tyler
-date: 2025/03/19
 ---
 
 When routing around a waku app, you may notice some surprisingly quick navigation speeds. This was my experience while starting development at least. Let's walk through the basic mechanics of the router to get a better shared understanding of how it works and maybe even how to speed things up even further 🚀.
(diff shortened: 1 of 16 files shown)


docs/guides/router-fetch-strategy.mdx
```diff
@@ -0,0 +1,37 @@
+---
+slug: router-fetch-strategy
+title: Waku Router Under the Hood
+description: A brief look into the mechanics of the Waku Router
+---
+
+When routing around a waku app, you may notice some surprisingly quick navigation speeds. This was my experience while starting development at least. Let's walk through the basic mechanics of the router to get a better shared understanding of how it works and maybe even how to speed things up even further 🚀.
+
+## Power of Client-Side Cacheing
```
hamlim, line 9: Small nit: I think the proper spelling is `Caching` (without the e)

```suggestion
## Power of Client-Side Caching
```

docs/guides/router-fetch-strategy.mdx
```diff
@@ -0,0 +1,37 @@
...
+<Link unstable_prefetchOnEnter to="/dynamic">
+  Dynamic
+</Link>
+```
+
+Now, when the user places their cursor over our link, we will fetch the RSC of that next page. Then when the user presses the link, if the promise for the RSC has resolved, we will switch to the next page, otherwise we will wait for the initial prefetch to finish. It's still a nice head start though!
+
+## Why is This Cool?
+
+None of this is specific to RSC, nor is it unique to Waku as a framework. However, the ergonomics of `prefetch` paired with the power of the client cacheing static RSC are quite nice features that will not be the default experience everywhere.
```
hamlim, line 35:
```suggestion
None of this is specific to RSC, nor is it unique to Waku as a framework. However, the ergonomics of `prefetch` paired with the power of the client caching static RSC are quite nice features that will not be the default experience everywhere.
```

dai-shi (approved):
Merging this for now. I think there should be room for improvement for website, like ordering guides, and styling.

rmarscher:
Maybe we should mention that Suspense and promises passed from the server can be used in layouts or pages to be able to display content while the rest is still being streamed down from the server. We could link out to the React docs about Suspense for people to learn more about it - https://react.dev/reference/react/Suspense

By default, a `useTransition` hook is used when navigating with the <Link> component. That causes the stale content inside Suspense to continue to display until the new content has loaded. There's an experimental `unstable_startTransition` to supply a different function to call. This can be the startTransition from your own useTransition hook call which can give you access to the isPending state. Or it's possible to pass a no-op function to skip useTransition entirely.

dai-shi:
> By default, a `useTransition` hook is used when navigating with the component. 

That's no longer the case. #1257

PR #1312, opened 2025-03-18
chore: rename dai-shi/waku => wakujs/waku for urls

tylersayshi:
the redirect works, but this seems good and easy enough to do


dai-shi (approved):
Thanks so much!

PR #1313, opened 2025-03-19
feat(create-pages): page part component

tylersayshi:
page part components open up support for multiple components per each page. These components are siblings in place of where a page component would be typically.

this is useful for marking _parts_ of a page static and other parts dynamic.


packages/waku/src/router/create-pages.ts
```diff
@@ -237,6 +253,7 @@ export const createPages = <
     createLayout: CreateLayout;
     createRoot: CreateRoot;
     createApi: CreateApi;
+    createPagePartComponent: CreatePagePartComponent;
```
dai-shi, line 256: Can we simply call `createPagePart`? Everything is component already (except for `createApi`).

packages/waku/src/router/create-pages.ts
```diff
@@ -135,12 +135,14 @@ export type CreatePage = <
     | {
         render: Extract<Render, 'dynamic'>;
         path: PathWithoutSlug<Path>;
-        component: FunctionComponent<PropsForPages<Path>>;
+        /** single page component. `null` would indicate the page will use `createPagePartComponent` instead. */
```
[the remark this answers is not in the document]
tylersayshi: It does feel strange as an API, I agree.

What should `createPage` get though when `createPagePart` is intended to be used for that page? 

```tsx
createPage({
  render: 'static',
  path: '/blog/:slug',
  component: // what here?
});

createPagePart({
  render: 'static',
  path: '/blog/:slug',
  order: 0,
  component: BlogPost,
});
```

maybe it is just that the separate functions are too awkward? Could we do this?

```tsx
createPage({
  render: 'dynamic',
  path: '/blog/:slug',
  component: [
    {
      render: 'static',
      component: BlogPost,
    },
    {
      render: 'dynamic',
      component: BlogPostPart,
      order: 1,
    },
  ]
});
```

it complicates the `createPage` implementation a bit, but I think the API is much smoother this way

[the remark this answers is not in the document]
dai-shi: > What should `createPage` get though when `createPagePart` is intended to be used for that page?

I was thinking about prohibiting using both `createPage` and `createPagePart` for the same path.

> Could we do this?

That's close to my original idea but then we agree that our design is to minimize fs-router assumption (meaning fs-router should be as simple wrapper as possible.)

packages/waku/src/router/create-pages.ts
```diff
@@ -504,10 +509,61 @@ export const createPages = <
     }
   };
 
+  const createPagePart: CreatePagePart = (params) => {
```
dai-shi, line 512:
```suggestion
  const createPagePart: CreatePagePart = (params) => {
    if (!import.meta.env.VITE_EXPERIMENTAL_WAKU_ROUTER) {
      console.warn('createPagePart is still experimental');
      return;
    }
```

tylersayshi:
ah page parts pages need to be added to the typegen /draft

packages/waku/src/router/create-pages.ts
```diff
@@ -237,6 +252,11 @@ export const createPages = <
     createLayout: CreateLayout;
     createRoot: CreateRoot;
     createApi: CreateApi;
+    /**
+     * Page Part pages are assumed to be dynamic in order to allow dynamic components.
```
dai-shi, line 256: This sounds like a limitation.
Can't we make all page parts static and make the entire page static?

tylersayshi: Technically you can, but why would you?

If your goal is to make a static page, shouldn't you just use createPage?

createPagePart only seems useful to me when you're mixing dynamic and static which then requires the page to be dynamic, right?

dai-shi: > shouldn't you just use createPage?

Technically, not always.

`getConfig` itself is dynamically configurable at build time.

```js
  return {
    render: Date.now() < SOME_POINT ? 'static' : 'dynamic',
  };
};
```

tylersayshi: I see what you mean. I can add logic for static page when all parts are static in that case. Thanks for the example.

dai-shi: marking draft for now.

tylersayshi:
@dai-shi the static support for this should basically be ready.

I am considering adding PRD only tests here: https://github.com/wakujs/waku/blob/page-part-component/e2e/create-pages.spec.ts#L7-L8

This would be to test to be sure that static pages reflect the time of the page build and stay the same on refresh. This sort of test wouldn't really be possible in dev mode afaik.

tylersayshi:
> I am considering adding PRD only tests here: [page-part-component/e2e/create-pages.spec.ts#L7-L8](https://github.com/wakujs/waku/blob/page-part-component/e2e/create-pages.spec.ts?rgh-link-date=2025-03-31T03%3A11%3A18Z#L7-L8)
> This would be to test to be sure that static pages reflect the time of the page build and stay the same on refresh. This sort of test wouldn't really be possible in dev mode afaik.

I am realizing now this isn't really possible to test. Since the parts are static, they'll stay static regardless of the page render type. We would need to check the cache somehow and I don't think that's possible, so I'm marking this PR ready

packages/waku/src/router/create-pages.ts
dai-shi: @tylersayshi Can you fix the comment?

dai-shi (commented):
Looks good otherwise.

PR #1318, opened 2025-03-24
feat(fs-router): page parts added

tylersayshi:
Adds fsRouter support for page parts


packages/waku/src/router/fs-router.ts
```diff
@@ -149,6 +156,15 @@ export function unstable_fsRouter(
             render: 'static',
             ...config,
           });
+        } else if (pathItems.at(-1)?.startsWith('_part')) {
+          const order = parseInt(pathItems.at(-1)!.split('-')[0]!.substring(5));
```
dai-shi, line 160: We prefer slice to substring. And, maybe `'_part'.length`.

packages/waku/src/router/fs-router.ts
```diff
@@ -149,6 +156,15 @@ export function unstable_fsRouter(
...
+          createPagePart({
+            path,
+            component: mod.default,
+            render: 'static',
```
dai-shi, line 164: It feels like we should make "render: 'static'" by default for `createPage`.

packages/waku/src/router/fs-router.ts
```diff
@@ -149,6 +156,15 @@ export function unstable_fsRouter(
...
+            order,
```
dai-shi, line 165: I think we should throw an error if `Number.isNaN(order)`.

packages/waku/src/router/fs-router.ts
```diff
@@ -157,7 +157,16 @@ export function unstable_fsRouter(
             ...config,
           });
         } else if (pathItems.at(-1)?.startsWith('_part')) {
-          const order = parseInt(pathItems.at(-1)!.split('-')[0]!.substring(5));
+          const order = parseInt(
+            pathItems.at(-1)!.split('-')[0]!.slice('_part'.length),
+          );
+          if (isNaN(order)) {
```
dai-shi, line 163: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN

> It is a more robust version of the original, global [isNaN()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN) function.

```suggestion
          if (Number.isNaN(order)) {
```

packages/waku/src/router/fs-router.ts
tylersayshi: So nextjs actually used this default and had a lot of user issues with it. Next 15 made pages dynamic by default (by popular demand): [link to announcement](https://arc.net/l/quote/imbnghlr)

I should make the pagePart default render be dynamic. It matches what seems to be intuitive for users.

Maybe we should make layouts dynamic by default as well if we want consistency though?

dai-shi: > Maybe we should make layouts dynamic by default as well if we want consistency though?

That sounds better for consistency.

Though, this is still a simple rule:
- layout, root: static by default
- page, page parts: dynamic by default

Hm, I may prefer everything dynamic or static by default. Let's discuss with @sandren after we finish v0.22.

dai-shi (approved):

PR #1319, opened 2025-03-24
fix(create-waku): fix cli and swap to clack + pico

### LINKED ISSUE #1289: create-waku confuses the target dir

tylersayshi:
prompts => @clack/prompts
kolorist => picocolors

these libraries are copying the pattern set by create-vite

fixes #1289

revival of #1290 since my fork was deleted


tylersayshi:
this needs to wait for v0.22.0 to be published, but then should be ready. The problem shown here is that waku@latest currently wants react ~19.0 and needs to be ~19.1

image

tylersayshi:
@dai-shi do you recognize any of the errors in CI? I'm confused where they came from honestly.

dai-shi:
Some of them are flaky tests: #1249 

Others are really failing? Do they work locally?

tylersayshi:
aha! I am missing the test changes from: https://github.com/wakujs/waku/pull/1290/files

I'll fix that now

packages/create-waku/package.json
```diff
@@ -23,16 +23,19 @@
...
     "@types/fs-extra": "^11.0.4",
     "@types/tar": "^6.1.13",
+    "execa": "^9.5.2",
     "fs-extra": "^11.3.0",
     "picocolors": "^1.1.1",
     "tar": "^7.4.3",
     "tsup": "^8.4.0",
-    "update-check": "^1.5.4"
+    "update-check": "^1.5.4",
+    "vitest": "3.0.7"
```
dai-shi, line 39:
```suggestion
    "vitest": "^3.1.1"
```

.gitignore
```diff
@@ -13,3 +13,4 @@ dist
 .node-version
 pages.gen.ts
 test-results
+.test
```
dai-shi, line 16: What creates this folder?

tylersayshi: I thought it was `pnpm e2e`, but now I'm not so sure. I'll remove it and add it again only if I see the directory reappear

packages/create-waku/package.json
```diff
@@ -23,17 +23,19 @@
     "dev": "pnpm build --watch",
     "compile": "rm -rf template dist *.tsbuildinfo && pnpm run template && pnpm run build",
     "template": "cp -r ../../examples template && rm -rf template/*/dist && rm -rf template/*/node_modules && rm -rf template/*.tsbuildinfo && (for d in template/*; do mv $d/.gitignore $d/gitignore || true; done)",
-    "build": "tsup"
+    "build": "tsup",
+    "test": "vitest run"
   },
   "devDependencies": {
+    "@clack/prompts": "^0.10.0",
```
dai-shi, line 30: It may not matter much, but let's move dependencies that are directly used from `./src` to `"dependencies"`.

tylersayshi: This makes it so the deps are installed by users instead of written into the bundle. Sounds good to me though 👍 - I have no preference.

dai-shi: > This makes it so the deps are installed by users instead of written into the bundle.

What? I thought there were no differences in our case.

tylersayshi: No, so this is the top of the bundled file when they're `dependencies`

image

When they're `devDependencies` the bundler takes the functions we use from `devDependencies` and bundles these into the `dist/index.js`. 

So when using `dependencies`:
- deps are imported to dist/index.js
- deps are installed to node_modules by users

When using `devDependencies`:
= deps are bundled into dist/index.js
- users have no need to install dependencies

This is expected behavior with the bundler as well. The devDeps approach does make for less moving pieces, but it should all be abstracted away from users.

dai-shi: Then, I guess having everything in `devDependenceis` was intentional. Sorry. Let's revert 98dd8e8bcf0a00ea7710fef033ab6d717fabbaf5.

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -0,0 +1,200 @@
...
+const projectName = 'test-waku-app';
+const genPath = path.join(__dirname, projectName);
+const genPathWithSubfolder = path.join(__dirname, '.test', projectName);
+
+const run = <SO extends SyncOptions>(
+  args: string[],
+  options?: SO,
+): SyncResult<SO> => {
+  // @ts-expect-error relies on exactOptionalPropertyTypes being false
+  return execaCommandSync(`node ${CLI_PATH} ${args.join(' ')}`, options);
```
dai-shi, line 18: I wonder what this brings over `child_process.execSync`.

tylersayshi: from the docs it looks like it helps with escaping and different forms of IO. I don't think this use necessitates it, so I could remove it, but it does seem useful as a way to exec things with a little bit of help

https://www.npmjs.com/package/execa

tylersayshi: I had simply seen this in the clack/prompts test and copied their setup

dai-shi: > I had simply seen this in the clack/prompts test and copied their setup

Ahh, I see.

I prefer execSync, that avoids mental overhead to read the code, if it works.
```suggestion
  return execSync(`node ${CLI_PATH} ${args.join(' ')}`, options);
```

tylersayshi: ah as I am reading now, the thing we get from this is the ease of checking `stdout` and setting options like cwd and timeout on the different `run`s. 

Changing to `execSync` would actually complicate the test running in this case quite a bit.

dai-shi: Alright then.

dai-shi (approved):
Haven't tested by myself, so 🤞

dai-shi:
@tylersayshi The test seems flaky:
- https://github.com/wakujs/waku/actions/runs/14287902055/job/40045278335
- https://github.com/wakujs/waku/actions/runs/14289810415/job/40049521593?pr=1353

tylersayshi:
> @tylersayshi The test seems flaky:

The flakiness is due to installing node modules as part of the test. I'll remove those? They're the only tests that can run long here.

Another option is upping the timeout to 30 or 60 seconds or more per each test to give time for install

dai-shi:
Are you sure it's due to installing? If it's expected, increasing the timeout to 30 sec sounds good.

tylersayshi:
I'll check tomorrow. But I do think it's just the install time. npm is slow, maybe I'll have the tests use pnpm somehow too.

PR #1321, opened 2025-03-25
chore(eslint): run rules on all files and add prefer-string-slice rule

tylersayshi:
I noticed not all the rules were running on all the files. I am still not sure this is 100% correct, but it runs the custom rules we had on all our files, before it wasn't.

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -9,7 +9,12 @@ const compat = new FlatCompat();
 
 export default tseslint.config(
   {
-    ignores: ['**/dist/', 'packages/create-waku/template/', '**/.cache/'],
+    ignores: [
+      '**/dist/',
+      '**/.cache/',
+      'packages/create-waku/template/',
+      'examples/07_cloudflare/.wrangler/',
+    ],
   },
   eslint.configs.recommended,
   tseslint.configs.recommended,
@@ -31,6 +36,11 @@ export default tseslint.config(
       parserOptions: {
         project: './tsconfig.eslint.json',
       },
+      globals: {
+        globalThis: 'readonly',
+        document: 'readonly',
+        setTimeout: 'readonly',
+      },
     },
     plugins: {
       unicorn,
@@ -43,7 +53,6 @@ export default tseslint.config(
         'error',
         { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
       ],
-      'no-undef': 'off',
       'react/prop-types': 'off',
       curly: ['error', 'all'],
       'unicorn/prefer-string-slice': 'error',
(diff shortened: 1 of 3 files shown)


eslint.config.ts
```diff
@@ -3,18 +3,23 @@ import { FlatCompat } from '@eslint/eslintrc';
...
+  },
   eslint.configs.recommended,
   tseslint.configs.recommended,
   importPlugin.flatConfigs.recommended,
   react.configs.flat.recommended,
   react.configs.flat['jsx-runtime'],
   ...compat.extends('plugin:react-hooks/recommended'),
   {
+    files: ['**/*.{ts,tsx,js,jsx}'],
+    ignores: ['**/dist/', 'packages/create-waku/template/', '.cache/'],
```
dai-shi, line 22: Do we need this? At least, the second line seems duplicated.

eslint.config.ts
```diff
@@ -36,8 +44,10 @@ export default tseslint.config(
         'error',
         { argsIgnorePattern: '^_', varsIgnorePattern: '^_' },
       ],
+      'no-undef': 'off',
```
dai-shi, line 47: Is this intentional?

eslint.config.ts
```diff
@@ -36,8 +44,10 @@ export default tseslint.config(
...
       'react/prop-types': 'off',
       curly: ['error', 'all'],
+      'unicorn/prefer-string-slice': 'error',
```
dai-shi, line 50: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-slice.md
Looks good.

eslint.config.ts
tylersayshi: it is intentional, though idk if it's how you'd want to solve it.

this rule triggers on `setTimeout`, `globalThis`, and `document` in our js/jsx files 

```
/Users/tyler/gitspace/waku/examples/32_minimal_js/src/components/app.jsx
  29:34  error  'setTimeout' is not defined  no-undef

/Users/tyler/gitspace/waku/examples/32_minimal_js/src/main.jsx
  13:5   error  'globalThis' is not defined  no-undef
  14:15  error  'document' is not defined    no-undef
  16:14  error  'document' is not defined    no-undef

✖ 4 problems (4 errors, 0 warnings)
```

we check the undefined status of variables with typescript. Including js files, so using eslint no-undef seemed redundant with typescript to me and that's why I just disabled the rule. 

I'm happy to change how to solve this though if you prefer another way.

eslint.config.ts
tylersayshi: good point, changed

Issue #1279, opened 2025-03-01
Add a Cloudflare Pages build option (again)


rmarscher:
We updated our `--with-cloudflare` build option from Cloudflare Workers to Cloudflare Pages and then back to Cloudflare Pages with Static Assets.

Related:
- https://github.com/dai-shi/waku/discussions/763
- https://github.com/dai-shi/waku/pull/793
- https://github.com/dai-shi/waku/pull/795
- https://github.com/dai-shi/waku/issues/944
- https://github.com/dai-shi/waku/pull/989

I found that while Cloudflare Pages' advanced mode function and Cloudflare Workers are almost the same, it seems that Cloudflare Pages' advanced mode function runs after Cloudflare's edge cache while a Cloudflare Worker always runs before the cache. https://github.com/cloudflare/cloudflare-docs/issues/20445

For sites that want to do SWR, it can be better to leverage Cloudflare's [cache rules](https://developers.cloudflare.com/cache/how-to/cache-rules/) instead of using Cloudflare's [cache API](https://developers.cloudflare.com/workers/runtime-apis/cache/) from within a worker.

dai-shi:
If it's a long-term decision, we could separate two options: `---with-cloudflare-workers` and `--with-cloudflare-pages`
Isn't the plan to integrate Workers and Pages in the future?

And, it depends on how much percentage of waku + cloudflare users need Pages.

Another option is to allow a user-defined deploy adapter.

rmarscher:
> Another option is to allow a user-defined deploy adapter

Could it be possible to pass a string in cli and then attempt to load that? Similar to the new string middleware config? 

```sh
waku build --with-adapter './my-build-adapter'
```

That would work with a local ./my-build-adapter.ts in the root of my project and also open up for community waku build adapters hosted on npm.

dai-shi:
Yeah, I like the custom build plugin approach.

Actually, it should already be possible with `unstable_viteConfigs` in waku.config.ts or vite.config.ts.
Would you give it a try and see if you face any problems?

jkhaui:
@dai-shi I would like to attempt this if you have any guidance (@rmarscher too - not sure if much has changed in Waku since we last implemented the CF pages plugin).

I like the idea of giving users a `--with-cloudflare-pages` and `--with-cloudflare-workers` option. Even if we can just use the previous page plugin as a short-term solution while thinking about a long-term build system, that'd be great - I can't deploy my Waku project to CF anymore until there's pages support :(

[...]

PR #1327, opened 2025-03-27
feat: extend build-deploy vite config

rmarscher:
I think we need this for #1279


dai-shi (approved):
LGTM

The reason I didn't was that I thought the deploy plugin was already controllable if it was developed by the same person. It might not always be the case.

PR #1330, opened 2025-03-28
update react 19.1

dai-shi:
https://github.com/facebook/react/releases/tag/v19.1.0


himself65 (approved):
Btw, should we use pnpm catalog?

dai-shi:
Maybe not. examples should just work as template outside our monorepo.

PR #1333, opened 2025-03-30
chore: cleanup after dropping support for node@18

tylersayshi:
follow up to #1331

### DIFF
--- a/packages/waku/src/lib/builder/build.ts
+++ b/packages/waku/src/lib/builder/build.ts
@@ -144,7 +144,7 @@ const analyzeEntries = async (rootDir: string, config: ConfigDev) => {
         build: {
           write: false,
           ssr: true,
-          target: 'node18',
+          target: 'node22',
           rollupOptions: {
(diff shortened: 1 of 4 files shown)


packages/waku/src/lib/builder/build.ts
```diff
@@ -144,7 +144,7 @@ const analyzeEntries = async (rootDir: string, config: ConfigDev) => {
         build: {
           write: false,
           ssr: true,
-          target: 'node18',
+          target: 'node22',
```
dai-shi, line 147: Ah, I was going to do this originally, but forgot about it.

Do they have `node22` for the target option? We still support Node 20, so it doesn't seem correct.

dai-shi: https://vite.dev/config/build-options#build-target

https://esbuild.github.io/api/#target

`node20` seems good.

dai-shi (approved):
LGTM

PR #1337, opened 2025-03-31
Bump Node.js to v22 for Vercel deployments

jonathanhefner:
Vercel [supports Node.js v22][node-js-versions] as the default version.

Similar to #685, this commit bumps Node.js to v22 for Vercel deployments.

[node-js-versions]: https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#default-and-available-versions

---

__Additional information:__ I tried setting `"engines": { "node": "22.x" }` in my package.json file, but Vercel ignored it due to this line of code.  Is it actually necessary to hard code the Node.js version here?


dai-shi:
> Is it actually necessary to hard code the Node.js version here?

Yeah, can we remove the line?

https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#default-and-available-versions

It seems like it's the way to _override_ the default.

packages/waku/src/lib/plugins/vite-plugin-deploy-vercel.ts
```diff
@@ -140,7 +140,7 @@ export function deployVercelPlugin(opts: {
           );
         }
         const vcConfigJson = {
-          runtime: 'nodejs20.x',
+          runtime: 'nodejs22.x',
```
dai-shi, line 143: Let's see if this works.

```suggestion
```

dai-shi (changes requested):

packages/waku/src/lib/plugins/vite-plugin-deploy-vercel.ts
dai-shi: Thanks for the experiment. Yeah, I think I followed the build-output-api docs.

Alright. Let's accept your original suggestion.

dai-shi (approved):

PR #1340, opened 2025-04-02
test(layouts): pathless group layouts with mixed static + dynamic

tylersayshi:
check to make sure static + dynamic nested layouts work as expected


dai-shi (approved):
Thanks!

PR #1344, opened 2025-04-02
Support installing Waku package from git repo

jonathanhefner:
Prior to this commit, when installing Waku from a git repository, the `dist` directory would not be built, and all `waku` CLI commands would throw an `ERR_MODULE_NOT_FOUND` error.

This commit adds a `prepare` command to `packages/waku/package.json`, which is just an alias for `pnpm run compile`.  pnpm will automatically run `prepare` when installing the package from a git repo, thus solving the problem.

Also, documentation about referencing a package inside a monorepo seems to be scarce and inaccurate (many sources specify incorrect syntax). Therefore, this commit adds instructions to the README on how to build an app that uses an experimental version of Waku.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1080,3 +1080,15 @@ Please join our friendly [GitHub discussions](https://github.com/wakujs/waku/dis
 ## Roadmap
 
 Waku is in active development and we’re seeking additional contributors. Check out our [roadmap](https://github.com/wakujs/waku/issues/24) for more information.
+
+## Contributing
+
+To build an app with an experimental version of Waku, change the `waku` dependency in the app's `package.json` to `"github:<REPO_OWNER>/waku#<GIT_REF>&path:/packages/waku"` and run `pnpm install`. For example:
+
+```json
+{
+  "dependencies": {
+    "waku": "github:your_username/waku#your_branch&path:/packages/waku"
+  }
+}
+```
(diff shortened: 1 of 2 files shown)


dai-shi:
> pnpm will automatically run `prepare` when installing the package from a git repo

TIL
Is it only pnpm capability?

dai-shi (approved):
LGTM
Thanks for your contribution!

jonathanhefner:
> Is it only pnpm capability?

It is also [supported by npm](https://docs.npmjs.com/cli/v11/using-npm/scripts#life-cycle-scripts), but since the `compile` script relies on pnpm, I mentioned just pnpm in the README.

Thank you for reviewing!

dai-shi:
> It is also [supported by npm](https://docs.npmjs.com/cli/v11/using-npm/scripts#life-cycle-scripts),

> Runs BEFORE the package is packed, i.e. during npm publish

Ah, I think I knew it.

> Runs on local npm install without any arguments

But not that.

Thanks for sharing.

README.md
```diff
@@ -1080,3 +1080,15 @@ Please join our friendly [GitHub discussions](https://github.com/wakujs/waku/dis
 ## Roadmap
 
 Waku is in active development and we’re seeking additional contributors. Check out our [roadmap](https://github.com/wakujs/waku/issues/24) for more information.
+
+## Contributing
```
dai-shi, line 1084: Actually, I misunderstood. This was added to README.md. It should have added to CONTRIBUTING.md.

@jonathanhefner Can you move this section from README.md to CONTRIBUTING.md please? Feel free to improve CONTRIBUTING.md too.

PR #1348, opened 2025-04-03
feat(api): support default export for handling all api methods

tylersayshi:
for api routes, this will support a single default export as a catch-all method handler.

it will be overridden by any named method export when both are present.

for #1345 cc @hamlim


tylersayshi:
I am realizing that we do not have api routes documented in the readme, so I'll add that first in a separate PR, then make the adjustments needed to support this one after.

packages/waku/src/router/create-pages.ts
```diff
@@ -278,7 +282,7 @@ export const createPages = <
     {
       render: 'static' | 'dynamic';
       pathSpec: PathSpec;
-      handlers: Partial<Record<Method, ApiHandler>>;
+      handlers: Partial<Record<Method | 'all', ApiHandler>>;
```
dai-shi, line 285: I don't dislike `'all'`. But, just another idea is `'*'`.

tylersayshi: hm it would seem weird to me to see if the object was called literally, like:

```ts
createApi({
  render: 'dynamic',
  handlers: {
    '*': handler,
    GET: getHandler
  }
})
```

What about using the name `default`? At first I thought it may be weird, but I think it might be intuitive?

dai-shi: "all" sounds better to me than "default". I don't have a strong opinion though.

tylersayshi: I'll leave it as is then?

dai-shi: 👍

PR #1354, opened 2025-04-04
fix: React component as server function argument

jonathanhefner:
This test currently fails with the following error:

  > React Element cannot be passed to Server Functions from the Client without a temporary reference set. Pass a TemporaryReferenceSet to the options. {children: </>} ^^^

---

@dai-shi  I tried to find the best place to add an e2e per https://github.com/wakujs/waku/issues/1350#issuecomment-2777283099.  Let me know if you prefer elsewhere.

---

(edit by @dai-shi) Added some commits and fixed it.


dai-shi:
> Cannot access constructor on the server. You cannot dot into a temporary client reference from a server component. You can only pass the value through to the client.

Hmm, somehow the temporary references don't work as expected on the server.

PR #1359, opened 2025-04-06
chore: use customConditions to support go to definition internally

tylersayshi:
Kapture 2025-04-05 at 21 29 21

This makes it so "go to definition" works from internal packages. I find it quite helpful for developing.

blog post describing this approach: https://colinhacks.com/essays/live-types-typescript-monorepo

### DIFF
--- a/packages/waku/package.json
+++ b/packages/waku/package.json
@@ -18,51 +18,63 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "waku-internal": "./src/main.ts",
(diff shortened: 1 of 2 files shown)


dai-shi:
I'm not very happy with changing something in the production build. Are there any other methods that work only for DEV?

tylersayshi:
There's a tsconfig path's technique but I've seen issues with that, so I'd just close this PR if the package.json change isn't ok.

PR #1363, opened 2025-04-07
test: change cli installer to pnpm

tylersayshi:
change to faster installer for tests with pnpm instead of npm


dai-shi:
> increase timeout and change to faster installer for tests with pnpm instead of npm

If pnpm is faster, I don't think we need to increase timeout.

PR #1366, opened 2025-04-08
fix(create-waku): do not ask package name by default

dai-shi:
We should keep the previous behavior.


dai-shi:
Yeah. More general idea is to ask as less as possible.

PR #1378, opened 2025-04-20
fix: update client module cache after n css module importee hmr

### LINKED ISSUE #1328: (dev-only) BUG: half-broken behavior of css modules in client components and any components inside them (class mismatch)
The element class becomes desynced with the actual css on the first page load after following the next steps.
This bug affects the dev mode.

Steps to reproduce the issue:
1. create a new waku project without changing anything
2. create a client component and a corresponding css module file, make a visibly styled element (for example a colored div)
3. run the dev server and open the page
4. change the style of the element and see it updating properly
5. reload the page, class mismatch appears
6. update the style again and see it becoming normal before you refresh the page again

This issue doesn't affect RSCs, they work completely fine.

Further digging:
7. create an RSC that's using the same css module file with the same element
8. repeat the steps above and see only the RSC working fine while the client component is unstyled.

Aslemammad:
Resolves #1328


packages/waku/src/lib/plugins/vite-plugin-rsc-hmr.ts
```diff
@@ -147,29 +147,19 @@ export function rscHmrPlugin(): Plugin {
...
       }
     },
-    handleHotUpdate({ file }) {
+    handleHotUpdate({ file, modules, server }) {
+      const module = server.moduleGraph.getModuleById(file);
+      if (module?.file?.endsWith('.module.css')) {
+        module.importers.forEach((importer) => {
+          handleModuleUpdate(importer.file!, server);
+        });
+      }
```
Aslemammad, line 156: This is just to track the css module files! I wonder if we can have a better code for this. I just did it because it's part of the fix.

dai-shi (approved):
Thanks!!

PR #1380, opened 2025-04-21
chore(create-waku): cloudflare template types update

rmarscher:
Update the cloudflare template to use the latest preferred method for generating cloudflare worker types. https://developers.cloudflare.com/workers/languages/typescript/#generate-types


examples/07_cloudflare/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "skipLibCheck": true,
     "noUncheckedIndexedAccess": true,
     "exactOptionalPropertyTypes": true,
-    "types": ["react/experimental", "./.wrangler/types/runtime.d.ts"],
+    "types": ["./worker-configuration.d.ts"],
```
dai-shi, line 12: If `"react/experimental"` is no longer required, we should remove it from all examples and fixtures (in a separate PR).

dai-shi: #1382

Issue #1338, opened 2025-04-01
While navigating to a route containing an async server component and encoded url part with a `Link` that has `unstable_loading`, the page get stuck


arianrhodsandlot:
This bug appears when these 3 factors exist:
+ an async server component
+ an encoded path inside the URL, eg: /page/a%20b
+ a `Link` component with `unstable_pending` being set

Then we can not navigate to the target route by clicking the `Link`.

reproduction:
https://stackblitz.com/edit/stackblitz-starters-sgxqfnou

dai-shi:
That sounds relatively rare, but definitely a bug. Would anyone like to look into it?

tylersayshi:
this bug sounds like a multi-step cheat code to a video game in the early 2000s 😂

anyways... I'm checking it out, I'll report back soon!

tylersayshi:
@dai-shi I'm seeing this error in the callstack

```
Error: react-stack-top-frame
          at fakeJSXCallSite (/Users/tyler/gitspace/waku/examples/001_broken/node_modules/.vite/waku-dev-server-main/deps_ssr/react-server-dom-webpack_client__edge.js?v=fbc1a2f0:4955:16)
          at Object.handleRoute (file:///Users/tyler/gitspace/waku/packages/waku/dist/router/create-pages.js:514:47)
          at async getEntries (file:///Users/tyler/gitspace/waku/packages/waku/dist/router/define-router.js:141:53)
          at async renderIt (file:///Users/tyler/gitspace/waku/packages/waku/dist/router/define-router.js:224:25)
          at async Object.handleRequest (file:///Users/tyler/gitspace/waku/packages/waku/dist/router/define-router.js:247:18)
          at async Array.<anonymous> (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/middleware/handler.js:120:23)
          at async run (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/hono/engine.js:27:13)
          at <anonymous> (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/hono/engine.js:30:21)
          at <anonymous> (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/middleware/dev-server-impl.js:397:13)
          at async run (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/hono/engine.js:27:13)
          at <anonymous> (file:///Users/tyler/gitspace/waku/packages/waku/dist/lib/hono/engine.js:30:21)
[...]

PR #1391, opened 2025-04-25
feat(router): routing events

### LINKED ISSUE #1372: support for events listening for waku/router
[original ask on discord](https://discord.com/channels/627656437971288081/1083774812129857566/1361387461766545559)

current idea for usage is something like:

```ts
const router = useRouter();
const cleanup = router.unstable_onRouteChangeStart(() => {
  // ...
});
``` 

Start and finished states are important to start with.

---

Notes on other frameworks:

nextjs/pages: https://nextjs.org/docs/pages/api-reference/functions/use-router#routerevents
nextjs/app: https://nextjs.org/docs/app/api-reference/functions/use-link-status
remix/react-router: https://reactrouter.com/api/hooks/useNavigation
tanstack-router: https://tanstack.com/router/latest/docs/framework/react/api/router/RouterEventsType

a helpful thread with francois related to this: https://bsky.app/profile/francoisbest.com/post/3lmuvooscos2z

@wesbos's current workaround: https://github.com/wesbos/wesbos/commit/1f0ee30788439a5f83bdcc4b42851e77777a97f9

Ideas, comments, and concerns are welcome 😃

tylersayshi:
uses ref and context for adding callbacks for when route change completes and starts.

emitting browser events still seems explorable.

fixes #1372 #1338

### DIFF
--- a/packages/waku/src/minimal/client.ts
+++ b/packages/waku/src/minimal/client.ts
@@ -8,7 +8,6 @@ import {
   useCallback,
   useEffect,
   useState,
-  Component,
 } from 'react';
 import type { ReactNode } from 'react';
 import RSDWClient from 'react-server-dom-webpack/client';
@@ -58,12 +57,6 @@ const checkStatus = async (
 
 type Elements = Record<string, unknown>;
 
-// HACK I'm not super happy with this hack
-const erroredElementsPromiseMap = new WeakMap<
-  Promise<Elements>,
-  Promise<Elements>
->();
-
 const getCached = <T>(c: () => T, m: WeakMap<object, T>, k: object): T =>
   (m.has(k) ? m : m.set(k, c())).get(k) as T;
 const cache1 = new WeakMap();
@@ -71,19 +64,12 @@ const mergeElementsPromise = (
   a: Promise<Elements>,
   b: Promise<Elements>,
 ): Promise<Elements> => {
-  const getResult = () => {
-    const p = Promise.all([erroredElementsPromiseMap.get(a) || a, b])
-      .then(([a, b]) => {
-        const nextElements = { ...a, ...b };
-        delete nextElements._value;
-        return nextElements;
-      })
-      .catch((err) => {
-        erroredElementsPromiseMap.set(p, a);
-        throw err;
-      });
-    return p;
-  };
+  const getResult = () =>
+    Promise.all([a, b]).then(([a, b]) => {
+      const nextElements = { ...a, ...b };
+      delete nextElements._value;
+      return nextElements;
+    });
   const cache2 = getCached(() => new WeakMap(), cache1, a);
   return getCached(getResult, cache2, b);
 };
@@ -140,10 +126,11 @@ export const unstable_callServerRsc = async (
           enhanceFetch(fetch)(url, { method: 'POST', body }),
         );
   const data = enhanceCreateData(createData)(responsePromise);
-  const value = (await data)._value;
-  // FIXME this causes rerenders even if data is empty
-  fetchCache[SET_ELEMENTS]?.((prev) => mergeElementsPromise(prev, data));
-  return value;
+  const dataWithoutErrors = Promise.resolve(data).catch(() => ({}));
+  fetchCache[SET_ELEMENTS]?.((prev) =>
+    mergeElementsPromise(prev, dataWithoutErrors),
+  );
+  return (await data)._value;
 };
 
 const prefetchedParams = new WeakMap<Promise<unknown>, unknown>();
@@ -259,8 +246,8 @@ export const Root = ({
       // clear cache entry before fetching
       delete fetchCache[ENTRY];
       const data = fetchRsc(rscPath, rscParams, fetchCache);
-      setElements((prev) => mergeElementsPromise(prev, data));
-      // TODO Should we move this `await data` before `setElements` like callServerRsc and let the caller handle the error?
+      const dataWithoutErrors = Promise.resolve(data).catch(() => ({}));
+      setElements((prev) => mergeElementsPromise(prev, dataWithoutErrors));
       await data;
     },
     [fetchCache],
@@ -281,27 +268,8 @@ export const useRefetch = () => use(RefetchContext);
 
 const ChildrenContext = createContext<ReactNode>(undefined);
 const ChildrenContextProvider = memo(ChildrenContext.Provider);
-const ErrorContext = createContext<
-  [error: unknown, reset: () => void] | undefined
->(undefined);
-const ErrorContextProvider = memo(ErrorContext.Provider);
 
 export const Children = () => use(ChildrenContext);
-
-export const ThrowError_UNSTABLE = () => {
-  const errAndReset = use(ErrorContext);
-  if (errAndReset) {
-    throw errAndReset[0];
-  }
-  return null;
-};
-
-export const useResetError_UNSTABLE = () => {
-  const errAndReset = use(ErrorContext);
-  if (errAndReset) {
-    return errAndReset[1];
-  }
-};
 
 export const useElement = (id: string) => {
   const elementsPromise = use(ElementsContext);
@@ -314,70 +282,6 @@ export const useElement = (id: string) => {
   }
   return elements[id];
 };
-
-const InnerSlot = ({
-  id,
-  children,
-  setValidElement,
-  unstable_fallback,
-}: {
-  id: string;
-  children?: ReactNode;
-  setValidElement?: (element: ReactNode) => void;
-  unstable_fallback?: ReactNode;
-}) => {
-  const element = useElement(id);
-  const isValidElement = element !== undefined;
-  useEffect(() => {
-    if (isValidElement && setValidElement) {
-      // FIXME is there `isReactNode` type checker?
-      setValidElement(element as ReactNode);
-    }
-  }, [isValidElement, element, setValidElement]);
-  if (!isValidElement) {
-    if (unstable_fallback) {
-      return unstable_fallback;
-    }
-    throw new Error('Invalid element: ' + id);
-  }
-  return createElement(
-    ChildrenContextProvider,
-    { value: children },
-    // FIXME is there `isReactNode` type checker?
-    element as ReactNode,
-  );
-};
-
-class GeneralErrorHandler extends Component<
-  { children?: ReactNode; errorHandler: ReactNode },
-  { error: unknown | null }
-> {
-  constructor(props: { children?: ReactNode; errorHandler: ReactNode }) {
-    super(props);
-    this.state = { error: null };
-    this.reset = this.reset.bind(this);
-  }
-  static getDerivedStateFromError(error: unknown) {
-    return { error };
-  }
-  reset() {
-    this.setState({ error: null });
-  }
-  render() {
-    const { error } = this.state;
-    if (error !== null) {
-      if (this.props.errorHandler) {
-        return createElement(
-          ErrorContextProvider,
-          { value: [error, this.reset] },
-          this.props.errorHandler,
-        );
-      }
-      throw error;
-    }
-    return this.props.children;
-  }
-}
 
 /**
  * Slot component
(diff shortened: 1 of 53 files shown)


rmarscher:
The actual server request happens in refetch which is separate from the route state update. It would probably be better if the route didn't change until the refetch request resolves. Maybe refetch can return a promise that changeRoute can await. https://github.com/wakujs/waku/blob/v0.22.4/packages/waku/src/minimal/client.ts#L258-L266

rmarscher:
Maybe like this? https://github.com/wakujs/waku/compare/routing-events...rmarscher:waku:routing-events

rmarscher:
Maybe refs should be used to hold the current callback instead of state. I'm testing and trying some things on my branch...

tylersayshi:
@rmarscher this seems to work really well actually. Thanks for the help again 🙌 

@dai-shi would you like to review the API and changes?

In the meantime, I'll work on adding some tests

packages/waku/src/minimal/client.ts
```diff
@@ -260,6 +260,7 @@ export const Root = ({
       delete fetchCache[ENTRY];
       const data = fetchRsc(rscPath, rscParams, fetchCache);
       setElements((prev) => mergeElementsPromise(prev, data));
+      return data;
```
dai-shi, line 263: Something feels itchy. We shouldn't leak `data`, which isn't a real promise. I'm trying to refactor, but it makes some tests to fail...

dai-shi: Also, I wonder if `refetch` is async, should the caller handle the fetch error? This code smells quite a bit.

dai-shi: Pushed 5e6ecce3 anyway. The goal is to hide `data`. Tests are failing. This makes me feel I don't understand something.

rmarscher: Hmm.. The purpose the changes I made was to make the startTransitionFn wait for the server response to be received and to start streaming. https://github.com/wakujs/waku/pull/1391/files#diff-489ee8978c94b27e9ecc82b21d9226a7cdda0a32d76527dddbd251bedb73d74bR324-R334

Doing that seems to fix other existing issues like #1338. Maybe there is another way we can rewrite it to achieve the same effect.

rmarscher: Maybe return `responsePromise` and `data` from fetchRsc and then return responsePromise from changeRoute?

dai-shi: I'm still skeptical about `refetch` being async because of the error handling perspective, but maybe it's correct if it fixes two issues. It's safer to return just `Promise<void>`, and we should fix it and understand how it changes the behavior.

I would propose to separate this into three PRs.
- async `refetch` (probably needs e2e tests, needs design decision if we should catch errors with refetch call, or ErrorBoundry should catch it.)
- async `changeRoute` (fix #1338)
- `unstable_onRouteChangeStart/Complete` (fix #1372, and I would like to refactor the impl)

dai-shi: > needs design decision if we should catch errors with refetch call

More to think about it, it feels correct especially if it removes hacky `mergeElements`. I'll give it a try.

tylersayshi: sounds good. Can you expand your thoughts on this though:

> I would like to refactor the impl

dai-shi: I haven't looked into it yet closely, but at a glance, I don't think we need the second context, and I want to avoid useRef if possible.

tylersayshi: trying to debug this now. It looks like there is a race condition with the elementsPromise. On navigation, I think we need to await the elements of the new route to be fetched before changing the route. 

This is currently not how it is handled and there is a split second between when the route changes and the page re-renders to have all the elements for the new route.

It seems to be specific to encoded urls, so I think there is a mismatch for where the refetch is being called for encoded paths. So I think the steps are something like

1. click link and fetch for `/foo/is encoded`
2. promise resolves, change route to `/foo/is encoded`
3. render for /foo/is%20encoded and find `missing element id: "/foo/is%20encoded"`
4. refetch for `page:/foo/is%20encoded`, when promise resolves, re-render

This doesn't explain why #1338 get's stuck, but my best guess is that the uncaught error in that case results in some sort of infinite loop. Maybe it keeps looking for encoded and then decoded and on and on?

Anyways, I'm looking to pinpoint where this mismatched fetch is for now because that should be quite fixable.

tylersayshi: There aren't subsequent network requests, so the logic of encoded and decoded being mismatched must be internal 🤔

tylersayshi: made a draft showing where I was looking for this: https://github.com/wakujs/waku/pull/1395

packages/waku/src/router/client.ts
```diff
@@ -110,11 +119,42 @@ const RouterContext = createContext<{
   prefetchRoute: PrefetchRoute;
 } | null>(null);
 
+const RouteChangeContext = createContext<{
+  unstable_onRouteChangeStartRef: RefObject<ChangeRouteCallback | null>;
+  unstable_onRouteChangeCompleteRef: RefObject<ChangeRouteCallback | null>;
```
dai-shi, line 124: If we use stable things, we don't need a separate context. Can you put them in RouterContext?

Second request: Let's pass `Set<ChangeRouteCallback>` instead of ref in the context.

dai-shi (changes requested):
1st review

packages/waku/src/router/client.ts
tylersayshi: ok. I made the suggested changes. It changes how the api is, so maybe there will be feedback for that, but the context is gone and sets are used.

should be ready for another review

tylersayshi:
modeled this off the nextjs pages events implementation: https://nextjs.org/docs/pages/api-reference/functions/use-router#routerevents

packages/waku/src/router/client.ts
```diff
@@ -102,19 +103,46 @@ type ChangeRoute = (
...
 ) => Promise<void>;
 
+type ChangeRouteEvent = 'start' | 'complete';
+
+type ChangeRouteCallback = (
+  route: RouteProps,
+  options: {
+    shouldScroll: boolean;
+    skipRefetch?: boolean;
+  },
```
dai-shi, line 113: This is internal thing. Is it necessary to export it?

packages/waku/src/router/client.ts
```diff
@@ -102,19 +103,46 @@ type ChangeRoute = (
   },
 ) => Promise<void>;
 
+type ChangeRouteEvent = 'start' | 'complete';
+
+type ChangeRouteCallback = (
+  route: RouteProps,
```
dai-shi, line 109: I wonder if we should expose `RoutePros`, should it be a string? Not sure tho.

packages/waku/src/router/client.ts
```diff
@@ -102,19 +103,46 @@ type ChangeRoute = (
...
+  options: {
+    shouldScroll: boolean;
+    skipRefetch?: boolean;
+  },
+) => void;
+
+const EMPTY_ROUTE_CHANGE_LISTENERS: Record<
+  ChangeRouteEvent,
+  Set<ChangeRouteCallback>
+> = { start: new Set(), complete: new Set() };
```
dai-shi, line 119: Defining module state is incorrect here. We need to create new sets for different routers (mostly hypothetical though).
```js
const createRouteChangeListeners = () => ({ start: new Set(), complete: new Set() });
```

packages/waku/src/router/client.ts
```diff
@@ -102,19 +103,46 @@ type ChangeRoute = (
...
+  routeChangeListeners: Record<ChangeRouteEvent, Set<ChangeRouteCallback>>;
 } | null>(null);
 
 export function useRouter() {
   const router = useContext(RouterContext);
   if (!router) {
     throw new Error('Missing Router');
   }
+
+  const unstable_events = useMemo(() => {
```
dai-shi, line 136: We are using `useRef` internally anyway, so technically, we can avoid useMemo here and useCallback for executeListeners.
I'm not sure if the final result is more readable or less.
I would do something like this:
```diff
- const routeChangeListenersRef = useRef(EMPTY_ROUTE_CHANGE_LISTENERS);
+ const routeChangeListenersRef = useRef(null);
+ if (routeChangeListenersRef.current === null) {
+   const listeners = { start: new Set(), complete: new Set() };
+   const events = ...;
+   const executeLisetners = ...;
+   routeChangeListenersRef.current = [events, executeListenrs];
+ }
```

dai-shi (commented):
2nd review

packages/waku/src/router/client.ts
tylersayshi: having path, query, and hash separately seems useful to me. We could pass it as just a full string, but that seems less useful.

packages/waku/src/router/client.ts
tylersayshi: changed

packages/waku/src/router/client.ts
tylersayshi: changed

tylersayshi (commented):
thanks

packages/waku/src/router/client.ts
```diff
@@ -498,11 +558,13 @@ const InnerRouter = ({
     });
   }, [initialRoute]);
 
+  const [, executeListeners] = routeChangeListenersRef.current;
```
dai-shi, line 561: This should be more readable.
```suggestion
  const [routeChangeEvents, executeListeners] = routeChangeListenersRef.current;
```

packages/waku/src/router/client.ts
```diff
@@ -102,19 +102,36 @@ type ChangeRoute = (
...
+
+type ChangeRouteCallback = (route: RouteProps) => void;
+
+const EMPTY_ROUTE_CHANGE_LISTENER: Record<
+  'on' | 'off',
+  (event: ChangeRouteEvent, handler: ChangeRouteCallback) => void
+> = {
+  on: () => {},
+  off: () => {},
+};
```
dai-shi, line 115: Something like this?
```suggestion
const DUMMY_ROUTE_CHANGE_LISTENER: Record<
  'on' | 'off',
  (event: ChangeRouteEvent, handler: ChangeRouteCallback) => void
> = {
  on: () => notAvailableInServer('routeChange:on'),
  off: () => notAvailableInServer('routeChange:off'),
};
```

Can we define this close to `INTERNAL_ServerRouter`?
`DUMMY_` sounds misleading. Suggestions?

packages/waku/src/router/client.ts
dai-shi: Sounds fair.

dai-shi (commented):
3rd review

packages/waku/src/router/client.ts
tylersayshi: mock?

dai-shi: > mock?

maybe.

Can we define this close to INTERNAL_ServerRouter?

packages/waku/src/router/client.ts
```diff
@@ -521,10 +584,11 @@ const InnerRouter = ({
         handleScroll();
       }
       setRoute(route);
+      executeListeners('complete', route);
       refetching.current[0]?.();
       refetching.current = null;
```
dai-shi, line 589:
```suggestion
      refetching.current[0]?.();
      refetching.current = null;
      executeListeners('complete', route);
```

By the way, "complete" doesn't mean "finish". It means, it's done to tell React render. So, it's the "start" of rendering.

packages/waku/src/router/client.ts
```diff
@@ -498,11 +558,14 @@ const InnerRouter = ({
     });
   }, [initialRoute]);
 
+  const [_routeChangeEvents, executeListeners] =
```
dai-shi, line 561:
```suggestion
  const [routeChangeEvents, executeListeners] =
```

packages/waku/src/router/client.ts
```diff
@@ -600,7 +664,14 @@ const InnerRouter = ({
   );
   return createElement(
     RouterContext.Provider,
-    { value: { route, changeRoute, prefetchRoute } },
+    {
+      value: {
+        route,
+        changeRoute,
+        prefetchRoute,
+        routeChangeEvents: routeChangeListenersRef.current[0],
```
dai-shi, line 672:
```suggestion
        routeChangeEvents,
```

packages/waku/src/router/client.ts
dai-shi, line 589:
```suggestion
      refetching.current[0]?.();
      refetching.current = null;
      executeListeners('complete', route);
```

By the way, "complete" doesn't mean "finish". It means, it's done to tell React render. So, it's the "start" of rendering.

packages/waku/src/router/client.ts
dai-shi: > mock?

maybe.

Can we define this close to INTERNAL_ServerRouter?

dai-shi (commented):
4th review

packages/waku/src/router/client.ts
```diff
@@ -709,7 +771,13 @@ export function Router({
...
   );
 }
-
+const MOCK_ROUTE_CHANGE_LISTENER: Record<
+  'on' | 'off',
+  (event: ChangeRouteEvent, handler: ChangeRouteCallback) => void
+> = {
+  on: () => notAvailableInServer('routeChange:on'),
+  off: () => notAvailableInServer('routeChange:off'),
+};
```
dai-shi, line 780:
```suggestion

const MOCK_ROUTE_CHANGE_LISTENER: Record<
  'on' | 'off',
  (event: ChangeRouteEvent, handler: ChangeRouteCallback) => void
> = {
  on: () => notAvailableInServer('routeChange:on'),
  off: () => notAvailableInServer('routeChange:off'),
};

```

dai-shi (approved):
LGTM

PR #1392, opened 2025-04-26
#1391 routing events updates

rmarscher:
I think these help the cause. @tylersayshi

### DIFF
--- a/examples/21_create-pages/src/components/RoutingHandler.tsx
+++ b/examples/21_create-pages/src/components/RoutingHandler.tsx
@@ -1,18 +1,15 @@
 'use client';
 
-import { useEffect } from 'react';
 import { useRouter } from 'waku/router/client';
 
 export const RoutingHandler = () => {
   const { unstable_onRouteChangeComplete, unstable_onRouteChangeStart } =
     useRouter();
-  useEffect(() => {
-    unstable_onRouteChangeStart((...args) => {
-      console.log('onRouteChangeStart', args);
-    });
-    unstable_onRouteChangeComplete((...args) => {
-      console.log('onRouteChangeComplete', args);
-    });
-  }, [unstable_onRouteChangeComplete, unstable_onRouteChangeStart]);
+  unstable_onRouteChangeStart((...args) => {
+    console.log('onRouteChangeStart', args);
+  });
+  unstable_onRouteChangeComplete((...args) => {
+    console.log('onRouteChangeComplete', args);
+  });
   return null;
 };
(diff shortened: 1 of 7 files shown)


tylersayshi:
@dai-shi how would you feel about adding this lint rule? https://typescript-eslint.io/rules/no-floating-promises/

it would help catch things like the floating refetch calls that rob added `void` to here

rmarscher:
> @dai-shi how would you feel about adding this lint rule? https://typescript-eslint.io/rules/no-floating-promises/
> 
> it would help catch things like the floating refetch calls that rob added `void` to here

I think that rule is already there. I had to fix it when I ran `pnpm run test`.

examples/21_create-pages/src/components/RoutingHandler.tsx
```diff
@@ -1,18 +1,18 @@
...
-    console.log('onRouteChangeComplete');
-  });
+  useEffect(() => {
+    unstable_onRouteChangeStart((...args) => {
+      console.log('onRouteChangeStart', args);
+    });
+    unstable_onRouteChangeComplete((...args) => {
+      console.log('onRouteChangeComplete', args);
+    });
+  }, [unstable_onRouteChangeComplete, unstable_onRouteChangeStart]);
```
tylersayshi, line 16: hm, if it looks like this, the better api would probably just be to give `useRouter().isPending` 

I personally like the `useRouter().events.on('routechangestart', someCallback)` api best from what I've seen and I think that would need to use emitted browser events to work?

rmarscher: Actually, since I changed it to use a RefObject, it probably doesn't have to still go in a useEffect to avoid over-triggering.

rmarscher:
It looks like I did something to break e2e. I'm looking into it.

tylersayshi:
> I think that rule is already there. I had to fix it when I ran `pnpm run test`.

ah, I see it now, nice!

examples/21_create-pages/src/components/RoutingHandler.tsx
rmarscher: Yup. I was able to simplify that to remove `useEffect`.

Re: `useRouter().isPending` -- I think we don't want to add startTransition or state updates though. Not sure. The event callback approach won't add anything to React unless the developer decides to - which is probably nice for this use case.

rmarscher:
> thanks for the assist here!

This actually solves a problem I have in a work app with the url updating and page scrolling to top before the server has responded. I've had that one sitting in the backlog for a few months... 😬

tylersayshi:
merging these to my branch to see where that leaves us on the other branch. thanks!

PR #1395, opened 2025-05-01
fix(create-pages): keep path encoded consistently when passed as a page prop

tylersayshi:
this is related to #1338

this fixes the inconsistent `path` sent to page components as a prop through using create-pages


dai-shi:
Nice investigation. 

> There seems to be a mechanism for properly waiting for non-encoded paths to be fetched from the elementsPromise that is not working for encoded paths.

Yeah, if it's a timing issue, it can happen with the normal path, like `/nested/normal-path`. So, it sounds like we miss doing encode/decode somewhere.

> so I'm hoping you'd have an idea.

Unfortunately, nothing at the moment.

tylersayshi:
@dai-shi this is the bug currently - on client side navigation the decoded path is shown, but on full browser refresh, the path is encoded

Kapture 2025-06-05 at 22 05 28

packages/waku/src/lib/middleware/handler.ts
```diff
@@ -35,9 +35,10 @@ const getInput = async (
   }
   const basePrefix = config.basePath + config.rscBase + '/';
   if (ctx.req.url.pathname.startsWith(basePrefix)) {
-    const rscPath = decodeRscPath(
+    let rscPath = decodeRscPath(
       decodeURI(ctx.req.url.pathname.slice(basePrefix.length)),
     );
+    rscPath = rscPath.split('/').map(encodeURIComponent).join('/');
```
tylersayshi: @dai-shi let me know how you'd like this to be cleaned up or fixed differently, but this makes it so the path passed through our router as a page prop is consistently encoded.

tylersayshi:
tests pass locally now, I think the failures may be related to partially failed install of deps due to the gcp outage today https://news.ycombinator.com/item?id=44260810

tylersayshi:
@himself65 do you have any ideas for why the windows tests seem to fail here? I didn't think I changed something here that would affect the broken-links test that is failing... also don't have a windows machine to check it out myself.

tylersayshi:
> @himself65 do you have any ideas for why the windows tests seem to fail here? I didn't think I changed something here that would affect the broken-links test that is failing... also don't have a windows machine to check it out myself.

ohh i see this now. I'll mark mine ready since i think the fails remaining are fixed here
https://github.com/wakujs/waku/pull/1471

packages/waku/src/router/create-pages.ts
```diff
@@ -771,7 +771,11 @@ export const createPages = <
       }
       const layoutMatchPath = groupPathLookup.get(routePath) ?? routePath;
       const pathSpec = parsePathWithSlug(layoutMatchPath);
-      const mapping = getPathMapping(pathSpec, path);
+      const mapping = getPathMapping(
+        pathSpec,
+        // ensure path is encoded for props of page component
+        path.split('/').map(encodeURIComponent).join('/'),
```
dai-shi: This looks like bad practice. Doesn't it work with `encodeURI(path)`?

tylersayshi: `encodeURI(decodeURI(path))` seems to work.

The issue is that sometimes `path` is already encoded and other times it is not. So I think if we always decode and then encode the logic is correct and better than the split, map, join approach like you're mentioning here.

dai-shi: It looks much better than split&join. However,

> The issue is that sometimes path is already encoded and other times it is not.

That sounds really strange. I wonder if other frameworks / routers deal with it.

That said, I'm fine to merge this and leave it for the future discussion. Let me know what you think.

tylersayshi: I agree, we can merge this. It is an improvement over current behavior.

It would be better to fix the inconsistency for how rscPath is encoded higher up in our callstack as a follow up. Like rscPath should either always be encoded or not encoded.

dai-shi: > Like rscPath should either always be encoded or not encoded.

Can you leave the source code comment **and** create a corresponding issue please?

dai-shi: @tylersayshi 

Hm, even if I remove decodeURI, all tests still seem to pass. Is it unexpected?

```diff
diff --git a/packages/waku/src/router/create-pages.ts b/packages/waku/src/router/create-pages.ts
index c92df908..54afb40c 100644
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -774,7 +774,7 @@ export const createPages = <
       const mapping = getPathMapping(
         pathSpec,
         // ensure path is encoded for props of page component
-        encodeURI(decodeURI(path)),
+        encodeURI(path),
       );
       const result: Record<string, unknown> = {};
       if (Array.isArray(pageComponent)) {
```

> The issue is that sometimes path is already encoded and other times it is not.

Doesn't the test cover such cases?

tylersayshi: I'll add a test, the tests currently miss the case I am talking about so adding one is a good idea.

dai-shi:
> The issue is that sometimes `path` is already encoded and other times it is not.

It turns out that it's rsc or ssr. #1489 should fix!

PR #1396, opened 2025-05-02
feat: async refetch

dai-shi:
ref: https://github.com/wakujs/waku/pull/1391#discussion_r2069602375

This is a breaking change behavior-wise. We need to catch errors with try-catch and the error doesn't go to the error boundary by default.

- [x] make refetch async
- [x] fix e2e
- [x] refactor
- [x] add refetch with transaction test

### DIFF
--- a/examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
+++ b/examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
@@ -80,7 +80,9 @@ export const BaseSyncAtoms = ({
         const rscParams = {
           jotai_atomValues: serializedAtomValues,
         };
-        refetch(...createRscPathAndRscParams(rscPath, rscParams));
+        refetch(...createRscPathAndRscParams(rscPath, rscParams)).catch((e) => {
+          console.error('Failed to refetch:', e);
+        });
       };
       const unsub = store.sub(atomValuesAtom, () => {
         callback(store.get(atomValuesAtom));
--- a/examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
+++ b/examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
@@ -80,9 +80,8 @@ export const BaseSyncAtoms = ({
         const rscParams = {
           jotai_atomValues: serializedAtomValues,
         };
-        refetch(...createRscPathAndRscParams(rscPath, rscParams)).catch((e) => {
-          console.error('Failed to refetch:', e);
-        });
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        refetch(...createRscPathAndRscParams(rscPath, rscParams));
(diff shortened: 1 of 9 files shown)


examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
```diff
@@ -80,6 +80,7 @@ export const BaseSyncAtoms = ({
         const rscParams = {
           jotai_atomValues: serializedAtomValues,
         };
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         refetch(...createRscPathAndRscParams(rscPath, rscParams));
```
tylersayshi, line 84:
```suggestion
        void refetch(...createRscPathAndRscParams(rscPath, rscParams));
```

examples/54_jotai/src/lib/waku-jotai/client.tsx
```diff
@@ -80,6 +80,7 @@ export const BaseSyncAtoms = ({
         const rscParams = {
           jotai_atomValues: serializedAtomValues,
         };
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
         refetch(...createRscPathAndRscParams(rscPath, rscParams));
```
tylersayshi, line 84: same void comment here

tylersayshi (commented):
looks like there was some good cleanup along with this... thanks!

examples/08_jotai-demo/src/lib/waku-jotai/client.tsx
dai-shi: This is intentional. To me, `void` implies "it's good." No, it's not good. We shouldn't hide errors (or they will be uncaught). So, in other places, I added `.catch`. For `waku-jotai`, I'll revisit and `eslint-disable` shows less "it's good."

dai-shi:
For context, when I started developing Waku, `startTransition` didn't accept async function as far as I recall. So, it seemed "fetch error" should be caught in the error boundaries.
Now that `startTransition` accepts or even recommends async functions, it feels more natural to catch errors there. I'm not 100% confident on this idea, but it eliminates all hacks in minimal/client required for error handlings for router/client.

A few concerns remain:
- Even we avoid using error boundaries, we still need Suspense to work (for a historical reason?), so we have `dataWithoutErrors` as a "workaround".
- And that means error boundary and suspense are not _paired_.

packages/waku/src/router/client.ts
```diff
@@ -464,14 +467,49 @@ const handleScroll = () => {
   });
 };
 
+const overridePathAndQuery = async (
```
dai-shi, line 470: This approach fails the use case of `unstable_rerenderRoute` which wasn't caught by tests.
#1406 is the fix for this, reverting the change in this PR.

PR #1399, opened 2025-05-04
fix: make scroll reset happen only after navigation

Raicuparta:
I noticed that currently scroll resetting happens immediately when a route change is requested, even even though it can take a while for the route change to be finalized. This is a bit jarring since the user will first see two events when clicking a Link component: first the page scrolls up, then the next page loads. This change makes it so the scroll reset only happens on the next render, so you just see a single change.

Not sure if this is the best way to fix it. After some tinkering, I think using the ref+effect combo is the way to go. I also moved handleScroll to be fully inside the effect because what happens inside handleScroll is important for stuff like figuring out what kind of effect to use, I think it's safer to have less indirection there.

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -453,17 +453,6 @@ class CustomErrorHandler extends Component<
 
 const getRouteSlotId = (path: string) => 'route:' + decodeURIComponent(path);
 
-const handleScroll = () => {
-  const { hash } = window.location;
-  const { state } = window.history;
-  const element = hash && document.getElementById(hash.slice(1));
-  window.scrollTo({
-    left: 0,
-    top: element ? element.getBoundingClientRect().top + window.scrollY : 0,
-    behavior: state?.waku_new_path ? 'instant' : 'auto',
-  });
-};
-
 const InnerRouter = ({
   routerData,
   initialRoute,
@@ -481,6 +470,7 @@ const InnerRouter = ({
     ...initialRoute,
     hash: '',
   }));
+  const shouldScroll = useRef(false);
   // Update the route post-load to include the current hash.
   useEffect(() => {
     setRoute((prev) => {
@@ -503,9 +493,7 @@ const InnerRouter = ({
         const rscParams = createRscParams(route.query);
         refetch(rscPath, rscParams);
       }
-      if (options.shouldScroll) {
-        handleScroll();
-      }
+      shouldScroll.current = options.shouldScroll;
       setRoute(route);
     },
     [refetch, staticPathSet],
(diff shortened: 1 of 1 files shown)


Raicuparta:
Oh cool we got auto previews being deployed. It's really easy to see what I mean on waku.gg too:
1. Disable cache in devtools.
2. Go To https://waku.gg/
3. Scroll down.
4. Enable throttling in devtools (I throttled to 50Kbps).
5. Click "Blog" on the navigation sidebar.
6. See how it first snaps up, then after a second loads the next page.

Then do the same on the Vercel preview linked above.

dai-shi:
> Not sure if this is the best way to fix it.

Hm, it doesn't seem very good. We are working on something (#1391, #1396), so let's discuss later about the actual fix after that.

In the meantime, adding an e2e test will be a great help. We'll be refactoring a lot in the future, and adding tests is important.

Raicuparta:
I'll try to look into it eventually, but might be a few weeks before I have the time.
I presume this means using `router.unstable_events.on('complete', etc)` and doing the scroll there.

dai-shi:
This will be an internal fix, so we shouldn't use `unstable_events`. I think there should be a better way.

Alright, @tylersayshi or myself might take it, but we have some issues to work on and it might take some weeks. So, at this point, we'll wait for you.

Raicuparta:
Actually I don't think it would have worked anyway since the "complete" event still seems to be called before rendering the new route? I'd need to look more into the code to understand it better. But yeah I'll see what could be the best approach.

tylersayshi:
> This will be an internal fix, so we shouldn't use unstable_events. I think there should be a better way.

Agreed, this logic should be internal to the router code. No need for using `unstable_events`. I can look into a proper fix this one of the next few days.

@Raicuparta if you have time to try sooner I think it might just be moving this set of lines to after line 579: https://github.com/wakujs/waku/blob/main/packages/waku/src/router/client.ts#L574-L576

rmarscher:
I didn't see this PR before, but I reported related issues in #1437. I just added an e2e test to continue this PR as part of [#1459](https://github.com/wakujs/waku/pull/1459/commits/f096bc0e50a310017cd07dfb5e7261b0a0adb9e3).

rmarscher:
> 4\. Enable throttling in devtools (I throttled to 50Kbps).

If we add this to our ['static long suspense'](https://github.com/wakujs/waku/blob/20a4259/e2e/create-pages.spec.ts#L214) test, we could show similar behavior with slow network requests that don't use a dynamic server or suspense.

```
const client = await page.context().newCDPSession(page);
await client.send('Network.enable');
await client.send('Network.emulateNetworkConditions', {
  offline: false,
  downloadThroughput: 50 * 1024 / 8, // 50Kbps
  uploadThroughput: 50 * 1024 / 8,
  latency: 500 // 500ms latency
});
```

https://playwright.dev/docs/api/class-cdpsession

I might have time early next week to continue working on this and #1459 but I need to put it down for the moment.

PR #1402, opened 2025-05-08
chore: add node 24 in tests

dai-shi:


fivemeyestoreserver:
The performance optimizations here are quite clever. Nice engineering work!

PR #1403, opened 2025-05-09
refactor(router): use getConfig for order instead of filename

tylersayshi:
convention is: 
```ts
type PagePartFilename = `_part${string}.tsx`
// order must be defined in `getConfig`
```


dai-shi (approved):
This looks clean and straightforward!

PR #1405, opened 2025-05-09
fix(tsconfig): isolate modules

### LINKED ISSUE #1404: import magic?
Can you please explain to me, how does one file know about type in another file without imports? Is it some magic which waku implements, or did i mess up my linter or something. Im using latest waku version.

Image

Image

tylersayshi:
this adds sane defaults for our example tsconfig
ref: https://www.totaltypescript.com/tsconfig-cheat-sheet

fixes #1404


dai-shi:
It doesn't update all examples/*/tsconfig.json. Why so?

tylersayshi:
> It doesn't update all examples/*/tsconfig.json. Why so?

ah, i'll add it to all of them good catch.

dai-shi (approved):
LGTM

Issue #756, opened 2024-06-20
Sometimes app crashes with "The render was aborted by the server without a reason" (Unable to enqueue)


emilgpa:
I don't know why it happens. I can't isolate it to reproduce this error. It just happens sometimes. I put the log:

```
Error: The render was aborted by the server without a reason.
    at abort (/Users/emgp/Projects/example_app/clients/browser/node_modules/.pnpm/react-dom@19.0.0-rc-fb9a90fa48-20240614_react@19.0.0-rc-fb9a90fa48-20240614/node_modules/react-dom/cjs/react-dom-server.edge.development.js:7074:17)
    at Object.cancel (/Users/emgp/Projects/example_app/clients/browser/node_modules/.pnpm/react-dom@19.0.0-rc-fb9a90fa48-20240614_react@19.0.0-rc-fb9a90fa48-20240614/node_modules/react-dom/cjs/react-dom-server.edge.development.js:8620:21)
    at readableByteStreamControllerCancelSteps (node:internal/webstreams/readablestream:3022:37)
    at [kCancel] (node:internal/webstreams/readablestream:1166:12)
    at ensureIsPromise (node:internal/webstreams/util:192:19)
    at readableStreamCancel (node:internal/webstreams/readablestream:1887:5)
    at node:internal/webstreams/readablestream:1439:15
    at complete (node:internal/webstreams/readablestream:1323:9)
    at shutdownWithAnAction (node:internal/webstreams/readablestream:1319:5)
    at node:internal/webstreams/readablestream:1438:14
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
node:internal/webstreams/transformstream:488
    throw new ERR_INVALID_STATE.TypeError('Unable to enqueue');
    ^

TypeError: Invalid state: Unable to enqueue
    at new NodeError (node:internal/errors:405:5)
    at transformStreamDefaultControllerEnqueue (node:internal/webstreams/transformstream:488:11)
    at TransformStreamDefaultController.enqueue (node:internal/webstreams/transformstream:301:5)
    at Timeout._onTimeout (file:///Users/emgp/Projects/example_app/clients/browser/node_modules/.pnpm/waku@0.21.0-alpha.2_react-dom@19.0.0-rc-fb9a90fa48-20240614_react-server-dom-webpack@19.0.0-r_loaivdwzpiimz52uywzu27muau/node_modules/waku/dist/lib/renderers/html-renderer.js:109:32)
[...]

PR #1407, opened 2025-05-11
fix(cli/dev only): workaround to ignore "TypeError [ERR_INVALID_STATE]: Invalid state: Unable to enqueue"

t6adev:
This PR introduces a workaround for issue #756.
I've also been facing the same issue often.

# What's happening

Please look at 00:35 to observe the error.
I just added new lines to the file, then restarted the process.

https://github.com/user-attachments/assets/bf6b038c-10b9-435a-b85c-97651fc7b24e

# Workaround

It seems that the error comes from TransformStream, but all of the code seems to have no issue, and it's too hard to investigate.
However, as you can see in the above video, the error might be ignorable.
My suggestion is to catch the error and ignore it in only the dev CLI.

https://github.com/user-attachments/assets/4964197e-7594-49ab-a71f-c8398f25e168

You can see the result at 00:42.

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -3,6 +3,7 @@ import { existsSync, readFileSync } from 'node:fs';
 import { pathToFileURL } from 'node:url';
 import { parseArgs } from 'node:util';
 import { createRequire } from 'node:module';
+import process from 'node:process';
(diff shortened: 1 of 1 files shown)


dai-shi (approved):
LGTM
Thanks for your contribution!

PR #1436, opened 2025-05-22
chore: use React.TransitionFunction type for unstable_startTransition def

rmarscher:
These functions can be async. The current `() => void` type shows an error if we pass an async function to `unstable_startTransition`.


dai-shi (approved):
Nice catch!

Issue #1437, opened 2025-05-22
Routing completes before new page component has sent anything


rmarscher:
I have an example project to test behavior when changing routes for dynamic RSC pages that take a long time to start sending data. It has a [static layout with `<Suspense>` wrapping the {children}](https://github.com/rmarscher/waku-long-ttfb/blob/main/src/pages/_layout.tsx#L20).

It is working as expected in v0.21.20. The suspense boundary preserves it's content while navigating and the url location and layout does not update to the new route until the server component has retrieved it's initial data and starts streaming content. https://waku-long-ttfb.martiandev.workers.dev/

When I update this project to v0.23.1, the route changes immediately and the layout shows the suspense fallback again until the new page starts sending content. https://waku-long-ttfb-updated.martiandev.workers.dev/  - this is using the [waku-0.23.1 branch](https://github.com/rmarscher/waku-long-ttfb/tree/waku-0.23.1).

I thought this was because the startTransition was removed from Link when there is no unstable_pending component. I thought that add a transition back should get it to have the same behavior as before - passing an [unstable_pending prop](https://github.com/rmarscher/waku-long-ttfb/pull/1/files#diff-8387d64f5b5e50e1e386a42e75a377d0d8bf6d40a149697aca42cfeda4c8c206R44) will get it to use startTransition like it did in older versions of waku. However, now it seems to show the last rendered version of the new page while it is loading if that page has a cached version. Almost like SWR behavior. In the https://waku-long-ttfb-updated.martiandev.workers.dev/  example - the page links in the header have `unstable_pending="..."` while the link between about and index in the body of the page do not.

It seems that the layout's children element slot is getting updated to the new page right away and maybe it didn’t used to do that? So now, when React startTransition is used, it shows the previously rendered version of the new page instead of continuing to show the previously rendered version of the current page?

I think this is unintended behavior. The routing completes before the server has sent any content for the new page. The new routing "complete" event fires and the promises accessible from the unstable_enhanceCreateData Router prop all resolve as soon as the server returns a response -- but that is well before the new page RSC has sent anything. Thanks a lot for helping to look into this.

dai-shi:
I'd need to have a closer look but I think I know what's happening. I don't have an immediate answer, but we need to find a solution anyway.
Meanwhile, would you be able to add a failing test?

rmarscher:
Yes, thank you. I will figure out a test case.

dai-shi:
@rmarscher Can you try #1441?

[...]

PR #1441, opened 2025-05-28
wip: continue transition with waku/router Link

dai-shi:
for #1437 

Tasks:
- [ ] Add failing test
- [ ] Fix

### DIFF
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -163,6 +163,15 @@ for (const mode of ['DEV', 'PRD'] as const) {
       ).toBeVisible();
     });
 
+    // https://github.com/wakujs/waku/issues/1437
+    test('static long suspense', async ({ page }) => {
+      await page.goto(`http://localhost:${port}/static-long-suspense/3`);
+      await expect(page.getByText('Long Suspense Page 3')).toBeVisible();
+      await page.click("a[href='/static-long-suspense/4']");
+      await expect(page.getByText('Loading...')).not.toBeVisible();
+      await expect(page.getByText('Long Suspense Page 4')).toBeVisible();
+    });
+
     test('api hi', async () => {
       const res = await fetch(`http://localhost:${port}/api/hi`);
       expect(res.status).toBe(200);
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -164,18 +164,82 @@ for (const mode of ['DEV', 'PRD'] as const) {
         page.getByRole('heading', { name: 'Long Suspense Page 1' }),
       ).toBeVisible();
       await page.click("a[href='/long-suspense/2']");
-      await expect(page.getByTestId('long-suspense-pending')).toHaveText(
-        'Pending...',
+      await page.waitForFunction(
+        () => {
+          const pathname = window.location.pathname;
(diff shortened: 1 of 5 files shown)


dai-shi:
@rmarscher The tests pass even without the change (= `main`). Don't we need a failing test?

rmarscher:
> @rmarscher The tests pass even without the change (= `main`). Don't we need a failing test?

OK. I assumed that it would fail. I'll try reverting the changes in #1441 and see if I can reproduce #1437.

rmarscher:
Sorry... I thought there were a few actual code changes in this branch. 🤔  I'll investigate more.

dai-shi:
1498b44 reverted the first change.

rmarscher:
> [1498b44](https://github.com/wakujs/waku/commit/1498b44cc66e7e72ed81be4963d6a8c6a59ec357) reverted the first change.

Thanks. I understand how to create a failing test now. I'm working on it.

dai-shi:
@rmarscher Now, the test doesn't pass with the fix.

rmarscher:
~~Yes, adding a client component in LongSuspenseLayout demonstrates more issues. In the example, I added it wrapping the entire layout, but it also happens if the RouterEvents are added as a sibling of the Suspense component - so this issue isn't limited to layouts with an outer client component.~~ edit: I was incorrect. The issues happen without client components too.

I think the main issue is that the elements promise resolves before the page component has sent content, but I'm not exactly sure why. ~~Maybe client components and server components are merged and the promise resolves because the client components are already loaded?~~ edit: no

Here's a screen recording running the create-pages e2e fixture manually with the rsc promise delay increased to 3500 to make it easier to watch the network requests:

https://github.com/user-attachments/assets/31c9d704-064c-4ebe-9718-ce133e45223e

edit: without the RouterEvents component, the early update of the url location and the showing of cached content when clicking back to /long-suspense/2 from /long-suspense/3 still occurs.

dai-shi:
> without the RouterEvents component, the early update of the url location and the showing of cached content when clicking back to /long-suspense/2 from /long-suspense/3 still occurs.

In this PR, I would like to fix the behavior unrelated to `unstable_events`, which is unstable. It would be nice if you could modify the test without `unstable_events` as we should avoid changing the test when we make some changes with `unstable_events`.

dai-shi:
> Here's a screen recording running the create-pages e2e fixture manually with the rsc promise delay increased to 3500 to make it easier to watch the network requests:

I'm not sure how to interpret the recording. Can you elaborate please?

rmarscher:
The recording was trying to show in the network tab that no data has been sent from the server yet, but the changeRoute promise has completed.

This PR was getting confusing for me and I think there are three separate issues that I am reporting. I opened #1458 to replace this PR. I'll let you close this issue @dai-shi if you agree we should move over to #1458. I think we can merge that and then discuss part 2 and 3 of what I noticed in #1437.

dai-shi:
> The recording was trying to show in the network tab that no data has been sent from the server yet, but the changeRoute promise has completed.

Does anything change on screen while there are no data? I'm not sure how to see the promise has completed besides "Loading..." disappears.

dai-shi:
> I'll let you close this issue

Sounds good.

PR #1444, opened 2025-05-29
fix: patch react refresh properly

sapphi-red:
While the current patch code works fine as `this` is `undefined` for current Vite, Vite 7+ will pass the context (https://github.com/vitejs/vite/pull/19936) and the current code doesn't work anymore.
This PR changes the code to call the hook function with the original `this`.

This PR should fix the ecosystem-ci failure.
https://github.com/vitejs/vite-ecosystem-ci/actions/runs/15291891294/job/43012757791#step:8:665


dai-shi:
> BTW, this hacky patch should go away, and I don't even know if it's still required.

I just confirmed that removing the patch plugin makes our e2e test fail.

PR #1450, opened 2025-05-31
wip: e2e tests for #1437 and #1441

rmarscher:
This updated fixture and tests demonstrate the intended behavior for pages that take a long time to load.

When I run the e2e locally, the "long response" test is failing in PRD. I'm not sure why. I built the create-pages fixture and run it with `pnpm run build && pnpm run start` and the behavior at https://localhost:8080/long-suspense/1 is correct.


rmarscher:
I'm pretty sure we can merge this and #1441 to fix #1437. Do you see any issues?

dai-shi (approved):
Thanks. Let me double check.

PR #1451, opened 2025-05-31
Add E2E test for wildcard route interception API route request

hamlim:


hamlim:
Figured I'd try to implement a fix also, pushed a change here: https://github.com/wakujs/waku/pull/1451/commits/0805f30d47dd8809b4c8327f2ac2db927ea49dc0 to sort wildcard path configs to the end of the array so we match on them last.

Not sure if this is a naive fix (and maybe leads to other issues), but it seems to make the most sense in that we should probably match explicit path specs first before falling back on wildcard route segments.

cc @tylersayshi

dai-shi (changes requested):
> Not sure if this is a naive fix (and maybe leads to other issues), but it seems to make the most sense in that we should probably match explicit path specs first before falling back on wildcard route segments.

It might make sense, but I would like to leave this responsibility for create-pages.ts. Can you modify create-pages.ts instead of define-router.ts?

hamlim:
@dai-shi got some time to take a look at the code again today - I don't think this can be solved for within `create-pages.ts` as the code is today. There's no exposed hook within `defineRouter` that allows `create-pages` to sort the complete list of path specs or to choose the matching path config.

One possible option could be to have `defineRouter` allow for integrations to pass in a custom `getPathConfigItem` function (keep the current implementation as a default parameter, maybe tweak it slightly to call `getPathConfigItem` with the `pathConfig` and the existing `pathname` argument), then `create-pages` could pass in a custom method to sort the list of path configs and then return a `.find` for the path. Would that change make more sense?

After thinking about my naive solve here a bit more, an even simpler solve could be to move the [api route spread here](https://github.com/wakujs/waku/blob/79efb5c6cd48ad03feae40cac3543055a404cd36/packages/waku/src/router/define-router.ts#L233-L243) up above the [page route spread](https://github.com/wakujs/waku/blob/79efb5c6cd48ad03feae40cac3543055a404cd36/packages/waku/src/router/define-router.ts#L204-L232).

dai-shi:
Ah I see what you mean. I think we should make some changes in define-router.ts. I can work on it later but not soon.

dai-shi:
> I can work on it later but not soon.

I haven't started. Give me time.

dai-shi:
> I haven't started. Give me time.

👉 #1487

dai-shi (changes requested):
@hamlim I'm sorry, but you misunderstand my request. The fix should be in create-pages.ts. Do not modify define-router.ts.

hamlim:
Sorry about that, updated the branch to apply the fix in `create-pages` instead!

If `defineRouter` will be more of a stable public API in the long run - we may want to consider documenting this quirk to let folks know that this issue may happen if they're not using the `create-pages` setup - but that can be done in a future change

dai-shi:
Record of flaky tests:

https://github.com/wakujs/waku/actions/runs/15853235995/job/44691999735?pr=1451

> Error:   1) [chromium-prd] › e2e/ssg-performance.spec.ts:17:3 › high volume static site generation › build and verify 
>     Test timeout of 60000ms exceeded.

https://github.com/wakujs/waku/actions/runs/15853235995/job/44691999777?pr=1451

> Error:   1) [chromium-prd] › e2e\ssg-performance.spec.ts:17:3 › high volume static site generation › build and verify 
>     Test timeout of 60000ms exceeded.

dai-shi (approved):
LGTM

@tylersayshi Can you have another review?

packages/waku/src/router/create-pages.ts
```diff
@@ -753,7 +753,19 @@ export const createPages = <
...
+      return (
+        [...routeConfigs, ...apiConfigs]
+          // Sort wildcard routes to the end - we want to match the most specific routes first
+          .sort((pathConfigA, pathConfigB) => {
+            if (pathConfigA.path.find((spec) => spec.type === 'wildcard')) {
+              return 1;
+            }
+            if (pathConfigB.path.find((spec) => spec.type === 'wildcard')) {
+              return -1;
+            }
```
tylersayshi:
```suggestion
            if (pathConfigA.path.at(-1)?.type === 'wildcard') {
              return 1;
            }
            if (pathConfigB.path.at(-1)?.type === 'wildcard') {
              return -1;
            }
```
this would work too, right?

hamlim: I realized that there was a possible edge case with the sorting logic (api wildcard routes ending up lower than standard api routes).

I've added another test case and updated the sorting logic - it's a bit more convoluted but I think it mostly makes sense.

fyi @dai-shi

packages/waku/src/router/create-pages.ts
```diff
@@ -753,7 +753,38 @@ export const createPages = <
...
+            let priorityA = 0;
+            if (pathConfigA.type === 'api') {
+              if (hasWildcardA) {
+                priorityA = 2;
+              } else {
+                priorityA = 1;
+              }
+            } else if (hasWildcardA) {
+              priorityA = 3;
+            }
```
dai-shi: Make it a function, incl. L760, and reuse the logic. It will be more readable.

hamlim: This commit abstracts this logic away into a function: https://github.com/wakujs/waku/pull/1451/commits/a4a6e63a1a754f481d68eccd4026767d91d14f19

packages/waku/src/router/create-pages.ts
```diff
@@ -753,7 +753,27 @@ export const createPages = <
...
+
+      return (
+        [...routeConfigs, ...apiConfigs]
+          // Sort routes by priority: "standard routes" -> api routes -> api wildcard routes -> standard wildcard routes
+          .sort((pathConfigA, pathConfigB) => {
+            const priorityA = getRoutePriority(pathConfigA);
+            const priorityB = getRoutePriority(pathConfigB);
+
+            return priorityA - priorityB;
+          })
```
dai-shi: Looks good.

My preference would be:
`sort((configA, configB) => getRoutePriority(configA) - getRoutePriority(configB))`

hamlim:
Thanks for all the reviews and feedback!

PR #1454, opened 2025-06-03
chore: failing e2e tests for #1437

rmarscher:


dai-shi (commented):
Does that mean that the test doesn't fail if we don't use `router.unstable_events`?
I think the behavior is unrelated to `router.unstable_events`, but maybe it's too hard to reproduce the behavior within the test.

Issue #1446, opened 2025-05-30
Support for forwarding/merging `ref` on Link component


hamlim:
I'm attempting to use a Waku `Link` component as a tooltip target for a shadcn/ui tooltip (radix tooltip), however it doesn't seem to work as expected.

Replacing the `Link` component with an anchor (`a`) seems to work as expected (the tooltip content will show on hover).

I believe this is because Radix expects a ref to be forwarded to the element - but the `Link` component manages it's own ref internally.

I'll try to make a minimal reproduction!

dai-shi:
> This will make is so that both `props.ref` and our internal ref can both be assigned from the underlying `<a>` tag.

Does it work for ref callbacks too?

PR #1455, opened 2025-06-03
feat(Link): Support forwarding ref on Link component

hamlim:
## Context:

This change should support forwarding a provided `ref` down to the `a` element rendered by the `Link` component, handling "merging" the ref with the internal managed ref using a strategy from [`@ds-pack/use-refs`](https://www.npmjs.com/package/@ds-pack/use-refs).

I couldn't find a good test (e2e or package `tests`) to validate this behavior - I'm open to suggestions! I was considering maybe a `e2e/fixtures/third-party` to test Waku with various third party dependencies (radix would be the example for this PR)

Closes: #1446

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -216,6 +216,30 @@ export function useRouter() {
   };
 }
 
+function useSharedRef<T>(
+  ref:
+    | React.RefObject<T | null>
+    | ((instance: T | null) => void)
+    | null
+    | undefined,
+): React.RefObject<T | null> {
+  const managedRef = useRef<T>(null);
+
+  return {
+    set current(value: T | null) {
+      managedRef.current = value;
+      if (typeof ref === 'function') {
+        ref(value);
+      } else if (ref) {
+        ref.current = value;
+      }
+    },
+    get current(): T | null {
+      return managedRef.current;
+    },
+  };
+}
+
 export type LinkProps = {
   to: InferredPaths;
   children: ReactNode;
@@ -231,6 +255,11 @@ export type LinkProps = {
   unstable_prefetchOnEnter?: boolean;
   unstable_prefetchOnView?: boolean;
   unstable_startTransition?: ((fn: TransitionFunction) => void) | undefined;
+  ref?:
+    | React.RefObject<HTMLAnchorElement>
+    | ((instance: HTMLAnchorElement | null) => void)
+    | null
+    | undefined;
 } & Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'href'>;
 
 export function Link({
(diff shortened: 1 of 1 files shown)


packages/waku/src/router/client.ts
```diff
@@ -216,6 +216,30 @@ export function useRouter() {
...
+    | null
+    | undefined,
+): React.RefObject<T | null> {
+  const managedRef = useRef<T>(null);
+
+  return {
+    set current(value: T | null) {
+      managedRef.current = value;
+      if (typeof ref === 'function') {
+        ref(value);
```
dai-shi: Ref callback may return a cleanup function. We should handle it too.

dai-shi (commented):
Also, please check CI errors.

packages/waku/src/router/client.ts
hamlim: This change should handle cleanup functions being returned by callback refs: https://github.com/wakujs/waku/pull/1455/commits/16c9015f07f828d1d8856ebb8f94d8451ceb911d

Unfortunately I couldn't find a way to preserve the return type of `useSharedRef` since `setters` can't return a value - so I changed it to look a bit like `useState`, returning the managed ref and a function to set the ref.

tylersayshi:
lgtm thanks!

packages/waku/src/router/client.ts
```diff
@@ -231,6 +255,11 @@ export type LinkProps = {
   unstable_prefetchOnEnter?: boolean;
   unstable_prefetchOnView?: boolean;
   unstable_startTransition?: ((fn: TransitionFunction) => void) | undefined;
+  ref?:
+    | RefObject<HTMLAnchorElement>
+    | ((instance: HTMLAnchorElement | null) => void)
+    | null
+    | undefined;
```
dai-shi:
```suggestion
  ref?: Ref<HTMLAnchorElement> | undefined;
```

packages/waku/src/router/client.ts
```diff
@@ -216,6 +217,29 @@ export function useRouter() {
...
+        const callbackRefResult = ref(node);
+        // if the result of the ref callback is a function, return it
+        // this handles the case where the ref callback returns a cleanup function
+        if (typeof callbackRefResult === 'function') {
+          return callbackRefResult;
+        }
+      } else if (ref) {
+        ref.current = node;
+      }
+    },
```
dai-shi: Can it be something like this?
```js
function handleRef(node) {
  managedRef.current = node;
  const isRefCallback = typeof ref === 'function';
  let cleanup;
  if (isRefCallback) {
    cleanup = ref(node);
  } else {
    ref.current = node;
  };
  return () => {
    managedRef.current = null;
    if (isRefCallback) {
      if (cleanup) {
        cleanup();
      } else {
        ref(null);
      }
    } else {
      ref.current = null;
    }
  };
}
```

dai-shi (commented):
Thanks for working on it. Here's the first review.

packages/waku/src/router/client.ts
dai-shi: Do we need to use `useCallback` for `handleRef`?

dai-shi (approved):
LGTM!

PR #1459, opened 2025-06-05
wip: update url after route change

rmarscher:
For #1437.


rmarscher:
Why does [9791f07](https://github.com/wakujs/waku/pull/1459/commits/9791f07d40d9ade473049b29976d787a670b3790) cause e2e tests like "broken-links: static server > redirect" to fail? Is there some behavior that depends on the history update?

I commented out the [new e2e tests for scroll position and url updating](https://github.com/wakujs/waku/pull/1459/commits/5fba52c6b3ce7d3f9bbee3f3e9c5cf464125f31f) so we can focus on that issue. Then maybe we can figure out how to fix the refetch promise in changeRoute and re-enable these e2e tests.

dai-shi:
@rmarscher Please merge main.

rmarscher:
I think I got a little closer by using `finally` to make sure it always reaches the code that updates the url. However with a client-side redirect, it seems to switch to the final url and then back to the originally requested url.

So with the broken-links example, running it with a dev server, click on "Correct redirect" and the url history will go to /exists and then /redirect. It would be nice to know that a redirect changed it somehow and skip updating.

I added RSC redirects in 4265f05 to be able to test it with a dev server. The RSC redirect was only in the public/serve.json.

dai-shi:
If things around redirect handling seem too hard to implement or too complicated to maintain, I think our idea should be to reduce the requirement. For example, my understanding around client redirect was not enough when we added the support for redirect config in serve.json, which is technically outside the framework and which depends on rsc path convention. Maybe, it's time to revisit. (and, should do before v1-alpha.)

dai-shi:
@rmarscher What's the current status?

rmarscher:
@dai-shi I think all of the tests are passing except one. Router events seem to be firing twice on the first navigation after loading a dynamic page. They do not do that on subsequent page loads or if the static page is loaded first.

cc/ @tylersayshi 

```txt
e2e/use-router.spec.ts:119:5 › useRouter › calls route change event handlers › on dynamic pages 

    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 2

      Array [
        "Route change started",
    +   "Route change started",
    +   "Route change completed",
        "Route change completed",
      ]
```

rmarscher:
> `changeRoute` is called twice on first navigation to /static after loading /dynamic in the `use-router` fixture.

rmarscher:
> `changeRoute` is called twice on first navigation to /static after loading /dynamic in the `use-router` fixture.

I found that this is because of the "locationListeners". After the RSC fetch returns, it is looking at the routeData and then running locationListeners, if necessary. I think this is to support server-side redirects or just an extra check to make sure that the window.location and current page elements are in sync?

Before this PR, the location was updated before this code was hit so the check to see if the location update passed. The one labeled - "FIXME this check here seems ad-hoc" ;-) 

Now it's failing so we'll need another way to determine if we are already in the middle of a route change.

rmarscher:
It's looking good now @dai-shi @tylersayshi. Maybe just a flaky windows test failure.

packages/waku/src/router/client.ts
```diff
@@ -157,18 +157,27 @@ export function useRouter() {
...
-      });
+      const currentPath = window.location.pathname;
+      const newPath = url.pathname !== currentPath;
+      try {
+        await changeRoute(parseRoute(url), {
+          shouldScroll: options?.scroll ?? newPath,
+        });
+      } catch (err) {
+        console.error('Error while navigating to new route:', err);
+        throw err;
```
dai-shi: I would like to avoid `console.error` if we throw it.

packages/waku/src/router/client.ts
```diff
@@ -157,18 +157,27 @@ export function useRouter() {
...
+        if (window.location.pathname === currentPath) {
+          window.history.pushState(
+            {
+              ...window.history.state,
+              waku_new_path: newPath,
+            },
+            '',
+            url,
+          );
+        }
```
dai-shi: Are you sure that we should pushState even if changeRoute throws an error?
I'm pretty sure this is intentional, but let me double check just in case.

dai-shi (commented):
@rmarscher Thanks for working on this! It looks pretty clean. Just left a few comments.

@tylersayshi Would you like to give a review with another eye?

packages/waku/src/router/client.ts
rmarscher: I think this was accidentally left from debugging. I'll remove the catch.

packages/waku/src/router/client.ts
rmarscher: IIRC, there were failing tests if it doesn't always update to the requested url.

I think pushState on error is the expected behavior. The new page would show that you hit an error trying to load that page. Browser back button would take you back to the previously working page.

rmarscher: I pushed a commit that removes this try / catch / finally from the `push` method. It looks like I did it for consistency with how other routing navigation methods work - but I understand that `push` is a bit different because it is an explicit API.

rmarscher: Yup. I was wrong. This still passes all tests.

e2e/fixtures/broken-links/src/redirects.ts
```diff
@@ -1,22 +1,23 @@
...
-      };
-      break;
-    default:
-      return await next();
+  const redirects = {
+    '/redirect': '/exists',
+    '/RSC/R/redirect.txt': '/RSC/R/exists.txt',
+    '/broken-redirect': '/broken',
+    '/RSC/R/broken-redirect.txt': '/RSC/R/broken.txt',
+  };
```
tylersayshi: should this be declared outside the function scope? or does it not really matter?

packages/waku/src/router/client.ts
```diff
@@ -482,6 +491,18 @@ const Redirect = ({ to, reset }: { to: string; reset: () => void }) => {
       })
```
tylersayshi, line 498: Just curious: Is the setTimeout hack still needed with the new route change behavior?

packages/waku/src/router/client.ts
```diff
@@ -805,27 +826,28 @@ const InnerRouter = ({ initialRoute }: { initialRoute: RouteProps }) => {
...
+            window.history.pushState(
+              {
+                ...window.history.state,
+                waku_new_path: url.pathname !== window.location.pathname,
+              },
+              '',
+              url,
+            );
+          }
         }
```
tylersayshi: why not use `.finally` here like the other use?

tylersayshi (commented):
rest lgtm 👍

packages/waku/src/router/client.ts
```diff
@@ -337,19 +343,29 @@ export function Link({
...
-        });
+        const currentPath = window.location.pathname;
+        const newPath = url.pathname !== currentPath;
+        try {
+          await changeRoute(route, {
+            shouldScroll: scroll ?? newPath,
+            unstable_startTransition: startTransitionFn,
+          });
+        } catch (err) {
+          console.error('Error while navigating to new route:', err);
```
dai-shi: We have console.error here too.

dai-shi (commented):
@rmarscher Can you clarify again when we need `.finally {}` and when not?

e2e/fixtures/broken-links/src/redirects.ts
rmarscher: 🤷  Yeah, probably. technically the `/RSC/R` prefix and `.txt` extensions should probably be imported from waku constants or use a function that converts a normal path to an RSC path.

packages/waku/src/router/client.ts
rmarscher: I just tried and there are e2e tests that break if I remove it.

packages/waku/src/router/client.ts
rmarscher: Good question. I should have not switched it to an async def. I'll fix.

rmarscher: Oh... it's so the error can fall through. I think we need it.

rmarscher:
> @rmarscher Can you clarify again when we need `.finally {}` and when not?

Now that `window.history.pushState` is called after awaiting `changeRoute()`, we need to use `finally` to make sure we update the url even if changeRoute throws an exception.

It is not necessarily needed for the `useRouter()` APIs because the developer can catch the error from changeRoute and handle it on their own.

The `try { await changeRoute(); } finally { pushState }` pattern is used for useTransition functions (`<Link>` ).

The `changeRoute().catch(() => { console.error }).finally(() => { pushState })` pattern is used for useEffect and startTransition functions (`<Redirect>`, `locationListeners` ).

`changeRoute` is called without pushState in the popstate window event listener - because it's reacting to the url already being updated. Similarly, `reload()` and `<NotFound>` do not call pushState.

I think that covers all of the uses of changeRoute in waku/router/client.

rmarscher:
Hmm... the locationListeners use `https://react.dev/reference/react/startTransition` - they probably should use the `.finally()` syntax instead of `async`. I'm going to change that.

e2e/fixtures/broken-links/src/redirects.ts
rmarscher: @dai-shi what do you think about exposing `encodeRscPath()` for this?

rmarscher:
OK. I think I have addressed all of the comments. Thanks for the reviews.

rmarscher:
> Thanks for working on it. It's a little bit unfortunate to see things become complicated, but can't be helped. Any ideas to fix/mitigate later are welcome.

Maybe updating history should move into changeRoute. 

`changeRoute(url, { history: 'push' });`

The history option could be `'push'`, `'replace'` or `false`.

I'm not sure about `unstable_startTransition` in changeRoute options. I'm not sure a transition function should happen inside changeRoute. The place that calls changeRoute should wrap it in a transition - which we are doing in Link onClick. And when devs use the `useRoute()` APIs, they can wrap those in a useTransition hook if they want.

dai-shi:
> I'm not sure about unstable_startTransition in changeRoute options.

Maybe we don't need it in the future. One motivation was to support view transition api, but it will be supported by React natively.

PR #1468, opened 2025-06-11
fix: remove `Provider` suffix on react context

himself65:
See https://react.dev/reference/react/createContext#provider


dai-shi (commented):
I don't think it's super important, but it should be fine as we only support React 19+.
Let's merge this after #1469 is resolved.

Issue #1464, opened 2025-06-09
Hot reload doesn't work with 2 Waku instance


ugurrdemirel:
Hello, I am working on a full stack project which is a monorepo. I found that when I run two Waku projects same time, second project says `WebSocket server error: Port 24678 is already in use` and hot reload doesn't working. I think this should be handled by framework level.

dai-shi:
Good point. I wonder if there's a way to customize the websocket port. Is anyone familiar with it?

hamlim:
I think you can customize the `server.hmr.port` option within the vite config, maybe assigning something like `'0'` will let vite/vites server automatically assign an available port?

rmarscher:
I'm not sure that vite will automatically assign a port but that does seem to be the correct setting. https://github.com/vitejs/vite/issues/14328#issuecomment-2153152319

dai-shi:
So, do we need to find a free port on our end?

hamlim:
Opened an initial PR here that fixes the issue for me locally: https://github.com/wakujs/waku/pull/1470

dai-shi:
I'll check it later. If it works, we can tweak our tests. (And, before that, we need to fix e2e tests for windows. #1469 )

Issue #1469, opened 2025-06-11
e2e CI crash on Windows


himself65:
right now I don't have a real windows machine so it's hard to debug for me.

But I think there're some windows setup issue on playwright. Because run waku build / dev is working fine without playwright. Idk what happened here.

```shell
stderr: Error: Missing Root component
    at useElementsPromise_UNSTABLE (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\waku\src\minimal\client.ts:319:11)
    at Slot (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\waku\src\minimal\client.ts:347:27)
    at react-stack-bottom-frame (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:9251:18)
    at renderWithHooks (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:4898:19)
    at renderElement (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:5333:23)
    at retryNode (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:6092:31)
    at renderNodeDestructive (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:6042:11)
    at renderElement (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:5779:15)
    at retryNode (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:6092:31)
    at renderNodeDestructive (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-  ✘  1 …2e\broken-links.spec.ts:19:7 › broken-links: normal server: DEV › server side navigation › existing page (19.4s)    at renderElement (C:\Users\himself65\AppData\Local\Temp\broken-linksoSITti\node_modules\react-dom\cjs\react-dom-server.edge.development.js:5512:13)
[...]

PR #1470, opened 2025-06-11
fix: automatically pick available hmr port

hamlim:
## Details:

Fixes: #1464 

This is maybe a bit of a naive fix - but it uses the underlying OS (or an implementation of Node.js / the JS runtime in use while running Waku) to select an available port and then we specify that within the Vite `server.hmr` options for the main vite server.

I tested applying a similar change to the two other places ([here](https://github.com/wakujs/waku/blob/0f5d360550a6c1deb9b34080a264b07f95467834/packages/waku/src/lib/utils/vite-loader.ts#L11-L19), and [here](https://github.com/wakujs/waku/blob/0f5d360550a6c1deb9b34080a264b07f95467834/packages/waku/src/lib/middleware/dev-server-impl.ts#L270-L325)) where we call `createServer` from `vite` - but it seems like those had no impact and making only this change seemed to fix the issue for me.

### Testing

I tested this a bit manually, specifically I used this relatively bare bones reproduction repo: https://github.com/hamlim/waku-conflicting-hmr-port-reproduction, and then did the following:

- Compiled changes to `waku` locally
- Clone and install deps for `waku-conflicting-hmr-port-reproduction` repo (bun i)
- `rm -rf ./waku-conflicting-hmr-port-reproduction/node_modules/waku`
- `cp -r ./waku/packages/waku ./waku-conflicting-hmr-port-reproduction/node_modules/waku`
- `rm -rf ./waku-conflicting-hmr-port-reproduction/node_modules/waku/node_modules`
- `bun run dev` (at root of `waku-conflicting-hmr-port-reproduction` repo)
- Open both `localhost:3000` and `localhost:3001`
- Check both consoles to see no errors about hmr
- Make a change to one app and save the file - see only that app refresh
- Make a change to the other app and save the file - see only the other app refresh

The reason for the convoluted way to pull in the local version of the `waku` package was that I was running into issues with conflicting react packages from waku's local `node_modules/react` package (for initial testing I used a `file:` dependency as a resolution in my root package.json). I maybe could have tested with a slightly different flow - but the hacky `rm -rf` and `cp -r` setup above seemed to work fine.

### DIFF
--- a/packages/waku/src/lib/middleware/dev-server-impl.ts
+++ b/packages/waku/src/lib/middleware/dev-server-impl.ts
@@ -1,6 +1,6 @@
 import { Readable, Writable } from 'node:stream';
 import { Server } from 'node:http';
-import type { AddressInfo } from 'node:net';
+import net, { type AddressInfo } from 'node:net';
 import { AsyncLocalStorage } from 'node:async_hooks';
 import { createServer as createViteServer } from 'vite';
 import viteReact from '@vitejs/plugin-react';
--- a/packages/waku/src/lib/middleware/dev-server-impl.ts
+++ b/packages/waku/src/lib/middleware/dev-server-impl.ts
@@ -29,6 +29,7 @@ import { rscDelegatePlugin } from '../plugins/vite-plugin-rsc-delegate.js';
 import type { ClonableModuleNode, Middleware } from './types.js';
 import { fsRouterTypegenPlugin } from '../plugins/vite-plugin-fs-router-typegen.js';
 import { hackTailwindcss4Stackblitz } from '../plugins/hack-tailwindcss4-stackblitz.js';
+import type { AddressInfo } from 'node:net';
 
 // TODO there is huge room for refactoring in this file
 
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -29,6 +29,7 @@ import { rscDelegatePlugin } from '../plugins/vite-plugin-rsc-delegate.js';
 import type { ClonableModuleNode, Middleware } from './types.js';
 import { fsRouterTypegenPlugin } from '../plugins/vite-plugin-fs-router-typegen.js';
 import { hackTailwindcss4Stackblitz } from '../plugins/hack-tailwindcss4-stackblitz.js';
+import type { AddressInfo } from 'node:net';
```
himself65: I think we prefer not to directly use node.js API here

hamlim: The file already pulls in `node:stream` and `node:http` up at the top of the imports (I could probably move this import to be up next to those too)

himself65: oh I didn't see that.

packages/waku/src/lib/middleware/dev-server-impl.ts
```diff
@@ -88,6 +89,20 @@ const splitByLastEndTag = (input: string): readonly [string, string] => {
   }
 };
 
+const findAvailablePort = async () => {
```
himself65: we already have a find port util here. could u copy paste here?

https://github.com/wakujs/waku/blob/370d2c9a7e0606be1c0d9ff98ef2706616c840e3/e2e/utils.ts#L52-L60

hamlim: Ah good callout - updated this one by copying and pasting that function!

dai-shi:
Thanks guys. I'll adjust our tests once #1469 is resolved.

dai-shi:
👉 #1478

PR #1473, opened 2025-06-14
chore: move pnpm setting to `pnpm-workspace.yaml`

himself65:
reduce some npm warnings when using like `npx prettier .`


dai-shi:
LGTM

PR #1474, opened 2025-06-14
ci: do not compile code automatically

himself65:


packages/waku/package.json
```diff
@@ -80,7 +80,7 @@
     "compile": "rm -rf dist *.tsbuildinfo && pnpm run compile:code && pnpm run compile:types && cp ../../README.md .",
     "compile:code": "swc src -d dist --strip-leading-paths",
     "compile:types": "tsc --project tsconfig.json",
-    "prepare": "pnpm run compile"
+    "prepublishOnly": "pnpm run compile"
```
dai-shi, line 83: @himself65 If it was causing the problem, we should remove it entirely. It was added by a contributor to ease the dev process. It's not for publish.

himself65: OK

dai-shi: #1479

PR #1476, opened 2025-06-15
test: let `waku` auto find tcp port & enable parallelism

himself65:
Since https://github.com/wakujs/waku/pull/1470, we should allow parallelism.

### DIFF
--- a/e2e/render-type.spec.ts
+++ b/e2e/render-type.spec.ts
@@ -10,7 +10,7 @@ test.describe('render type', () => {
     'Browsers are not relevant for this test. One is enough.',
   );
   test.skip(
-    ({ mode }) => mode !== 'PRD',
+    ({ mode }) => mode === 'PRD',
     'This test is only relevant in production mode.',
   );
 
--- a/e2e/utils.ts
+++ b/e2e/utils.ts
@@ -24,7 +24,7 @@ export type TestOptions = {
 
 export async function findWakuPort(cp: ChildProcess): Promise<number> {
   return new Promise((resolve, reject) => {
-    function listener (data: unknown) {
+    function listener(data: unknown) {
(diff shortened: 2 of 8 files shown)


playwright.config.ts
```diff
@@ -69,8 +69,7 @@ const config = defineConfig<TestOptions>({
     },
   ]),
   forbidOnly: !!process.env.CI,
-  // no parallelization, otherwise the `waku` command will have race conditions
-  workers: 1,
+  workers: process.env.CI ? 1 : 3,
```
himself65, line 72: playwright recommend use worker 1 for stable runner, I will only run for local dev

e2e/utils.ts
```diff
@@ -85,7 +95,6 @@ export function debugChildProcess(cp: ChildProcess, sourceFile: string) {
       return;
     }
     info(`(${sourceFile}) stdout: ${str}`);
-    console.log(`(${sourceFile}) stdout: ${str}`);
```
himself65, line 88: just realized that there will have double log

e2e/render-type.spec.ts
```diff
@@ -10,7 +10,7 @@ test.describe('render type', () => {
     'Browsers are not relevant for this test. One is enough.',
   );
   test.skip(
-    ({ mode }) => mode !== 'PRD',
+    ({ mode }) => mode === 'PRD',
```
dai-shi, line 13: @himself65 Why this change?

dai-shi: #1479

dai-shi: oops, it was a wrong file.

dai-shi: #1486

PR #1484, opened 2025-06-17
e2e: move to pnpm submodule

himself65:


dai-shi:
What's the benefit of this?

himself65:
> What's the benefit of this?

I tought this will fix the window test? but seems like it's unrelated.

Im closing this for now, maybe this is too breaking

PR #1486, opened 2025-06-19
fix(e2e): skip conditions and etc.

dai-shi:
There was a wrong fix in #1479.

Also, fixes the problem with fixed test options.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -9,9 +9,6 @@ const config = defineConfig<TestOptions>({
   testDir: './e2e',
   fullyParallel: true,
   timeout: process.env.CI ? 120_000 : 30_000,
-  expect: {
-    timeout: process.env.CI ? 10_000 : 5_000,
-  },
   use: {
     viewport: { width: 1440, height: 800 },
     locale: 'en-US',


playwright.config.ts
```diff
@@ -77,8 +61,4 @@ const config = defineConfig<TestOptions>({
   reporter: process.env.CI ? 'github' : 'list',
 });
 
-if (process.env.CI) {
-  config.retries = 3;
-}
```
dai-shi, line 82: @himself65 We should fix flaky tests (maybe with hacky waits), rather than ignoring them.

playwright.config.ts
```diff
@@ -1,32 +1,16 @@
...
  * See https://playwright.dev/docs/test-configuration.
  */
 const config = defineConfig<TestOptions>({
   testDir: './e2e',
   fullyParallel: true,
-  timeout: process.env.CI ? 60_000 : 30_000,
-  expect: {
-    timeout: process.env.CI ? 10_000 : 5_000,
-  },
+  timeout: process.env.CI ? 120_000 : 30_000,
```
dai-shi, line 11: This is the crucial change for windows.

e2e/utils.ts
```diff
@@ -111,7 +117,7 @@ export function debugChildProcess(cp: ChildProcess, sourceFile: string) {
 }
 
 export const test = basicTest.extend<TestOptions>({
-  mode: 'DEV',
+  mode: ['DEV', { option: true }],
```
dai-shi, line 120: We didn't test `PRD` without this.

playwright.config.ts
```diff
@@ -1,32 +1,16 @@
...
   use: {
-    browserName:
-      (process.env.BROWSER as PlaywrightWorkerOptions['browserName']) ??
-      'chromium',
```
dai-shi, line 27: We didn't test firefox and webkit without this.

Issue #1448, opened 2025-05-30
Bug: Catch-all route segment matches before API route


hamlim:
If a Waku app using the pages router has a catch-all route segment, e.g. `[...notFound].tsx`, this segment will match before an API route.

Minimal reproduction: https://github.com/hamlim/waku-catch-all-route-catches-api-route

For example, if we have the following app:

```
src/
  pages/
    api/
      greet.ts # should be available under `/api/greet`
    index.tsx
    [...notFound].tsx
```

Fetching the API route (via `fetch('/api/greet')`) seems to instead respond with the html from the `[...notFound].tsx` route.

I think this is because we spread in API routes __after__ all other route segments - which can include the wildcard/catch-all route segments here:

https://github.com/wakujs/waku/blob/5e63316f710b204974e596f5c09bd81930095645/packages/waku/src/router/define-router.ts#L203-L244

and then we do a `.find` on that array to find the correct route handler here - which matches on the first found occurrence from the beginning of the array:

https://github.com/wakujs/waku/blob/5e63316f710b204974e596f5c09bd81930095645/packages/waku/src/router/define-router.ts#L250-L252

dai-shi:
@tylersayshi can take a look?

PR #1487, opened 2025-06-22
feat(define-router): combined getConfig API

dai-shi:
Technically breaking change for unstable defineRouter.

Related: #1451

### DIFF
--- a/e2e/fixtures/define-router/src/server-entry.tsx
+++ b/e2e/fixtures/define-router/src/server-entry.tsx
@@ -13,13 +13,14 @@ const PATH_PAGE: Record<string, unknown> = {
 };
 
 const router: ReturnType<typeof defineRouter> = defineRouter({
-  getRouteConfig: async () =>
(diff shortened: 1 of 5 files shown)


dai-shi:
@hamlim @tylersayshi Can you please review this?

hamlim:
Won't this change still result in the same bug/behavior? Where wildcard routes will be found in the array before API routes?

dai-shi:
This doesn't solve #1448. In that sense, this PR is just a refactor.
I would like you to modify #1451 based on this to fix the issue.

hamlim:
> This doesn't solve https://github.com/wakujs/waku/issues/1448. In that sense, this PR is just a refactor.
> I would like you to modify https://github.com/wakujs/waku/pull/1451 based on this to fix the issue.

Ah ok that makes sense, thanks!

hamlim (approved):
Changes look good to me 👍

hamlim:
I'll update my PR a bit later today 👍

PR #1490, opened 2025-06-25
BREAKING: update vite 7

dai-shi:
This is breaking as some old node.js versions are no longer supported.


packages/waku/tests/vite-plugin-fs-router-typegen.test.ts
```diff
@@ -43,6 +43,7 @@ async function runTest(
   }
   // @ts-expect-error - we're not passing the full Vite config
   await plugin.configResolved?.({ root });
+  // @ts-expect-error - FIXME
```
dai-shi: @tylersayshi Can you look into it?

PR #1498, opened 2025-06-27
e2e: add build table test

himself65:


dai-shi:
This test doesn't seem to reproduce a bug described in #1496.
We slightly changed our way to write an e2e test, so please open a new PR.

PR #1500, opened 2025-06-29
fix(docs): use absolute URL for example

yeskunall:
It’s a small fix. Found it while going through your docs.

The issue also exists in `docs/minimal-api.mdx` but it doesn’t look like it’s accessible publicly so I left it untouched. LMK if you want the URLs fixed there as well.

---

💖


dai-shi (approved):
It was intentional, but at that time, we didn't have guides in website.

PR #1502, opened 2025-06-30
fix(docs): drop experimental flag and update script name

yeskunall:
The Cloudflare example [uses](https://github.com/wakujs/waku/blob/main/examples/07_cloudflare/package.json#L18) `v4.x` of `wrangler` which by default sets the `--include-runtime` flag.

Also, the example [uses](https://github.com/wakujs/waku/blob/main/examples/07_cloudflare/package.json#L7) a script called `cf-typegen`, not `build-cf-types`.

---

💖


dai-shi:
@rmarscher Can you review?

Issue #1507, opened 2025-07-04
API catch-all route precedence


dswbx:
Let's say I have this structure:

```
src/pages
├── _layout.tsx
├── about.tsx
├── api
│   ├── [...api].ts
│   └── another.ts  <--
└── index.tsx
```

I would expect the router to resolve like this:
* `GET /api` → should be handled by `[...api].ts`
* `GET /api/another` → should be handled by `another.ts`
* `GET /api/anything-else` → should be handled by `[...api].ts`

But it routes all requests starting with `/api/*` to the catch all (except `/api` which just returns 404, and I can't create an `index.ts` like on normal page catch-all routes). 

Is this something that could be adjusted, or am I doing something wrong?

tylersayshi:
We should be able to handle this sort of prioritized behavior. You're correct to notice it's not currently handled.

We just had a similar issue focused on routing prioritization if you'd like to see how that ended up getting resolved: https://github.com/wakujs/waku/issues/1448

cc @hamlim

hamlim:
Interesting, I had thought that I accounted for this edge case in #1448 / #1451, maybe I didn't though!

hamlim:
I was able to reproduce the bug here: https://github.com/hamlim/waku-catch-all-api-route-bug

hamlim:
I wonder if there's subtly two bugs here at play:

- API route matching order between catch all and non-catch all routes
- catch-all routes don't match "index" routes
  - this one happens for regular pages for me as well, e.g. `pages/[...all].tsx` without `pages/index.tsx` results in a broken page for me in dev when visiting `/`

hamlim:
Starting with some failing E2E's here: https://github.com/wakujs/waku/pull/1508 - will try and work on finding a fix for both issues this weekend!

dai-shi:
@hamlim Thanks for taking care of it. Let me know if I need to fix the define-router part.
For the API design, @tylersayshi should be open to discuss.

dai-shi:
fixed by #1508

PR #1508, opened 2025-07-05
fix(create-pages): Fix API catch-all routes matching before standard API routes

hamlim:
### Details:

This change adds e2e test cases for both issues:

- catch-all api route incorrectly matching before standard api route #1507 
- ~~catch-all route not handling index route~~ -> moved this to it's own issue: #1518 

and also adds a fix to the first issue above by sorting the routes first and then finding the first match. The sorting logic now accounts for "specificity" of the route (e.g. number of path specs) and also if the path spec is a wildcard.


packages/waku/src/router/create-pages.ts
```diff
@@ -312,7 +326,15 @@ export const createPages = <
...
-    for (const [p, v] of apiPathMap.entries()) {
+    const apiConfigEntries = Array.from(apiPathMap.entries()).sort(
+      ([__, a], [___, b]) => {
+        return (
+          getRoutePriority({ path: a.pathSpec, type: 'api' }) -
+          getRoutePriority({ path: b.pathSpec, type: 'api' })
+        );
+      },
+    );
+    for (const [p, v] of apiConfigEntries) {
```
hamlim: The main bug here is that the `apiPathMap` is sorted by insertion order rather than priority order - a potentially more efficient solution here is to use a different data structure that preserves priority order on insertion (e.g. some kind of tree or trie)

tylersayshi: > The main bug here is that the apiPathMap is sorted by insertion order rather than priority order

what exactly do you mean by this? The `Map` isn't meant to be ordered

tylersayshi: Oh I see what you mean, sorry. 

with the current data structures the options are either:

1. sort by priority then match and return early
2. find all matches, sort them based on priority, return the highest priority match

another idea could be to generate a sortedPath map as the last step of `configure()`. That way it awaits all `create*` calls then we synchronously create the sorted by priority apiConfigEntries so that they're ready by the time we call `getApiRoutePath`

hamlim: Yea at the moment i'm leaning towards the first option (keep maps as-is, sort when doing path matching) to be the least invasive change, haven't had much time to get back to this but hopefully this weekend i will have some time.

hamlim:
I'm going to split out the `/[...catchAll]` route not matching `/` paths bug into it's own issue -> #1518

packages/waku/tests/create-pages.test.ts
```diff
@@ -1096,7 +1096,7 @@ describe('createPages pages and layouts', () => {
     await expect(getConfig).rejects.toThrowError('Duplicated path: /test');
   });
 
-  it('creates a complex router', async () => {
+  it.only('creates a complex router', async () => {
```
hamlim:
```suggestion
  it('creates a complex router', async () => {
```

e2e/wildcard-api-routes.spec.ts
```diff
@@ -14,26 +14,46 @@ test.describe(`wildcard api routes`, async () => {
     await stopApp();
   });
 
+  // @TODO: re-enable when root wildcard route can match index route
```
hamlim:
```suggestion
  // @TODO: re-enable when root wildcard route can match index route - #1518
```

tylersayshi:
lgtm, thanks!

tylersayshi (approved):

packages/waku/src/router/create-pages.ts
```diff
@@ -248,6 +248,52 @@ type ComponentList = {
...
+  // - No wildcard = more specific than wildcard at the end
+  let specificityScore = pathLength * 1000; // Base score from path length
+
+  if (!hasWildcard) {
+    // No wildcard - most specific
+    specificityScore += 100;
+  } else {
+    // Wildcard at the end - less specific
+    specificityScore += 0;
+  }
```
dai-shi: I don't prefer those magic numbers. You should be able to directly implement the logic in `routePriorityComparator`.

dai-shi (changes requested):
Can you refactor it please?

packages/waku/src/router/create-pages.ts
hamlim: Pushed up this change to inline the logic from `getRoutePriority` into `routePriorityComparator`: https://github.com/wakujs/waku/pull/1508/commits/f2799be4c31ec00a59778009c9bb1cb9f1396740

packages/waku/src/router/create-pages.ts
```diff
@@ -312,7 +344,15 @@ export const createPages = <
...
     method: string,
   ) => string | undefined = (path, method) => {
-    for (const [p, v] of apiPathMap.entries()) {
+    const apiConfigEntries = Array.from(apiPathMap.entries()).sort(
+      ([__, a], [___, b]) => {
+        return routePriorityComparator(
+          { path: a.pathSpec, type: 'api' },
+          { path: b.pathSpec, type: 'api' },
+        );
+      },
```
dai-shi:
```suggestion
      ([, a], [, b]) => routePriorityComparator(
          { path: a.pathSpec, type: 'api' },
          { path: b.pathSpec, type: 'api' },
        ),
```

Does this work? Format it please.

Issue #1504, opened 2025-07-03
[Edge case] console error by Fragment with script tag in root layout


t6adev:
Hi, I'm facing an edge case issue that shows the error in the browser console when using Fragment and adding a script tag to the root layout in a production build only.

# Reproduction

1. Run `npm create waku@latest` (waku v0.23.3)
2. Apply [this diff](https://github.com/t6adev/repro-waku-hydration-error/commit/971a4bdc28f695fc8aaaa4ef97fcb69c68d0f7cb)
3. Run `npm run build && npm run start`
4. Open your browser and console tab to see the error

```
Uncaught Error: Minified React error #418; visit https://react.dev/errors/418?args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at vu (client-XN82YjJy.js:24:24705)
    at ki (client-XN82YjJy.js:24:25666)
    at sh (client-XN82YjJy.js:24:77890)
    at vf (client-XN82YjJy.js:24:107019)
    at r1 (client-XN82YjJy.js:24:106280)
    at Th (client-XN82YjJy.js:24:105320)
    at zc (client-XN82YjJy.js:24:105158)
    at K1 (client-XN82YjJy.js:24:101957)
    at fv (client-XN82YjJy.js:24:113586)
    at MessagePort.Au (client-XN82YjJy.js:9:1657)
```

It hasn't happened with waku v0.22.4.

dai-shi:
Thanks for reporting.

We need to create a failing test.

> It hasn't happened with waku v0.22.4.

Can you confirm if it happens with waku v0.23.0?

t6adev:
Sure. It seems that the issue occurred from v0.23.2.

dai-shi:
Can you help with creating a failing test case?
If it's not related with the router, we should be able to reproduce it with `e2e/ssr-basic.spec.ts`.

dai-shi:
> Sure. It seems that the issue occurred from v0.23.2.

Probably related PRs are #1438 and #1453.

PR #1514, opened 2025-07-09
test: add a failing test case for #1504

t6adev:
For https://github.com/wakujs/waku/issues/1504#issuecomment-3050945997


dai-shi:
Thanks. I'm almost sure how it's happening, but not sure how to fix. Let me tackle it later when I have a chance.

PR #1515, opened 2025-07-09
chore: refactor router client history update

rmarscher:
Follow up to #1459


rmarscher:
It looks like there are still failing tests. Changing to draft while I fix.

rmarscher:
I'm just learning about the transition inside a transition pattern. That is probably why the startTransitionFn is also called after `await changeRoute`. I should probably revert the changes to unstable_startTransition in this PR and reconsider.
https://react.dev/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition

rmarscher:
Closing as stale

PR #1524, opened 2025-07-13
fix monorepo with pnpm and yarn

dai-shi:
Problem: monorepo spec doesn't pass locally. It looks like, when it seems to work, it reuses `standaloneDir` wrongly.


packages/waku/src/lib/plugins/vite-plugin-rsc-resolve-rsdw.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function rscResolveRsdwPlugin(): Plugin {
+  return {
+    name: 'rsc-resolve-rsdw-plugin',
+    enforce: 'pre',
+    async resolveId(id) {
+      if (id === 'react-server-dom-webpack/server.edge') {
+        const resolved = await this.resolve(id);
+        return resolved?.id;
+      }
```
dai-shi, line 11: This is a workaround for "pnpm monorepo" spec, but I wonder if @hi-ogawa has a nicer solution or #1493 has already fix it.

PR #1543, opened 2025-07-18
feat(create-pages): add createSlice to createPages api

tylersayshi:
createSlice implementation with tests


packages/waku/src/router/create-pages.ts
```diff
@@ -149,6 +150,7 @@ export type CreatePage = <
      * This is intended for extending support to create custom routers.
      */
     exactPath?: ExactPath;
+    slices?: Slices;
```
dai-shi, line 157:
```suggestion
    /**
    * List of slice ids used in the component.
    * This is _required_ to send the slices along with the component.
    */
    slices?: Slices;
```

PR #1548, opened 2025-07-20
feat(fs-router): slice support for managed mode

tylersayshi:
currently trying to overcome this error

```
Error: Invalid element: slice:slice001
    at Slot (file:///Users/tyler/gitspace/waku/packages/waku/dist/minimal/client.js:221:15)
```

it's confusing because `handleSlice` is not being called. And from checking logs both `createSlice` and `createPage` are called in the same way as the create-pages spec.

I'm probably missing something obvious so if someone sees the issue that'd be very helpful 🙌

### DIFF
--- a/e2e/fixtures/fs-router/src/pages/_layout.tsx
+++ b/e2e/fixtures/fs-router/src/pages/_layout.tsx
@@ -76,6 +76,9 @@ const HomeLayout = ({ children }: { children: ReactNode }) => (
       <li>
         <Link to="/page-parts">Page Parts</Link>
       </li>
+      <li>
+        <Link to="/slices">Slices</Link>
+      </li>
     </ul>
     {children}
   </div>
--- a/e2e/fixtures/fs-router/src/pages/slices/index.tsx
+++ b/e2e/fixtures/fs-router/src/pages/slices/index.tsx
@@ -1,0 +1,17 @@
+import { Slice } from 'waku/router/client';
+
+export default function Slices() {
+  return (
+    <>
+      <h2>Slices</h2>
+      <Slice id="slice001" />
+      <Slice id="slice002" />
+    </>
+  );
+}
+
+export const getConfig = () => {
+  return {
+    slices: ['slice001', 'slice002'],
+  };
+};
--- a/packages/waku/src/router/fs-router.ts
+++ b/packages/waku/src/router/fs-router.ts
@@ -32,6 +32,7 @@ export function unstable_fsRouter(
       createRoot,
       createApi,
       createPagePart,
+      createSlice,
     }) => {
       let files = await unstable_getPlatformData<string[]>('fsRouterFiles');
       if (!files) {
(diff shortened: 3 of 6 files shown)


packages/waku/src/router/fs-router.ts
```diff
@@ -93,6 +94,17 @@ export function unstable_fsRouter(
           .replace(/\.\w+$/, '')
           .split('/')
           .filter(Boolean);
+        // must come before isIgnoredPath check to include slices from inside _components
+        if (pathItems.at(-1)?.startsWith('_slice')) {
+          createSlice({
+            component: mod.default,
+            render: 'dynamic',
+            // Q: should we default to using the file name as the id?
```
dai-shi: Yeah, I was thinking of it. As long as we can detects duplicates and warns it, it feels safe.
I wasn't sure about using the parent path for the slice id, but maybe it's too confusing.

e2e/fixtures/fs-router/src/pages/slices/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <Slice id="slice002" />
+    </>
+  );
+}
+
+export const getConfig = () => {
+  return {
+    slices: ['slice001', 'slice002'],
+  };
+};
```
dai-shi: The duplication of the ids is slightly less than ideal in DX.

How about mitigating it with this pattern?

```tsx
const slices = ['slice001', 'slice002'] as const;

export default function Slices() {
  return (
    <>
      <h2>Slices</h2>
      <Slice id={slices[0]} />
      <Slice id={slices[1]} />
    </>
  );
}

export const getConfig = () => {
  return {
    slices,
  };
};
```

packages/waku/src/router/fs-router.ts
tylersayshi: I think only using the filename is simplest

e2e/fixtures/fs-router/src/pages/_layout.tsx
```diff
@@ -76,6 +76,9 @@ const HomeLayout = ({ children }: { children: ReactNode }) => (
       <li>
         <Link to="/page-parts">Page Parts</Link>
       </li>
+      <li>
+        <Link to="/slices">Slices</Link>
```
dai-shi: This naming is slightly confusing when reading.

How about something like this?
```suggestion
        <Link to="/page-with-slices">Page with Slices</Link>
```

e2e/fixtures/fs-router/src/pages/slices/index.tsx
tylersayshi: seems fine, updated to this

packages/waku/src/router/fs-router.ts
```diff
@@ -93,6 +94,16 @@ export function unstable_fsRouter(
           .replace(/\.\w+$/, '')
           .split('/')
           .filter(Boolean);
+        // must come before isIgnoredPath check to include slices from inside _components
+        if (pathItems.at(-1)?.startsWith('_slice')) {
+          createSlice({
+            component: mod.default,
+            render: 'dynamic',
+            id: pathItems.at(-1)!.slice('_slice'.length),
```
dai-shi: Another idea is to keep `slice`? Not sure which is more intuitive.
```suggestion
            id: pathItems.at(-1)!.slice('_'.length),
```

tylersayshi: I think that would make the id's read better. i'll change to that 👍

dai-shi (approved):
We may need more e2e tests for various use cases and we may revisit some details of API later, but it looks good to me at this point.

PR #1556, opened 2025-07-24
docs(slices): examples for fs-router and create-pages

tylersayshi:


examples/11_fs-router/src/pages/_components/_slice-one.tsx
```diff
@@ -0,0 +1,16 @@
...
+        Slice One {new Date().toLocaleString()}
+      </h2>
+    </div>
+  );
+}
+
+export const getConfig = () => {
+  return {
+    render: 'static',
+    id: 'slice-one',
```
dai-shi, line 14: Should we remove this or keep it as an example?

dai-shi (approved):
LGTM otherwise.

PR #1557, opened 2025-07-24
test(slices): added e2e tests for slices

tylersayshi:
@dai-shi the `lazy` test doesn't work. Am I missing something obvious? Or is it a bug?

### DIFF
--- a/e2e/fixtures/create-pages/src/server-entry.tsx
+++ b/e2e/fixtures/create-pages/src/server-entry.tsx
@@ -347,7 +347,7 @@ const pages: ReturnType<typeof createPages> = createPages(
     createPage({
       render: 'dynamic',
       path: '/slices',
-      slices: ['slice001', 'slice002', 'slice003'],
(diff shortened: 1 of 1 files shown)


e2e/fixtures/create-pages/src/server-entry.tsx
```diff
@@ -344,14 +345,19 @@ const pages: ReturnType<typeof createPages> = createPages(
     }),
 
     createPage({
-      render: 'static',
+      render: 'dynamic',
       path: '/slices',
-      slices: ['slice001', 'slice002'],
+      slices: ['slice001', 'slice002', 'slice003'],
```
dai-shi: lazy slice doesn't need to be sent with the page. if it's sent, it's not lazy.

```suggestion
      slices: ['slice001', 'slice002'],
```

tylersayshi: Oooo gotcha. Thanks for the explanation.

PR #1558, opened 2025-07-24
Add --choose to readme

nojaf:


README.md
```diff
@@ -27,6 +27,13 @@ Start a new Waku project with the `create` command for your preferred package ma
 npm create waku@latest
 ```
 
+Alternatively, you can choose a template via
+
+```sh
+npm create waku@latest -- --choose
+bun create waku@latest --choose
```
dai-shi: We don't take bun specifically.
```suggestion
```

dai-shi:
@sandren What do you think?

This is kind of an advanced usage, and an alternative approach I have is to create a dedicated doc in ./docs/guides.

README.md
nojaf: I put it there because replacing `npm` with `bun` didn't work for me.
`bun` only works without the extra `--`, but `npm` didn't work without `--`.

sandren:
> @sandren What do you think?
> 
> This is kind of an advanced usage, and an alternative approach I have is to create a dedicated doc in ./docs/guides.

Yeah, I think it's a good idea to document it somewhere, but should be in guides instead of the main readme.

README.md
dai-shi: If we were to ad `bun`, we would need to add others... I wonder if `bun` would accept extra `--` in the future...

dai-shi:
> Please create a new guide in ./docs/guides/, instead of adding this in readme.md.

If anyone is interested, feel free to take it.

PR #1563, opened 2025-07-25
docs(slices): blog post for 0.25 and slice api

tylersayshi:
new slices api blog post is here 🎉

### DIFF
--- a/packages/website/private/contents/post-012.mdx
+++ b/packages/website/private/contents/post-012.mdx
@@ -1,15 +1,15 @@
 ---
 slug: rethinking-fine-grained-components
 title: Introducing “slice components”
-description: We’ve re-thought the api for fine-grained component rendering.
+description: We’ve re-thought the API for fine-grained component rendering.
 author: tyler
 release: v0.25
 date: 2025/08/12
 ---
 
-We’re happy to announce Waku v0.25 with a brand new API: the Slice Components!
+We’re happy to announce Waku v0.25 with a brand new API: Slice Components!
 
-We heard the feedback around the Page Part components and feel like the Slices API will both solve a lot of the issues, while also unlocking brand new flexibility with render patterns!
+We feel the Slices API solves issues reported with the now deprecated Page Part components while also unlocking brand new flexibility with render patterns!
 
 ## Components are re-useable across pages
 
@@ -63,11 +63,11 @@ export const getConfig = () => {
 };
 ```
 
-We're really happy with the new approach and how it will allow simply importing and assembling your components like normal components.
+We’re really happy with the new approach and how it will allow simply importing and assembling your components like normal components, while also specifying alternate rendering.
 
 ## Lazy Slice Components
 
-`lazy` Slice Components are a new addition with the Slices API. They are Waku's take on the [server islands feature](https://astro.build/blog/astro-4120/) from Astro. `lazy` allows components to be requested independent from the page they are used on. This is useful for components will be dynamically rendered on otherwise static pages and layouts.
+`lazy` Slice Components are a new addition with the Slices API. They are Waku’s take on the [server islands feature](https://astro.build/blog/astro-4120/) from Astro. `lazy` allows components to be requested independent from the page they are used on. This is useful for components will be dynamically rendered on otherwise static pages and layouts.
 
 If we re-use the example above, it could be written like this:
 


packages/website/private/contents/post-011.mdx
```diff
@@ -0,0 +1,99 @@
...
+An example for `_slice-one.tsx`:
+
+```tsx
+export default function SliceOne() {
+  return <p>🍕</p>;
+}
+
+export const getConfig = () => {
+  return {
+    id: 'slice-one',
```
dai-shi: This can be omitted, right?

packages/website/private/contents/post-011.mdx
```diff
@@ -0,0 +1,99 @@
...
+  );
+}
+
+export const getConfig = () => {
+  return {
+    render: 'static',
+    slices: ['slice-one'], // Note: `slice-two` is lazy so it is not included with the page config
+  };
+};
+```
```
dai-shi: Let's be super clear that this example allows slince-two to be dynamic.

packages/website/private/contents/post-011.mdx
tylersayshi: it can be, but do we want the example to show how it's configurable, or no?

dai-shi: > And by default, the slice id will be the filename without the `_` prefix.

As it's instructed, I thought we wanted to show the example. Or, comment it:

```suggestion
    id: 'slice-one', // can be omitted if it can be taken from the file name
```

tylersayshi: added the comment and one with the default render on the next line

tylersayshi:
should this version be changed to 0.25?

packages/waku/src/router/fs-router.ts
```diff
@@ -95,11 +95,11 @@ export function unstable_fsRouter(
           .split('/')
           .filter(Boolean);
         // must come before isIgnoredPath check to include slices from inside _components
-        if (pathItems.at(-1)?.startsWith('_slice')) {
+        if (pathItems.at(-1)?.startsWith('_slice-')) {
           createSlice({
             component: mod.default,
             render: 'dynamic',
-            id: pathItems.at(-1)!.slice('_'.length),
+            id: pathItems.at(-1)!.slice('_slice-'.length),
```
dai-shi: @tylersayshi If we were to modify the code, it should be a separate PR, because merge timings are different.

examples/11_fs-router/src/pages/slice-page.tsx
```diff
@@ -13,6 +13,6 @@ export default function SlicePage() {
 export const getConfig = () => {
```
himself65: small nit: we should have a type for `getConfig` function. Now it includes too many properties

dai-shi: There's a huge debate on it. 😄

himself65: I think Next.js has a clearer API for this. like

https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config

where is debate thead

tylersayshi: There was more than one (in the discord), but I'll try to summarize:

1. a hard requirement is to not add a function call
2. Explicitly typing the function signature or the return type will widen the result type beyond what we could infer with `as const` or `satisfies`. They'll give type hints and checking to the return type though which is good.
e.g.
```ts
export const getConfig: GetConfig = () => {}
export const getConfig = (): GetConfigResult => {}
```
3. `as const` was decided as the preference over `satisfies` primarily due to the newness of `satisfies` as a keyword iirc

---

> I think Next.js has a clearer API for this. like
https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config

I'm curious... in what way(s) is this clearer? From what I can tell, they're documenting special & separate export names that are all untyped?

packages/waku/src/router/fs-router.ts
tylersayshi: created #1592 separately

tylersayshi:
> Thanks for working on it. I would like to see the lazy Slice to be compared with (or promoted as) "islands", because it's the solution for #816 that we discussed before.

added a mention, let me know what you think or want to change/add.

packages/website/private/contents/post-011.mdx
```diff
@@ -0,0 +1,101 @@
+---
+slug: rethinking-fine-grained-components
+title: Introducing “slice components”
+description: We’ve re-thought the api for fine-grained component rendering.
+author: tyler
+release: v0.25
+date: 2025/08/15
```
dai-shi: Rename the file to `post-012.mdx` and
```suggestion
date: 2025/08/12
```

packages/website/private/contents/post-011.mdx
```diff
@@ -0,0 +1,101 @@
...
+    render: 'static', // default is 'dynamic'
+  };
+};
+```
+
+We're really happy with the new approach and how it will allow simply importing and assembling your components like normal components.
+
+## Lazy Slice Components
+
+`lazy` Slice Components are a new addition with the Slices API. They are waku's take on the [server islands feature](https://astro.build/blog/astro-4120/) from Astro. `lazy` allows components to be requested indepent from the page they are used on. This is useful for components that are expensive to render or that could be hosted separately (e.g. on a CDN).
```
dai-shi: waku->Waku

> requested indepent from the page they are used on.

I'm not sure if it makes sense. "requested after the page rendered (or hydrated)"??

> that are expensive to render

The point is not "expensive" but "dynamic."

tylersayshi: > "requested after the page rendered (or hydrated)"??

sounds like it underscores a network waterfall which seems like a bad way to celebrate the feature imo. Maybe something like: "requested separate from the rest of the page"? (I don't love the sound of this either, but just trying to think out loud here)

> The point is not "expensive" but "dynamic."

👍 I'll rephrase

tylersayshi: > I'm not sure if it makes sense.

also there's a typo... indepent is meant to be independent. I'll fix that for now at least 🤷

dai-shi: > sounds like it underscores a network waterfall

That's fair. I'll leave it to @sandren then.

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+```ts
+export const getConfig = () => {
+  return {
+    render: 'dynamic',
+  } as const;
+};
+```
+
+If your full build is static, this change won't impact you since the static build mode makes every page only render once at build time.
```
dai-shi: I don't think this is true. Full static build must be with all static layouts/pages/...

tylersayshi: so it errors when dynamic is defined?

dai-shi: I think it just passes, but the result in `dist/public` isn't statically deployable. I might be wrong. I haven't tried it lately.

tylersayshi: I just removed this since this isn't really helpful anyways

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+date: 2025/08/12
+---
+
+We’re happy to announce Waku v0.25 with a brand new API: the Slice Components!
+
+We heard the feedback around the Page Part components and feel like the Slices API will both solve a lot of the issues, while also unlocking brand new flexibility with render patterns!
+
+## Components are re-useable across pages
+
+Slices are components that are defined in the `src/pages/slices` directory. This allows us to re-use slices across pages as any other component would be used.
```
dai-shi:
```suggestion
Slices are components that are defined in the `src/pages/_slices` directory. This allows us to re-use slices across pages as any other component would be used.
```

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+```
+ src/pages
+ ├── slices
```
dai-shi:
```suggestion
 ├── _slices
```

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+    slices: ['one', 'two'],
+  };
+};
+```
+
+Please notice: `slices` must be specified with each of the slice ids in order for Waku to know how to bundle your page without any compiler magic.
+
+By default, the slice id will be the filename. If slices are nested like `src/pages/slices/one/two.tsx`, the id will be `one/two`.
+
+An example for `src/pages/slices/one.tsx`:
```
dai-shi:
```suggestion
An example for `src/pages/_slices/one.tsx`:
```

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+  return {
+    render: 'static',
+    slices: ['one', 'two'],
+  };
+};
+```
+
+Please notice: `slices` must be specified with each of the slice ids in order for Waku to know how to bundle your page without any compiler magic.
+
+By default, the slice id will be the filename. If slices are nested like `src/pages/slices/one/two.tsx`, the id will be `one/two`.
```
dai-shi:
```suggestion
By default, the slice id will be the filename. If slices are nested like `src/pages/_slices/one/two.tsx`, the id will be `one/two`.
```

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+An example for `src/pages/_slices/one.tsx`:
+
+```tsx
+export default function SliceOne() {
+  return <p>🍕</p>;
+}
+
+export const getConfig = () => {
+  return {
+    id: 'one', // can be omitted if it can be taken from the file name
```
dai-shi: Can we remove this example, or rather leave it?

dai-shi (commented):
A few more ideas if we have a chance to incorporate.

- Mention this is inspired by Gatsby Slice API (or rather not?)

- Show more use cases (or maybe those are for guides?)

```
a) mix static and dynamic "components" in a single page
b) share "components" across pages (only meaningful with static components?)
c) add dynamic "components" to static pages that can be deployed on CDN. (using lazy prop)
```

packages/website/private/contents/post-012.mdx
```diff
@@ -0,0 +1,115 @@
...
+  return {
+    render: 'static',
+    slices: ['one', 'two'],
+  };
+};
+```
+
+Please notice: `slices` must be specified with each of the slice ids in order for Waku to know how to bundle your page without any compiler magic.
+
+By default, the slice id will be the filename. If slices are nested like `src/pages/_slices/one/two.tsx`, the id will be `one/two`.
```
dai-shi:
```suggestion
The slice id will be the filename. If slices are nested like `src/pages/_slices/one/two.tsx`, the id will be `one/two`.
```

packages/website/private/contents/post-012.mdx
dai-shi:
```suggestion
```

dai-shi (approved):

dai-shi (approved):
LGTM

PR #1567, opened 2025-07-30
feat(website): support llm.txt

himself65:
add a vite plugin to generate llm.txt to help AI to use waku

### DIFF
--- a/packages/website/src/lib/vite-plugin-llmstxt.ts
+++ b/packages/website/src/lib/vite-plugin-llmstxt.ts
@@ -1,0 +1,137 @@
+import type { Plugin } from 'vite';
+import { readdir, readFile, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { compileMDX } from 'next-mdx-remote/rsc';
+
+interface LLMTextPluginOptions {
+  hostname?: string;
+  contentsPath?: string;
+  outputPath?: string;
+}
+
+interface BlogPost {
+  slug: string;
+  title: string;
+  description?: string;
+  author?: string;
+  date?: string;
+  content: string;
+}
+
+export function llmTextPlugin(options: LLMTextPluginOptions = {}): Plugin {
+  const {
+    hostname = 'https://waku.gg',
+    contentsPath = './private/contents',
+    outputPath = './public/llm.txt',
+  } = options;
+
+  async function generateLLMText() {
+    const posts: BlogPost[] = [];
+
+    // Read all MDX files
+    const dirents = await readdir(contentsPath);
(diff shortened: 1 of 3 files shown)


himself65:
after I implement this plugin, I realized that this plugin is ran 4 times during the dev / build. How can we make this only run only once?

himself65:
wait for #1493 for a stable plugin API

rmarscher:
Yeah, maybe wait for #1493. I've been adding plugins to `"build-client"` section of `unstable_viteConfigs` that I only want to run once near the end of the build.

packages/website/src/lib/vite-plugin-llmstxt.ts
```diff
@@ -0,0 +1,137 @@
...
+  author?: string;
+  date?: string;
+  content: string;
+}
+
+export function llmTextPlugin(options: LLMTextPluginOptions = {}): Plugin {
+  const {
+    hostname = 'https://waku.gg',
+    contentsPath = './private/contents',
+    outputPath = './public/llm.txt',
```
rmarscher: I think it's `llms.txt` and not `llm.txt`. https://llmstxt.org/

himself65: whoops

himself65:
I will pick this up again since plugin API is stable now

himself65:
I might do this PR first https://github.com/wakujs/waku/pull/1576

nojaf:
I would be great if the llm.txt file could be shipped as part of the npm package.
That way you could point your LLM to a version that is matching our Waku version.

dai-shi:
> as part of the npm package.

Just curious, but what's wrong with `README.md`?

nojaf:
There is nothing wrong with the README but my naïeve thinking is that the llm.txt would contain more information in a better way for an llm to process.

dai-shi:
For now, what I can think of the content for llms.txt is the same as README.md.

rmarscher:
I saw this - https://www.better-auth.com/llms.txt

They are linking to raw markdown docs - https://www.better-auth.com/llms.txt/docs/adapters/postgresql.md

That seems like a good idea too.

himself65:
> I saw this - https://www.better-auth.com/llms.txt
> 
> They are linking to raw markdown docs - https://www.better-auth.com/llms.txt/docs/adapters/postgresql.md
> 
> That seems like a good idea too.

Oh, by chance, I'm also a better auth maintainer. Yeah, it follows AI llmtxt guideline

dai-shi:
> I will pick this up again since plugin API is stable now

Please open a new PR then.

dai-shi:
https://waku.gg/api/llms-full.txt is available now.

PR #1568, opened 2025-07-30
fix: fouc in website

dai-shi:
There's a FOUC regression. It's caused by my change in #1514.
This feels like an ad-hoc fix. I wonder if #1493 fixes completely.

### DIFF
--- a/packages/waku/src/lib/renderers/html.ts
+++ b/packages/waku/src/lib/renderers/html.ts
@@ -61,7 +61,7 @@ globalThis.__WAKU_HYDRATE__ = true;
 `;
   const arr = parseHtml(htmlHead) as ReactElement<any>[];
   const headModules: string[] = [];
-  const headElements: ReactElement[] = [];
+  const headElements: ReactElement<{ async?: unknown }>[] = [];
   for (const item of arr) {
     if (item.type === 'script') {
       if (item.props?.src) {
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/renderers/html.ts
```diff
@@ -164,14 +164,15 @@ export async function renderHtml(
...
+      ...headElements.filter((ele) => ele.props?.async !== true),
       // See: https://github.com/wakujs/waku/pull/1545
       // isolate `React.use` in its own component
       // https://github.com/facebook/react/issues/33937#issuecomment-3091349011
       createElement(async () => {
         const resolvedHtmlNode = await htmlNode;
         return createElement(HtmlNodeWrapper, null, resolvedHtmlNode);
       }),
-      ...headElements,
+      ...headElements.filter((ele) => ele.props?.async === true),
```
tylersayshi, line 175: why are these added separately?

himself65: I believe it will have some performance improvement when putting sync script on top. and async script won't be ran immediately

dai-shi: It seems like some head elements should come before to avoid FOUC and some should come after to avoid hydration mismatch errors.

PR #1576, opened 2025-08-01
feat: allow read private folder during rsc

himself65:
Fixes: https://github.com/wakujs/waku/issues/1569


packages/waku/src/lib/plugins/vite-plugin-rsc-private.ts
```diff
@@ -17,8 +17,11 @@ export function rscPrivatePlugin({
       privatePath = joinPath(config.root, privateDir);
     },
     load(id) {
+      if (this.environment.name === 'rsc') {
+        return;
+      }
```
dai-shi: This doesn't work anyways. So, I think we should remove it to avoid confusion.

dai-shi (approved):

PR #1591, opened 2025-08-03
fix: remove deprecated `unstable_viteConfigs`

himself65:


dai-shi:
Let's do this when we will completely drop legacy cli.

PR #1592, opened 2025-08-03
fix(slices): change slice prefix to remove _slice then any leading non-alphanumeric characters

tylersayshi:
for review from @sandren

### DIFF
--- a/packages/waku/src/router/fs-router.ts
+++ b/packages/waku/src/router/fs-router.ts
@@ -11,6 +11,20 @@ import { isIgnoredPath } from '../lib/utils/fs-router.js';
 
 const DO_NOT_BUNDLE = '';
 
+const SLICE_PREFIX = '_slice';
+
+/**
+ * Helper to remove the `_slice-` prefix from a slice id.
+ * It will also remove any non-alphanumeric characters until one is found
+ */
+const removeStart = (s: string) => {
+  let withoutStart = s.slice(SLICE_PREFIX.length);
+  while (withoutStart.match(/[^a-zA-Z0-9]/)) {
+    withoutStart = withoutStart.slice(1);
+  }
+  return withoutStart;
+};
+
 export function unstable_fsRouter(
   importMetaUrl: string,
   loadPage: (file: string) => Promise<any> | undefined,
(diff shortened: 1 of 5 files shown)


packages/waku/src/router/fs-router.ts
```diff
@@ -11,6 +11,20 @@ import { isIgnoredPath } from '../lib/utils/fs-router.js';
...
 
+const SLICE_PREFIX = '_slice';
+
+/**
+ * Helper to remove the `_slice-` prefix from a slice id.
+ * It will also remove any non-alphanumeric characters until one is found
+ */
+const removeStart = (s: string) => {
+  let withoutStart = s.slice(SLICE_PREFIX.length);
+  while (withoutStart.match(/[^a-zA-Z0-9]/)) {
```
himself65: move regex on top level

```ts
const NON_ALPHANUMERIC_START = /^[^a-zA-Z0-9]+/;
```

packages/waku/src/router/fs-router.ts
```diff
@@ -11,6 +11,20 @@ import { isIgnoredPath } from '../lib/utils/fs-router.js';
 
 const DO_NOT_BUNDLE = '';
 
+const SLICE_PREFIX = '_slice';
+
+/**
+ * Helper to remove the `_slice-` prefix from a slice id.
+ * It will also remove any non-alphanumeric characters until one is found
+ */
+const removeStart = (s: string) => {
```
himself65, line 20: I realized this will skip the non-English character (like Chinese, Japanese)

dai-shi: Maybe we should remove just delimiters. Maybe only `-` and `_` for the start?

packages/waku/src/router/fs-router.ts
dai-shi: I don't think we need to move regex on top level.
What I care more is the while loop. Let's just use one `replace`.
```js
withoutStart = withoutStart.replace(/^[-_]*/g, '');
```

packages/waku/src/router/fs-router.ts
```diff
@@ -11,6 +11,16 @@ import { isIgnoredPath } from '../lib/utils/fs-router.js';
...
+  return s.slice(SLICE_PREFIX.length).replace(/^[-_]*/g, '');
```
dai-shi: my bad, we don't need `g`.
```suggestion
  return s.slice(SLICE_PREFIX.length).replace(/^[-_]*/, '');
```
or maybe we should remove the first one only?
```suggestion
  return s.slice(SLICE_PREFIX.length).replace(/^[-_]/, '');
```

dai-shi (approved):
I'm not familiar with the function name `removeStart`, but otherwise LGTM.

tylersayshi:
> I'm not familiar with the function name `removeStart`, but otherwise LGTM.

it was just a name I made up. Happy to change it or even remove the util if that makes it clearer.

i'll merge this now, but please comment if you'd like to see a change and I'll come back to update it

dai-shi:
@tylersayshi How about `removeSlicePrefix`?

PR #1593, opened 2025-08-03
fix(create-pages): clean path mapping of pathless groups

### LINKED ISSUE #1574: Segment Routes not working as expected in Grouped Routes
Hello, I'm trying to access segmented route's data but it just says undefined every time. I noticed that my segment route inside of grouped route. I tried to move my segmented route to the outside of grouped route, it fixed the problem so I guess this should be a bug.

I'm trying to get postid with using this structure and it fails

```
pages
├── (auth)
│   └── ...pages
└── (withSidebarLayout)
    ├── post
    │   └── [postid].tsx
    └── ...pages

```

But if I move post route outside of grouped route it works and I can get postid

```
pages
├── (auth)
│   └── ...pages
├── (withSidebarLayout)
│   └── ...pages
└── post
    └── [postid].tsx

```

tylersayshi:
this generates the correct mapping for paths using pathless groups

fixes #1574

### DIFF
--- a/packages/waku/src/lib/utils/path.ts
+++ b/packages/waku/src/lib/utils/path.ts
@@ -155,16 +155,20 @@ export const getPathMapping = (
   pathname: string,
 ): Record<string, string | string[]> | null => {
   const actual = pathname.split('/').filter(Boolean);
-  if (pathSpec.length > actual.length) {
-    const hasWildcard = pathSpec.some((spec) => spec.type === 'wildcard');
+  // remove pathless groups
+  const cleanPathSpec = pathSpec.filter(
+    (spec) => !(spec.type === 'literal' && spec.name.startsWith('(')),
+  );
+  if (cleanPathSpec.length > actual.length) {
+    const hasWildcard = cleanPathSpec.some((spec) => spec.type === 'wildcard');
     if (!hasWildcard || actual.length > 0) {
       return null;
     }
   }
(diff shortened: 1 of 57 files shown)


packages/waku/src/lib/utils/path.ts
```diff
@@ -155,16 +155,20 @@ export const getPathMapping = (
   pathname: string,
 ): Record<string, string | string[]> | null => {
   const actual = pathname.split('/').filter(Boolean);
-  if (pathSpec.length > actual.length) {
-    const hasWildcard = pathSpec.some((spec) => spec.type === 'wildcard');
+  // remove pathless groups
+  const cleanPathSpec = pathSpec.filter(
+    (spec) => !(spec.type === 'literal' && spec.name.startsWith('(')),
```
dai-shi: I don't prefer this. Can you separate utils for define-router and create-pages? I hope to have separate files for each. Or only utils.ts should be for the define-router notion.

tylersayshi: how would you feel about moving the functions that are only used for create-pages just into the create-pages.ts file?

dai-shi: Yes, that will be good.

tylersayshi: reorganized to move create-pages specific logic into create pages file

dai-shi (approved):
LGTM

PR #1616, opened 2025-08-10
feat(router): change from prefix for slices to directory

tylersayshi:
new logic is to have `src/pages/_slices` directory. subdirectories lead to ids joined on `/`

e.g.

`src/pages/_slices/one.tsx` => `id='one'`
`src/pages/_slices/foo/bar.tsx` => `id='foo/bar'`


packages/waku/src/config.ts
```diff
@@ -32,6 +32,11 @@ export interface Config {
    * Defaults to "api".
    */
   apiDir?: string;
+  /**
+   * The slices directory inside pagesDir.
+   * Defaults to "slices".
```
dai-shi:
```suggestion
   * Defaults to "_slices".
```

packages/waku/src/lib/config.ts
```diff
@@ -21,6 +21,7 @@ export async function resolveConfigDev(config: Config) {
     distDir: 'dist',
     pagesDir: 'pages',
     apiDir: 'api',
+    slicesDir: 'slices',
```
dai-shi:
```suggestion
    slicesDir: '_slices',
```

packages/waku/src/router/fs-router.ts
```diff
@@ -32,6 +22,8 @@ export function unstable_fsRouter(
      * is `"foo"`, then it will detect pages in `src/foo/api`.
      */
     apiDir: string;
+    /** e.g. `"slices"` will detect slices in `src/pages/slices`. */
```
dai-shi:
```suggestion
    /** e.g. `"_slices"` will detect slices in `src/pages/_slices`. */
```

PR #1622, opened 2025-08-14
docs(readme): add slices info

tylersayshi:
updating readme to document slices api

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,113 @@ export const getConfig = async () => {
 };
 ```
 
+### Slices
+
+Slices are reusable components that are defined in the `src/pages/_slices` directory. They allow you to compose pages by assembling components like normal React components while specifying alternate rendering patterns.
+
+#### Creating slices
+
+Slices are created by placing files in the `src/pages/_slices` directory. The slice ID corresponds to the filename, and nested slices use the full path as the ID.
+
+```
+ src/pages
+ ├── _slices
+ │   ├── one.tsx
+ │   ├── two.tsx
+ │   └── nested
+ │       └── three.tsx
+ └── some-page.tsx
+```
+
+Each slice file exports a default React component and a `getConfig` function that specifies the render method.
+
+```tsx
+// ./src/pages/_slices/one.tsx
+
+// Create slice component
+export default function SliceOne() {
+  return <p>🍕</p>;
+}
+
+export const getConfig = () => {
+  return {
+    render: 'static', // default is 'static'
+  };
+};
+```
+
+```tsx
+// ./src/pages/_slices/nested/three.tsx
+
+// Create nested slice component
+export default function SliceThree() {
+  return <p>🍰</p>;
+}
+
+export const getConfig = () => {
+  return {
+    render: 'dynamic',
+  };
+};
+```
+
+#### Using slices
+
+Slices are used in pages and layouts by importing the `Slice` component from Waku and specifying the slice ID. The `slices` array in the page's `getConfig` must include all slice IDs used on that page.
+
+```tsx
+// ./src/pages/some-page.tsx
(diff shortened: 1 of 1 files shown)


dai-shi:
@tylersayshi I just noticed we didn't have group routes either.

tylersayshi:
> @tylersayshi I just noticed we didn't have group routes either.
You want that this in this PR or a new one?

tylersayshi:
> @tylersayshi I just noticed we didn't have group routes either.

You want that this in this PR or a new one?

dai-shi:
Maybe a new one. We need eyes from @sandren and probably a separate PR would be easier.

tylersayshi:
> Maybe a new one. We need eyes from @sandren and probably a separate PR would be easier.

https://github.com/wakujs/waku/pull/1629

README.md
```diff
@@ -513,6 +513,113 @@ export const getConfig = async () => {
 };
 ```
 
+### Slices
```
dai-shi, line 617: I think we should explain more about the use cases in the Guides.
For README, we keep it minimal, and it looks good at this point.

dai-shi (commented):
@tylersayshi Can you also update `./docs/create-pages.mdx` ? It can be a separate PR, but we can use this PR, and hopefully get @sandren 's eyes.

docs/create-pages.mdx
```diff
@@ -397,3 +405,21 @@ if (globalThis.__WAKU_HYDRATE__) {
...
+
+export const Slice = () => {
+  return (
+    <div>
+      <h2>Slice</h2>
+    </div>
+  );
+};
+```
+
```
tylersayshi, line 426: @dai-shi could you give an example use case or two for this?

dai-shi: https://newsletter.daishikato.com/p/waku-gets-slices-api-inspired-by-gatsby
Does it help?

> 1. Shared Components
We could use the slice in both / and /foo routes. When we navigate from / to /foo, we can reuse the cached slice component. This makes sense if either / or /foo is dynamic, and the slice component is static. It works only with client navigation (or soft navigation).
> 
> 2. Mixing Dynamic and Static Parts
While it is technically possible with layouts and the (group) syntax, using the Slices API can be more straightforward. If your page is static, you can embed a dynamic slice to mix static and dynamic parts. If your page is dynamic, you can embed a static slice. You can also embed multiple static or dynamic slices regardless of the page type.
> 
> 3. Lazy Slices

Or, maybe you want more realistic examples.

tylersayshi: I think this is good info, but one more realistic example would be helpful for illustrating "what a slice really is". 

I'll add the reference to the prior art with gatsby, but for devs who are new to the concept, a real world example would probably be very valuable.

dai-shi: I think a real world example is something like a "static" modal dialog, which we can use in a various components regardless of the routes.

tylersayshi: Would it be better if the static component was something that was more expensive to render?

dai-shi: Well, technically, static pages/layouts/slices should not be re-computed at runtime. Currently, I don't know how it can be implemented, so static things are re-computed in dynamic routes and then eliminated before sending back the response.

So, if we solved it, there'd be such a benefit, but it's tricky to say it as of now.

tylersayshi: hm maybe this could just go in the expanded docs later then and what's here is good for the readme?

dai-shi: @tylersayshi I think that's good. @sandren may suggest to write a brief note about the use cases in readme. Can you also check failing CIs?

PR #1629, opened 2025-08-16
docs(readme): add group routes info

tylersayshi:
explains group routes

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,6 +415,113 @@ export const getConfig = async () => {
 };
 ```
 
+#### Group routes
+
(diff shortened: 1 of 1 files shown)


dai-shi:
@tylersayshi Please format README.md.

dai-shi (approved):
As @sandren isn't responding, let's move this forward.

It feels a bit lengthy, but it should be okay compared to segment routes.
We should probably revisit this and make README more minimal, moving and adding more examples in Guides.

README.md
```diff
@@ -415,6 +415,107 @@ export const getConfig = async () => {
 };
 ```
 
+#### Group routes
```
dai-shi, line 418: I'm not sure if we should put this after `#### Other layouts`. If necessary, let's do it later.

PR #1623, opened 2025-08-14
chore: separate createEngine from createHonoHandler

rmarscher:
Alternate solution to #1619

PR #1633, opened 2025-08-18
drop unstable_hono in favor of hono/context-storage

dai-shi:
I tried it once, but didn't work well. Now with vite-rsc, this should be more straightforward.


dai-shi:
> that is a different version number

Oh, that's probably it > but didn't work well
(And, waku/middleware/context has a hack with `globalThis`.)

I wonder if we can make hono a peer dependency, or maybe injectable.

rmarscher:
> maybe injectable

Yes, I think the correct solution is to allow custom server.entry.tsx file. As mentioned in the other draft PRs we are working on (#1619, #1623), we could update Waku's [server.entry.tsx](https://github.com/wakujs/waku/blob/main/packages/waku/src/vite-rsc/entry.server.tsx) to look something like this:

```ts
import { Hono } from 'hono';
import { createEngine } from 'waku/server';
import { INTERNAL_setAllEnv, unstable_builderConstants } from 'waku/server';
import { flags, config, isBuild } from 'virtual:vite-rsc-waku/config';
import path from 'node:path';
import fs from 'node:fs';

const { DIST_PUBLIC } = unstable_builderConstants;

createEngine();

const app = new Hono();
INTERNAL_setAllEnv(process.env as any);
if (flags['experimental-compress']) {
  app.use(compress());
}
if (isBuild) {
  app.use(serveStatic({ root: path.join(config.distDir, DIST_PUBLIC) }));
}
app.use((c) => {
  const ctx = handler({ req });
  if (ctx.res.body || ctx.res.status) {
      const status = ctx.res.status || 200;
      const headers = ctx.res.headers || {};
      if (ctx.res.body) {
        return c.body(ctx.res.body, status as never, headers);
      }
      return c.body(null, status as never, headers);
    }
    await next();
});
app.notFound((c) => {
  const file = path.join(config.distDir, DIST_PUBLIC, '404.html');
  if (fs.existsSync(file)) {
    return c.html(fs.readFileSync(file, 'utf8'), 404);
  }
  return c.text('404 Not Found', 404);
});

export default app.fetch;
```

Then... a dev could copy that file into their project and it would automatically use the "injected" Hono anywhere it's used in the app. Waku doesn't need to access Hono at all, so there won't be a conflict.

This reminds me a little bit of ejecting from a create-react-app. By default, Waku bundles Hono as a default server, but you can create your own server.

dai-shi:
I like the basic idea. It's also related with #1388, right?
This is huge, and will take time.
For now, I would like to use unstable_honoEnhancer to merge this PR specifically.
And, follow up later after more refactors. Give me time.
(My concern is if this can be integrated into `server-entry.tsx` somehow. Otherwise, it's too complicated. If it goes well, we can drop `unstable_honoEnhancer`.)

dai-shi:
I know this is a suboptimal solution. We'll definitely revisit this (or more customizable server entry) and provide a better solution.

The key message here is that Waku core isn't responsible for hono context (or hono entirely.)

PR #1644, opened 2025-08-20
fix(create-waku): Template chosen through prompt was always overwritten

jakekasan:
This fixes a bug where if you ran `npm create waku@latest -- --choose`, whatever option you chose for the template was ignored and the first template, "01_template", was used.


dai-shi:
Nice catch!

@tylersayshi I wonder if you have time for a quick review.

tylersayshi:
@jakekasan it should be possible to add a test for this I think?

https://github.com/wakujs/waku/blob/main/packages/create-waku/src/tests/cli-with-args.test.ts

our tests for the cli run only using the cli args, but it looks from your change like this bug would be reproducible with both the args and the interactive mode.

jakekasan:
@tylersayshi I'm not sure it can be reproduced outside of interactive mode? The issue is that the option chosen in the prompt is ignored, but the prompt is bypassed if a choice is already provided in the cli args.

I'll add a test for the interactive mode.

tylersayshi:
> I'll add a test for the interactive mode.

I've tried this earlier and was unable to test it in a non-flakey way reliably.

packages/create-waku/src/index.ts
```diff
@@ -172,7 +172,6 @@ async function doPrompts() {
     return {
       ...results,
       packageName: results.packageName || toValidPackageName(targetDir),
-      templateName: values.template ?? templateNames[0]!,
```
tylersayshi, line 175: is it maybe supposed to be this?

```ts
      templateName: results.templateName ?? templateNames[0]!,
```

I believe by deleting this line the template_01 default is also removed?

jakekasan:
> > I'll add a test for the interactive mode.
> 
> I've tried this earlier and was unable to test it in a non-flakey way reliably.

One approach might be to do what clack do themselves and just mock stdin and stdout for the whole thing and test the prompts directly (https://github.com/bombshell-dev/clack/blob/main/packages/prompts/test/select.test.ts#L28) but that seems extreme.

packages/create-waku/src/index.ts
jakekasan: It could be but is there ever a case where `results.templateName` would be false-y?

tylersayshi:
@jakekasan that was my exact conclusion when looking into it the first time around. 

At that point we're retesting clack itself just to test our own interactive mode.

Maybe theres another decent way to do it somehow... but idk

packages/create-waku/src/index.ts
tylersayshi: I think... `values.choose` can be falsy while template and example are both undefined which would make `results.templateName` undefined and falsy

jakekasan: Any chance it would be a better fit here? https://github.com/wakujs/waku/blob/061a60a8b58ad901e5282c7bddb5c1249ace3443/packages/create-waku/src/index.ts#L156-L166

Changing line 158 to
```javascript
return new Promise.resolve(values.template || values.example || templateNames[0]);
```
should cover the case where `values.choose` is false/undefined and `values.template` and `values.example` are both undefined and means that all the logic for what template to use is in this one function?

jakekasan:
@tylersayshi So definitely not decent and probably very flaky, but this is the only way I could imagine testing this. Basically just send "up" and "enter" to the subprocess and then check for dependencies which aren't in the default template's `package.json`.

Works only so long as the final template doesn't change. Alternatively could read whatever is last in the template's directory at test-time and then check for that. It's very hacky, so I'm happy to just remove it if it's unsuitable.

packages/create-waku/src/index.ts
tylersayshi: ya that seems cleaner to me too! thanks!

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -83,6 +83,20 @@ describe('create-waku CLI with args', () => {
     expect(stdout).toContain('Choose a starter template');
   });
 
+  test('choosing an option chooses that option', async () => {
+    const cmd = execa({
+      cwd: import.meta.dirname, timeout: 2000, reject: false
+    })`node ${CLI_PATH} --project-name ${projectName} --choose --skip-install`
+    cmd.stdin.write("k\r\n")
```
tylersayshi: can you leave a comment explaining what this is typing?

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -83,6 +83,20 @@ describe('create-waku CLI with args', () => {
     expect(stdout).toContain('Choose a starter template');
   });
 
+  test('choosing an option chooses that option', async () => {
+    const cmd = execa({
+      cwd: import.meta.dirname, timeout: 2000, reject: false
+    })`node ${CLI_PATH} --project-name ${projectName} --choose --skip-install`
```
tylersayshi: would it make sense to define the executor at the top of this file after the imports so that it can be reused/read a lil easier?

```suggestion
const executor = execa({
  cwd: import.meta.dirname, timeout: 2000, reject: false
})
    //...
    
    const cmd = executor`node ${CLI_PATH} --project-name ${projectName} --choose --skip-install`
```

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -83,6 +83,20 @@ describe('create-waku CLI with args', () => {
...
+    cmd.stdin.write("k\r\n")
+    await cmd
+    const packageJsonPath = path.join(genPath, "package.json")
+    expect(fs.existsSync(packageJsonPath)).toBe(true)
+    const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath).toString())
+    const dependencies = Object.keys(packageJsonContent.dependencies || {})
+    expect(dependencies).toContain("jotai")
+    expect(dependencies).toContain("waku-jotai")
+  }, 15000)
```
tylersayshi: this is failing prettier. can you run `prettier --write .`?

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -83,6 +83,20 @@ describe('create-waku CLI with args', () => {
...
+      cwd: import.meta.dirname, timeout: 2000, reject: false
+    })`node ${CLI_PATH} --project-name ${projectName} --choose --skip-install`
+    cmd.stdin.write("k\r\n")
+    await cmd
+    const packageJsonPath = path.join(genPath, "package.json")
+    expect(fs.existsSync(packageJsonPath)).toBe(true)
+    const packageJsonContent = JSON.parse(fs.readFileSync(packageJsonPath).toString())
+    const dependencies = Object.keys(packageJsonContent.dependencies || {})
+    expect(dependencies).toContain("jotai")
+    expect(dependencies).toContain("waku-jotai")
```
tylersayshi: this will potentially fail if we re-order the templates, right?

can you either comment the assumptions this test makes wrt order or make it resiliant to re-ordering.

packages/create-waku/src/tests/cli-with-args.test.ts
jakekasan: Went a slightly different route to try to preserve the logging of stdout and stderr in the event of a failed test - see new changes

packages/create-waku/src/tests/cli-with-args.test.ts
jakekasan: Yeah this was far too brittle. I've changed it now to match the expected dependencies to the index of the template in the directory, and then "press" down the amount of times corresponding to that index.

This way the ordering shouldn't really matter (unless `fs.readdirSync` returns a different order in the test compared to the CLI...)

The benefit here with this approach is that the choosing can be parameterised. The downside is that most of the templates have the same dependency list, so I filtered out the duplicates but there might be a more elegant way of doing that.

There's also the question of whether the dependencies in package.json are the best way to determine which template actually got created... 🤷‍♂️

packages/create-waku/src/index.ts
```diff
@@ -277,6 +278,7 @@ async function init() {
 
 init()
   .then(notifyUpdate)
+  .then(() => process.exit(0))
```
tylersayshi: is this to make it so the tests don't hang and they exit?

dai-shi: If it's test only, can we avoid it?
```suggestion
```

jakekasan:
I think I solved it... with a running child process, you have to explicitly close STDIN so that it knows nothing more will come in and it can end. This means there is no need for a timeout in the test or an explicit `process.exit` in the actual script. Latest commit fixes this.

packages/create-waku/src/tests/cli-with-args.test.ts
```diff
@@ -189,4 +253,29 @@ describe('create-waku CLI with args', () => {
...
+          cwd: import.meta.dirname,
+          reject: false,
+        },
+      );
+      // We input 'j' (vimspeak for down) x times depending on the template's index
+      // \r\n simulates enter
+      const keyStrokes = new Array(index).fill('j').join('');
+      cmd.stdin.write(`${keyStrokes}\r\n`);
+      // close stdin otherwise the process will hang and test will timeout
+      cmd.stdin.end();
```
dai-shi, line 273: Nice! I thought it's related with something with stdio, but didn't notice this.

PR #1647, opened 2025-08-21
chore: update deps

dai-shi:


packages/waku/package.json
```diff
@@ -78,17 +78,17 @@
     "node": "^24.0.0 || ^22.12.0 || ^20.19.0"
   },
   "dependencies": {
-    "@hono/node-server": "1.18.1",
-    "@swc/core": "1.13.3",
+    "@hono/node-server": "1.19.0",
+    "@swc/core": "1.13.4",
     "@vitejs/plugin-react": "5.0.1",
-    "@vitejs/plugin-rsc": "0.4.19",
+    "@vitejs/plugin-rsc": "0.4.21",
```
nojaf, line 84: There is a 0.4.22 version here, which has a fix for my project. Could we have a `^` maybe for this dependency?

dai-shi: oh, was it published after this PR was created?

dai-shi: #1649

nojaf: Yes, as part of https://www.npmjs.com/package/@vitejs/plugin-rsc/v/0.4.22

PR #1668, opened 2025-09-04
use route path for layout path

### LINKED ISSUE #1665: layouts should use router props as pages do
I'm trying to create a static page with Waku. It will have a language switcher, which will stay in the header of the page, so it should be moved into the layout. The current language is the first path segment. Unfortunately this path segment will not be usable in the layout and the layout will not render differently based on this path segment.

So the structure is:

pages
| [lang]
| | _layout.tsx <-- this should react to the lang path segment

There are ways to solve this problem:

1. Making the language switcher a client component, that changes based on the navigation state.
2. Making the layout render dynamic and fetch the current language manually from the path.
3. Moving the language switcher into every page component.

I'd really like to have a solution for doing this with static rendering in the layout and without a client component.

tylersayshi:
initial attempt, this will need testing

fixes #1665

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -837,13 +837,11 @@ export const createPages = <
           dynamicLayoutPathMap.get(segment)?.[1] ??
           staticComponentMap.get(joinPath(segment, 'layout').slice(1)); // feels like a hack
 
-        const isDynamic = dynamicLayoutPathMap.has(segment);
-
         if (layout && !Array.isArray(layout)) {
           const id = 'layout:' + segment;
           result[id] = createElement(
             layout,
-            isDynamic ? { path } : null,
+            { path: routePath },
             createElement(Children),
           );
         } else {


rmarscher:
This is a good start :+1: It would  be nice to pass if we passed the parent route segments to the layout in addition to `path`.

I noticed there is an issue with static layouts that have route segments (slugs). I think the layout stays cached on the client instead of being re-rendered with the route segment changed. 

https://github.com/rmarscher/waku-examples/blob/main/examples/layout-route-props/src/pages/%5Blang%5D/_layout.tsx#L7

^ in this example, the layout doesn't re-render when the [lang] changes.

https://github.com/user-attachments/assets/b1a4b2b9-1e8a-4f1d-910c-cd5fb49fb66e

rmarscher:
Based on further discussion in #1665, we want to fix the cache issue and also pass route segments to the layout component.

We'll need to adjust the typescript [CreateLayout](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L174C13-L188) definition to accept staticPaths.

We'll need to update [createLayout](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L530-L546) to handle staticPaths similar to [createPages](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L530-L546). Maybe there is a way to refactor createPage and createLayout to share code.

I think we'll also need to separate the page static paths from the layout static paths. See [getOriginalStaticPathSpec](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L393C9-L398) which is using the staticPathMap set during createPages.

With the issue I noted above, I think [getLayouts](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L615-L629) is the function that needs to be adjusted. Or maybe the defineRouter [getConfig](https://github.com/wakujs/waku/blob/layout-path/packages/waku/src/router/create-pages.ts#L648-L675).

We should create a breaking test first. @fuma-nama offered to help. I probably won't have much time to contribute until next weekend.

tylersayshi:
> We should create a breaking test first. @fuma-nama offered to help. I probably won't have much time to contribute until next weekend.

Thanks for the help on this! I'm also limited on contribution time, I'll be happy to review that once it's ready 🙌

PR #1671, opened 2025-09-06
breaking: customizable server entry

dai-shi:
- [x] introduce createApp and unstable_honoMiddleware
- [x] replace honoEnhancer and middleware
- [x] fix cloud deploy adapters
- I wonder if I should consider #1388 idea 👉 #1707

### DIFF
--- a/e2e/fixtures/broken-links/src/server-entry.tsx
+++ b/e2e/fixtures/broken-links/src/server-entry.tsx
@@ -1,9 +1,8 @@
 /// <reference types="vite/client" />
 import { Hono } from 'hono';
-import { unstable_fsRouter as fsRouter } from 'waku/router/server';
-import { unstable_defineServer as defineServer } from 'waku/minimal/server';
 import { unstable_honoMiddleware as honoMiddleware } from 'waku/internals';
-
+import { unstable_defineServer as defineServer } from 'waku/minimal/server';
+import { unstable_fsRouter as fsRouter } from 'waku/router/server';
 import redirectsMiddleware from './middleware/redirects';
 
 export default defineServer({
--- a/packages/waku/src/lib/types.ts
+++ b/packages/waku/src/lib/types.ts
@@ -1,5 +1,4 @@
 import type { ReactNode } from 'react';
-
 import type { Config } from '../config.js';
 
 type Elements = Record<string, unknown>;
--- a/packages/waku/src/lib/vite-rsc/plugin.ts
+++ b/packages/waku/src/lib/vite-rsc/plugin.ts
@@ -18,8 +18,8 @@ import { INTERNAL_setAllEnv, unstable_getBuildOptions } from '../../server.js';
 import {
   DIST_PUBLIC,
   SRC_CLIENT_ENTRY,
-  SRC_SERVER_ENTRY,
   SRC_PAGES,
+  SRC_SERVER_ENTRY,
 } from '../constants.js';
 import {
   getManagedClientEntry,
--- a/e2e/fixtures/broken-links/package.json
+++ b/e2e/fixtures/broken-links/package.json
@@ -9,7 +9,7 @@
     "start": "waku start"
   },
   "dependencies": {
-    "hono": "4.9.10",
+    "hono": "4.9.11",
     "react": "19.2.0",
     "react-dom": "19.2.0",
     "react-server-dom-webpack": "19.2.0",
--- a/e2e/fixtures/ssr-catch-error/package.json
+++ b/e2e/fixtures/ssr-catch-error/package.json
@@ -9,7 +9,7 @@
     "start": "waku start"
   },
   "dependencies": {
-    "hono": "4.9.10",
+    "hono": "4.9.11",
     "react": "19.2.0",
     "react-dom": "19.2.0",
     "react-error-boundary": "^6.0.0",
--- a/e2e/fixtures/ssr-catch-error/src/server-entry.tsx
+++ b/e2e/fixtures/ssr-catch-error/src/server-entry.tsx
@@ -1,9 +1,8 @@
 /// <reference types="vite/client" />
 import { Hono } from 'hono';
-import { unstable_fsRouter as fsRouter } from 'waku/router/server';
-import { unstable_defineServer as defineServer } from 'waku/minimal/server';
 import { unstable_honoMiddleware as honoMiddleware } from 'waku/internals';
-
+import { unstable_defineServer as defineServer } from 'waku/minimal/server';
+import { unstable_fsRouter as fsRouter } from 'waku/router/server';
 import validatorMiddleware from './middleware/validator';
 
 export default defineServer({
--- a/examples/07_cloudflare/src/server-entry.tsx
+++ b/examples/07_cloudflare/src/server-entry.tsx
@@ -1,10 +1,9 @@
 /// <reference types="vite/client" />
 import { Hono } from 'hono';
 import { contextStorage, getContext } from 'hono/context-storage';
-import { unstable_fsRouter as fsRouter } from 'waku/router/server';
-import { unstable_defineServer as defineServer } from 'waku/minimal/server';
 import { unstable_honoMiddleware as honoMiddleware } from 'waku/internals';
-
+import { unstable_defineServer as defineServer } from 'waku/minimal/server';
+import { unstable_fsRouter as fsRouter } from 'waku/router/server';
 import cloudflareMiddleware from './middleware/cloudflare';
 
 export default defineServer({
--- a/packages/waku/src/lib/hono/middleware/not-found.ts
+++ b/packages/waku/src/lib/hono/middleware/not-found.ts
@@ -1,8 +1,8 @@
-import path from 'node:path';
 import fs from 'node:fs';
+import path from 'node:path';
 import type { MiddlewareHandler } from 'hono';
-import type { Unstable_CreateAppArgs as CreateAppArgs } from '../../types.js';
 import { DIST_PUBLIC } from '../../constants.js';
+import type { Unstable_CreateAppArgs as CreateAppArgs } from '../../types.js';
 
 export function notFoundMiddleware(args: CreateAppArgs): MiddlewareHandler {
   const { config } = args;
--- a/packages/waku/src/lib/hono/middleware/static.ts
+++ b/packages/waku/src/lib/hono/middleware/static.ts
@@ -1,8 +1,8 @@
 import path from 'node:path';
-import type { MiddlewareHandler } from 'hono';
 import { serveStatic } from '@hono/node-server/serve-static';
-import type { Unstable_CreateAppArgs as CreateAppArgs } from '../../types.js';
+import type { MiddlewareHandler } from 'hono';
 import { DIST_PUBLIC } from '../../constants.js';
+import type { Unstable_CreateAppArgs as CreateAppArgs } from '../../types.js';
 
 export function staticMiddleware(args: CreateAppArgs): MiddlewareHandler {
   const { config, isBuild } = args;
--- a/packages/waku/src/lib/utils/managed.ts
+++ b/packages/waku/src/lib/utils/managed.ts
@@ -1,4 +1,4 @@
-import { SRC_PAGES, EXTENSIONS } from '../constants.js';
+import { EXTENSIONS, SRC_PAGES } from '../constants.js';
 
 export const getManagedServerEntry = (config: { srcDir: string }) => {
   const globBase = `/${config.srcDir}/${SRC_PAGES}`;
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -97,8 +97,8 @@ importers:
   e2e/fixtures/broken-links:
     dependencies:
       hono:
-        specifier: 4.9.10
-        version: 4.9.10
+        specifier: 4.9.11
+        version: 4.9.11
       react:
         specifier: 19.2.0
         version: 19.2.0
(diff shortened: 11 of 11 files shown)


packages/waku/src/lib/vite-rsc/deploy/aws-lambda/entry.ts
```diff
@@ -29,4 +29,4 @@ export const handler: any = import.meta.env.WAKU_AWS_LAMBDA_STREAMING
   ? honoAwsLambda.streamHandle(app)
   : honoAwsLambda.handle(app);
 
-export { handleBuild } from '../../build.js';
+export { processBuild } from '../../build.js';
```
dai-shi: @hi-ogawa I'm not sure how this export is used. Could you explain?

hi-ogawa: The code is here. It imports `dist/server/index.js` after the build and runs `handleBuild` export.
(Rsc plugin's `buildApp` runs first, then Waku's `buildApp` runs 2nd, so `dist/server/index.js` is ready at this point.)

https://github.com/wakujs/waku/blob/d3d84cf94cc360c103795172f092b3e64f4f150a/packages/waku/src/lib/vite-rsc/plugin.ts#L394-L409

dai-shi: Thanks! And, I missed `pathToFileURL(entryPath).href`... Yes, it makes totally sense.

packages/waku/src/lib/types.ts
```diff
@@ -44,31 +46,26 @@ export type Unstable_HandleBuild = (utils: {
   generateDefaultHtml: (pathname: string) => Promise<void>;
 }) => Promise<void>;
 
+export type Unstable_CreateAppArgs = {
+  processRequest: (req: Request) => Promise<Response | null>;
```
dai-shi, line 49: Now, I wonder if this could go away.

e2e/fixtures/broken-links/src/server-entry.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { unstable_defineServer as defineServer } from 'waku/minimal/server';
+import { unstable_honoMiddleware as honoMiddleware } from 'waku/server';
+
+import redirectsMiddleware from './middleware/redirects';
+
+export default defineServer({
+  ...fsRouter(
+    import.meta.glob('/src/pages/**/*.{tsx,ts}', { base: '/src/pages' }),
+    { apiDir: 'api', slicesDir: '_slices' },
+  ),
```
dai-shi: This doesn't feel a very nice developer experience. 🤔

hi-ogawa: How about exporting manged server entry? something like

- server-entry.tsx

```js
import mangedServer from "waku/router/managed/server";

export default defineServer({
  ...managedServer,
  createApp() { ... },
}
```

- waku/router/managed/server.js

```js
export { default } from "virtual:vite-rsc-waku/manged-server-entry"
```

and this virtual can be implemented same as this

https://github.com/wakujs/waku/blob/4585faa63756180beb1998550ae0a5257da6ff04/packages/waku/src/lib/vite-rsc/plugin.ts#L276-L278

packages/waku/src/lib/vite-entries/entry.server.tsx
```diff
@@ -1,35 +1,21 @@
...
 
-export default app.fetch;
+const app = new Hono();
+app.use(staticMiddleware(args));
+const newApp = createApp(args);
+app.use(notFoundMiddleware(args));
+
+export const fetch = newApp.fetch;
 
 export { processBuild } from '../vite-rsc/build.js';
```
hi-ogawa: One idea I have is to (partially) move out `processBuild` logic from server entry (i.e. runtime) to plugin `buildApp` logic, which should make it unnecessary to export this as build output. Instead, server entry only needs to export `default.handleBuild`, which probably makes it easier for custom server entry.

hi-ogawa: I just tried a different idea https://github.com/wakujs/waku/pull/1692, which also eliminates the need for custom server entry to re-export `processBuild`.

e2e/fixtures/broken-links/src/server-entry.tsx
dai-shi: Thanks for the suggestion. Though, I'm not a big fan of `waku/router/managed/server` or any "managed" things visible. 🙃  

I'm currently thinking about #1388, which would eliminate this issue entirely.

hi-ogawa: I'm not entirely sure about the idea there. Is it something like convention for `src/server.js` in addition to current `src/server-entry.js`?

dai-shi: Yeah, I thought the single entry point `src/server-entry.js` would be good, which this PR does. It added `createApp` along with handleRequest, which is not what I originally planned.

Yes, what I have in mind and I'd try with another PR based on this will be `src/server.js` or `src/start-server.js` (in addition to `src/server-entry.js`).

hi-ogawa: Though I thought those two are confusing, it might be fine since users needing both custom entries are unlikely.
So, for honoEnhancer and custom middleware use case, do they go to this new kind of server entry?

dai-shi: Yes. I'm not 100% sure if the separate approach would be good.

I generally prefer the single entry. So, this PR tries that.

Maybe my question to you should be this:

```js
import { unstable_fsRouter as fsRouter } from "waku/router/server";

export default defineServer({
  ...fsRouter(),
  createApp() { ... },
}
```

```js
import {
  unstable_getSrcPagesFiles as getSrcPagesFiles,
  unstable_getConfig as getConfig,
} from 'waku/server';

export unstable_fsRouter(
  pagesFiles = getSrcPagesFiles(),
  config = getConfig(),
) { ... }
```

```js
export { getSrcPagesFiles as unstable_getSrcPagesFiles } from 'virtual:...';
export { getConfig as unstable_getConfig } from 'virtual:...';
```

Is that possible? I hope so. `getSrcPagesFiles` feels too specific, though.

dai-shi: > I generally prefer the single entry. So, this PR tries that.

On second thought, if we were to eliminate deploy adapters, having single entry file would be impossible as the requirement of deploy adapters are different. (cloudflare to export `fetch` and, node to start the server, for instance.)

dai-shi: > would be impossible

It's not entirely true. We could do something like:
```js
export const entries = fsRouter(...);

// for cloudflare case
export default { fetch: ... }

// for node
import { serve } from '@hono/node-server';
serve(...);
```

Hmm, feels tricky. I think having two separate files would be more straightforward.

packages/waku/src/lib/types.ts
dai-shi: This is necessary to isolate hono from vite. We may revisit it in follow-up PRs.

packages/waku/src/lib/vite-rsc/plugin.ts
```diff
@@ -275,59 +274,21 @@ if (import.meta.hot) {
...
-      `;
-    }),
     createVirtualPlugin('vite-rsc-waku/config', async function () {
+      const globBase = `/${config.srcDir}/${config.pagesDir}/`;
+      const globPattern = `${globBase}**/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`;
       return `
         export const config = ${JSON.stringify({ ...config, vite: undefined })};
         export const flags = ${JSON.stringify(flags)};
         export const isBuild = ${JSON.stringify(this.environment.mode === 'build')};
+        export const globSrcPages = import.meta.glob(${JSON.stringify(globPattern)}, { base: ${JSON.stringify(globBase)} });
```
hi-ogawa: One concern about exporting `import.meta.glob` here is that Vite will try to process the `pages/**` files for non managed mode users. They might not even be tree-shaken depending on configuration or other plugins.

That's why I was suggesting to expose from new dedicated entry like `waku/router/managed/server`.

dai-shi: Yeah, I had the same concern. Isn't there any other solutions? Let's assume we don't have managed mode, but we have two fs routers, `fsRouter` and `fsRouter2`.

hi-ogawa: I don't think it's possible. If `import.meta.glob` is reachable from `waku/router/server`, then any users (or internally) importing `waku/router/server` will get `pages/**` processed.

dai-shi: How about injecting it at the second iteration?
```js
export const getModulesInSrcPages = () => globalThis.__WAKU_MODULES_IN_SRC_PAGES__;
```
and replace it _if_ it isn't  tree-shaked.

hi-ogawa: tree-shaking is decided after all transform is finished, so for example, it's too late to try to replace `__WAKU_MODULES_IN_SRC_PAGES__` during `renderChunk` hook. If we replaced it during `transform`, then it's too early and the situation is the same as writing `import.meta.glob` directly in the code.

dai-shi: Previously, I think I tried tree-shaking before vite transform happens. It's a bit too tricky though.

> They might not even be tree-shaken depending on configuration

Hm, like if we pass the first argument to fsRouter(), the glob will remain. So, tree-shaking doesn't help.

image

dai-shi: I think I'd go with this.

image

hi-ogawa: I think there's a gap in the concerns we are thinking.

Tree-shaking away the modules from `import.meta.glob` in the final output should be possible by tweaking exports from a single file. But, it's not possible to not let those modules processed through Vite / Rollup's transforms. 

The latter might not cause an actual issue except it's causing to process more unnecessary modules during build (which eventually tree-shaken in the end). It's like a similar phenomena as how barrel exports do.

dai-shi: Hm, okay, can you explain your concern again? When it happens and what the downside is.

hi-ogawa: Honestly, I think what I'm thinking is a non-issue and no downside. :sweat_smile: It would be same as, when importing library with barrel exports, all the modules will be processed even though some are tree shaken in the final output.

examples/11_fs-router/src/server-entry.tsx
```diff
@@ -2,6 +2,5 @@
 import { unstable_fsRouter as fsRouter } from 'waku/router/server';
 
 export default fsRouter(
-  import.meta.glob('/src/pages/**/*.{tsx,ts}', { base: '/src/pages' }),
-  { apiDir: 'api', slicesDir: '_slices' },
+  import.meta.glob('./pages/**/*.tsx', { base: './pages' }),
```
dai-shi: @hi-ogawa I would like to make this to work as this is technically not a virtual module. Is it possible?

dai-shi: https://vite.dev/guide/features#base-path
Hm, I think I know the workaround.

packages/waku/src/lib/vite-rsc/plugin.ts
dai-shi: I just reverted `globSrcPages`.

examples/11_fs-router/src/server-entry.tsx
dai-shi: `import.meta.glob('./pages/**/*.tsx', { base: './pages' })` doesn't work,
but `import.meta.glob('./**/*.tsx', { base: './pages' })` works.

packages/waku/src/lib/vite-rsc/deploy/cloudflare/entry.ts
```diff
@@ -1,10 +1,15 @@
 import { Hono } from 'hono';
-import { createHonoHandler } from '../../engine.js';
-import { honoEnhancer } from 'virtual:vite-rsc-waku/hono-enhancer';
+import { config, isBuild } from 'virtual:vite-rsc-waku/config';
+import serverEntry from 'virtual:vite-rsc-waku/server-entry';
```
rmarscher, line 3: I think we can create a solution for defining [durable objects](https://developers.cloudflare.com/durable-objects/api/base/) and additional [handlers](https://developers.cloudflare.com/workers/runtime-apis/handlers/)? Durable Objects have to be defined as top level exports of this file and the additional handlers need to be added to the default export of this file.

Maybe
```
import cloudflareEntry from 'virtual:vite-rsc-waku/cloudflare-entry';
export * from 'virtual:vite-rsc-waku/cloudflare-entry';
```

and we build src/cloudflare-entry.tsx to that virtual module via a vite plugin?

https://github.com/wakujs/waku/issues/1352

Once we merge #1655, those features should work in dev mode too.

packages/waku/src/lib/vite-rsc/deploy/cloudflare/entry.ts
```diff
@@ -29,7 +34,7 @@ export default {
   fetch: (request: Request, env: any, ctx: any) => {
```
rmarscher, line 34: `...(cloudflareEntry || {})` could be placed before this line to merge any default exports defined in the cloudflareEntry file. https://developers.cloudflare.com/workers/runtime-apis/handlers/

packages/waku/src/lib/vite-rsc/deploy/cloudflare/entry.ts
dai-shi: @rmarscher Please check #1707, which I would like to merge after this one.

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/18367933575

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/18430240953

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/18430788415

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/18438647871

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/18440564876

PR #1680, opened 2025-09-08
fix: route not found error when access to static segment route in group route

### LINKED ISSUE #1669: Segment route under group route returns "Route not found"
Hello,

I’d like to report an issue where using a Segment Route inside a Group Route results in "Route not found."

For example, suppose we have the following file:

```tsx
// src/pages/(hoge)/blog/[slug].tsx

import type { PageProps } from 'waku/router';

// Create blog article pages
export default async function BlogArticlePage({
  slug,
}: PageProps<'/blog/[slug]'>) {
  return <div>{slug}</div>;
}

  return {
    render: 'static',
    staticPaths: ['introducing-waku', 'introducing-pages-router'],
  } as const;
};
```

In this state, accessing `/blog/introduction-waku` results in `Route not found: /blog/introducing-waku`

This only happens when the Segment Route is inside a Group Route. If I move the file above to `src/pages/blog/[slug].tsx`, the issue is resolved.

Minimal reproducible repo: https://stackblitz.com/~/github.com/r4ai/waku-segment-routes-issue

r4ai:
close #1669

- [x] add a failing e2e test
- [x] fix the bug
- [x] add unit tests
- [x] refactor the code


packages/waku/src/router/create-pages.ts
```diff
@@ -501,7 +501,7 @@ export const createPages = <
         if (pagePath !== definedPath) {
           groupPathLookup.set(pagePath, definedPath);
         }
-        const id = joinPath(...pathItems, 'page');
+        const id = joinPath(pagePath, 'page').replace(/^\//, '');
```
r4ai, line 504: I fixed it to use groupless path.
This should be the same as what the static page without slugs does:

https://github.com/r4ai/waku/blob/01b1d3c8bcf833fd9c128d7c534536adedc90e75/packages/waku/src/router/create-pages.ts#L451-L456

As a result, the route path should be retrieved by the following `if` statement, and the error should be fixed.

https://github.com/r4ai/waku/blob/01b1d3c8bcf833fd9c128d7c534536adedc90e75/packages/waku/src/router/create-pages.ts#L343-L345

PR #1683, opened 2025-09-09
Support accessing segment params from layouts

fuma-nama:
addressing #1665 

- pass the props from `createPages()`

need to ensure caching & other stuffs first. Ideally, also make the params available to `useRouter()`, or Next-style `useParams()` if needed.

### DIFF
--- a/packages/waku/src/router/create-pages.ts
+++ b/packages/waku/src/router/create-pages.ts
@@ -18,7 +18,7 @@ import type {
   PropsForPages,
 } from './create-pages-utils/inferred-path-types.js';
 import { Children, Slot } from '../minimal/client.js';
-import { ErrorBoundary } from '../router/client.js';
+import { ErrorBoundary } from './client.js';
 
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
 export const METHODS = [
@@ -428,7 +428,7 @@ export const createPages = <
     }
 
     const pathSpec = parsePathWithSlug(page.path);
-    const { numSlugs, numWildcards } = getSlugsAndWildcards(pathSpec);
+    const { numSlugs, numWildcards } = countSlugsAndWildcards(pathSpec);
     if (page.unstable_disableSSR) {
       noSsrSet.add(pathSpec);
     }
@@ -612,20 +612,36 @@ export const createPages = <
     await ready;
   };
 
-  const getLayouts = (spec: PathSpec): string[] => {
-    const pathSegments = spec.reduce<string[]>(
-      (acc, _segment, index) => {
-        acc.push(pathSpecAsString(spec.slice(0, index + 1)));
-        return acc;
-      },
-      ['/'],
-    );
+  const getLayouts = (spec: PathSpec) => {
+    const layoutInfos: {
+      path: string;
+      spec: PathSpec;
+      component: ComponentEntry | FunctionComponent<any>;
+      isDynamic: boolean;
+    }[] = [];
 
-    return pathSegments.filter(
-      (segment) =>
-        dynamicLayoutPathMap.has(segment) ||
-        staticComponentMap.has(joinPath(segment, 'layout').slice(1)), // feels like a hack
-    );
+    for (let i = -1; i < spec.length; i++) {
+      const subSpec = spec.slice(0, i + 1);
+      const key = pathSpecAsString(subSpec);
+      const dynamicLayout = dynamicLayoutPathMap.get(key)?.[1];
+      const staticLayout = staticComponentMap.get(
+        joinPath(key, 'layout').slice(1),
+      );
+      const component = dynamicLayout ?? staticLayout;
+
+      if (!component) {
+        continue;
+      }
+
+      layoutInfos.push({
+        path: key,
+        spec: subSpec,
+        component,
+        isDynamic: dynamicLayout !== undefined,
+      });
+    }
+
+    return layoutInfos;
   };
 
   const definedRouter = unstable_defineRouter({
@@ -644,21 +660,15 @@ export const createPages = <
       const rootIsStatic = !rootItem || rootItem.render === 'static';
       for (const [path, { literalSpec, originalSpec }] of staticPathMap) {
         const noSsr = noSsrSet.has(literalSpec);
-
-        const layoutPaths = getLayouts(originalSpec ?? literalSpec);
-
         const elements = {
-          ...layoutPaths.reduce<Record<string, { isStatic: boolean }>>(
-            (acc, lPath) => {
-              acc[`layout:${lPath}`] = {
-                isStatic: !dynamicLayoutPathMap.has(lPath),
-              };
-              return acc;
-            },
-            {},
-          ),
           [`page:${path}`]: { isStatic: staticPathMap.has(path) },
         };
+
+        for (const layout of getLayouts(originalSpec ?? literalSpec)) {
+          elements[`layout:${layout.path}`] = {
+            isStatic: !layout.isDynamic,
+          };
+        }
 
         routeConfigs.push({
           type: 'route',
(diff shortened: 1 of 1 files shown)


fuma-nama:
I may overlook some internal details of Waku router's behaviours, lmk know if there's any problems in the design @dai-shi

rmarscher:
I think part of the problem is that the originalSpec is always used over the literalSpec in getConfig.
https://github.com/wakujs/waku/blob/main/packages/waku/src/router/create-pages.ts#L648
```ts
const layoutPaths = getLayouts(originalSpec ?? literalSpec);
```

If originalSpec contains segments in the layout part of the path, then it should use the corresponding literalSpec path for the layout id.

The layout ID was updated [here](https://github.com/wakujs/waku/pull/1683/files#diff-aa5fd10e2174215f8433b37357a7b91805c9a9d4d00c232fd563ce9f1119651dR818-R819) but I think we want the path part of the ID to include the replaced segments when it's not dynamic. And then we don't need the `(${trimmed.segment})` part of the layout ID.

For example,

```txt
/[lang]/about.tsx
/[lang]/_layout.tsx
```

If both are defined as static, and about.tsx has staticPaths of `['en', 'fr', 'ja']`,  then we want there to be

```txt
layout:/en
layout:/fr
layout:/ja
page:/en/about
page:/fr/about
page:/ja/about
```

in the elements map. 

Adding test coverage to https://github.com/wakujs/waku/blob/main/packages/waku/tests/create-pages.test.ts and https://github.com/wakujs/waku/blob/main/e2e/fs-router.spec.ts might help uncover other issues.

fuma-nama:
> If both are defined as static, and about.tsx has staticPaths of ['en', 'fr', 'ja'], then we want there to be

I will follow the current decisions in #1665 because otherwise /en/_layout.tsx will conflict with /[lang]/_layout.tsx

and need to update getConfig later

fuma-nama:
Just finished some necessary refactors to reduce duplication, should be consistent with all current behaviors (checked by unit tests).

fuma-nama:
note that there's some e2e test cases affected by `path` prop being removed, perhaps we can consider some backward compatibility layers? @dai-shi

dai-shi:
Can we simply drop `path` prop?
In that case, please open a new PR which will be merged before this one.
@tylersayshi What do you think?

tylersayshi:
> Can we simply drop `path` prop? In that case, please open a new PR which will be merged before this one. @tylersayshi What do you think?

What is the downside of keeping `path`? It seems still helpful to have the complete `path` prop?

packages/waku/src/router/create-pages.ts
```diff
@@ -1,24 +1,23 @@
...
 import type { PathSpec } from '../lib/utils/path.js';
 import type {
   AnyPage,
   GetSlugs,
+  PropsForLayouts,
   PropsForPages,
 } from './create-pages-utils/inferred-path-types.js';
 import { Children, Slot } from '../minimal/client.js';
-import { ErrorBoundary } from '../router/client.js';
+import { ErrorBoundary } from './client.js';
```
tylersayshi, line 20: was this moved?

packages/waku/src/router/create-pages.ts
```diff
@@ -256,12 +258,26 @@ const createNestedElements = (
...
+interface LayoutInfo {
+  path: string;
+  spec: PathSpec;
   component: FunctionComponent<any>;
-}[];
+  isDynamic: boolean;
 
-type ComponentEntry = FunctionComponent<any> | ComponentList;
+  staticSlotIds: Set<string>;
+  staticPaths?: string[][] | undefined;
```
tylersayshi, line 279: the `?` seems unnecessary given that `| undefined` is here

fuma-nama:
I have tried to minimise the change (most added loc is from unit tests), but yeah at certain degree, a bigger redesign will eventually happen as @dai-shi said in the original issue about the future definition of static and dynamic pages & layouts.

This PR only made the necessary adjustments for dynamic segments, and a little of foundation for pre-rendering (PageInfo API). 

Also yes then I will update the e2e tests instead

dai-shi:
@tylersayshi https://github.com/wakujs/waku/pull/1683#issuecomment-3301319702

> What is the downside of keeping `path`? It seems still helpful to have the complete `path` prop?

I thought it's that it's impossible to always provide "correct" `path` prop.
Did you not agree with the change? https://github.com/wakujs/waku/issues/1665#issuecomment-3276851026

packages/waku/src/router/create-pages.ts
fuma-nama: It refers to same file

packages/waku/src/router/create-pages.ts
fuma-nama: Waku enabled strict undefined hence optional parameter (?) does not accept explicit undefined value

fuma-nama:
also I think as a consequence, layouts need to be dynamic by default

tylersayshi:
> Did you not agree with the change? 

ah, i just missed this, no issues with it from me.

tylersayshi:
> also I think as a consequence, layouts need to be dynamic by default

this would be a quite unfortunate break imo. it was a big decision for us to center around static as the default for everything except api routes. is there reasoning for this?

dai-shi:
> also I think as a consequence, layouts need to be dynamic by default

If you think so, I might be misunderstanding your design and impl. Could you elaborate?
We want to make all pages/layouts/slices static by default and our big target is building static sites.

fuma-nama:
If you create a layout under segment path, by default, it will cause error. It is quite counterintuitive if the dev must define a config and specify static paths, otherwise a new created layout without additional configuration cannot work.

And there's the same backward compatibility issue discussed before, it's a big problem for existing projects because many static layouts would need migration, this can be possibly avoided if we can make it loose, like auto including needed static paths from child pages (which reduces the layouts that require migration.

dai-shi:
I know the backward incompatibility, which seems okay and reasonable to me.

As I said, consider our primary target is static sites. So, our doc should strictly mention that `staticPaths` are required if static layout is under a slug. And, it shouldn't build if `staticPath` are missing or "wrong"; `waku build` should fail with such errors. This applies to non-static sites too. Does it make sense?

pmk1c:
One thought I had: Waku could allow static layouts in a segmented path not define `staticPaths`. In that case the layout does not receive the segments as props and there is only one layout id to cache, as waku did before, right? So the doc would say: If you want to access segment props in a static layout, you need to define staticPaths. Not sure if that's a good option to have.

I'd go with making the backwards incompatible change for now and maybe iterate on other options (like the one described above) later if they are helpful and not to complex to implement (removing the necessity of defining staticPaths, will not be breaking).

dai-shi:
Hey @pmk1c, thanks for jumping in.

> Waku could allow static layouts in a segmented path not define `staticPaths`. In that case the layout does not receive the segments as props and there is only one layout id to cache, as waku did before, right?

We had a little discussion in #1665 and my answer is no. It feels too complicated implementation-wise and for me api-design-wise too (changing the behavior based on the existence of a prop doesn't feel right). If we were to do it, I would like to have another option (or another convention) to be explicit.

e2e/fixtures/create-pages/src/components/DynamicLayout.tsx
```diff
@@ -1,15 +1,9 @@
...
-  path: string;
-}) {
+export default function DynamicLayout({ children }: { children: ReactNode }) {
   return (
     <>
-      <nav>Current path: {path}</nav>
+      <nav>Current path: /dynamic</nav>
       {children}
     </>
   );
```
rmarscher, line 9: I think this should use

```ts
const path = getContext().req.url.pathname;
```

instead of hard-coded `/dynamic`. 

It will show people how to migrate if they are using `path` in a dynamic layout.

rmarscher (commented):
I think the backwards incompatible changes are ok. If you need static layouts in a path with route segments, you have to define staticPaths.

I think `new URL(getContext().req.url).pathname` should work to replace `path` for people that need it in a dynamic layout.

e2e/fixtures/create-pages/src/components/DynamicLayout.tsx
rmarscher: Oops. Is req.url a string? I guess it's `new URL(getContext().req.url).pathname`

dai-shi:
@fuma-nama Whenever you have time,

Here're action items:
- Open a separate PR to drop `path` prop (I'll merge it first.)
- Follow https://github.com/wakujs/waku/pull/1683#issuecomment-3310110280 and create e2e tests (Those will be my primary review points.)

e2e/fixtures/create-pages/src/components/DynamicLayout.tsx
dai-shi: Yes, `req` is now a plain `Request`.

dai-shi:
Thanks for working on this, but we took a different approach. https://github.com/wakujs/waku/issues/1665#issuecomment-3423859931

PR #1686, opened 2025-09-10
docs(readme): add a section about ignored routes

r4ai:


dai-shi:
@sandren Please review this.

PR #1687, opened 2025-09-10
fix: basePath

yuusheng:
fix: https://github.com/wakujs/waku/issues/698

### DIFF
--- a/packages/waku/src/router/client.ts
+++ b/packages/waku/src/router/client.ts
@@ -45,6 +45,7 @@ import {
 import type { RouteProps } from './common.js';
 import type { RouteConfig } from './base-types.js';
 import { getErrorInfo } from '../lib/utils/custom-errors.js';
+import { withoutTrialSlash } from '../lib/utils/path.js';
 
 type AllowPathDecorators<Path extends string> = Path extends unknown
   ? Path | `${Path}?${string}` | `${Path}#${string}`
@@ -57,12 +58,20 @@ type InferredPaths = RouteConfig extends {
   : string;
 
 const normalizeRoutePath = (path: string) => {
+  let _path = path;
+
+  const basePath = import.meta.env.WAKU_CONFIG_BASE_PATH;
+  if (_path.startsWith(basePath)) {
+    _path =_path.slice(basePath.length) || '/';
+  }
+
   for (const suffix of ['/', '/index.html']) {
-    if (path.endsWith(suffix)) {
-      return path.slice(0, -suffix.length) || '/';
+    if (_path.endsWith(suffix)) {
+      return _path.slice(0, -suffix.length) || '/';
     }
   }
-  return path;
+
+  return _path;
 };
 
 const parseRoute = (url: URL): RouteProps => {
@@ -136,6 +145,7 @@ const RouterContext = createContext<{
     (event: ChangeRouteEvent, handler: ChangeRouteCallback) => void
   >;
   fetchingSlices: Set<SliceId>;
+  basePath: string
 } | null>(null);
 
 export function useRouter() {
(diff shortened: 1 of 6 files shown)


yuusheng:
It's working now. But with some problems.
1. e2e tests
2. It should be localhost:8080/test/RSC/ instead of localhost:8080/testRSC/ when using Link component image

packages/waku/src/router/client.ts
```diff
@@ -57,12 +58,20 @@ type InferredPaths = RouteConfig extends {
   : string;
 
 const normalizeRoutePath = (path: string) => {
+  let _path = path;
```
dai-shi: Coding convention: We don't use `_` for "internal" thing. Please rename it.

packages/waku/src/router/client.ts
```diff
@@ -286,7 +296,6 @@ export type LinkProps = {
 } & Omit<AnchorHTMLAttributes<HTMLAnchorElement>, 'href'>;
 
 export function Link({
-  to,
```
dai-shi, line 289: We shouldn't remove this.

dai-shi (commented):
Just a quick review. I didn't read fully, so this is not a real review.

packages/waku/src/router/client.ts
yuusheng: I use const to = withoutTrialSlash(`${router?.basePath}${props.to}`) to redefine it. Or we must change everywhere using `to`.

packages/waku/src/router/client.ts
yuusheng: Done

packages/waku/src/router/client.ts
dai-shi: We use `props` in L398, and I don't think `to` should be included there.

yuusheng: Oh, I see. You're right

yuusheng: Done

yuusheng:
Something I would like to discuss. Now in production, when basePath = '/test', we can correctly render the page when opening http://localhost:8080/test. But we can also open http://localhost:8080. Do you think we should redirect to `/{basePath}` when the user enters `/`?

dai-shi:
I think it should be `404 Not Found`.

yuusheng:
Also make sense.

packages/waku/src/router/client.ts
```diff
@@ -928,6 +935,7 @@ const InnerRouter = ({
         prefetchRoute,
         routeChangeEvents,
         fetchingSlices: useRef(new Set<SliceId>()).current,
+        basePath: import.meta.env.WAKU_CONFIG_BASE_PATH
```
hi-ogawa, line 938: `import.meta.env.WAKU_CONFIG_BASE_PATH` is a constant, so passing it around as prop seems odd. What about defining this constant at the top of this file?

```js
const BASE_PATH = import.meta.env.WAKU_CONFIG_BASE_PATH;
```

packages/waku/src/minimal/client.ts
```diff
@@ -29,7 +30,7 @@ const DEFAULT_HTML_HEAD = [
   createElement('meta', { name: 'generator', content: 'Waku' }),
 ];
 
-const BASE_RSC_PATH = `${import.meta.env?.WAKU_CONFIG_BASE_PATH}${
+const BASE_RSC_PATH = `${import.meta.env?.WAKU_CONFIG_BASE_PATH && withTrialSlash(import.meta.env?.WAKU_CONFIG_BASE_PATH)}${
```
hi-ogawa, line 33: If any easier, we can normalize and ensure trailing slash here, so runtime don't need to worry about it

https://github.com/wakujs/waku/blob/4585faa63756180beb1998550ae0a5257da6ff04/packages/waku/src/lib/vite-rsc/plugin.ts#L106-L108

EDIT: Well, you indeed already ensure this in `packages/waku/src/lib/vite-rsc/plugin.ts`, so don't need this?

PR #1708, opened 2025-09-26
refactor: browser preamble

dai-shi:
Does this work?


dai-shi:
@hi-ogawa Is this valid?

dai-shi:
Let's assume it is. Otherwise, let's revert it later.

PR #1710, opened 2025-09-27
ci: fix window-latest

himself65:
Fixes: https://github.com/wakujs/waku/issues/1696


himself65:
SOOO SLOW. let's see second commit

dai-shi:
🎉

PR #1725, opened 2025-10-18
fix(cli): top level await error

### LINKED ISSUE #1724: "Detected unsettled top-level await" while running the `build` command
Hey,

while testing waku together with [renoun](https://github.com/souporserious/renoun) as a potential replacement for our nextjs based documentation, I have created a simple blog example to test if everything would work as expected.

Good message: It does!

But while running the `build` command, I get the following error message:

```
Warning: Detected unsettled top-level await at file:///<project-path>/node_modules/.pnpm/waku@0.26.1_@types+node@24.7.2_jiti@2.6.1_lightningcss@1.30.1_react-dom@19.2.0_react@19_97196f4ce0df8eba400e36359f4c5601/node_modules/waku/dist/cli.js:72
    await cli(cmd, values);
    ^

 ELIFECYCLE  Command failed with exit code 13.
```

But the static pages are generated as expected and I can see them in the `dist/public` directory and also running `pnpm preview` shows the expected content.

## Step to reproduce 

https://github.com/noxify/waku-renoun-ssg

> Tested it with `node@22.19.0` and `node@24.10.0`

```
git clone git@github.com:noxify/waku-renoun-ssg.git
cd waku-renoun-ssg
pnpm i
pnpm build
```

-----

If you need additional information or something else, please let me know.

Thanks!

tylersayshi:
I'm not 100% on the full why of this ocurring. This works as a way to fix the error in #1724 though. My best guess is that it has something to do with the top level await being called inside a conditional statement, but this seems like an easy enough small fix to just make ¯\_(ツ)_/¯
    
fixes #1724

### DIFF
--- a/packages/waku/src/cli.ts
+++ b/packages/waku/src/cli.ts
@@ -55,7 +55,7 @@ const { values, positionals } = parseArgs({
 
 const cmd = positionals[0];
 
-async function run()  {
+async function run() {
   if (values.version) {
     const { version } = require('../package.json');
     console.log(version);


hi-ogawa (approved):
I suspect there's a hanging promise bug either waku ssg or OP's tools, but I've commonly see this cli pattern, so I suppose it's safe to do.

packages/waku/src/cli.ts
```diff
@@ -91,3 +93,8 @@ Options:
       --help            Display this help message
 `);
 }
+
+run().catch((error) => {
+  console.error(error);
+  process.exit(1);
```
dai-shi, line 99: We might want to improve something in the future, but okay for now.

PR #1730, opened 2025-10-20
fix(create-pages): remove path prop from all layouts

tylersayshi:
related to #1665 and #1683

BREAKING CHANGE: path prop is no longer available for layouts


tylersayshi:
marked draft while we resolve: https://github.com/wakujs/waku/issues/1665#issuecomment-3423859931

tylersayshi:
should be ready - from the conversation linked above, not having this refactor is a dealbreaker for static layouts

dai-shi (approved):
This looks good and reasonable.

PR #1748, opened 2025-11-04
fix(typegen): generate in managed mode when server-entry.tsx exists

### LINKED ISSUE #1747: Typegen breaks when server-entry.ts exists
Since middleware was removed from `defineConfig` in waku.config.ts, the only remaining way to provide middleware as i see is by creating a server-entry.ts with `middlewareFns`. However, doing so breaks typegen (pages.gen.ts) because of the `entriesFilePossibilities` condition in fs-router-typegen.ts — it simply checks for the existence of the server-entry file and returns void if it exists.

As a result, there is currently no way to provide middleware without breaking the typegen (pages.gen.ts).

tylersayshi:
fixes #1747


packages/waku/src/lib/vite-plugins/fs-router-typegen.ts
```diff
@@ -73,18 +73,52 @@ export const fsRouterTypegenPlugin = (opts: { srcDir: string }): Plugin => {
...
+            readFileSync(existingServerEntry, 'utf8'),
+            {
+              syntax: 'typescript',
+              tsx: true,
+            },
+          );
+
+          const usesCreatePages = file.body.some((node) => {
+            if (node.type === 'ImportDeclaration') {
+              if (node.source.value !== 'waku') {
```
dai-shi: It's technically possible and valid to do `import { fsRouter } from 'waku/router/server'`.

dai-shi (approved):
Looks good otherwise.

What's your opinion?

packages/waku/src/lib/vite-plugins/fs-router-typegen.ts
tylersayshi: That should still generate types I think because it's still using filesystem routing. 

With this logic that would still run the generation because we only check for createPages being used to disable generation.

dai-shi: Oh, this is `usesCreatePages`. Got it.

Then, my concern is that we shouldn't run generation for this case: https://github.com/wakujs/waku/blob/fe8fb224abbaba0fa962782a25690264945e1114/examples/31_minimal/src/server-entry.tsx

And, also `import { createPages } from 'waku/router/server'` should be detected.

dai-shi (changes requested):
https://github.com/wakujs/waku/pull/1748#discussion_r2496665067

GitHub doesn't have a capability to "unapprove".

packages/waku/src/lib/vite-plugins/fs-router-typegen.ts
tylersayshi: should this check be `!node.source.value.startsWith('waku')` then you think?

dai-shi: @tylersayshi 
I think we should have `usesFsRouter` instead of `!usesCreatePages`.
And check `import { fsRouter } from 'waku'` or `import { fsRouter } from 'waku/router/server'`.

tylersayshi: if it uses fsRouter wouldn't we still want to run the typegen?

dai-shi: We want to run the typegen, if fsRouter is used. Otherwise, we don't. No need to check the existence of `createPages`.

tylersayshi: thanks! updated to match that logic

dai-shi (approved):

PR #1802, opened 2025-11-28
refactor: define-router

dai-shi:
#1744 changed our structure and some assumptions. Now, we should make our define-router slightly more straightforward.

### DIFF
--- a/packages/waku/src/router/define-router.tsx
+++ b/packages/waku/src/router/define-router.tsx
@@ -312,13 +312,12 @@ export function unstable_defineRouter(fns: {
       pathname: decodedPathname,
       query: pathConfigItem.isStatic ? undefined : query,
     };
-    const slices = pathConfigItem.slices;
     const myConfig = await getMyConfig();
     const sliceConfigMap = new Map<
       string,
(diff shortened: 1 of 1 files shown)


packages/waku/src/router/create-pages.tsx
```diff
@@ -389,6 +352,111 @@ export const createPages = <
     }
   };
 
+  // FIXME this should be refactored
```
dai-shi, line 355: This is just a hack to keep the logic with copy&paste code.
I'm pretty sure there should be a better way to implement this.
@tylersayshi When you have a chance, would you please take some time to properly implement create-pages with the new define-router api?

packages/waku/src/router/create-pages.tsx
```diff
@@ -628,35 +696,44 @@ export const createPages = <
   };
 
   const definedRouter = unstable_defineRouter({
-    getConfig: async () => {
+    getConfigs: async () => {
       await configure();
```
dai-shi, line 700: This is just called once, so actually it doesn't make much sense.

PR #1811, opened 2025-12-01
internal: expose full server-entry

dai-shi:
ref: https://github.com/wakujs/waku/pull/1795#discussion_r2575799797

### DIFF
--- a/packages/waku/src/lib/vite-entries/entry.server.tsx
+++ b/packages/waku/src/lib/vite-entries/entry.server.tsx
@@ -2,10 +2,11 @@ import serverEntry from 'virtual:vite-rsc-waku/server-entry';
 import * as serverEntryExports from 'virtual:vite-rsc-waku/server-entry';
 import { INTERNAL_setAllEnv } from '../../server.js';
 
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/vite-entries/entry.server.tsx
```diff
@@ -1,7 +1,11 @@
 import serverEntry from 'virtual:vite-rsc-waku/server-entry';
+import * as serverEntryExports from 'virtual:vite-rsc-waku/server-entry';
 import { INTERNAL_setAllEnv } from '../../server.js';
 
-export { serverEntry as unstable_serverEntry };
+export const unstable_serverEntryExports =
+  serverEntryExports as typeof serverEntryExports & {
+    [someOtherExport: string]: unknown;
+  };
```
dai-shi: Hmm, I wonder if we could simply re-export named exports

```suggestion
export { serverEntry as unstable_serverEntry };
export * from 'virtual:vite-rsc-waku/server-entry';
```

But, the there are no types...

PR #1826, opened 2025-12-05
refactor(create-pages): remove createRouterRenderers

tylersayshi:
follow up to: https://github.com/wakujs/waku/pull/1802#discussion_r2573330730

### DIFF
--- a/packages/waku/src/router/create-pages.tsx
+++ b/packages/waku/src/router/create-pages.tsx
@@ -321,6 +321,8 @@ export const createPages = <
   const staticComponentMap = new Map<string, FunctionComponent<any>>();
   const getStaticLayout = (id: string) =>
     staticComponentMap.get(joinPath(id, 'layout').slice(1));
+  const getDynamicLayout = (segment: string) =>
+    dynamicLayoutPathMap.get(segment)?.[1];
   const slicePathMap = new Map<string, string[]>();
   const sliceIdMap = new Map<
     string,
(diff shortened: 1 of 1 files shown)


packages/waku/src/router/create-pages.tsx
```diff
@@ -319,6 +319,8 @@ export const createPages = <
     }
   >();
   const staticComponentMap = new Map<string, FunctionComponent<any>>();
+  const getStaticLayout = (id: string) =>
+    staticComponentMap.get(joinPath(id, 'layout').slice(1));
```
tylersayshi, line 323: @dai-shi i saw your comment about this feeling like a hack, but i think it's the best way to get the layout components. It was in two places, so I just made the little helper to make it a bit clearer.

packages/waku/src/router/create-pages.tsx
```diff
@@ -352,26 +354,10 @@ export const createPages = <
     }
   };
 
-  // FIXME this should be refactored
```
dai-shi, line 355: Thanks for working on it. 
What I wondered is that we don't need createRouterRenders in the first place.
In particular, 
```ts
        if (!renderer) {
          throw new Error('Renderer not found for ' + id);
        }
```
shouldn't be required. (I don't mean to just remove it, but I see if we restructure our logic, we don't even need it.)

tylersayshi: ok! this should be closer now then and ready for a review

dai-shi: it looks good.

PR #1831, opened 2025-12-07
feat: add useServerInsertedHTML for CSS-in-JS support

nihgwu:
Attempt to address https://github.com/wakujs/waku/discussions/1211

I know CSS-in-JS is dead, but we have tons of old projects still using it and can't be migrated quickly, this PR followed NextJS's take on `useServerInsertedHTML` to add streaming SSR support to CSS-in-JS libs

I've explored different approaches before submitting this PR, like exposing SSR hooks with wrapElement and transformHtml, it's simple and works great with SSG but doesn't support steaming

Happy to discuss about the API design, like should we prefix with `unstable_` and should we just expose it from `waku`

### DIFF
--- a/packages/waku/src/server-html/client.tsx
+++ b/packages/waku/src/server-html/client.tsx
@@ -1,0 +1,45 @@
+'use client';
+
+import type { ReactNode } from 'react';
+import * as React from 'react';
+import { ServerInsertedHTMLContext } from './server-inserted-html.js';
+
+export type ServerInsertedHTMLHook = (callback: () => ReactNode) => void;
+
+/**
+ * Hook to register HTML content to be inserted during server-side rendering.
+ * This is similar to Next.js's `useServerInsertedHTML` hook.
+ *
+ * The callback will be called during streaming and its result will be
+ * inserted into the HTML stream before the closing </head> tag.
+ *
+ * @example
+ * ```tsx
+ * 'use client';
+ *
+ * import { useState } from 'react';
+ * import { useServerInsertedHTML } from 'waku/server-html';
+ *
+ * export function StyleRegistry({ children }) {
+ *   const [styles, setStyles] = useState<string[]>([]);
+ *
+ *   useServerInsertedHTML(() => {
+ *     if (styles.length === 0) return null;
+ *     const html = styles.join('');
+ *     setStyles([]);
+ *     return <style dangerouslySetInnerHTML={{ __html: html }} />;
+ *   });
+ *
+ *   return <>{children}</>;
+ * }
+ * ```
+ */
+export function useServerInsertedHTML(callback: () => ReactNode): void {
+  const addInsertedServerHTMLCallback = React.useContext(
+    ServerInsertedHTMLContext,
+  );
+  // Should have no effects on client where there's no provider
+  if (addInsertedServerHTMLCallback) {
+    addInsertedServerHTMLCallback(callback);
+  }
+}
--- a/examples/60_styled-components/package.json
+++ b/examples/60_styled-components/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "60_styled-components",
+  "version": "0.1.0",
+  "type": "module",
+  "private": true,
+  "scripts": {
+    "dev": "waku dev",
+    "build": "waku build",
+    "start": "waku start"
+  },
+  "dependencies": {
+    "react": "~19.2.1",
+    "react-dom": "~19.2.1",
+    "react-server-dom-webpack": "~19.2.1",
+    "styled-components": "^6.1.19",
+    "waku": "0.27.3"
+  },
+  "devDependencies": {
+    "@types/react": "^19.2.7",
+    "@types/react-dom": "^19.2.3",
+    "typescript": "^5.9.3"
+  }
+}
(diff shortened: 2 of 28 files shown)


dai-shi:
This looks like a good addition, but there's no description about anything and the coding style is far different from what I would like to see.
I would like to understand the problem space better. Would you please open a new discussion first and help us understand it?
For now, closing this, but we may reopen it later.

nihgwu:
@dai-shi sorry I thought the title is pretty clear and the examples tell everything, I don't want to paste the description generated by AI which looks good but just rubbish, and weird to see it's closed directly, we can discuss on the PR and I can add more details

I spent hours to come with this solution, although most code is done by Claude, but I've tried different approaches and eventually submitted this one
I know CSS-in-JS is dead, but we have tons of old projects still using it and can't be migrated quickly

> coding style is far different from what I would like to see.

Want to hear more, I've manually changed the examples to align with others and adjusted code structures, I know it's definitely different from the creator's style, so I try to isolate the added code into a different space

dai-shi:
If I'm not mistaken, this shouldn't be added to the core. It should be a library. But, I'm not sure. That's why I asked the explanation, not a solution yet. I don't need the AI generation. I wanted the background and motivation (which is closer to the prompt to the AI). The link to #1211 is helpful.

Is it possible to turn those examples into e2e tests? That's more helpful, but let's first discuss.

> I spent hours to come with this solution

If you are not happy with it, don't spend time for solutions before discussing. Assume PRs can be rejected.

nihgwu:
I understand PR can be rejected, but didn't expected it's rejected so hastily
I was just experimenting at background, and I think this approach is promising so I created the PR for discussion :) 
I don't think it's feasible to do it via a lib without change in core, I tried various ways, and my principle is always minimal change on Waku

dai-shi:
> rejected so hastily

Sorry about that. I kind of worried that you are rushed.

> without change in core

I didn't strictly mean not to add any changes, but we carefully design it to make changes that enables to do css-in-js as libs. But, again, I'm not sure and I need to learn the problem better. Currently, I believe the solution is not acceptable.

packages/waku/src/server-html/client.tsx
```diff
@@ -0,0 +1,45 @@
+'use client';
```
dai-shi, line 1: We have never done something like this in src/lib

nihgwu: Don't understand what you are talking about
Are you suggesting not using the directive here?
Or the code style stuff?

dai-shi: Never mind then. It's not on you. I'll find out the way to support css-in-js as a lib.

nihgwu: I really don't get it, I found this https://github.com/wakujs/waku/blob/main/packages/waku/src/router/client.tsx#L1, initially the code was in src/lib, but I think as an optional feature only for CSS-in-JS, I think it's better to isolate it to another sub entry

nihgwu:
Don't mean to be offensive, I had really weird feeling on the contribution experience, it was my fault didn't mark it as draft
to make it clear it's created for discussion, then it was closed without discussion, till now I still don't see any meaningful discussion, just saying it's not aligned with the code style or vision, I tried to follow the code style, I don't have the full vision as I'm not the creator of this project, I do hope I can see more useful and clearer comments that I don't need to guess and can take actions, otherwise seems I just need to raise requests and creator will do everything
> I'll find out the way to support css-in-js as a lib.

dai-shi:
I had a really weird feeling too. It almost looked like a spam.
This PR is opened with no description and I failed to understand it's related with #1211.
The code is really against my design, but as it's not well defined, it can't be helped. But, I'm very sure this is not the starting point for the discussion. Move to 👉 https://github.com/wakujs/waku/discussions/1211#discussioncomment-15194238

dai-shi:
> > I'll find out the way to support css-in-js as a lib.

> I just need to raise requests and creator will do everything

One thing I want to make it clear. I'll take care of the core part, but I'm counting on you for the css-in-js part. So, I meant "I'll find out the way to support * as a lib." But, I'm not familiar with the problem, so I can't tell concrete ideas yet.

nihgwu:
> I had a really weird feeling too. It almost looked like a spam.

It's my fault not making it as draft and I thought the title tells everything, and the examples demonstrate the purpose

> I'm very sure this is not the starting point for the discussion. Move to 👉 https://github.com/wakujs/waku/discussions/1211#discussioncomment-15194238

There are discussions there but no actions taken, if no one take actions, it will hang over there to unclear end, so I decide to move the first step, and I expect to discuss in the PR to make clearer what's expected and how it can be done, otherwise it will be still floating over the surface

I'm not criticising, just sharing my feeling as a contributor, I've been watching this project since it's created, I do hope it's flourishing, we do need to establish a better contribution flow, when I created this PR, there was no todo or anything else from the template(I guess there is no PR template)

Also based on my observation, you do tend to do most of the things on your way, I do understand that's your project and you want to control the direction(and I like the design overall)

You don't need to reply to my PR in rush like 
> I didn't strictly mean not to add any changes, but we carefully design it to make changes that enables to do css-in-js as libs. But, again, I'm not sure and I need to learn the problem better. Currently, I believe the solution is not acceptable.
>
> This is not only the one, but easy one.

this is helpless, just take time to have a better picture of the issue, not just negative comments without actual meanings(at least I don't understand)

nihgwu:
Thanks to Waku's flexible API, I managed to make it work without touching waku, will create another PR to only add examples

PR #1845, opened 2025-12-14
test: styled-components fixture and spec

dai-shi:
For #1211

Hacks:
- use AsyncLocalStorage to pass ServerStyleSheet from "ssr" to "server" with globalThis
- Convert web stream to node stream
- patch globalThis.require for https://github.com/styled-components/styled-components/blob/a21089e1cde9d2492349088787c59dd85358a337/packages/styled-components/src/models/ServerStyleSheet.tsx#L90

TODOs:
- [x] DEV
- [x] PRD
- Reusable library (not for this PR)


e2e/fixtures/styled-components/src/components/StyledComponentsRegistry.tsx
```diff
@@ -0,0 +1,39 @@
...
+export function StyledComponentsRegistry({
+  children,
+}: {
+  children: ReactNode;
+}) {
+  // Only create stylesheet once with lazy initial state
+  const [styledComponentsStyleSheet] = useState(() => new ServerStyleSheet());
+
+  // On the client, just render children directly
+  if (typeof window !== 'undefined') {
```
nihgwu: JFYI this can be even simpler with Vite

Image
we even don't need `useState` as in SSR it only runs once

e2e/fixtures/styled-components/src/server-entry.ts
```diff
@@ -0,0 +1,46 @@
...
+          statusText: res.statusText,
+          headers: res.headers,
+        });
+        return newRes;
+      };
+      return router.handleRequest(input, { ...utils, renderHtml });
+    });
+  },
+  handleBuild: (utils) => {
+    return router.handleBuild(utils);
```
nihgwu: you'll have to inject `utils.renderHtml` as well here for SSG, actually we just need to inject `renderHTML` once

e2e/fixtures/styled-components/src/components/StyledComponentsRegistry.tsx
nihgwu: Just realise I don't need that suppression
image

dai-shi: Thanks.

e2e/fixtures/styled-components/src/server-entry.ts
dai-shi: Will try.

e2e/fixtures/styled-components/src/server-entry.ts
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+const router = fsRouter(import.meta.glob('./**/*.tsx', { base: './pages' }));
+
+export default adapter({
+  handleRequest: async (input, utils) =>
+    serverStyleSheetStorage.run({}, () =>
+      router.handleRequest(input, {
+        ...utils,
+        renderHtml: patchRenderHtml(utils.renderHtml),
```
nihgwu, line 42: I did exactly the same as you (injectRenderHtml), but eventually moved to ssr hack to avoid duplication :)

e2e/fixtures/styled-components/src/server-entry.ts
```diff
@@ -0,0 +1,52 @@
...
+      }),
+    ),
+  handleBuild: (utils) =>
+    router.handleBuild({
+      ...utils,
+      renderHtml: patchRenderHtml(utils.renderHtml),
+      withRequest: <T>(req: Request, fn: () => T): T =>
```
nihgwu, line 49: it works for me without injecting `withRequest`

e2e/fixtures/styled-components/src/components/StyledComponentsRegistry.tsx
```diff
@@ -0,0 +1,36 @@
...
+ * @see https://styled-components.com/docs/advanced#server-side-rendering
+ */
+export function StyledComponentsRegistry({
+  children,
+}: {
+  children: ReactNode;
+}) {
+  let sheet: ServerStyleSheet | undefined;
+  if (import.meta.env.SSR) {
+    const store = serverStyleSheetStorage?.getStore();
```
nihgwu, line 27: I'd suggest to use the callback pattern so the solution is not coupled with SC https://github.com/wakujs/waku/discussions/1211#discussioncomment-15252184

Image

e2e/fixtures/styled-components/src/server-entry.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Readable } from 'node:stream';
+import type { ServerStyleSheet } from 'styled-components';
+import { fsRouter } from 'waku';
+import adapter from 'waku/adapters/default';
+
+const serverStyleSheetStorage = new AsyncLocalStorage<{
+  sheet?: ServerStyleSheet;
+}>();
+(globalThis as any).__SERVER_STYLE_SHEET_STORAGE__ = serverStyleSheetStorage;
+globalThis.require = createRequire(import.meta.url);
```
nihgwu, line 12: leftover?

dai-shi: How did you solve it without patching require? This is react-server environment.

nihgwu: I don't have any issue without it, tried dev/build/start, I mean my own project
maybe because you are exposing sheet to asyncLS while I don't

dai-shi: Yeah, I don't have a solution to have everything in ssr environment without touching waku yet.

dai-shi:
The implementation in this fixture will be replaced with #1847 soon.

e2e/fixtures/styled-components/src/server-entry.ts
dai-shi: So, it was because my approach was to make use of `interleaveWithNodeStream`.

PR #1847, opened 2025-12-15
test: simplify styled-components fixture

nihgwu:
Follow up of https://github.com/wakujs/waku/discussions/1211
putting everything together

Alternative of https://github.com/wakujs/waku/pull/1831 without touching Waku. Key diffs:
* use `async:hooks` instead of React Context
* insert string instead of React elements

This works for styled-components and emotion, for styled-jsx, it doesn't provide API to get style tag but only style element, so we need to use `renderToStaticMarkup` to convert to string first (didn't try)

If we want to provide better support we should expose a way to interpolate `renderHTML` from SSR entry, which is used by both `handleRequest` and `handleBuild`

Maybe it's already possible via middleware to interpolate the response steam? @dai-shi 

----

edit(daishi): #1845 has been merged and this PR is updated to apply the changes to the e2e test fixture.

### DIFF
--- a/e2e/fixtures/create-pages/package.json
+++ b/e2e/fixtures/create-pages/package.json
@@ -7,7 +7,7 @@
     "dev": "waku dev",
     "build": "waku build",
     "start": "waku start",
-    "start:static": "serve dist/public -p 8080"
+    "start:static": "serve dist/public"
   },
   "dependencies": {
     "react": "~19.2.3",
--- a/e2e/fixtures/render-type/package.json
+++ b/e2e/fixtures/render-type/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "waku dev",
     "build": "waku build",
-    "start:dynamic": "waku start",
+    "start": "waku start",
     "start:static": "serve dist/public"
   },
   "dependencies": {
--- a/e2e/fixtures/styled-components/package.json
+++ b/e2e/fixtures/styled-components/package.json
@@ -6,12 +6,14 @@
   "scripts": {
     "dev": "waku dev",
     "build": "waku build",
-    "start": "waku start"
+    "start": "waku start",
+    "start:static": "serve dist/public"
   },
   "dependencies": {
     "react": "~19.2.3",
     "react-dom": "~19.2.3",
     "react-server-dom-webpack": "~19.2.3",
+    "serve": "^14.2.5",
     "styled-components": "^6.1.19",
     "waku": "latest"
   },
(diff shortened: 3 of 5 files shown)


dai-shi:
Looks promising. Nice work.
Would you please make changes to the fixture in #1845? If you like, I can merge #1845 first so that your contribution can be more visible as a fixing PR.

dai-shi:
> it's already possible via middleware to interpolate the response steam?

Unfortunately, middleware only works with server, not with SSG.

> we should expose a way to interpolate renderHTML from SSR entry

Yes, I started to think about it. But, it'll take time. Meanwhile, the solution in this PR seems fairly good.

nihgwu:
@dai-shi I think this change is basically a standalone example, can be merged first, and add fixture as followup, but I can do it in this PR later if you prefer

dai-shi:
I would like to see the test passing in the PR. Okay, let me merge #1845 first. Please make this PR to change the fixture instead of adding an example, but I can do that for you if you feel you've done enough.

nihgwu:
@dai-shi no worries, I'll do it

dai-shi:
#1845 has been merged!

nihgwu:
😢

nihgwu:
Now it's hard to understand the PR's description 🫠

e2e/fixtures/styled-components/src/server-html/context.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Request-scoped storage for server-inserted HTML using AsyncLocalStorage.
+ */
+
+import { AsyncLocalStorage } from 'async_hooks';
```
dai-shi:
```suggestion
import { AsyncLocalStorage } from 'node:async_hooks';
```

e2e/fixtures/styled-components/src/server-html/ssr.ts
```diff
@@ -0,0 +1,45 @@
+import { ReactNode } from 'react';
```
dai-shi:
```suggestion
import type { ReactNode } from 'react';
```

e2e/fixtures/styled-components/src/server-html/stream.ts
```diff
@@ -0,0 +1,60 @@
...
+        }
+        controller.enqueue(chunk);
+        return;
+      }
+
+      if (content) {
+        toInsert += content;
+      }
+
+      const decoded = decoder.decode(chunk);
```
hi-ogawa, line 33: binary `chunk` doesn't necessary contain full utf-8 boundary, in which case this causes garbled character.
`rsc-html-stream` uses `decode(chunk, { stream: true })` and read remaining by `decode()` within single `setTimeout` task https://github.com/devongovett/rsc-html-stream/blob/9b858445f4f5817470f373ae266dea04d5fcfac3/server.js#L18

But not sure maybe if this transformer comes after `rsc-html-stream`, then there might not be such chunks reaching here anymore.

PR #1848, opened 2025-12-15
Fix catch-all create-pages routing precedence bug

hamlim:
## Details:

> [!NOTE]
> I refer to `"static"` in this PR description to mean their filename/file path and thus the route segment name are known at build time, as opposed to being dynamic or catch-all segments like `/[foo].tsx` or `/[...catchAll].tsx`

This change implements both an e2e test for the `fsRouter` where catch-all route segments seem to be matched **before** "static" route segments are, and also a fix to address that failing issue.

The fix checks that we don't prefer catch-all segments in cases where there are static routes that that should be matched first, but previously had a shorter "route length" and so would come after the catch-all segments.

For example the `/` route has a length of 0, and the `/[...catchAll]` route has a length of one, so previously we'd sort `/[...catchAll]` before the `/` route and match it before the "static" `/` route.

In the example fixture we have a few "static" routes, that we expect to be matched when visiting them in the app, and then a single catch-all segment that is acting as a not found experience for the application.

The desired state is that "static" routes are matched first, then we match dynamic routes, and finally we fallback to catch-all route segments, the fix implements the desired state.


tylersayshi:
could the other fs-router spec be re-used and we just run the test on a subroute like `/subroute/[...catchAll]` and `/subroute/` ?

hamlim:
Yea I can update the existing fs-router spec 👍

tylersayshi (approved):
lgtm, thanks!

hamlim:
Thanks for the quick reviews and merge!

PR #1853, opened 2025-12-17
experimental renderHtmlEnhancer

dai-shi:
Like #1847, if we need to change renderHtml behavior, it requires a hack with `globalThis`.
This is an attempt to provide a way to customize renderHtml function.


nihgwu:
it works just like a charm, is `RenderHtml` exported

nihgwu:
the change
image

nihgwu:
I don't like API, it's too low level and couple with internal details, like the API contract of `renderHtml`

I think we can just expose something like `transformResponse` or (`transformHtmlResponse` to make it explicit is about HTML not API), and `response.body` is the steam to manipulate

nihgwu:
my take https://github.com/wakujs/waku/pull/1855, now we don't expose internal details of `renderHtml`, it simply manipulate response without extra knowledge

dai-shi:
Yeah, it's low level. But, our middleware api is already low level and I'm thinking of something the same level. Let me think about various ideas. I'll be back in a few days.

nihgwu:
my approach is also low level, it's more about hiding details which is more likely to change
but the response won't change

dai-shi:
My definition is this is part of the "minimal" API (so I don't actually like using `waku.config.ts`), so depending on internals is technically okay. I'm thinking of a use case to patch `html` before passing to `renderHtml`.

dai-shi:
It was my bad choice to use waku.config.ts. I hope #1858 would work.

PR #1854, opened 2025-12-17
chore: use `viteRsc.loadModule` in styled-components demo

hi-ogawa:
- Related https://github.com/wakujs/waku/discussions/1211#discussioncomment-15277402

I'm not sure this approach has been brought up. This isn't clean either, but just sharing that importing "ssr" environment module is possible in this way.


nihgwu:
The idea is to avoid `import.meta.viteRsc.loadModule` from user space which is a heavy mental model, we explored this approach https://github.com/wakujs/waku/discussions/1211#discussioncomment-15277882

PR #1855, opened 2025-12-17
refactor enhanceRenderHtml

nihgwu:
addressing https://github.com/wakujs/waku/pull/1853#issuecomment-3664876854


packages/waku/src/lib/vite-entries/entry.ssr.tsx
```diff
@@ -1,14 +1,15 @@
...
-export async function INTERNAL_enhanceRenderHtml(renderHtml: RenderHtml) {
+export async function INTERNAL_enhanceRenderHtml(
+  render: () => Promise<Response>,
+): Promise<Response> {
   if (renderHtmlEnhancer) {
-    renderHtml = renderHtmlEnhancer(renderHtml);
+    return renderHtmlEnhancer(render);
   }
-  return renderHtml;
+  return render();
```
dai-shi, line 14: I don't completely dislike this idea. This is limited, but simpler.
It's just naming what's weird. As it's no longer enhancing, something like this would be good:
```js
export async function INTERNAL_transformRenderHtmlResponse(
  res: Response
): Promise<Response> {
  // ...
}
```

nihgwu: yes I was using `transformHtmlResponse`, but I have to use factory to assign globalThis early, while `transformHtmlResponse` indicates a method to manipulate response directly

I main purpose is hiding details, we don't need to care `elementsStream` nor `html` nodes, not saying `options` I have no idea how to do with them

and the only demand is manipulating the response, I'll think about the name

dai-shi:
https://github.com/wakujs/waku/pull/1853#issuecomment-3673420835

PR #1858, opened 2025-12-19
experimental loadSsrModule

dai-shi:
Another take for #1853.

### DIFF
--- a/packages/waku/src/lib/vite-plugins/combined-plugins.ts
+++ b/packages/waku/src/lib/vite-plugins/combined-plugins.ts
@@ -12,6 +12,7 @@ import { mainPlugin } from './main.js';
 import { notFoundPlugin } from './not-found.js';
 import { patchRsdwPlugin } from './patch-rsdw.js';
 import { privateDirPlugin } from './private-dir.js';
+import { ssrLoaderPlugin } from './ssr-loader.js';
 import { userEntriesPlugin } from './user-entries.js';
 import { virtualConfigPlugin } from './virtual-config.js';
 
@@ -19,6 +20,7 @@ export function combinedPlugins(config: Required<Config>): PluginOption {
   return [
     extraPlugins(config),
     allowServerPlugin(), // apply `allowServer` DCE before "use client" transform
+    ssrLoaderPlugin(), // compile `loadSsrModule` hints
     cloudflarePlugin(),
     rsc({
       serverHandler: false,


packages/waku/src/lib/vite-plugins/ssr-loader.ts
```diff
@@ -0,0 +1,238 @@
...
+  `ssr_${crypto.createHash('sha256').update(key).digest('hex').slice(0, 12)}`;
+
+const getOrCreateSsrInputMap = (ssrConfig: EnvironmentOptions) => {
+  ssrConfig.build ??= {};
+  ssrConfig.build.rollupOptions ??= {};
+  ssrConfig.build.rollupOptions.input ??= {};
+  const input = ssrConfig.build.rollupOptions.input;
+  if (typeof input === 'string' || Array.isArray(input)) {
+    throw new Error(
+      '[waku] ssrLoaderPlugin expects environments.ssr.build.rollupOptions.input to be an object',
```
nihgwu, line 21: should `rollupOptions` be hardcoded? I'm using rolldown and I see warnings from Vite ask me to migrate to `rolldownOptions`

e2e/fixtures/styled-components/src/waku.server.ts
```diff
@@ -1,6 +1,9 @@
 import { fsRouter } from 'waku';
 import adapter from 'waku/adapters/default';
-import { injectRenderHtml } from './server-html/ssr';
+import { unstable_loadSsrModule as loadSsrModule } from 'waku/server';
+
+const { injectRenderHtml } =
+  await loadSsrModule<typeof import('./server-html/ssr')>('./server-html/ssr');
```
nihgwu, line 6: if we do this seems there is no need to expose `loadSsrModule` at all, we can achieve the same without it, or we can do https://github.com/wakujs/waku/discussions/1211#discussioncomment-15277882

The purpose of a new API is to avoid duplication in both handleRequest and handleBuild

And also avoid looking into waku's internal, currently we have to construct type of `renderHtml` in user space, it's already the problem for us as we are using 0.27.0 but the type has been changed in latest version, so we need to update, I do hope we can just provide a way to transform the SSR steam

packages/waku/src/lib/vite-plugins/ssr-loader.ts
dai-shi: Good point. We could support both rollupOptions and rolldownOptions, right?

e2e/fixtures/styled-components/src/waku.server.ts
dai-shi: Unfortunately, I'm pretty sure that my goal is different from your hope. This is about exposing the minimal API in user space which includes `renderHtml` (which is intentionally unstable). My goal is to make it possible to create a third-party library like `waku-css-in-js` without using waku.config.ts.

nihgwu: I mean CSS-in-JS is already possible without any change from waku side, of we can use vite-rsc's API to load SSR module, why do we need it from Waku? Just to avoid the rollupOptions config?

And my approach also support CSS-in-JS already, and yesterday I just did another thing with it `no_js` mode
image

dai-shi: > without using waku.config.ts.

means "to avoid the rollupOptions config" as I understand.

>  it's already the problem for us as we are using 0.27.0 but the type has been changed in latest version

So, you don't have any problems beside that, do you?

I was hoping to ask you to develop `waku-css-in-js`, but maybe it's out of your interest.

e2e/fixtures/styled-components/src/components/StyledComponentsRegistry.tsx
```diff
@@ -12,11 +11,15 @@ export const StyledComponentsRegistry = ({
   if (import.meta.env.SSR) {
     const sheet = new ServerStyleSheet();
 
-    insertServerHTML(() => {
-      const styles = sheet.getStyleTags();
-      sheet.instance.clearTag();
-      return styles;
-    });
+    import('../server-html/context')
```
dai-shi, line 14: Not a big fan of this, but would there be other ways to avoid `globalThis`?

nihgwu:
I feel the biggest problem of extending Waku is sometimes you have to override 2 places: `waku-server.ts` and `waku.config.ts` , if we can just do it in a single place, we just need to expose `withStyledComponents(wakuConfig)`, similar how to extend Next.js, or maybe it's already possible with `unstable_adaptor` 🤔 , then we don't need any change from Waku

dai-shi:
> if we can just do it in a single place

Yes, I would like to do it in `waku.server.tsx`, and this PR is an effort for it.
```js
export default adapter(cssInJs(router));
```

> maybe it's already possible with `unstable_adaptor` 🤔

Oh, that's an interesting idea. If we extend the scope of adapters, `unstable_adapters` might be possible, or new concept with `unstable_enhancers` or something.

hi-ogawa:
I'm thinking to provide something like this on `@vitejs/plugin-rsc` side as `import.meta.viteRsc.import("...", { environment: "ssr" })`. Please check out the PR for the detail https://github.com/vitejs/vite-plugin-react/pull/1063

nihgwu:
I also believe this should be done at rsc-plugin level

e2e/fixtures/styled-components/src/server-html/ssr.ts
```diff
@@ -43,3 +32,20 @@ export function injectRenderHtml(renderHtml: RenderHtml): RenderHtml {
     });
   };
 }
+
+export function cssInJs(handlers: Handlers) {
+  return defineHandlers({
```
hi-ogawa, line 37: Importing `waku/minimal/server`'s `defineHandlers` in `ssr` environment feels off since Waku "server" API is meant to be `rsc` environment and otherwise Waku "client" (though in this case, `defineHandlers` is merely a type helper, so it won't technically matter).

dai-shi:
Let me give up this idea for now. I agree that it's not great mental model-wise, and it's for too edge cases. For now, globalThis hack seems better for cssInJs hack.

PR #1863, opened 2025-12-19
fix: batch flight data

### LINKED ISSUE #1861: RSC flight data should be consolidated
It currently pollutes the DOM with separate `<script>` tags even on statically prerendered routes (chunks only make sense for streaming).

Image

dai-shi:
close #1861

### DIFF
--- a/packages/waku/src/lib/utils/stream.ts
+++ b/packages/waku/src/lib/utils/stream.ts
@@ -61,7 +61,12 @@ export function batchReadableStream(
     clearTimeout(timer);
     timer = undefined;
     if (buffer.length) {
-      controller.enqueue(concatUint8Array(buffer));
+      try {
+        controller.enqueue(concatUint8Array(buffer));
+      } catch {
+        // ignore errors
+        // ref: https://github.com/wakujs/waku/pull/1863#discussion_r2634546953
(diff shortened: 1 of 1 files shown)


packages/waku/src/lib/utils/stream.ts
```diff
@@ -34,3 +34,49 @@ export const base64ToStream = (base64: string): ReadableStream => {
...
+  const buffer: Uint8Array[] = [];
+  let timer: ReturnType<typeof setTimeout> | undefined;
+
+  const flushBuffer = (
+    controller: TransformStreamDefaultController<Uint8Array>,
+  ): void => {
+    clearTimeout(timer);
+    timer = undefined;
+    if (buffer.length) {
+      controller.enqueue(concatUint8Array(buffer));
```
hi-ogawa: Batching makes sense to me, but doing `controller.enqueue` inside `setTimeout` can crash server with unhandled error with aborted stream etc. I think it's safe to silence any error by `try { ... } catch { }` similar to https://github.com/devongovett/rsc-html-stream/pull/10

dai-shi: Thanks for the catch!

PR #1864, opened 2025-12-20
feat: export vite plugins (unstable)

jxom:
This PR proposes to export `rscPlugin` via the internal entrypoint. I intentionally did not add it in `src/internals.ts` as it would conflict with the virtual modules exposed in `createServerEntryAdapter`.

Rationale: Exploring a new version of [Vocs](https://github.com/wevm/vocs) that will use Waku internally (in a Vite plugin). As such, it would be amazing to have access to the `rscPlugin`!

----

edit(daishi): export all internal plugins instead of `rscPlugin`.

### DIFF
--- a/packages/waku/package.json
+++ b/packages/waku/package.json
@@ -38,9 +38,9 @@
       "types": "./dist/internals.d.ts",
       "default": "./dist/internals.js"
     },
-    "./internals/*": {
-      "types": "./dist/internals/*.d.ts",
-      "default": "./dist/internals/*.js"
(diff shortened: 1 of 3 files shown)


dai-shi:
We used to do something similar. I think it's better to export individual plugins instead of the combined one, as it gives more control. I want to rename something. May I add commits here?

jxom:
That will work better for me actually! Go ahead for it

dai-shi:
Merging this. Hope your project goes well!

PR #1882, opened 2026-01-01
feat: dedupe head tags

jxom:
When using React's native head tag hoisting, duplicate `<title>`, `<meta>`, and `<link>` tags can appear in the document when both a layout and page component define the same tags. For example:

```tsx
// _layout.tsx
<title>My Site</title>
<meta name="description" content="Default description" />

// about.tsx  
<title>About</title>
<meta name="description" content="About page description" />
```

Results in duplicate tags in the rendered HTML.

This PR adds a streaming HTML transform which deduplicates head tags during SSR, keeping only the last occurrence of each tag (so descendant tags override ancestor tags).

Other frameworks such as [Tanstack Router](https://tanstack.com/router/v1/docs/framework/react/guide/document-head-management##deduping) are also known to dedupe head tags, so it would be nice if Waku was smart enough to do the same!

**Rules:**

- `<title>`: Only one allowed
- `<meta charset>`: Only one allowed
- `<meta name="…">`: Dedupe by `name` attribute
- `<meta property="…">`: Dedupe by `property` attribute
- `<meta http-equiv=".…">`: Dedupe by `http-equiv` attribute
- `<link>`: Dedupe by `rel` + `href` attributes

### DIFF
--- a/packages/waku/src/lib/utils/dedupe-head.ts
+++ b/packages/waku/src/lib/utils/dedupe-head.ts
@@ -20,9 +20,12 @@ export const parseAttributes = (html: string) => {
 const getMetaIdentity = (attrs: Record<string, string>) => {
   if ('charset' in attrs) return 'meta:charset';
   if (attrs['name']) return `meta:name:${attrs['name'].toLowerCase()}`;
-  if (attrs['property']) return `meta:property:${attrs['property'].toLowerCase()}`;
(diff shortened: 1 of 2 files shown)


jxom:
Actually, I figured I could just create a middleware in consumer code for this. Closing for now.

dai-shi:
I thought about it too before but I think it's the responsibility of the React library, not of Waku. Yeah, user land solution makes sense.

PR #1889, opened 2026-01-06
Move llms.txt and rss.xml for _api for webside 

deepso7:
I was going through the docs and noticed `https://waku.gg/api/llms-full.txt` route was broken, then I saw, we've moved api to _api directory

Fixed broken `/api/llms-full.txt` route by moving it to the `_api` directory structure on the website.

Changes:
- Moved route from `api/` to `_api/` directory
- Renamed `llms-full.txt` to `llms.txt` to align with standard conventions
- Updated route path to reflect new structure


dai-shi (approved):
Ohhhhh, I knew I needed this change, but then forgot! Thanks for the catch!

PR #1896, opened 2026-01-11
test: add race-condition e2e

Master-Hash:
fix https://github.com/wakujs/waku/issues/1880

### DIFF
--- a/e2e/fixtures/router-client/src/components/skip-nav.tsx
+++ b/e2e/fixtures/router-client/src/components/skip-nav.tsx
@@ -1,0 +1,16 @@
+'use client';
+
+import { useRouter } from 'waku';
+
+export function SkipNavToBButton() {
+  const router = useRouter();
+  return (
+    <button
+      data-testid="skip-go-b"
+      onClick={() => router.push('/skip/b')}
+      type="button"
+    >
+      Go skip/b
+    </button>
+  );
+}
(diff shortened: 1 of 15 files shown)


e2e/fixtures/race-condition/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "fs-router",
```
dai-shi:
```suggestion
  "name": "race-condition",
```

dai-shi (commented):
Thanks. I'll have a closer look later.

dai-shi:
marking draft for now.

e2e/fixtures/race-condition/src/pages/_components/FastNavigate.tsx
```diff
@@ -0,0 +1,13 @@
+'use client';
+
+import { useCallback } from 'react';
+import { useRouter } from 'waku';
+
+export function FastNavigate() {
+  const { push } = useRouter();
+  const onClick = useCallback(() => {
+    void push('/about');
+    void push('/bar');
```
dai-shi: If the reproduction is such a simple way, I would probably consider putting the test case in an existing spec and fixture.

dai-shi: f3d887df

PR #1910, opened 2026-01-14
fix(router): inconsistent URI encoded path between SSR and client

Master-Hash:
fix https://github.com/wakujs/waku/issues/1901

### DIFF
--- a/packages/waku/src/router/fs-router.ts
+++ b/packages/waku/src/router/fs-router.ts
@@ -63,18 +63,6 @@ export function fsRouter(
           GET?: (req: Request) => Promise<Response>;
         };
 
-        try {
-          const decodedFile = decodeURI(file);
-          if (decodedFile !== file) {
-            console.warn(
-              `[waku] Using encoded URI in file name is not supported.
-       Consider renaming the file: ${decodedFile}`,
-            );
-          }
-        } finally {
-          // make eslint happy
-        }
-
         // Use WHATWG URL encoding for the file path (different from RFC2396-based encoding)
         file = new URL(file, 'http://localhost:3000').pathname.slice(1);
         const config = await mod.getConfig?.();


bhaktatejas922:
## 🤖 AI Preview Test

Hey! We noticed you're doing amazing open source work. We built an AI model that's really good at testing web applications, so we ran it on your preview deployment to help catch any issues.

**Preview URL:** https://waku-git-fork-master-hash-unicode-daishis-projects.vercel.app

### What We Found
✅ Preview deployment detected and queued for testing.

---
**Want unlimited free access?** We'd love to support your project with free AI-powered testing!

If you'd rather we not comment on your PRs, just 👎 this message and we'll remove you from our list.

<sub>Automated by [Morph](https://morphllm.com) • [Learn more](https://morphllm.com/docs/preview-testing)</sub>

Master-Hash:
todo:

`/route/[param]`

$$(dynamic + static)(space + 漢字)(SSR + navigation)$$

This is 8 situations. ((((((((

Master-Hash:
This can be merged before upstream fixes their issue. After upstream is done, we just need to remove the warning.

dai-shi:
Let's have @tylersayshi review for this.

packages/waku/src/router/fs-router.ts
```diff
@@ -62,8 +62,24 @@ export function fsRouter(
...
+            );
+          }
+        } finally {
+          // make eslint happy
+        }
+        // Always use encoded URI paths
+        file = new URL(file, 'http://example.com').pathname;
         // strip "./" prefix
-        file = file.replace(/^\.\//, '');
+        file = file.replace(/^\//, '');
```
tylersayshi: to make sure i understand here: this works around the vite issue by removing the characters that vite doesn't support? 

and we also plan to support these characters after the vite issue is resolved?

tylersayshi:
> This can be merged before upstream fixes their issue. After upstream is done, we just need to remove the warning.

can the comments and code make it explicit which code should be removed once the upstream issue is resolved?

changing the regex and `file = new URL(file, 'http://example.com').pathname;` feel like non-obvious changes to revert

Master-Hash:
> changing the regex and file = `new URL(file, 'http://example.com').pathname;` feel like non-obvious changes to revert

Do you know a better way to convert path to encoded if needed, and do nothing if already?

tylersayshi:
> Do you know a better way to convert path to encoded if needed, and do nothing if already?

oh sorry no. I thought this was part of the workaround. I see what it does now, this seems good 👍

packages/waku/src/router/fs-router.ts
```diff
@@ -62,8 +62,27 @@ export function fsRouter(
...
+            );
+          }
+        } finally {
+          // make eslint happy
+        }
+        // workaround ends
+
+        // Always use encoded URI paths
+        file = new URL(file, 'http://example.com').pathname;
         // strip "./" prefix
```
dai-shi: The comment and the code doesn't match.

packages/waku/src/router/fs-router.ts
```diff
@@ -62,8 +62,27 @@ export function fsRouter(
...
+              decodedFile,
+            );
+          }
+        } finally {
+          // make eslint happy
+        }
+        // workaround ends
+
+        // Always use encoded URI paths
+        file = new URL(file, 'http://example.com').pathname;
```
dai-shi: Doesn't this work?
```suggestion
        file = new URL(file, 'http://localhost:3000').pathname.slice(1);
```

dai-shi: On second thought, I think this is better and more readable without unnecessary `example.com`.
```suggestion
        file = encodeURI(decodeURI(file));
```

Master-Hash: > On second thought, I think this is better and more readable without unnecessary `example.com`.
> ```suggestion
>         file = encodeURI(decodeURI(file));
> ```

This will not work, because if file name is already encoded, it will be encoded twice then.

Master-Hash: > On second thought, I think this is better and more readable without unnecessary `example.com`.
> ```suggestion
>         file = encodeURI(decodeURI(file));
> ```

Suppose there is a file named `%a`.

Client requests `%25a` and should get something.

but this way results in error. `Uncaught URIError: URI malformed`

Of course we can try, compare, but this is too complex.

dai-shi:
```
> new URL("%a", "http://localhost:3000").pathname.slice(1)
'%a'
> encodeURI(decodeURI("%a"))
Uncaught URIError: URI malformed
    at decodeURI (<anonymous>)
```
I see.

dai-shi: In this case, do we want `file` to be `%a` or `%25a`?

dai-shi: Is this what we want?
```
> const safeDecodeURI = (s) => { try { return decodeURI(s) } catch { return s } }
undefined
> encodeURI(safeDecodeURI("%a"))
'%25a'
```

Master-Hash: > In this case, do we want `file` to be `%a` or `%25a`?
> 

I want to support both, on error when both exist and conflict.

Master-Hash: > In this case, do we want `file` to be `%a` or `%25a`?
> 

I want to support both, on error when both exist and conflict.

btw this is about vite, we can do nothing here.

Master-Hash: > Is this what we want?
> ```
> > const safeDecodeURI = (s) => { try { return decodeURI(s) } catch { return s } }
> undefined
> > encodeURI(safeDecodeURI("%a"))
> '%25a'
> ```

Maybe... It's okay to me if it passes the ci.

dai-shi: We prefer minimalistic. Please do not support both. Pick just one that is more reasonable.

Master-Hash: > We prefer minimalistic. Please do not support both. Pick just one that is more reasonable.

We will support encoded ones.

But, there is another problem. 

```ts
> encodeURI('nested/[name].tsx')
'nested/%5Bname%5D.tsx'
> new URL('nested/[name].tsx', "http://localhost").pathname.slice(1)
'nested/[name].tsx'
```

`safeDecodeURI` is still broken.

dai-shi: Oh, so the comment `// Always use encoded URI paths` is not accurate, right?

Master-Hash: > Oh, so the comment `// Always use encoded URI paths` is not accurate, right?

Yes... I'm amazed by the encoding algorithm :(

Can you come up with a better comment?

dai-shi: Let's call it
```
// Use browser URL encoding for the file path (different from RFC2396-based encoding)
```

packages/waku/src/router/fs-router.ts
```diff
@@ -62,8 +62,21 @@ export function fsRouter(
...
+              `[waku] Using encoded URI in file name is not supported.
+       Consider renaming the file: ${decodedFile}`,
+            );
+          }
+        } finally {
+          // make eslint happy
+        }
+
+        // Always use encoded URI paths
+        file = new URL(file, 'http://localhost').pathname.slice(1);
```
dai-shi: Let's keep 3000 as it's our convention in define-router.tsx. Not important. Just to be consistent.

```suggestion
        file = new URL(file, 'http://localhost:3000').pathname.slice(1);
```

packages/waku/src/router/fs-router.ts
Master-Hash: LGTM.

btw: browser url-> WHATWG URL

dai-shi: whichever works for me.

packages/waku/src/router/fs-router.ts
```diff
@@ -62,8 +62,21 @@ export function fsRouter(
...
+        try {
+          const decodedFile = decodeURI(file);
+          if (decodedFile !== file) {
+            throw new Error(
+              `[waku] Using encoded URI in file name is not supported.
+       Consider renaming the file: ${decodedFile}`,
+            );
+          }
+        } finally {
+          // make eslint happy
```
dai-shi: Do we still need this?

Master-Hash: Yes, because eslint disallow empty block.

Encoded filename is now unsupported. I think explicit failure is better than silent misfunction.

Master-Hash: Yes, because eslint disallow empty block.

Encoded filename is now unsupported. I think explicit failure is better than silent misfunction.

Master-Hash: Yes, because eslint disallow empty block.

Encoded filename is now unsupported. I think explicit failure is better than silent misfunction.

Master-Hash: Yes, because eslint disallow empty block.

Encoded filename is now unsupported. I think explicit failure is better than silent misfunction.

dai-shi: My question was the entire try-finally block. Sorry for the confusion.
Now, I get that try is for decodeURI, like safeDecodeURI.

My next question is whether we should use WHATWG URL decoding, instead of RSC2396 decoding.

Master-Hash: * The entire `try {}` block makes the case error, instead of silent malfunction. Removing it will not break anything: it's only my preference on explicit failure.
* But `new URL` can only encode something, and cannot decode anything... And `decodeURI` works here. It will not stop anything which should work. (It may let something broken pass, but It's enough for now?)


---

Why GitHub repeat my word for 4 times?????

dai-shi: Ah, it's not try-catch. 🤭 Can we remove the try block then? What do you mean by "making the case error"?

Master-Hash: Oh sorry. I added a warning inside the try block; you said you only want to support one usage, so I changed it into throw; but in try block it will do nothing.

I'm tired of fixing this. If removing this can make merging sooner, I'll remove it.

Master-Hash: Oh sorry. I added a warning inside the try block; you said you only want to support one usage, so I changed it into throw; but in try block it will do nothing.

I'm tired of fixing this. If removing this can make merging sooner, I'll remove it.

Master-Hash: Oh sorry. I added a warning inside the try block; you said you only want to support one usage, so I changed it into throw; but in try block it will do nothing.

I'm tired of fixing this. If removing this can make merging sooner, I'll remove it.

Master-Hash:
If no `try`: filename `%a` will throw: I don't know if this is legal.

So I choose removing it entirely.

dai-shi:
I don't know either. To clarify, defining `safeDedodeURI` is acceptable. But, I can merge as-is too.

Master-Hash:
please merge it as is.

PR #1917, opened 2026-01-22
fix: prefetch client assets also on dyn -> static pages

Master-Hash:


packages/waku/src/router/define-router.tsx
```diff
@@ -755,6 +752,11 @@ export function unstable_defineRouter(fns: {
       });
     }
 
+    await generateFile(
+      '__waku_preload_manifest.js',
```
dai-shi: I would like to put this in assets dir. But, I'm not sure how to make "assets" configurable. Maybe we could hard-code for now.

packages/waku/src/router/client.tsx
```diff
@@ -958,6 +958,7 @@ const InnerRouter = ({
   const rootElement = (
     <Slot id="root">
       <meta name="httpstatus" content={httpStatus} />
+      <script type="module" async src="/__waku_preload_manifest.js" />
```
dai-shi: We need to support `basePath`.

packages/waku/src/router/define-router.tsx
Master-Hash: I put it in root dir because __waku_build_metadata.js and __vite_rsc_assets_manifest.js did so (although they are in server).

If you want to refactor it later, I'll resolve it now?

Master-Hash:
One problem: __waku_preload_manifest.js contains no hash in file name, so cache rules may break it.

~~Manually add a hash may solve it.~~

I wish ~~there were a way to add it to vite module graph~~ but rsc are populated after trunks rendered

Similarly, it's impossible to calculate hash manually and pass it to client.tsx

packages/waku/src/router/define-router.tsx
dai-shi: > __waku_build_metadata.js and __vite_rsc_assets_manifest.js

They are in `dist`. This is about `dist/public`.

hi-ogawa:
I think this can naturally achieved through build metadata. Have we considered following approach?
- build (`handleBuild`)
  - ssg: inject `path2moduleIds` script directly to ssg html (that's what https://github.com/wakujs/waku/pull/1915 does)
  - save build metadata `saveBuildMetadata(path2moduleIds)`
- prod ssr (`handleRequest)`
  - load build metadata `loadBuildMetadata(path2moduleIds)`
  - inject `path2moduleIds` to ssr html

dai-shi:
> Similarly, it's impossible to calculate hash manually and pass it to client.tsx

Yeah, I think this should be done somehow in the vite build process.
Currently it's not possible. We need to find a way and design the minimal api.
Could we run another vite build process after `buildApp` hook? cc @hi-ogawa

dai-shi:
> prod ssr (`handleRequest)`
> * load build metadata `loadBuildMetadata(path2moduleIds)`
> * inject `path2moduleIds` to ssr html

Oh!! Sounds good.

👉 https://github.com/wakujs/waku/pull/1915/commits/95a722fe0bce28849d3eb8b838559d5292fa7f7d

dai-shi:
I don't dislike the approach. Now it's a public api, I need to consider the option name (I don't like `prefetchCode` actually, blame me.)

@hi-ogawa What do you think about the approach?

hi-ogawa:
> I don't dislike the approach. Now it's a public api, I need to consider the option name (I don't like `prefetchCode` actually, blame me.)
> 
> @hi-ogawa What do you think about the approach?

I think @Master-Hash's motivation is about avoiding extra inline script for nonce and we should probably get to know that first because Waku technically doesn't support `nonce` on its API surface.

Master-Hash:
Yes, but this still simplify my patch for nonce in my own website, or potential support in waku. I would appreciate it if waku can consider my own need.

hi-ogawa:
That said, if we put that unofficial nonce aside and aim for just avoiding the extra inline script for this new logic, this might be possible by:
- during buildApp
  - calculate "entire client build hash" based on all client assets (for example by grabbing `generateBundle(bundle)` and summing all chunk hash)
  - pass that hash through `handleBuild` (👈 this looks like new API though)
  - during handleBuild
    - collect `path2moduleIds` like this now,
    - `renderHtml` with `<script href="/__waku_prefetch_(hash).js">`
    - `generateFile("__waku_prefetch_<hash>.js", generatePrefetchCode(path2moduleIds))`
    - `saveBuildMetadata("prefetch-hash", hash)`
 - during prod ssr
   - during handleRequest
     - `loadBuildMetadata("prefetch-hash")` to restore hash
     - `renderHtml` with  `<script href="/__waku_prefetch_(hash).js">`

(EDIT: actually if we are delaying `renderHtml` until `path2moduleIds` is ready, then we can just calculate object hash of `path2moduleIds` inside `define-router` on its own, so no need to pass through `handleBuild` :thinking: )

dai-shi:
Thank you both.

If the goal is to support csp and if patching is still required after this, it feels suboptimal and I wouldn't accept this approach for now. (Although it feels better with avoiding suppressHydrationWarning hack.)

> if waku can consider my own need.

your own need = csp, right? Then, it's not a need for a single person. I would like to design minimal api so that people don't need to patch for csp.
Would you please explain the problem (not a solution) in the current `main` before introducing #1915?

Master-Hash:
> Would you please explain the problem (not a solution) in the current main before introducing https://github.com/wakujs/waku/pull/1915?

See https://discord.com/channels/627656437971288081/627656438436724738/1452473838007877726

> It's not supported yet but it should be possible to something similar as Next or other SSR frameworks https://nextjs.org/docs/app/guides/content-security-policy. Please feel free to create a feature request or issue on Github.
> At least,nonce needs to be able to passed to here, which is currently just always undefined https://github.com/wakujs/waku/blob/c6915b0bf5468410b6f71796c0af1e876d3cce96/packages/waku/src/lib/vite-rsc/ssr.tsx#L98

dai-shi:
I saw it, but thought you might have a better explanation. Let me try questions from my side:
- Is passing `nonce` option everything we need?
- Do people need `nonce` value on user end?

Master-Hash:
> Is passing `nonce` option everything we need?

Basic function only needs to pass `nonce` option to both inline script (including `renderToReadableStream`) and request header.

For static html rendering, We can't use nonce, so a way to extract inline script or adding hash value is needed.

(For me, I simply not uploading html to cloudflare workers assets.)

> Do people need nonce value on user end?

No. It can be generated inside waku framework.

(My patch: generate nonce internally, pass a default csp header, and modify it with a hono middleware)

dai-shi:
Thanks for your comments!
 
> Basic function only needs to pass `nonce` option to both inline script (including `renderToReadableStream`) and request header.

You mean response header, right?

> For static html rendering, We can't use nonce, so a way to extract inline script or adding hash value is needed.

Do we have such a case for now?

> No. It can be generated inside waku framework.

I see. I thought there might be the case where people could add inline script on user code.

> (My patch: generate nonce internally, pass a default csp header, and modify it with a hono middleware)

I hope with #1922, you could do it in waku.server.tsx without patching.

Master-Hash:
> You mean response header, right?

Yes, sorry for typo.

> I thought there might be the case where people could add inline script on user code.

In fact, end user can use custom client entry and add everything they wanted to inline.

Or, a `useEffect` in Root component is also an equivalent. 

> Do we have such a case for now?

What does it mean?

dai-shi:
> What does it mean?

I mean, since nonce is dynamic page only, for static page, there's nothing we could do. I wasn't sure what is "so a way to extract inline script or adding hash value is needed" about.

dai-shi:
Merging this into my branch.

Master-Hash:
> I mean, since nonce is dynamic page only, for static page, there's nothing we could do. I wasn't sure what is "so a way to extract inline script or adding hash value is needed" about.

Not **nothing** literally. If we want both good of SSG HTML and no XSS, we can:

1. Move inline script to assets when building, and disable all inline script;
2. Calculate SHA384 of every inline script, and allow them like `Content-Security-Policy: script-src 'sha384-oqVuAfXRKap7fdgcCY5uykM6+R9GqQ8K/uxy9rx7HNQlGYl1kPzQho1wx4JwY8wC' 'sha256-fictional_value'`

Both approach is insane complected, so I would simply give up SSG HTML.

PR #1918, opened 2026-01-22
chore: use react-dom.preloadModule instead of import

Master-Hash:


packages/waku/src/router/client.tsx
```diff
@@ -891,7 +892,13 @@ const InnerRouter = ({
...
     const rscParams = createRscParams(route.query);
     prefetchRsc(rscPath, rscParams);
-    (globalThis as any).__WAKU_ROUTER_PREFETCH__?.(route.path);
+    for (const asset of (globalThis as any).__WAKU_ROUTER_PREFETCH__?.(
+      route.path,
+    ) ?? []) {
+      preloadModule(asset, {
+        as: 'script',
+      });
+    }
```
dai-shi: This is okay, but I may prefer this:
```js
(globalThis as any).__WAKU_ROUTER_PREFETCH__?.(route.path, (id) => {
  preloadModule(id, { as: 'script' });
});
```
Does it work?

dai-shi (approved):
Love it! Nice minimal change.
Thanks for the suggestion!

PR #1920, opened 2026-01-23
fix(managed): exclude test and spec files from middleware auto-discovery

struong:
## Summary

Excludes `.test.*` and `.spec.*` files from middleware auto-discovery in managed mode.

## Motivation

When developers place test files (e.g., `auth.test.ts`, `logging.spec.ts`) alongside their middleware in the `src/middleware/` directory, these files are incorrectly picked up by `import.meta.glob` and loaded as middleware modules. This causes runtime errors like `mod.default is not a function`.

## Changes

- Modified `middlewareGlob` in `getManagedServerEntry()` to use an array pattern with negation
- Added exclusion patterns for `*.test.*` and `*.spec.*` files

## Testing

Vite's `import.meta.glob` supports array patterns with negation prefixes per the [Vite documentation](https://vite.dev/guide/features#multiple-patterns).

### DIFF
--- a/packages/waku/src/lib/utils/managed.ts
+++ b/packages/waku/src/lib/utils/managed.ts
@@ -4,7 +4,10 @@ export const getManagedServerEntry = (srcDir: string) => {
   const globBase = `/${srcDir}/${SRC_PAGES}`;
   const exts = EXTENSIONS.map((ext) => ext.slice(1)).join(',');
   const globPattern = `${globBase}/**/*.{${exts}}`;
-  const middlewareGlob = [`/${srcDir}/${SRC_MIDDLEWARE}/*.{${exts}}`, `!/${srcDir}/${SRC_MIDDLEWARE}/*.{test,spec}.{${exts}}`];
+  const middlewareGlob = [
+    `/${srcDir}/${SRC_MIDDLEWARE}/*.{${exts}}`,
+    `!/${srcDir}/${SRC_MIDDLEWARE}/*.{test,spec}.{${exts}}`,
+  ];
   return `
 import { fsRouter } from 'waku';
 import adapter from 'waku/adapters/default';


packages/waku/src/lib/utils/managed.ts
```diff
@@ -3,7 +3,11 @@ import { EXTENSIONS, SRC_MIDDLEWARE, SRC_PAGES } from '../constants.js';
 export const getManagedServerEntry = (srcDir: string) => {
   const globBase = `/${srcDir}/${SRC_PAGES}`;
   const globPattern = `${globBase}/**/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`;
-  const middlewareGlob = `/${srcDir}/${SRC_MIDDLEWARE}/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`;
+  const middlewareGlob = [
+    `/${srcDir}/${SRC_MIDDLEWARE}/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`,
+    `!/${srcDir}/${SRC_MIDDLEWARE}/*.test.*`,
+    `!/${srcDir}/${SRC_MIDDLEWARE}/*.spec.*`,
```
dai-shi: Could we use `{test,spec}` syntax?

struong: Yes

struong:
Hi @dai-shi can I get the workflow runs approved?

Thank you for reviewing!

packages/waku/src/lib/utils/managed.ts
```diff
@@ -3,7 +3,8 @@ import { EXTENSIONS, SRC_MIDDLEWARE, SRC_PAGES } from '../constants.js';
 export const getManagedServerEntry = (srcDir: string) => {
   const globBase = `/${srcDir}/${SRC_PAGES}`;
   const globPattern = `${globBase}/**/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`;
-  const middlewareGlob = `/${srcDir}/${SRC_MIDDLEWARE}/*.{${EXTENSIONS.map((ext) => ext.slice(1)).join(',')}}`;
+  const exts = EXTENSIONS.map((ext) => ext.slice(1)).join(',');
```
dai-shi, line 5: If we do this, should we use it in globPattern too?

struong: Yes, missed that

PR #1922, opened 2026-01-24
feat: support for csp

dai-shi:
ref: https://github.com/wakujs/waku/pull/1917#issuecomment-3793345963

TODOs
- docs/guides

### DIFF
--- a/packages/waku/src/lib/utils/render.ts
+++ b/packages/waku/src/lib/utils/render.ts
@@ -1,18 +1,8 @@
-import { getContext } from '../context.js';
 import type {
   Unstable_ParseRsc,
   Unstable_RenderHtml,
   Unstable_RenderRsc,
 } from '../types.js';
-
-const getNonce = () => {
-  try {
-    const context = getContext();
-    return context.nonce;
-  } catch {
-    return undefined;
-  }
-};
 
 export function createRenderUtils(
   temporaryReferences: unknown,
@@ -63,7 +53,7 @@ export function createRenderUtils(
       const htmlResult = await renderHtmlStream(elementsStream, rscHtmlStream, {
         rscPath: options.rscPath,
         formState: options.formState as never,
-        nonce: getNonce(),
+        nonce: options.nonce,
         extraScriptContent: options.unstable_extraScriptContent,
       });
       return new Response(htmlResult.stream, {
--- a/e2e/fixtures/ssr-nonce/src/waku.server.tsx
+++ b/e2e/fixtures/ssr-nonce/src/waku.server.tsx
@@ -1,6 +1,5 @@
 import adapter from 'waku/adapters/default';
 import { Slot } from 'waku/minimal/client';
-import { unstable_getContext as getContext } from 'waku/server';
 import App from './components/App.js';
 
 // Fixed nonce for testing purposes
@@ -12,13 +11,12 @@ export default adapter({
       return renderRsc({ App: <App /> });
     }
     if (input.type === 'custom' && input.pathname === '/') {
-      const context = getContext();
-      context.nonce = TEST_NONCE;
       const response = await renderHtml(
         await renderRsc({ App: <App /> }),
         <Slot id="App" />,
         {
           rscPath: '',
+          nonce: TEST_NONCE,
         },
       );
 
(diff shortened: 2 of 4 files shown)


Master-Hash:
I'm confused by too many function call levels. Do adapters also need patching?

dai-shi:
> I'm confused by too many function call levels. Do adapters also need patching?

Yeah, this is very low level. Let me create an example or something.

From the minimal API point of view, adapters are user code. Creating your own adapter would be nice. But, for this one, I don't think we need a custom adapter, but not sure yet.

packages/waku/src/lib/utils/render.ts
```diff
@@ -51,9 +61,10 @@ export function createRenderUtils(
         onError,
       });
       const htmlResult = await renderHtmlStream(elementsStream, rscHtmlStream, {
-        formState: options.formState as never,
         rscPath: options.rscPath,
-        nonce: undefined,
+        formState: options.formState as never,
+        nonce: getNonce(),
```
hi-ogawa: I wasn't sure `render.ts` relying on `context.ts` is good, so I favored more higher level `define-router.ts` to take care of it. If this coupling is okay for you, this looks cleaner to me as well.

packages/waku/src/lib/vite-rsc/ssr.tsx
```diff
@@ -16,6 +16,7 @@ type RenderHtmlStream = (
     rscPath: string | undefined;
     formState: ReactFormState | undefined;
     nonce: string | undefined;
+    extraScriptContent: string | undefined;
```
hi-ogawa, line 19: As @Master-Hash said, I don't think this is needed anymore.

dai-shi:
https://github.com/wakujs/waku/pull/1923#issuecomment-3793987720

> Let me grab your examples into my branch and try the first one.

Done.

hi-ogawa:
> Do people need nonce value on user end?
> https://github.com/wakujs/waku/pull/1917#issuecomment-3793390247

About this, I think this is a valid use case and mentioned in https://nextjs.org/docs/app/guides/content-security-policy#reading-the-nonce. Not just next's `Script`, but I think it's also possible for some 3rd party component might have `nonce=` prop if it somehow relies on inline script for its internal. With this PR's `getContext().nonce`, users can easily grab it similar to Next's example in their doc.

packages/waku/src/lib/utils/render.ts
dai-shi: I wasn't sure either, and to be honest, I'm not 100% confident. (which means, I may change mind in the future...)
But based on you guys discussions, I think I need the use of context anyway, particularly if we want to set the value from hono middleware.

packages/waku/src/lib/vite-rsc/ssr.tsx
dai-shi: I think we still need this if we go with #1917 approach.

dai-shi:
> but I think it's also possible for some 3rd party component might have `nonce=` prop if it somehow relies on inline script for its internal.

Ah, I see.

> With this PR's `getContext().nonce`, users can easily grab it similar to Next's example in their doc.

Yeah!

packages/waku/src/lib/utils/render.ts
dai-shi: On second thought, I'm not very happy with the coupling. So, I took the half of you idea. 1f29c452

Master-Hash:
> > Do people need nonce value on user end?
> > [#1917 (comment)](https://github.com/wakujs/waku/pull/1917#issuecomment-3793390247)
> 
> About this, I think this is a valid use case and mentioned in [nextjs.org/docs/app/guides/content-security-policy#reading-the-nonce](https://nextjs.org/docs/app/guides/content-security-policy#reading-the-nonce). Not just next's `Script`, but I think it's also possible for some 3rd party component might have `nonce=` prop if it somehow relies on inline script for its internal. With this PR's `getContext().nonce`, users can easily grab it similar to Next's example in their doc.

This can be solved by `strict-dynamic`.

dai-shi:
Merging this.

@Master-Hash If you can draft a guide doc and open a PR, it will be more than helpful.

PR #1923, opened 2026-01-24
feat: add nonce support for CSP

hi-ogawa:
## Background

CSP (Content Security Policy) nonces are used to allow inline scripts and styles while maintaining security. Each request requires a unique nonce that must be:
1. Generated per-request
2. Added to the CSP header
3. Applied to all inline `<script>` tags

### How Other Frameworks Handle This

**Next.js** (convention-based):
- User generates nonce in middleware
- User sets CSP header on the **request** (not response)
- Next.js extracts nonce from the request's CSP header
- Reference: https://nextjs.org/docs/app/guides/content-security-policy

This is conceptually unintuitive - CSP is a **response** header, but Next.js uses the request header as a side channel to pass the nonce value to the rendering layer. The framework reads from the request what should be a response concern.

**Redwood SDK** (framework-managed):
- Framework generates nonce internally
- Framework uses it automatically in SSR
- User configures CSP header if needed
- Reference: https://docs.rwsdk.com/core/security/

### Waku's Approach

Waku provides explicit APIs for nonce support without hidden magic:

- `unstable_setNonce(nonce)` - set nonce in middleware
- `unstable_getNonce()` - read nonce (used internally by router)
- Router passes nonce to `renderHtml` automatically

### Layer 1: Minimal API (explicit option)

Add `nonce` to `renderHtml` options. Consistent with existing pattern - we already expose `formState` (via `actionResult`) and `rscPath`.

**Changes:**
- `packages/waku/src/lib/types.ts` - Add `nonce?: string` to `Unstable_RenderHtml` options
- `packages/waku/src/lib/utils/render.ts` - Pass `nonce` through to `renderHtmlStream`

**User Usage:**
```typescript
// waku.server.tsx
export default adapter({
  handleRequest: async (input, { renderRsc, renderHtml }) => {
    const nonce = crypto.randomUUID();

    if (input.type === 'custom' && input.pathname === '/') {
      const response = await renderHtml(
        await renderRsc({ App: <App /> }),
        <Slot id="App" />,
        { rscPath: '', nonce },
      );

      response.headers.set(
        'Content-Security-Policy',
        `script-src 'self' 'nonce-${nonce}';`
      );

      return response;
    }
  },
});
```

### Layer 2: Router API (router users)

For router users who don't call `renderHtml` directly, use `unstable_setNonce` in middleware.

**Changes:**
- `packages/waku/src/router/define-router.tsx` - Add `unstable_setNonce`/`unstable_getNonce`, read nonce and pass to `renderHtml`
- `packages/waku/src/router/server.ts` - Export the APIs

**User Usage (simple):**
```typescript
// src/middleware/nonce.ts
import type { MiddlewareHandler } from 'hono';
import { unstable_setNonce } from 'waku/router/server';

const nonceMiddleware = (): MiddlewareHandler => {
  return async (c, next) => {
    const nonce = crypto.randomUUID();
    unstable_setNonce(nonce);
    await next();
    c.res.headers.set(
      'Content-Security-Policy',
      `script-src 'self' 'nonce-${nonce}';`
    );
  };
};

export default nonceMiddleware;
```

**User Usage (with Hono's secureHeaders):**
```typescript
// src/middleware/nonce.ts
import type { MiddlewareHandler } from 'hono';
import { secureHeaders, NONCE } from 'hono/secure-headers';
import { unstable_setNonce } from 'waku/router/server';

const nonceMiddleware = (): MiddlewareHandler => {
  const secure = secureHeaders({
    contentSecurityPolicy: {
      scriptSrc: ["'self'", NONCE],
    },
  });

  return async (c, next) => {
    await secure(c, async () => {
      // Bridge Hono's nonce to Waku
      unstable_setNonce(c.get('secureHeadersNonce'));
      await next();
    });
  };
};

export default nonceMiddleware;
```

## Notes

- **SSG/Build time**: Nonces are per-request and not applicable during static generation
- **Explicit API**: No hidden magic - user explicitly sets nonce via `unstable_setNonce`
- **Flexible**: Works with any nonce generation approach (simple `crypto.randomUUID()` or Hono's `secureHeaders`)

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/waku/src/lib/types.ts
+++ b/packages/waku/src/lib/types.ts
@@ -14,7 +14,12 @@ export type Unstable_ParseRsc = (
 export type Unstable_RenderHtml = (
   elementsStream: ReadableStream,
   html: ReactNode,
-  options: { rscPath: string; formState?: unknown; status?: number; nonce?: string },
+  options: {
+    rscPath: string;
+    formState?: unknown;
+    status?: number;
+    nonce?: string;
+  },
 ) => Promise<Response>;
 
 export type Unstable_EmitFile = (


dai-shi:
> expose `formState` (via `actionResult`)

off topic: I think I should rename the option.

dai-shi:
> ```ts
>       // User sets CSP header themselves
>       response.headers.set(
>         'Content-Security-Policy',
>         `script-src 'nonce-${nonce}' 'strict-dynamic';`
>       );
> ```

Ahhh, I didn't realize this. So, we don't need headers option in #1922.

hi-ogawa:
Sorry, I found some issues and changing some approaches. New approach is pushed and PR description updated.

hi-ogawa:
New approach feels more coupled weirdly, so it's not anything better though. But I think it would be nice to support Hono's `secureHeaders` middleware out of the box, so that means we need to grab nonce in `rscMiddleware` (maybe not necessary here?) through `c.get('secureHeadersNonce')` and that means `nonce` need to get passed through multiple layers i.e. `processRequest -> handleRequest -> renderHtml`.

dai-shi:
> But I think it would be nice to support Hono's `secureHeaders` middleware out of the box

I don't disagree with it. Will see if there's a way that I'm comfortable with.

hi-ogawa:
Tried another approach. This now changed to explicit define-router level API `unstable_setNonce`.

dai-shi:
Tried similar approach in #1922 to support `secureHeaders` ootb.

hi-ogawa:
> Tried similar approach in #1922 to support `secureHeaders` ootb.

I ended up with something different. I think explicit waku router level API makes sense since csp setup shouldn't need to actually rely on hono's buildin middleware. Please check two examples in PR description.

dai-shi:
My approach doesn't technically rely on hono middleware.
- I'm fine to drop the default nonceMiddleware in favor of your wrapper approach.
- I'm fine to add option for non context usage, but it feels against the minimalist philosophy.
- I'm hesitant to give a solution from define-router (conflicting with your mental model).

Let me grab your examples into my branch and try the first one.

hi-ogawa:
Sorry, I feel like comments are scattered in multiple PRs and hard to track 😅 I'm fine to close mine and will leave comment on others.

PR #1925, opened 2026-01-24
docs: support for csp (nonce)

Master-Hash:
ref: https://github.com/wakujs/waku/pull/1922


docs/guides/nonce.mdx
```diff
@@ -0,0 +1,103 @@
...
+  },
+  handleBuild: async () => {},
+});
+```
+
+## What's next
+
+In the case of SSG, a nonce is not secure because it must be randomly generated for every response. Since SSG produces static HTML, this is not possible. If your security requirements prioritize a nonce over static HTML benefits, consider disabling HTML pre-rendering.
+
+(Will we explore any alternative for case of SSG?)
```
dai-shi: I think it's fine. We probably won't explore it soon. Maybe change the section title then?

docs/guides/csp.mdx
```diff
@@ -0,0 +1,103 @@
...
+          rscPath: '',
+          nonce: NONCE,
+        },
+      );
+
+      // Set CSP header with the nonce
+      response.headers.set(
+        'Content-Security-Policy',
+        `script-src 'self' 'nonce-${NONCE}';`,
+      );
```
hi-ogawa, line 95: The test used this minimal csp, but if there's more sensible default setup in practice, I think it might be good to mention (like general recommendation)?

or putting it in another way, can this document be named as `csp.mdx` and then explain `nonce` use case as current solution? (we can mention non-nonce / ssg isn't supported now and also we can append it if we ever figure it out)

docs/guides/csp.mdx
```diff
@@ -0,0 +1,108 @@
...
+  },
+  handleBuild: async () => {},
+});
+```
+
+We set a minimal CSP header here. In practice, CSP should be as minimal or script as possible. We recommend hono middleware approach as it provides a [good default value](https://hono.dev/docs/middleware/builtin/secure-headers#supported-options).
+
+### Limitation
+
+In the case of SSG, a nonce is not secure because it must be randomly generated for every response. Since SSG produces static HTML, this is not possible. If your security requirements prioritize a nonce over static HTML benefits, consider disabling HTML pre-rendering.
```
Ariel-GonzAguer: I think it would be good to add in this section that it's possible to generate nonces for SSG-generated scripts using an Edge Function (like those from Netlify, for example). The HTML is generated statically, the nonce is generated dynamically on **each request**, which means that each user receives a unique, random, and secure nonce.
This link provides an overview of the topic I created → https://aprenderwaku.netlify.app/temas/15-seguridad
(It's in Spanish, but your browser can translate it)

Master-Hash: Your site seems broken.

```ts
Uncaught SyntaxError: missing ) after argument list [15-seguridad:101:3711](https://aprenderwaku.netlify.app/temas/15-seguridad/)
Uncaught Error: Connection closed.
    hn https://aprenderwaku.netlify.app/assets/index-omzISGE8.js:2
    h https://aprenderwaku.netlify.app/assets/index-omzISGE8.js:2
    promise callback*h https://aprenderwaku.netlify.app/assets/index-omzISGE8.js:2
    promise callback*h https://aprenderwaku.netlify.app/assets/index-omzISGE8.js:2
[index-omzISGE8.js:2:34364](https://aprenderwaku.netlify.app/assets/index-omzISGE8.js)
```

Master-Hash:
@Ariel-GonzAguer his approach await the full stream, which is not ideal. In fact, I have written a `TransformStream` nonce middleware.

https://github.com/Master-Hash/hashland/blob/vite/land/src/middleware/csp.ts#L6-L54

I don't really like this, but if this is useful, it's ok to place it onto guide.

P.S. Please edit my branch freely when I'm sleeping, for typo or whatever

docs/guides/csp.mdx
```diff
@@ -0,0 +1,108 @@
...
+      );
+
+      return response;
+    }
+  },
+  handleBuild: async () => {},
+});
+```
+
+We set a minimal CSP header here. In practice, CSP should be as minimal or script as possible. We recommend hono middleware approach as it provides a [good default value](https://hono.dev/docs/middleware/builtin/secure-headers#supported-options).
```
Ariel-GonzAguer:
```suggestion
We set a minimal CSP header here. In practice, CSP should be as minimal or **strict** as possible. We recommend hono middleware approach as it provides a [good default value](https://hono.dev/docs/middleware/builtin/secure-headers#supported-options).
```

PR #1928, opened 2026-01-26
chore: use hono/tiny preset

Master-Hash:
Waku doesn't rely on hono's router (if I understand correctly), so use a minimal router can reduce server bundle size by 20k.

### DIFF
--- a/packages/waku/src/adapters/netlify-build-enhancer.ts
+++ b/packages/waku/src/adapters/netlify-build-enhancer.ts
@@ -4,6 +4,7 @@ import path from 'node:path';
 export type BuildOptions = {
   distDir: string;
   privateDir: string;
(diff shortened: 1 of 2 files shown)


dai-shi:
Looks good to me! Could anyone else have a second review?

Master-Hash:
... nobody?

packages/waku/src/adapters/deno.ts
```diff
@@ -1,7 +1,7 @@
 import path from 'node:path';
 // FIXME hopefully we should avoid bundling this
```
dai-shi: Please also move this comment.

dai-shi (approved):
LGTM
ref: https://hono.dev/docs/api/presets

Issue #1860, opened 2025-12-19
env change isn't triggering RSC hot reload or React fast refresh


sandren:
Vite server restarts, but the Waku app's state isn't updated.

```
[vite] .env changed, restarting server...
[vite] server restarted.
[vite] connected.
```

dai-shi:
@hi-ogawa Do you think it's fixable?

hi-ogawa:
It's full server restart, so it won't be HMR but browser full reload at best.

Currently Waku CLI only expliictly watch `waku.config.ts`, but there are other files vite server automatically restarts itself, which are .env and also tsconfig.json. Waku would need to intercept Vite's own server restart to handle it, which is what Astro CLI dev server does if I remember correctly. 

https://github.com/wakujs/waku/blob/e4e49d44e08dfb56bdc394053667417ccc14b22d/packages/waku/src/lib/vite-rsc/cli.ts#L40-L54

dai-shi:
> there are other files vite server automatically restarts itself, which are .env and also tsconfig.json.

I see. What does vite do for restarting. It means it's not enough for waku, right?

> to intercept Vite's own server restart to handle it,

sounds like the right direction.

hi-ogawa:
> What does vite do for restarting. It means it's not enough for waku, right?

Yes, I was expecting Vite restarting itself can break Waku since it resets Vte dev server (and all internal plugins) but it doesn't reset Waku plugins.

PR #1929, opened 2026-01-26
feat(cli): restart Waku dev server on Vite server restart

yunho1017:
rel: #1860 

https://github.com/vitejs/vite/issues/17689

This issue causes Vite to fail to detect the change in the value of a previously defined environment variable.
but, if add a new environment variable, it is detected normally.

### DIFF
--- a/packages/waku/src/lib/vite-rsc/cmd-dev.ts
+++ b/packages/waku/src/lib/vite-rsc/cmd-dev.ts
@@ -6,39 +6,32 @@ import { loadConfig, loadEnv, overrideNodeEnv } from './loader.js';
 
 loadEnv();
 
-// Track if a restart is currently in flight to prevent concurrent restarts
-let restartInFlight = false;
+function createServerRestartHandler() {
+  let restartInFlight = false;
 
-async function withRestartLock<T>(
-  operation: () => Promise<T>,
-): Promise<T | undefined> {
-  if (restartInFlight) {
-    console.log('Server restart already in progress, skipping...');
-    return undefined;
-  }
-  restartInFlight = true;
-  try {
-    return await operation();
-  } finally {
-    restartInFlight = false;
-  }
-}
-
-async function handleServerRestart(
-  host: string | undefined,
-  port: number,
-  server: vite.ViteDevServer,
-) {
-  await withRestartLock(async () => {
-    console.log('Restarting server with fresh plugin configuration...');
-    const previousUrls = server.resolvedUrls;
-    await server.close();
-    const freshConfig = await loadConfig();
-    const newServer = await startDevServer(host, port, freshConfig, true);
-    if (previousUrls) {
-      server.resolvedUrls = newServer.resolvedUrls;
+  return async (
+    host: string | undefined,
+    port: number,
+    server: vite.ViteDevServer,
+  ) => {
+    if (restartInFlight) {
+      console.log('Server restart already in progress, skipping...');
+      return;
     }
-  });
+    restartInFlight = true;
+    try {
+      console.log('Restarting server with fresh plugin configuration...');
+      const previousUrls = server.resolvedUrls;
+      await server.close();
+      const freshConfig = await loadConfig();
+      const newServer = await startDevServer(host, port, freshConfig, true);
+      if (previousUrls) {
+        server.resolvedUrls = newServer.resolvedUrls;
+      }
+    } finally {
+      restartInFlight = false;
+    }
+  };
 }
 
 async function startDevServer(
@@ -55,6 +48,7 @@ async function startDevServer(
     plugins: [combinedPlugins(config)],
     server: host ? { host, port } : { port },
   });
+  const handleServerRestart = createServerRestartHandler();
   // Override Vite's restart to intercept automatic restarts (.env, tsconfig, etc.)
   server.restart = async () => {
     console.log('Vite server restart intercepted, reloading Waku plugins...');


dai-shi:
Does this fix #1860?

yunho1017:
It's not perfect
If a new env is added, it will be detected.
However, if an existing env changed, the server restarts but fails to reference the changed env value
This is a Vite problem (https://github.com/vitejs/vite/issues/17689)

To perfect, we need logic to delete the previous env variable
https://github.com/hi-ogawa/vite-plugins/pull/657/files

hi-ogawa:
Thanks for initiating this! I think this is a valid fix, but for overall architecture wise, I'm wondering if we should bring `vite.config.ts` as the source of truth and let Vite do more work instead of Waku dealing our own restart flow on its own. Our current approach is that we take over Vite initialization here:

https://github.com/wakujs/waku/blob/ccbe11049719d2ce07c872adec6d542ccbd55ca9/packages/waku/src/lib/vite-rsc/command.ts#L26-L28

If we move to `vite.config.ts` and let `vite.createServer()` load the config file and waku plugin from there, I think there's nothing for us to deal with regarding restart logic (except Vite's .env reload bug https://github.com/vitejs/vite/issues/17689). We should still be able to provide same `waku` cli as vite dev/build wrapper.

This will be a big breaking change though in terms of configuration. @dai-shi Are you open to this kind of change?

yunho1017:
Sounds like the right direction.
Are you saying that we should move the logic to Vite.config instead of override the restart of Vite?

hi-ogawa:
> Sounds like the right direction.
> 
> Are you saying that we should move the logic to Vite.config instead of override the restart of Vite?

I haven't fully figured out, but just the rough direction I would be more familiar with. Sveltekit / React-router takes similiar approach. Both have separate `svelte.config` and `react-router.config`, but how they get loaded is from there plugins in `vite.config`. React-router has a wrapper cli `react-router dev/build`, so what I'm suggesting in https://github.com/wakujs/waku/pull/1932 would be closer to that.

If Waku will have dedicated `waku.config` similar to svelte.config/react-router.config, we still need to manually watch the change and trigger `server.restart`, but this will be done from one of waku plugin's `configureServer(server)` hook, I think.

yunho1017:
I understand. Thank you. 
As you said, it's a big change in the direction of waku.config

I'll be waiting for other people's comment

dai-shi:
> This will be a big breaking change though in terms of configuration. @dai-shi Are you open to this kind of change?

This is a big non-technical decision.
We have discussed this internally and decided to have `waku.config.ts` as the single source of truth.
I still don't see a big enough reason to move away from it. So, my answer is no.

Just in case, adding @sandren in the loop: How are we comfortable with the change like #1932 ?
cc: @tylersayshi @himself65 @Aslemammad

yunho1017:
@dai-shi Then is this PR valid?

dai-shi:
> @dai-shi Then is this PR valid?

Yeah.

Have you looked into Sveltekit / React-router solutions?

Wouldn't it be possible to solve #1860 fully without fixing Vite?

hi-ogawa:
Since Waku uses `dotenv` on its own, Waku's server side `process.env` likely doesn't have issues (i.e. any change in `.env` should get applied to `process.env`). The issue could be only `import.meta.env.VITE_...` style constant, which is what Vite's builtin `.env` support does, which relies on problematic Vite's `loadEnv` utility (which doesn't pick up changed value of `.env` when they are loaded in `process.env`).

dai-shi:
> Since Waku uses `dotenv` on its own, Waku's server side `process.env` likely doesn't have issues (i.e. any change in `.env` should get applied to `process.env`). The issue could be only `import.meta.env.VITE_...` style constant, which is what Vite's builtin `.env` support does, which relies on problematic Vite's `loadEnv` utility (which doesn't pick up changed value of `.env` when they are loaded in `process.env`).

Sounds like this PR is the best effort solution for #1860 at the moment, right?

dai-shi:
I'll make some refactors later.

yunho1017:
Is there any improvement you have in mind?
If you tell me, I'll work on it

dai-shi:
> If you tell me, I'll work on it

Let me try then.

In the meantime, please check CI errors.

e2e/hot-reload.spec.ts
```diff
@@ -19,9 +19,31 @@ function modifyFile(
...
+  standaloneDir: string,
+  file: string,
+  content: string,
+) {
+  const filePath = join(standaloneDir, file);
+  if (existsSync(filePath) && !originalFiles[filePath]) {
+    originalFiles[filePath] = readFileSync(filePath, 'utf-8');
+  } else if (!existsSync(filePath)) {
+    // Mark as non-existent so we can delete it in cleanup
+    originalFiles[filePath] = '';
```
dai-shi: Let's make it `false` or something.
```ts
const originalFiles: { [key: string]: string | false } = {};
```

e2e/hot-reload.spec.ts
```diff
@@ -19,9 +19,31 @@ function modifyFile(
   writeFileSync(filePath, content.replace(search, replace));
 }
 
+function createOrModifyFile(
```
dai-shi: Just createFile? or createOrOverwriteFile?

e2e/hot-reload.spec.ts
```diff
@@ -19,9 +19,31 @@ function modifyFile(
...
+  standaloneDir: string,
+  file: string,
+  content: string,
+) {
+  const filePath = join(standaloneDir, file);
+  if (existsSync(filePath) && !originalFiles[filePath]) {
+    originalFiles[filePath] = readFileSync(filePath, 'utf-8');
```
dai-shi:
```suggestion
  if (existsSync(filePath)) {
    originalFiles[filePath] ??= readFileSync(filePath, 'utf-8');
```

e2e/hot-reload.spec.ts
```diff
@@ -19,9 +19,31 @@ function modifyFile(
...
+  }
+  writeFileSync(filePath, content);
+}
+
 test.afterAll(() => {
   for (const [file, content] of Object.entries(originalFiles)) {
-    writeFileSync(file, content);
+    if (content === '') {
+      // File didn't exist before, delete it
+      if (existsSync(file)) {
```
dai-shi: Do we need it?

packages/waku/src/lib/vite-rsc/cmd-dev.ts
dai-shi, line 1: Don't love the `dev/` folder.

packages/waku/src/lib/vite-rsc/dev/restart.ts
dai-shi, line 1: Can we combine this file?

packages/waku/src/cli.ts
```diff
@@ -1,11 +1,11 @@
...
 import process from 'node:process';
 import { parseArgs } from 'node:util';
-import * as dotenv from 'dotenv';
+import loadEnv from './lib/utils/env.js';
 
-const require = createRequire(new URL('.', import.meta.url));
+loadEnv();
 
-dotenv.config({ path: ['.env.local', '.env'], quiet: true });
+const require = createRequire(new URL('.', import.meta.url));
```
dai-shi, line 5: Can we preserve the order?

dai-shi (commented):
First review mostly from the preference.

Issue #1927, opened 2026-01-25
Dynamic Routes Failing on Netlify (even with waku-netlify-starter)


Ariel-GonzAguer:
## Description
Dynamic routes (configured with `render: 'dynamic'`) are not working correctly when deployed to Netlify, even using the netlify-starter template (https://github.com/wakujs/waku-netlify-starter). The pages return 404 errors when using the starter and return: **`This function has crashed
An unhandled error in the function code triggered the following message: TypeError -  w is not a function`** when not using the official starter.

## Steps to Reproduce
1. Clone the `waku-netlify-starter` template
2. Create a dynamic route file like `src/pages/otra.tsx` with:
```tsx
export default async function Otra() {
  return (
    <div>
      Otra página
    </div>
  );
}

export const getConfig = async () => {
  return {
    render: 'dynamic',
  } as const;
};
```
3. Deploy to Netlify (NETLIFY=1 pnpm run build
netlify deploy)
4. Try to access the dynamic route (e.g., `/otra`)
5. Observe that the page fails to load or returns a 404 error

## Expected Behavior
Dynamic routes should be properly rendered on Netlify, with the page content being served correctly.

## Actual Behavior
The dynamic route is not working on Netlify deployment. The page either:
- Returns a 404 error
- Is not properly generated during the build process
- Does not respect the `render: 'dynamic'` configuration

## Environment
- **Waku**: Latest version from waku-netlify-starter
- **Deploy Platform**: Netlify
- **Node.js**: 22.x (in Netlify)
- **Package Manager**: pnpm
- **OS**: Windows

## Additional Context
- Static routes work correctly
- The configuration follows the Waku documentation for dynamic rendering
- `netlify.toml` configuration is being used for build settings

## Related Files
- `src/pages/otra.tsx` - Dynamic route example
- `waku.config.ts` - Waku configuration
- `netlify.toml` - Netlify build configuration

You can watch the code of the repo here → https://github.com/Ariel-GonzAguer/test-otra-waku-net
Here is the deployed repo to checked live → https://test-waku-net-dyn.netlify.app/
Here the same repo **but** deployed in vervel → https://test-otra-waku-ver.vercel.app/

Note: If I remember correctly, this error didn't exist a few months ago, but I can't give an exact date.

dai-shi:
Oh, no! I don't think we've tested dynamic routing much and I'm not even sure if it worked before. Would you like to dig into it further? I thought trailing slash issue originally, but the error message doesn't seem so.

[...]

PR #1931, opened 2026-01-27
fix: netlify build enhancer path config

yunho1017:
rel #1927 
Add to config to allow rsc path


Ariel-GonzAguer:
> May I assume this is already reviewed by @Ariel-GonzAguer ?

Everything looks good.
I haven't tried it yet, but it should be fine.

PR #1932, opened 2026-01-27
demo: vite.config.ts as source of truth

### LINKED ISSUE #1860: env change isn't triggering RSC hot reload or React fast refresh

hi-ogawa:
## Summary

**This is a demo/exploration PR for discussion purposes.**

Demonstrates an alternative architecture where `vite.config.ts` is the source of truth instead of `waku.config.ts`. This lets Vite own the entire initialization lifecycle.

Related:
- #1929 (comment about architecture)
- #1860 (.env restart issue - this approach would fix it)

## Benefits of vite.config.ts as source of truth

1. **Functional config** - switch based on `env.command`/`env.mode` (not supported in waku.config.ts)
2. **Waku config lives inline** - no separate waku.config.ts needed
3. **Standard Vite plugin pattern** - familiar to Vite users
4. **Vite owns the lifecycle** - restart on .env change just works (fixes #1860)

## Option A: Explicit waku plugin (this PR)

```typescript
import waku, { defineConfig } from 'waku/vite-plugins';

export default defineConfig((env) => ({
  plugins: [
    tailwindcss(),
    waku({
      // srcDir: 'src',
      // privateDir: 'private',
    }),
  ],
}));
```

## Option B: Implicit waku plugin via top-level `waku` field

```typescript
import { defineConfig } from 'waku/vite-plugins';

export default defineConfig({
  plugins: [tailwindcss()], // waku() auto-added
  waku: {
    // srcDir: 'src',
    // privateDir: 'private',
  },
});
```

## Option C: Waku config shape wrapper (mirrors current waku.config.ts)

```typescript
import { defineConfig } from 'waku/vite-plugins';

export default defineConfig({
  // Same shape as current waku.config.ts
  srcDir: 'src',
  privateDir: 'private',
  vite: {
    plugins: [tailwindcss()],
  },
});
```

`defineConfig` transforms waku config shape to vite config internally.

## Option D: will keep `waku.config.ts` for Waku config (sveltekit, react-router style)

```typescript
// waku.config.ts

export default defineConfig({
  privateDir: 'private',

  // Vite config won't be here any more
  // vite: { ... }
});

// vite.config.ts
import waku, { defineConfig } from 'waku/vite-plugins';

export default defineConfig({
  // waku plugin explicitly in `vite.config.ts`
  plugiins: [waku()],
  // and any other vite configuration
  // ...
})
```

---

## Changes in this PR

- Add default export `waku(config?)` to `waku/vite-plugins`
- Re-export `defineConfig` from `waku/vite-plugins`
- Add `--experimental-vite-config` flag to CLI (for demo purposes)
- Load dotenv in plugin (so `vite dev` works without CLI)
- Add example `vite.config.ts` in `examples/01_template`

## Known issues to solve / TBD

- React plugin may be added twice (`extra-plugins.ts` auto-adds react, but user also adds it in vite.config.ts)
- Need to decide on migration path if this direction is adopted
- Should we still support zero config?
- Should we officially support `vite dev` and `vite build`? (requires `vite` instllation at user project level)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/01_template/vite.config.ts
+++ b/examples/01_template/vite.config.ts
@@ -1,36 +1,43 @@
 import tailwindcss from '@tailwindcss/vite';
-import react from '@vitejs/plugin-react';
 import waku, { defineConfig } from 'waku/vite-plugins';
 
-// export default defineConfig({
-//   plugins: [
-//     tailwindcss(),
-//     // react({
-//     //   babel: {
-//     //     plugins: ['babel-plugin-react-compiler'],
-//     //   },
-//     // }),
-//     waku(),
-//   ],
-// });
+// Benefits of vite.config.ts as source of truth:
+//
+// 1. Functional config - switch based on command/mode
+//    (not supported in waku.config.ts)
+//
+// 2. Waku config lives inline in plugin call
+//    (no separate waku.config.ts needed)
+//
+// 3. Standard Vite plugin pattern
+//    (familiar to Vite users)
+//
+// 4. Vite owns the lifecycle
+//    (restart on .env change just works - fixes #1860)
 
 export default defineConfig((env) => ({
-  // NOTE
-  // functional forms allow env.command === 'build' etc.
-  // to switch configuration easily.
-
   plugins: [
+    // User plugins - you control the order
     tailwindcss(),
-    // TODO: need solution for auto react plugin config from waku plugin
+
+    // Waku plugin with inline config (previously in waku.config.ts)
+    waku({
+      // srcDir: 'src',
+      // distDir: 'dist',
+      // privateDir: 'private',
+    }),
+
+    // React plugin is auto-added by waku if not present
+    // Uncomment to customize:
     // react({
     //   babel: {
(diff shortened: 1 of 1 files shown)


tylersayshi:
I'd be in favor of something like this from the perspetive of centering on a config pattern that is common to the rest of vite frameworks I've come across.

waku.config.ts as our own thing doesn't seem to serve a large amount of utility (over just vite.config.ts), but I may be missing things.

dai-shi:
We decided to use `waku.config.ts` as the source of truth.
Conceptually, Waku uses Vite internally, and beginner users shouldn't notice if it depends on Vite.

Thanks for the suggestion, and sorry if this decision doesn't align with some expectations.

hi-ogawa:
No worries at all. This was just something I've been wondering and happy to have clear answer :+1:

PR #1351, opened 2025-04-04
test(fs-router): e2e test for alt click to use alt key

tylersayshi:
to fix the flakiness

from the docs it shows that macos is the one to use the meta key, which might help explain why the flakiness is only on mac

image

changing to alt should eliminate this as a possibility

PR #1934, opened 2026-01-27
chore: update deps

dai-shi:
including `@vitejs/plugin-rsc`


e2e/fs-router.spec.ts
```diff
@@ -137,7 +137,7 @@ test.describe('fs-router', () => {
     });
     await expect(page.getByRole('heading', { name: 'Home' })).toBeVisible();
     await page.click("a[href='/foo']", {
-      modifiers: ['Alt'],
+      modifiers: ['ControlOrMeta'],
```
dai-shi, line 140: This is reverting #1351.
cc @tylersayshi

PR #1936, opened 2026-01-29
fix(vite-plugins): hack parser options

dai-shi:
for https://github.com/vitejs/vite-ecosystem-ci/actions/runs/21466607412/job/61829997900


dai-shi:
Even with this change, vite-ecosystem-ci still fails for `07_cloudflare` example with:
```
[ERROR] service core:user:waku-project: Uncaught Error: No such module "node:module".
```

Is it a known issue?

PR #1938, opened 2026-01-30
feat(router): typed api request handling as a utility

### LINKED ISSUE #1906: Access route parameters within API function
It would be really useful if when defining a route with parameters that you can access the parameter from within the API function.

For example for a route defined at `/_api/api/thing/[...slug].ts`, it would be great to be able to easily (and maybe even with typesafety) access the value of the `slug` or for `/_api/thing/:a/:b` to be able to acess the `a` and `b` values.

It's possible this exists already and I am just missing it but I couldn't find any examples in the repo or docs when I searched.

Currently i've been manually passing the URL from the request which works but is not great for long term maintaince. Eg. it will break if I move the file around because it hardcodes the path.

Thanks!

tylersayshi:
Decisions:

1. Attach parsed parameters to the request object
2. Allow use of utility type to make the api request handler typesafe

resolves #1906

### DIFF
--- a/packages/waku/src/router/base-types.ts
+++ b/packages/waku/src/router/base-types.ts
@@ -6,12 +6,10 @@ import type {
 // PathsForPages collects the paths from the `createPages` result type
 // GetConfigResponse is a helper for generating types with fs-router
 export type {
+  ApiContext,
   PathsForPages,
   GetConfigResponse,
 } from './create-pages-utils/inferred-path-types.js';
-
-// ApiContex for typed API route handlers
-export type { ApiContext } from './common.js';
 
 export interface RouteConfig {
   // routes to be overridden by users
--- a/packages/waku/src/router/create-pages.tsx
+++ b/packages/waku/src/router/create-pages.tsx
@@ -11,13 +11,13 @@ import {
 import type { PathSpec } from '../lib/utils/path.js';
 import { Children, Slot } from '../minimal/client.js';
 import { ErrorBoundary } from '../router/client.js';
-import type { ApiHandler } from './common.js';
 import type {
   AnyPage,
   GetSlugs,
   PropsForPages,
 } from './create-pages-utils/inferred-path-types.js';
 import { unstable_defineRouter } from './define-router.js';
+import type { ApiHandler } from './define-router.js';
 
 // https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods
 export const METHODS = [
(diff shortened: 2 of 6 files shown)


packages/waku/src/router/define-router.tsx
```diff
@@ -451,6 +451,12 @@ export function unstable_defineRouter(fns: {
       const url = new URL(input.req.url);
       url.pathname = input.pathname;
       const req = new Request(url, input.req);
+      const params = getPathMapping(pathConfigItem.path, input.pathname) ?? {};
+      Object.defineProperty(req, 'params', {
+        value: params,
+        writable: false,
+        enumerable: true,
+      });
```
tylersayshi: This runtime addition parses the params and attaches them to the request object. 

I opted for this over adding a second parameter to the handlers to avoid a breaking change. This also just feels cleaner to me.

tylersayshi:
so for nextjs they use a second parameter which is a typed context object. Similar to what you recommended on discord

https://nextjs.org/docs/app/api-reference/file-conventions/route#route-context-helper

For Tanstack, the first parameter includes `{request, params}` and that seems like an unnecessary breaking change for us to make.

https://tanstack.com/start/latest/docs/framework/react/guide/server-routes

I'd vote for copying nextjs on this with the typed second parameter, named as "context". @dai-shi do you agree?

dai-shi:
@tylersayshi Yes, I like the second parameter. I'm not sure if I like the name `context`, but it doesn't matter as we can change it at anytime, and usually we destrucuture the parameter. Let's call it `apiContext` for now.

packages/waku/src/router/base-types.ts
```diff
@@ -10,6 +10,9 @@ export type {
   GetConfigResponse,
 } from './create-pages-utils/inferred-path-types.js';
 
+// ApiContext and ApiParams for typed API route handlers
+export type { ApiContext, ApiParams } from './common.js';
```
dai-shi: Do we need to export `ApiParams`? I hope `ApiContext<...>['params']` works.

tylersayshi: That does work. I'm fine with only exporting the context. I'll change that

packages/waku/src/router/create-pages.tsx
```diff
@@ -11,6 +11,7 @@ import {
 import type { PathSpec } from '../lib/utils/path.js';
 import { Children, Slot } from '../minimal/client.js';
 import { ErrorBoundary } from '../router/client.js';
+import type { ApiHandler } from './common.js';
```
dai-shi: I'm not sure if I like the responsibility of common.ts, but it's on me.

dai-shi (approved):
LGTM

PR #1944, opened 2026-02-10
fix(internals): expose unstable_resolveConfig separating adapter builders

nihgwu:
it's exposed from internals via https://github.com/wakujs/waku/pull/1877/changes#diff-3419b09b5a26a11c0269bfd7e085ea2a9eccc0d0ddf918c45bb940439b5293a8
but `waku/internals` re-exports handler.js containing virtual: imports that Node's native ESM loader cannot handle

probably we can just call resolveConfig in `defineConfig` to merge with defaults?

### DIFF
--- a/e2e/fixtures/custom-library-adapter/modules/custom-adapter/src/index.js
+++ b/e2e/fixtures/custom-library-adapter/modules/custom-adapter/src/index.js
@@ -1,4 +1,4 @@
-import { unstable_createServerEntryAdapter as createServerEntryAdapter } from 'waku/internals';
+import { unstable_createServerEntryAdapter as createServerEntryAdapter } from 'waku/adapter-builders';
 
 export default createServerEntryAdapter(
   ({ processRequest, processBuild, config }) => {


packages/waku/src/lib/utils/config.ts
```diff
@@ -1,4 +1,45 @@
-import type { Config } from '../../config.js';
+import type { UserConfig } from 'vite';
+
+export interface Config {
```
dai-shi: Why do you need to move the interface definition?

nihgwu: closer to where it's used, and otherwise there will be circular imports

dai-shi:
> handler.js containing virtual: imports that Node's native ESM loader cannot handle

gotcha.

packages/waku/src/lib/utils/config.ts
dai-shi: I see.

dai-shi:
Refactored to split the entry points.

PR #1947, opened 2026-02-11
feat(config): support callback form in defineConfig

2wheeh:
## Description

- Allow `defineConfig` to accept a callback `({ cmd }) => Config` similar to Vite's `defineConfig` pattern
- ~Rename internal `loadEnv` (dotenv) to `loadDotenv` to avoid confusion with Vite `loadEnv`.~ 👉 #2024

This enables common practice with Vite's `defineConfig`:

```ts
// waku.config.ts

export default defineConfig(({ cmd }) => {
  return {
    basePath: cmd === 'build' ? '/prod/' : '',
  };
});
```

### DIFF
--- a/.github/workflows/canary-ci.yml
+++ b/.github/workflows/canary-ci.yml
@@ -4,6 +4,9 @@ on:
   issue_comment:
     types: [created]
 
+permissions:
+  contents: read
+
 jobs:
   parse-version:
     if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/canary-ci run') }}
@@ -13,8 +16,10 @@ jobs:
     steps:
       - name: Get React version from comment
         id: get-version
+        env:
+          COMMENT_BODY: ${{ github.event.comment.body }}
         run: |
-          comment="${{ github.event.comment.body }}"
+          comment="$COMMENT_BODY"
           # Extract version after '/canary-ci run', defaulting to 'canary' if not specified
           version=$(echo "$comment" | sed -n 's/^\/canary-ci run\s*\(\S*\).*/\1/p')
           if [ -z "$version" ]; then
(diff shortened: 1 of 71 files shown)


dai-shi:
Thanks for the suggestion, but unfortunately, I'm not a big fan of this complexity. loadEnv isn't for public api and I'd like to avoid mode abstraction and hide env. Only what I think is acceptable is to pass command. But, I'm not sure if there are many use cases. Do you really want to change basePath with command?

dai-shi:
While I understand the general suggestion, let's first discuss the use cases. Feel free to open a new discussion.

2wheeh:
@dai-shi Sorry for the late reply here! Agreed - basePath example is not really a common case, and for loadEnv, dotenv on the consumer side would be enough.

I still think the command option could be useful for some cases. I'll open a discussion first. Thanks for the review!

dai-shi:
> I still think the command option could be useful for some cases.

#2000 is the real use case reported. Let me add some commits to simplify things.

hi-ogawa:
Since Waku forces `process.env.NODE_ENV`, checking `development vs production` inside `waku.config.ts` should indicate the same thing.

dai-shi:
> Since Waku forces `process.env.NODE_ENV`, checking `development vs production` inside `waku.config.ts` should indicate the same thing.

Oh, really? Then, I don't think we need this.

hi-ogawa:
Haven't verified but I think so. Technically both `build` and `start` becomes `production`, but if I remember correctly, `start` doesn't really make use of `waku.config` except `distDir`.

dai-shi:
I will reopen this if https://github.com/wakujs/waku/issues/2000#issuecomment-4266427214 doesn't work.

PR #1971, opened 2026-02-27
test: add failing test with createPortal

dai-shi:
for https://github.com/wakujs/waku/issues/1937


dai-shi:
I don't think it's a reproduction.

Issue #1912, opened 2026-01-14
Remove fully static routes from production bundle


oscartbeaumont:
I wonder if it would be possible for a route marked as `render: "static"` to be fully omitted from the final build. Currently the output of the static generation is included in the `dist/public` directory as expected but in `dist/server/assets` I am also seeing the wasm blob for my OpenGraph library which is a fully static route.

In my usecase I am generating all of my OpenGraph images statically and deploying to Cloudflare Workers. Cloudflare have a fixed code size limit and my project is currently *just* under the 3MB free tier limit. I am expecting as I add more content to the site this is going to become a problem and removing the OpenGraph code from the final worker would be the easiest way to free up some headroom as all I need are the final `og.png`'s that are static built into `dist/public` during build.

I wonder if this is something that would be concidered adding to Waku as from my understanding if the route is marked as static it won't be needed at runtime?

I assume supporting this would complicate the build process thought as all endpointswould require bundling for static generation so maybe Vite would need to generate two different outputs, one for prerendering and one for production so maybe this feature could add too much complexity.

I have been loving Waku, it's awesome!

dai-shi:
> if the route is marked as static it won't be needed at runtime?

It's not true unless it's fully static site (meaning we can delete dist/server).

> I assume supporting this would complicate the build process

That's also another concern.

(I wonder if some generalization could solve this and #1790 at the same time.)

dai-shi:
> It's not true unless it's fully static site (meaning we can delete dist/server).

Wait, I might be wrong. Now after #1744, it might not be the case. 🤔 

Yeah, I guess it's only that the implementation will be very complicated.

dai-shi:
@oscartbeaumont If you have a chance, please try #2044 and see if it solves your issue.

oscartbeaumont:
Awesome! That's super exciting! Ended up having to upgrade my Cloudflare account because I quickly went over the limit.

Would you expect it to work when doing `CLOUDFLARE=1 waku build`?

I am seeing the log output about pruning when doing `waku build` but not when building for Cloudflare?

Haven't had a chance to look into if this is an issue on my end so could just be my setup (it's [open source](https://github.com/specta-rs/website) if useful.

dai-shi:
> I am seeing the log output about pruning when doing `waku build` but not when building for Cloudflare?

Ahhh, nice catch. cloudflare adapter is tricky. Let me work on it. Thanks.

PR #1974, opened 2026-03-01
feat: split runtime/build server entries to exclude static routes from runtime bundle

TiwariLokesh:
## Summary

This PR addresses #1912 by introducing a proper architectural separation between runtime and build server entries.

Instead of pruning static route modules after build, this change prevents fully static routes from entering the runtime module graph in the first place.

---

## Motivation

Currently, routes marked with `render: "static"` are still bundled into the runtime server output even though they are only needed during SSG. This increases the production bundle size (e.g. large OpenGraph libraries, WASM blobs) and can cause deployment size constraints in environments like Cloudflare Workers.

The previous workaround relied on post-build pruning. This PR removes that approach and replaces it with a cleaner graph-level separation.

---

### 1. Runtime / Build Entry Split

Introduced two separate virtual server entries:

- `virtual:vite-rsc-waku/server-entry-runtime`
- `virtual:vite-rsc-waku/server-entry-build`

Runtime and build now generate independent module graphs.

---

### 2. Managed Mode Runtime Exclusion

In managed `fsRouter` mode:

- Runtime entry excludes fully static page modules using `import.meta.glob` negative patterns.
- Build entry keeps the full route graph to allow SSG to function normally.

---

### 3. Mode-Aware Router Resolution

- `getConfigs({ mode })` is now propagated through `define-router`.
- Runtime paths resolve configs using `mode: "runtime"`.
- Build paths resolve configs using `mode: "build"`.

---

### 4. Removed Post-Build Pruning

`build-static-files.ts` no longer performs file-level pruning.
Static route exclusion is now handled at graph-construction time instead of cleanup time.

---

## Result

- Static routes are still generated during SSG.
- Static-only route modules are no longer included in the runtime server bundle (in managed mode).
- No public API changes.
- Backward compatible.
- Cleaner architecture.

---

## Notes

Static route detection in managed mode is heuristic-based (literal `render: "static"` detection) and conservative to avoid accidentally excluding dynamic routes.

Custom server entries may still bundle static modules if explicitly imported.

### DIFF
--- a/packages/waku/src/lib/build-output-pruning.ts
+++ b/packages/waku/src/lib/build-output-pruning.ts
@@ -1,5 +1,5 @@
-import path from 'node:path';
 import { rm } from 'node:fs/promises';
+import path from 'node:path';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import {
   BUILD_METADATA_FILE,
--- a/packages/waku/src/lib/vite-plugins/build-static-files.ts
+++ b/packages/waku/src/lib/vite-plugins/build-static-files.ts
@@ -7,10 +7,10 @@ import { pathToFileURL } from 'node:url';
 import pc from 'picocolors';
 import type { Plugin } from 'vite';
 import {
+  type BuildOutputGraph,
   applyBuildOutputPrunePlan,
   collectBuildOutputGraph,
   computeBuildOutputPrunePlan,
-  type BuildOutputGraph,
 } from '../build-output-pruning.js';
 import { joinPath } from '../utils/path.js';
 import { createProgressLogger } from '../utils/progress-logger.js';


packages/waku/src/lib/utils/managed.ts
```diff
@@ -1,9 +1,113 @@
+import fs from 'node:fs';
+import { readdir, readFile } from 'node:fs/promises';
+import path from 'node:path';
 import { EXTENSIONS, SRC_MIDDLEWARE, SRC_PAGES } from '../constants.js';
 
-export const getManagedServerEntry = (srcDir: string) => {
```
dai-shi, line 3: I'm not sure if I 100% follow the idea, but it shouldn't be managed mode only. So, changing code here seems not ideal.

packages/waku/src/router/create-pages.tsx
```diff
@@ -641,7 +641,8 @@ export const createPages = <
   };
 
   const definedRouter = unstable_defineRouter({
-    getConfigs: async () => {
+    getConfigs: async (context?: { mode?: 'runtime' | 'build' }) => {
+      const mode = context?.mode || 'runtime';
```
dai-shi: The change here is not nice and not acceptable. only define-router should be responsible to this capability.

pnpm-workspace.yaml
dai-shi, line 1: Do not change this in this PR.

dai-shi (changes requested):
Thanks for working on it.

> Instead of pruning static route modules after build, this change prevents fully static routes from entering the runtime module graph in the first place.

Sounds good.

We have waku/router on top of waku/minimal api. I wonder which is responsible to this capability. If it's waku/minimal related, we should work on it first, without caring waku/router much. (At least, the PR must be separated.)

dai-shi:
Actually, while not sure, this might be somewhat related with #1790, which I'm hoping to tackle sometime soon.

TiwariLokesh:
Thanks for the feedback!

You're right - the current implementation mixes router-level responsibility with managed mode logic.

I'll revise the PR to:
- revert the pnpm-workspace.yaml change
- avoid modifying create-pages.tsx
- move the runtime/build mode handling fully into define-router so that the capability is router-driven rather than managed-mode specific.

I'll push an updated version shortly.

packages/waku/src/router/create-pages.tsx
TiwariLokesh: Thanks for pointing this out!

I've updated the implementation so that runtime/build filtering is handled entirely in `define-router`.
`create-pages.tsx` is now mode-agnostic as suggested.

Let me know if this looks better.

TiwariLokesh:
Thanks for the detailed review!

I've pushed an update addressing the feedback:

* Reverted changes to `pnpm-workspace.yaml`
* Removed mode-aware logic from `create-pages.tsx`
* Removed managed-mode filtering
* Moved runtime/build filtering entirely into `define-router`

Now `create-pages` simply constructs the route configs and `define-router` handles runtime/build separation internally.

Build verification:
pnpm --filter waku compile:code
pnpm --filter waku compile:types

Please let me know if there’s anything else I should adjust.

packages/waku/src/lib/utils/managed.ts
```diff
@@ -1,5 +1,4 @@
 import { EXTENSIONS, SRC_MIDDLEWARE, SRC_PAGES } from '../constants.js';
-
```
dai-shi, line 2: remove changes in this file.

packages/waku/src/lib/vite-plugins/user-entries.ts
```diff
@@ -11,9 +11,23 @@ export function userEntriesPlugin({ srcDir }: { srcDir: string }): Plugin {
...
-      if (source === 'virtual:vite-rsc-waku/server-entry-inner') {
+      if (source === 'virtual:vite-rsc-waku/server-entry-runtime-inner') {
+        const resolved = await this.resolve(
+          `/${srcDir}/${SRC_SERVER_ENTRY}`,
+          undefined,
+          options,
+        );
+        return resolved ? resolved : '\0' + source;
+      }
+      if (source === 'virtual:vite-rsc-waku/server-entry-build-inner') {
```
dai-shi: Any difference between `-runtime-inner` and `-build-inner`?

dai-shi (changes requested):
> Please let me know if there’s anything else I should adjust.

Don't expect this to be merged very soon. There are many things to consider some design decisions as I noted.

Meanwhile, make this PR minimal. Please don't change unrelated files. Review changes before pushing. For now, the quality feels AI generation level.

dai-shi:
marking draft until we choose some design decisions.

dai-shi:
Hmm, I don't think this approach is good, which only works for managed mode. We often use non-managed mode, which should be fully supported and managed mode is a special case.

I'll try "pruning static route modules after build" approach, because essentially the build time doesn't change.

dai-shi:
I thought the alternative approach could be done more simply. However, I don't like the implemented complexity. I don't think OP is happy with this anyway. post-build pruning workaround should be more controllable.
Let me close this, as we don't plan this feature for now.

PR #1979, opened 2026-03-06
fix(router): empty router issue and added mapping

### LINKED ISSUE #1978: Static API route with wildcard causes file-system conflicts
when there's an empty path in `staticPaths`, the error can be reproduced.

```ts
// _api/llms.mdx/docs/[...slugs].ts
import { ApiContext } from 'waku/router';
import { unstable_notFound } from 'waku/router/server';

export async function GET(_: Request, { params }: ApiContext<'/llms.mdx/docs/[...slugs]'>) {
  const slugs = params.slugs;
  console.log(slugs);
  return new Response('hello world', {
    headers: {
      'Content-Type': 'text/markdown',
    },
  });
}

  return {
    render: 'static' as const,
    staticPaths: [[], ['test']],
  } as const;
}
```

Image

tylersayshi:
fixes #1978

### DIFF
--- a/packages/waku/src/router/create-pages.tsx
+++ b/packages/waku/src/router/create-pages.tsx
@@ -483,9 +483,9 @@ export const createPages = <
           throw new Error('staticPaths does not match with slug pattern');
         }
         if (staticPath.length === 0 && numWildcards > 0) {
-          throw new Error(
+          console.warn(
             `Empty staticPaths entry is not supported for wildcard routes. ` +
-              `Route "${page.path}" has a wildcard segment, so each staticPaths entry must contain at least one element.`,
+              `Route "${page.path}" has a wildcard segment, so each staticPaths entry should contain at least one element.`,
           );
         }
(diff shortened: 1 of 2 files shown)


packages/waku/src/router/define-router.tsx
```diff
@@ -665,11 +675,21 @@ export function unstable_defineRouter(fns: {
       if (!pathname) {
         continue;
       }
+      // If this pathname is also a directory prefix for another static API
+      // route, emit as pathname/index to avoid file-system conflicts.
```
dai-shi: I'm not sure about this. What use cases is this trying to cover?

tylersayshi: The one from #1978. It supports the empty staticPath as a / on /[...wildcard]

dai-shi: Hmm, but adding `/index` here seems wrong. Any other solution?

tylersayshi: this should be better now - just warning based on the issue conversation

PR #1981, opened 2026-03-07
feat: support no-trailing-slash middleware in template

dai-shi:
following #1980


examples/01_template/src/middleware/no-trailing-slash.ts
```diff
@@ -0,0 +1,3 @@
+import { trimTrailingSlash } from 'hono/trailing-slash';
+
+export default () => trimTrailingSlash({ alwaysRedirect: true });
```
dai-shi, line 3: This might cause some troubles in edge cases, but let's see how it goes.

PR #1982, opened 2026-03-13
feat: migrate to Vite 8

Master-Hash:
Main changes:

* `@vitejs/plugin-react@6` now doesn't include a babel options. https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-react#readme
* ~~Vite 8 use lightningcss instead of postcss, so I use `@stylexjs/unplugin` instead of babel + `@stylexjs/postcss`~~
* Remove deprecated API `parseAstAsync` and `transformWithEsbuild`
* Rename `rollupOptions` -> `rolldownOptions`

### DIFF
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -244,11 +244,57 @@ test.describe(`create-pages`, () => {
     ).toBeVisible();
     const pendingSeen = waitForSelectorSeen(page, PENDING_SELECTOR);
     await clickClientLink(page, '/long-suspense/2');
+    await page.waitForFunction(
+      ([pendingSel, sel]) => {
+        const pathname = window.location.pathname;
+        const pendingElement = document.querySelector(pendingSel);
+        const heading = document.querySelector(sel);
+        return (
+          pendingElement?.textContent === 'Pending...' &&
+          pathname === '/long-suspense/1' &&
+          heading?.textContent === 'Long Suspense Page 1'
+        );
+      },
+      [PENDING_SELECTOR, SELECTOR] as const,
+      { timeout: 1000 },
+    );
     await pendingSeen;
     await waitForSelectorText(page, SELECTOR, 'Long Suspense Page 2');
+    const pendingSeen2 = waitForSelectorSeen(page, PENDING_SELECTOR);
     await clickClientLink(page, '/long-suspense/3');
+    await page.waitForFunction(
+      ([pendingSel, sel]) => {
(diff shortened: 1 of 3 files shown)


Master-Hash:
The change about stylex doesn't depend on vite 8, and could be separate.

dai-shi:
> Don't merge

I plan to merge this after v1-alpha.6 is released.

Meanwhile, I think v1-alpha.5 should work with vite 8. Would anyone like to confirm?



> The change about stylex doesn't depend on vite 8, and could be separate.

Please send a separate pr then.

Master-Hash:
> Meanwhile, I think v1-alpha.5 should work with vite 8. Would anyone like to confirm?

End user can use 

```json
{
  "pnpm": {
    "overrides": {
      "vite": "8.0.0",
    }
  }
}
```

and it works perfectly, only with some warning about deprecated APIs.

> Please send a separate pr then.

Okay.

dai-shi:
> and it works perfectly, only with some warning about deprecated APIs.

Cool. The warning is expected.

docs/guides/react-compiler.mdx
```diff
@@ -13,20 +13,14 @@ npm install -D @vitejs/plugin-react babel-plugin-react-compiler
 ```ts
 // waku.config.ts
-import react from '@vitejs/plugin-react'; // make sure v5.0.1 or above is installed
+import react, { reactCompilerPreset } from '@vitejs/plugin-react'; // make sure v6.0.0 or above is
```
dai-shi:
```suggestion
import react, { reactCompilerPreset } from '@vitejs/plugin-react'; // make sure v6.0.0 or above is installed
```

nihgwu:
currently it works with vite8, we have to use resolution, it would be nice to have this released

dai-shi:
> currently it works with vite8, we have to use resolution, it would be nice to have this released

Have you tested the pkg-pr-new preview of this PR?

examples/41_path-alias/waku.config.ts
```diff
@@ -1,8 +1,10 @@
-import tsconfigPaths from 'vite-tsconfig-paths';
 import { defineConfig } from 'waku/config';
 
 export default defineConfig({
   vite: {
-    plugins: [tsconfigPaths()],
+    resolve: {
+      tsconfigPaths: true,
+    },
+    plugins: [],
```
dai-shi:
```suggestion
```

PR #1983, opened 2026-03-13
chore(examples/37_css-stylex): use `@stylexjs/unplugin`

Master-Hash:
Vite 8 use lightningcss instead of postcss, so original approach will break. This PR is the preparation.

This approach is also used by StyleX upstream: https://github.com/facebook/stylex/tree/main/examples/example-waku

Split from https://github.com/wakujs/waku/pull/1982

### DIFF
--- a/examples/37_css-stylex/src/pages/index.tsx
+++ b/examples/37_css-stylex/src/pages/index.tsx
@@ -1,7 +1,7 @@
 import '../index.css';
 import * as stylex from '@stylexjs/stylex';
-import { Counter } from '../components/counter.js';
-import { DevStyleXInject } from '../components/stylex-dev-injection.js';
+import { Counter } from '../components/counter';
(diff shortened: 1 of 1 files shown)


examples/37_css-stylex/src/pages/index.tsx
```diff
@@ -1,6 +1,7 @@
 import '../index.css';
 import * as stylex from '@stylexjs/stylex';
-import { Counter } from '../components/counter';
+import { Counter } from '../components/counter.js';
+import { DevStyleXInject } from '../components/stylex-dev-injection.js';
```
dai-shi: We currently use no extensions in examples.

```suggestion
import { Counter } from '../components/counter';
import { DevStyleXInject } from '../components/stylex-dev-injection';
```

dai-shi (approved):
LGTM

mellyeliu:
cool stuff :D

PR #1984, opened 2026-03-13
feat: support performance tracks for server components

dai-shi:
## Summary

Support performance tracks for server components and server requests.

See: https://react.dev/reference/dev-tools/react-performance-tracks

## Notes

- reuses Vite's HMR websocket for the debug transport
- patches RSDW in dev so Waku's plain-object payloads keep `_debugInfo`

### DIFF
--- a/packages/waku/src/lib/utils/ssr.ts
+++ b/packages/waku/src/lib/utils/ssr.ts
@@ -19,6 +19,8 @@ Promise.resolve(new Response(new ReadableStream({
   .map((line) => line.trim())
   .join('');
 
+// These constants are defined in packages/waku/src/minimal/client.tsx.
+// TODO(daishi): We should avoid duplicating definitions.
 const KEY_RESPONSE = 'r';
 const KEY_DEBUG_ID = 'd';
 


hi-ogawa:
I was curious about this feature, but couldn't figure out (also haven't research that much). Is there something `@vitejs/plugin-rsc` need to provide or requires only framework side wiring? If possible, I think I'd also add the wiring in standalone example in `@vitejs/plugin-rsc`.

dai-shi:
@hi-ogawa Thanks. I'm currently working and struggling with it.

- The biggest issue is that Waku's flight protocol is unique and not supported by React, so I need a hack. Still work in progress
- The debug channel is implemented with Vite's websocket. I need some refactors, but once done, this might be something that can be ported to `@vitejs/plugin-rsc`. Will see.

dai-shi:
> Is there something `@vitejs/plugin-rsc` need to provide

In addition to using HMR ws, I implemented part of the feature as a vite plugin, so probably it can be extracted for other framework. And, it will be able to hide the `globalThis` hack. One of the trickiest parts is to set the custom http header, which I gave up with a hack.

packages/waku/src/lib/utils/ssr.ts
```diff
@@ -19,14 +19,21 @@ Promise.resolve(new Response(new ReadableStream({
   .map((line) => line.trim())
   .join('');
 
+const KEY_RESPONSE = 'r';
+const KEY_DEBUG_ID = 'd';
```
dai-shi, line 25:
```suggestion
// These constants are defined in packages/waku/src/minimal/client.tsx.
// TODO(daishi): We should avoid duplicating definitions.
const KEY_RESPONSE = 'r';
const KEY_DEBUG_ID = 'd';
```

packages/waku/src/minimal/client.tsx
```diff
@@ -173,12 +196,21 @@ const fetchRscInternal: FetchRscInternal = (
...
+      [KEY_TEMPORARY_REFERENCES]: temporaryReferences,
+    };
     return undefined as never;
   }
   return createFromFetch<Elements>(checkStatus(responsePromise), {
     callServer: (funcId: string, args: unknown[]) =>
       unstable_callServerRsc(funcId, args, () => fetchRscStore),
+    ...(import.meta.hot && {
+      debugChannel: debug?.debugChannel,
+      findSourceMapURL,
```
hi-ogawa, line 212: I thought `@vitejs/plugin-rsc/browser` bakes-in `findSourceMapURL` always. Is this customizing something beyond what `@vitejs/plugin-rsc` provides?

dai-shi: Oh, I didn't realize that. So, we can simply delete it here, right?

dai-shi: https://github.com/wakujs/waku/pull/1993

PR #1986, opened 2026-03-17
feat(router): support prefixed slugs

### LINKED ISSUE #1972: File-based routing does not support prefixed dynamic segments
Hi team, thanks for building Waku.

I’m seeing a limitation in file-based routing: dynamic segments appear to work only when the segment is exactly bracketed (e.g. `[param]`), but not when a static prefix is attached (e.g. `prefix[param]`).

## Expected behavior

File-based routing should allow dynamic segments with static prefixes, so URLs like:

- `/@value`
- `/u-value`
- `/org-value`

can be represented directly in route segment naming.

## Actual behavior

Prefixed dynamic segment patterns are not recognized correctly.
Only plain dynamic segments (without prefix) seem to be supported.

## Why this matters

Many URL conventions rely on prefixed identifiers (for example handle-like URLs). Without prefix support, developers must change URL structure or add extra routing workarounds.

## Request

Could Waku support static-prefix + dynamic-segment patterns in file-based routing, or provide an official recommended pattern for this use case?

tylersayshi:
resolves #1972

### DIFF
--- a/packages/website/private/contents/post-006.mdx
+++ b/packages/website/private/contents/post-006.mdx
@@ -6,7 +6,7 @@ author: sophia
 date: 2024/12/31
 ---
 
-Continuing our winter retrospective of Waku's recently added features, today we're excited to highlight type-safe routing contributed by [Tyler Lawson](https://tylur.dev/). This addition brings compile-time safety to your routing and navigation logic while staying true to Waku's minimalist principles.
+Continuing our winter retrospective of Waku's recently added features, today we're excited to highlight type-safe routing contributed by [Tyler Lawson](https://tyler.fun/). This addition brings compile-time safety to your routing and navigation logic while staying true to Waku's minimalist principles.
 
 ## Type-safe routing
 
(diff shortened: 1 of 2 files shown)


e2e/fs-router.spec.ts
```diff
@@ -335,6 +335,22 @@ test.describe('fs-router', () => {
...
   });
 
+  test('prefixed dynamic segment @[username]', async ({ page }) => {
+    await page.goto(`http://localhost:${port}/@alice`);
+    await expect(
+      page.getByRole('heading', { name: 'Profile / alice' }),
+    ).toBeVisible();
+  });
+
+  test('prefixed dynamic segment @[username] with different value', async ({
```
dai-shi, line 345: "with different value"?

e2e/fixtures/fs-router/src/pages/@[username].tsx
```diff
@@ -0,0 +1,15 @@
+import type { PageProps } from 'waku/router';
+
+const Page = ({ username }: PageProps<'/@[username]'>) => (
```
Comment (dai-shi) [inline e2e/fixtures/fs-router/src/pages/@[username].tsx:3]: If we have a file `./src/pages/@foo.tsx`, it will be used with more priority than `./src/pages/@[username].tsx`, right? Could you add the test case for it?

tylersayshi:
the copilot review seems unhelpful to me for what it's pointing out. @dai-shi if you see anything that looks particularly important I can reconsider and make fixes, but otherwise I'm tempted to disregard it.

dai-shi:
I'm not familiar with every details, so copilot review might be more helpful than me. But, if you see they are pointless, feel free to ignore.

dai-shi (approved):
My review is mainly on e2e tests.

PR #1987, opened 2026-03-17
fix(cmd): use correct variable in ready log message

snimshchikov:
Right now when setting hostname via HOST env (which seems to be supported, but not documented), the console output is:
`ready: Listening on http://localhost:8080/`
The correct output should be (if HOST env is 0.0.0.0):
`ready: Listening on http://0.0.0.0:8080/`

This pr changes output when host flag is NOT set and HOST env variable IS set
This pr doesn't change the program's behaviour


dai-shi:
> Right now when setting hostname via HOST env (which seems to be supported, but not documented)

Uh, I wasn't aware it's supported. `if(host)` does the trick... It was by accident in #1815.

dai-shi:
Let's make it explicit: #1994

PR #1990, opened 2026-03-18
fix(router): restore cross-page hash scrolling after navigation

nihgwu:
**Bug**

- Client-side navigation to a different route with a hash tried to scroll before the destination route had committed.
- The target element was not in the document yet, so the hash scroll could miss the anchor and fall back to the top of the page.
- Same-page hash navigation still worked because the target element was already mounted.

**Changes**

- Defer route hash scrolling until after the next route has committed.
- Add regression coverage for cross-page hash navigation using a destination anchor rendered by the target route.

### DIFF
--- a/examples/11_fs-router/src/pages/_layout.tsx
+++ b/examples/11_fs-router/src/pages/_layout.tsx
@@ -66,6 +66,9 @@ const HomeLayout = ({ children }: { children: ReactNode }) => (
       <li>
         <Link to="/slice-page">Slice Page</Link>
       </li>
+      <li>
+        <Link to="/debug">Debug</Link>
+      </li>
     </ul>
     {children}
   </div>
--- a/examples/11_fs-router/src/pages/debug.tsx
+++ b/examples/11_fs-router/src/pages/debug.tsx
@@ -1,10 +1,19 @@
+import { Suspense } from 'react';
 import { getRouterConfigs } from '../waku.server.js';
+
+const SlowServerInfo = async () => {
+  await new Promise<void>((resolve) => setTimeout(resolve, 500));
+  return <p>Server delay complete at {new Date().toLocaleTimeString()}</p>;
+};
 
 export default async function Debug() {
   const configs = await getRouterConfigs();
   return (
     <div>
       <h4>Route inspection</h4>
+      <Suspense fallback={<p>Loading server info...</p>}>
+        <SlowServerInfo />
+      </Suspense>
       <pre>{JSON.stringify(configs, null, 2)}</pre>
     </div>
   );
--- a/packages/waku/src/lib/react-types.d.ts
+++ b/packages/waku/src/lib/react-types.d.ts
@@ -35,6 +35,7 @@ declare module 'react-server-dom-webpack/server.edge' {
   interface TemporaryReferenceSet {}
 
   type Options = {
+    debugChannel?: { readable?: ReadableStream; writable?: WritableStream };
     environmentName?: string;
     identifierPrefix?: string;
     signal?: AbortSignal;
@@ -75,6 +76,10 @@ declare module 'react-server-dom-webpack/client' {
 
   type Options<T> = {
     callServer?: CallServerCallback<T>;
+    debugChannel?:
+      | { writable?: WritableStream; readable?: ReadableStream }
+      | undefined;
+    findSourceMapURL?: (filename: string, environmentName: string) => string;
     temporaryReferences?: TemporaryReferenceSet;
   };
 
--- a/packages/waku/src/lib/types.ts
+++ b/packages/waku/src/lib/types.ts
@@ -10,6 +10,11 @@ export type Unstable_RenderRsc = (
   },
 ) => Promise<ReadableStream>;
 
+// Experimental: render RSC only for parse/copy flows without a debug channel.
+export type Unstable_RenderRscForParse = (
+  elements: Elements,
+) => Promise<ReadableStream>;
+
 export type Unstable_ParseRsc = (
   rscStream: ReadableStream,
 ) => Promise<Elements>;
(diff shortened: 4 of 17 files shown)


packages/waku/src/router/client.tsx
```diff
@@ -819,6 +820,24 @@ const InnerRouter = ({
     setRoute((prev) => (isSameRoute(prev, initialRoute) ? prev : initialRoute));
   }, [initialRoute]);
   const [err, setErr] = useState<unknown>(null);
+  const pendingScrollRef = useRef<{
```
dai-shi: We generally avoid using useRef. (AI often overuses it.)
I think we can use useState here.
`route` isn't needed.
I don't think we need both `behavior` and `scrollTopForMissingHash`.
What we really need is `pathChanged`.
```js
const [pathChanged, setPathChanged] = useState(false);
```
(Or, I wonder if we re-use `routeRef` to keep the previous route, we may not need `pathChanged` 🤔 )

packages/waku/src/router/client.tsx
```diff
@@ -819,6 +820,24 @@ const InnerRouter = ({
...
+    behavior: ScrollBehavior;
+    route: RouteProps;
+    scrollTopForMissingHash: boolean;
+  } | null>(null);
+
+  useLayoutEffect(() => {
```
dai-shi, line 826: Can we use useEffect here? Or does it break the behavior?

packages/waku/src/router/client.tsx
```diff
@@ -819,6 +820,24 @@ const InnerRouter = ({
     setRoute((prev) => (isSameRoute(prev, initialRoute) ? prev : initialRoute));
   }, [initialRoute]);
```
dai-shi, line 821: Now, I'm not sure why this is not this:

```suggestion
  }, [route]);
```

packages/waku/src/router/client.tsx
nihgwu: I would keep `useLayoutEffect`. `useEffect` should still work, but risks a visible paint at the wrong scroll position before correcting it. For router scroll,`useLayoutEffect` is the better timing.

dai-shi: > risks a visible paint at the wrong scroll position

Are you sure? Do you mean screen paint before scroll?

> For router scroll,useLayoutEffect is the better timing.

Could you find some evidences? For example, how other React frameworks deal with it?

packages/waku/src/router/client.tsx
```diff
@@ -939,15 +930,17 @@ const InnerRouter = ({
...
-              scrollTopForMissingHash: pathChanged,
-            }
-          : null;
         routeRef.current = nextRoute;
         setRoute(nextRoute);
+        if (options.shouldScroll) {
+          if (pathChanged && nextRoute.hash) {
+            setPendingHashScroll(true);
+          } else {
+            scrollToRoute(nextRoute, scrollBehavior, pathChanged);
```
dai-shi: Can we delay this scroll too? It's a bit delayed, but I'm not sure if there's a good reason to run in sync.

packages/waku/src/router/client.tsx
nihgwu: btw that reply was from myself not AI
I think in this case we should do it as early as possible, so useLayoutEffect is better, otherwise user would see layout shift, but here we should just present the latest state(offset) at very beginning

packages/waku/src/router/client.tsx
nihgwu: isn't this from existing logic, I don't think it's a concern here

packages/waku/src/router/client.tsx
dai-shi: I can't believe the user see the layout shift. Let's do useEffect for now for this PR and see how people see the layout shift.

packages/waku/src/router/client.tsx
dai-shi: Yes, this is from the existing logic, but I don't think that's very appropriate now that you raised this PR.

dai-shi:
Let me add some commits and see how they work for you.

packages/waku/src/router/client.tsx
dai-shi: Actually, I was probably wrong. And for https://github.com/wakujs/waku/pull/1990/changes#r2963107697, I think useLayoutEffect is necessary.

packages/waku/src/router/client.tsx
dai-shi: Noticed we update the route ref in a callback.

dai-shi:
Just one commit. Could you check if it works for you?

nihgwu:
@dai-shi I tested it yesterday but saw some issue with scrolling margin, and I can't confirm if it's on our side of from this change because of cache issue, and I didn't continue because of too late
I'll check again
the hash scrolling issue is gone for sure

dai-shi:
I believe this solves #1399 too. If not, please open a new issue. @Raicuparta

nihgwu:
re https://github.com/wakujs/waku/pull/1990#issuecomment-4103565739

@dai-shi confirmed the issue is at https://github.com/wakujs/waku/blob/076c49994a7528a2903d4094e68105b29d7ac272/packages/waku/src/router/client.tsx#L727
it doesn't take margin into account, why we use margin to preserve space to avoid be covered by header, but now it's always covered by header

dai-shi:
> re [#1990 (comment)](https://github.com/wakujs/waku/pull/1990#issuecomment-4103565739)
> 
> @dai-shi confirmed the issue is at
> 
> https://github.com/wakujs/waku/blob/076c49994a7528a2903d4094e68105b29d7ac272/packages/waku/src/router/client.tsx#L727
> 
> 
> it doesn't take margin into account, why we use margin to preserve space to avoid be covered by header, but now it's always covered by header

Thanks for confirming. Does the issue exist even before #1990? Hope to see your solution.

nihgwu:
> Does the issue exist even before https://github.com/wakujs/waku/pull/1990?

I believe so, that's the nature of `scrollTo`, a better way is `element.scrollIntoView`, but this could be problematic when the element is in a scrollable container not body, and then body is scrolled unexpectedly, I don't have the answer yet as currently we use `scrollIntoView` where it's safe to do so

PR #1993, opened 2026-03-20
fix(minimal/client): do not pass findSourceMapURL unnecessarily

dai-shi:
https://github.com/wakujs/waku/pull/1984#discussion_r2963836507

### DIFF
--- a/packages/waku/src/minimal/client.tsx
+++ b/packages/waku/src/minimal/client.tsx
@@ -71,18 +71,6 @@ const mergeElementsPromise = (
   return getCached(getResult, cache2, b);
 };
 
-// Keep this local for now.
-// Importing the helper from @vitejs/plugin-rsc pulls in its browser runtime,
-// which breaks Node-side tests.
-// TODO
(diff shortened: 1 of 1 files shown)


hi-ogawa (approved):
Yes, this looks like a same thing as https://github.com/vitejs/vite-plugin-react/blob/87e505bedaa60bf3f061cb0d4ecd0cdc550b5567/packages/plugin-rsc/src/react/browser.ts#L53-L65

packages/waku/src/minimal/client.tsx
hi-ogawa, line 13: I noticed `react-server-dom-webpack/client` is imported but this always gets swapped `@vitejs/plugin-rsc/browser` so it should be fine. 

https://github.com/wakujs/waku/blob/2ce74ee2381f6c0593b8246f33043434706889fe/packages/waku/src/lib/vite-plugins/patch-rsdw.ts#L26-L42

We should be able to actually rewrite this line directly, but think one issue is I need to type `@vitejs/plugin-rsc/browser` more thoroughly.

PR #1995, opened 2026-03-21
refactor: remove `react-server-dom-webpack/client` alias

hi-ogawa:
Follow-up to https://github.com/wakujs/waku/pull/1993#discussion_r2964040502

Waku was still using `react-server-dom-webpack/client` instead of accessing `@vitejs/plugin-rsc/browser` API even though they were forcefully aliased to the same thing via plugin.

`@vitejs/plugin-rsc` now has a proper options types for RSC API surfaces (see https://github.com/vitejs/vite-plugin-react/pull/1154) and it allows replacing directly without losing types.


dai-shi:
> Waku was still using `react-server-dom-webpack/client` instead of accessing `@vitejs/plugin-rsc/browser` API

That's actually intentional, and I hoped react-server-dom-vite to be real. Though, I agree the aliasing is counter-intuitive and not very straightforward. So, yeah, I can change that.

dai-shi:
Per https://github.com/wakujs/waku/pull/2043#discussion_r3171686463, I think it makes sense to drop patch-rsdw.

But, for now, there are two major issues:
- tests don't pass.
- we still need the patch for performance tracks for server components.

So, unless we have a complete solution to drop patch-rsdw, let's avoid partial change. Closing this for now, but I'd love to see the way to drop patch-rsdw completely.

PR #2002, opened 2026-03-27
fix(define-router): Cache bytes instead of promises

jamesopstad:
Fixes https://github.com/wakujs/waku/issues/1975

Fixes cross-request promise resolution errors in Cloudflare Workers (workerd runtime) by changing `cachedElementsForRequest` to store serialized RSC bytes (Uint8Array) instead of `Promise<ReactNode>`.

See https://developers.cloudflare.com/workers/observability/errors/#cannot-perform-io-on-behalf-of-a-different-request for more info about these errors.

Note that while debugging this I noticed that HMR updates trigger two identical RSC updates, which made this issue particularly apparent.

### DIFF
--- a/packages/waku/tests/stream.test.ts
+++ b/packages/waku/tests/stream.test.ts
@@ -1,7 +1,7 @@
 import { describe, expect, test } from 'vitest';
 import {
-  batchReadableStream,
   base64ToBytes,
+  batchReadableStream,
   bytesToBase64,
   bytesToStream,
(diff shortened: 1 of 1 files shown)


dai-shi:
Thanks. It looks generally good. I'll add some commits for refactor later.

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/24074438174

PR #2003, opened 2026-03-27
fix(router): respect scroll-margin-top in hash scrolling

nihgwu:
## Summary
- make hash scrolling in `scrollToRoute` account for the target element's `scroll-margin-top`
- refactor missing-element branch to avoid repeated `element` null checks

## Why
Anchored navigation currently scrolls to the raw element top, which can hide content under sticky headers even when `scroll-margin-top` is set.

## Changes
- if hash target is missing:
  - return early when `scrollTopForMissingHash` is false
  - otherwise scroll to top and return
- if hash target exists:
  - read `scrollMarginTop` via `getComputedStyle(element).scrollMarginTop`
  - subtract it from the final top offset

## Notes
No behavior change for non-hash route navigation.

### DIFF
--- a/docs/create-pages.mdx
+++ b/docs/create-pages.mdx
@@ -580,3 +580,63 @@ export default adapter(pages);
 ```
 
 This allows you to have a `dynamic` slice component while keeping the rest of the page static.
+
+#### Slug slices
+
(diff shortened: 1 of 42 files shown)


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/24083122685

PR #2005, opened 2026-04-03
fix(router): fallback to document navigation on refetch failures

nihgwu:
## Summary
- fall back to full document navigation when a client route refetch fails with a non-Waku error
- keep the existing abort, 404, and structured redirect handling intact
- cover the push and replace fallback behaviour in router tests

---

We have IAP enabled, when clicking a link, then it triggers RSC payload fetch, then IAP decides the session is not valid enough for that request, so it returns a redirect to Google OAuth, and the fetch will fail because of CORS and crash the app

### DIFF
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -222,18 +222,24 @@ test.describe(`create-pages`, () => {
     ({ port, stopApp } = await startApp(mode));
   });
 
-  test('server page unreachable', async ({ page, mode, browserName }) => {
+  test('server page unreachable', async ({ page, mode }) => {
     await page.goto(`http://localhost:${port}`);
     await waitForHydration(page);
     await stopApp();
+    const targetUrl = `http://localhost:${port}/error`;
+    const navigationFailure = page.waitForEvent(
+      'requestfailed',
+      (request) =>
+        request.isNavigationRequest() &&
+        request.resourceType() === 'document' &&
+        request.url() === targetUrl,
+    );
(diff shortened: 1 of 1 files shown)


dai-shi:
Could anyone review?

nihgwu:
@dai-shi added the background manually 😄

packages/waku/src/router/client.tsx
```diff
@@ -924,6 +937,10 @@ const InnerRouter = ({
             return;
           }
           routeChangeAbortRef.current = null;
+          if (!getErrorInfo(e)) {
```
dai-shi, line 940: While I'm not very sure about the problematic use case, this feels too narrow.

dai-shi:
@rmarscher @tylersayshi 
Any thoughts about the problem and possible solutions?

tylersayshi:
> @rmarscher @tylersayshi Any thoughts about the problem and possible solutions?

this seems like a reasonable fix as far as I can tell 🤷 - I'm not too confident on how to handle this really.

packages/waku/src/router/client.tsx
dai-shi: I think it means, we will never catch errors in the user error boundary, which isn't the original intention.

dai-shi:
I belive the current canonical solution is to have a dedicated error boundary on the user end. I might misunderstand the problem, and the current error boundary support might not be perfect, so please open a new discussion with a very small reproduction.

PR #2011, opened 2026-04-10
fix(router): account for empty catch all routes

### LINKED ISSUE #2008: Catch-all routes do not match zero segments (breaks optional locale prefix pattern)
## Description

Catch-all route segments (`[...param]`) do not match when the param has zero segments. This prevents common patterns like optional locale prefixes in i18n routing.

## Reproduction

File structure:
```
src/pages/
  [...locale]/
    about.tsx
```

Expected routing:
- `/about` → `locale = []` (default locale, no prefix)
- `/zh/about` → `locale = ["zh"]`

Actual behavior:
- `/about` → **404 Not Found**
- `/zh/about` → works correctly

## Additional context

- `staticPaths` also rejects empty arrays for wildcard routes: *"Empty staticPaths entry is not supported for wildcard routes"*
- Using `render: 'dynamic'` does not help — the route matcher itself does not match zero segments

## Use case

This is the standard i18n optional locale prefix pattern used by frameworks like Next.js and Astro, where the default locale has no URL prefix:
- `/about` (English, default)
- `/zh/about` (Chinese)

## Current workaround

Use `[locale]` (single segment) instead of `[...locale]`, requiring all locales to have explicit prefixes (`/en/about`, `/zh/about`).

## Environment

- waku: 1.0.0-alpha.6
- OS: macOS

tylersayshi:
fixes #2008

### DIFF
--- a/e2e/fixtures/wildcard-api-routes/src/pages/[...locale]/about.tsx
+++ b/e2e/fixtures/wildcard-api-routes/src/pages/[...locale]/about.tsx
@@ -1,0 +1,14 @@
+const AboutPage = ({ locale }: { locale: string[] }) => (
+  <div>
+    <h1>About</h1>
+    <p data-testid="locale">{locale.join('/')}</p>
+  </div>
+);
+
+export const getConfig = async () => {
+  return {
+    render: 'dynamic',
+  };
+};
+
+export default AboutPage;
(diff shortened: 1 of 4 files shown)


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/24226713741

dai-shi (approved):

PR #2015, opened 2026-04-11
fix(ssr): don't output unnecessary console.error with handled exceptions like NotFound and Redirect

### LINKED ISSUE #2013: unstable_redirect() logs unnecessary console.error
## Summary

Hi,
When calling `unstable_redirect()` in dev, the redirect works correctly, but the terminal outputs a spurious error:

```
Error during rendering: Error: Redirect
    at info (.../waku/src/lib/utils/custom-errors.ts:23:15)
    at unstable_redirect (.../waku/src/router/define-router.tsx:157:9)
    ... {
      digest: '__WAKU_CUSTOM_ERROR__;{"status":307,"location":"/target"}'
    }
```

Is this intended?

## Cause

The `onError` handler in [src/lib/utils/render.ts](https://github.com/wakujs/waku/blob/2d4974c9b16f0d390d3d4e7afbd656e804b63afb/packages/waku/src/lib/utils/render.ts#L31) calls `console.error` **before** checking the error's `digest` property:

```ts
  const onError = (e: unknown) => {
    console.error('Error during rendering:', e); // here
    if (
      e &&
      typeof e === 'object' &&
      'digest' in e &&
      typeof e.digest === 'string'
    ) {
      return e.digest;
    }
  };
```

Meanwhile, [src/lib/vite-rsc/ssr.tsx](https://github.com/wakujs/waku/blob/2d4974c9b16f0d390d3d4e7afbd656e804b63afb/packages/waku/src/lib/vite-rsc/ssr.tsx#L84) correctly checks digest first:

```tsx
      onError: (e: unknown) => {
        if (
          e &&
          typeof e === 'object' &&
          'digest' in e &&
          typeof e.digest === 'string'
        ) {
          return e.digest;
        }
        console.error('[SSR Error]', captureOwnerStack?.() || '', '\n', e); // here
      },
```

## Reproduction
Repository: https://github.com/t6adev/repro-waku-redirect-error/tree/main/baseline

```
pnpm install
pnpm dev
# Visit http://localhost:3000/ and check terminal output
```

t6adev:
I have modified it to check the digest first and then output the log, following the same approach used in [src/lib/vite-rsc/ssr.tsx](https://github.com/wakujs/waku/blob/2d4974c9b16f0d390d3d4e7afbd656e804b63afb/packages/waku/src/lib/vite-rsc/ssr.tsx#L84).

Close #2013


e2e/ssr-redirect.spec.ts
```diff
@@ -70,4 +73,20 @@ test.describe(`ssr-redirect`, () => {
...
   });
+
+  test('redirect should not log "Error during rendering" to server console', async ({
+    page,
+  }) => {
+    serverOutput.length = 0;
+
+    // TODO: async redirection on dev is flaky, so wrap with retry for now
+    // https://github.com/wakujs/waku/pull/1586
+    await expect(async () => {
```
dai-shi: I don't think we need this, unless we really see it.

e2e/ssr-redirect.spec.ts
```diff
@@ -70,4 +73,20 @@ test.describe(`ssr-redirect`, () => {
     await page.click('text=Redirect Action');
     await expect(page.getByRole('heading')).toHaveText('Destination Page');
   });
+
+  test('redirect should not log "Error during rendering" to server console', async ({
+    page,
+  }) => {
+    serverOutput.length = 0;
```
dai-shi: Is this AI? I don't think we do this pattern to clear an array. Do we need to clear them?

e2e/utils.ts
```diff
@@ -170,7 +170,10 @@ export const prepareNormalSetup = (fixtureName: string) => {
   let built = false;
   const startApp = async (
     mode: 'DEV' | 'PRD' | 'STATIC',
-    options?: { cmd?: string | undefined },
+    options?: {
+      cmd?: string | undefined;
+      onServerOutput?: (data: string) => void;
```
dai-shi, line 175: It looks to me like, if we want to assert server output, we should pipe stdout/stderr anyways, and ignore some expected errors in the spec.

dai-shi (changes requested):
Let's see if we can simply pipe stdout/err from the child process.

e2e/ssr-redirect.spec.ts
t6adev: Yeah, I don't like this pattern either, but I've seen it used in a few places in this project.
Anyway, I wasn't sure if this was necessary, so it's fine to remove it if it's not needed.

e2e/ssr-redirect.spec.ts
t6adev: If removed this, got this error:
```
    Error: expect(locator).toHaveText(expected) failed

    Locator: getByRole('heading')
    Expected: "Destination Page"
    Timeout: 5000ms
    Error: element(s) not found

    Call log:
      - Expect "toHaveText" with timeout 5000ms
      - waiting for getByRole('heading')
```

e2e/utils.ts
t6adev: How about this way?
[ab1f634](https://github.com/wakujs/waku/pull/2015/commits/ab1f6342383eebe486ed1d8d388bb39119a21971)
[58eee89](https://github.com/wakujs/waku/pull/2015/commits/58eee896ee8f6cd8d6d6a519f7c8cf38981ee91a)

e2e/ssr-redirect.spec.ts
dai-shi: I see. Then, that's my misunderstanding.

e2e/utils.ts
dai-shi: Oh..., that was my bad. I got confused with Vitest setup.
I like the original `onServerOutput` option better. Sorry about the false request.

e2e/ssr-redirect.spec.ts
dai-shi: Actually, this can be resolved. See #2016.

e2e/ssr-redirect.spec.ts
dai-shi: > I've seen it used in a few places

I remember, it's used in one place, which is performance critical.

e2e/utils.ts
t6adev: Reverted all!

e2e/ssr-redirect.spec.ts
```diff
@@ -60,4 +63,14 @@ test.describe(`ssr-redirect`, () => {
     await page.click('text=Redirect Action');
     await expect(page.getByRole('heading')).toHaveText('Destination Page');
   });
+
+  test('redirect should not log "Error during rendering" to server console', async ({
+    page,
+  }) => {
+    await page.goto(`http://localhost:${port}/async`);
```
dai-shi, line 70:
```suggestion
    await page.goto(`http://localhost:${port}/async`);
    await waitForHydration(page);
```

dai-shi (commented):
the test looks good otherwise.

dai-shi (approved):
LGTM!

PR #2020, opened 2026-04-15
fix(router): warning was wrong on pages

### LINKED ISSUE #1978: Static API route with wildcard causes file-system conflicts

tylersayshi:
fixes https://github.com/wakujs/waku/issues/1978#issuecomment-4243461983

### DIFF
--- a/packages/waku/src/router/create-pages.tsx
+++ b/packages/waku/src/router/create-pages.tsx
@@ -492,12 +492,6 @@ export const createPages = <
         if (staticPath.length !== numSlugs && numWildcards === 0) {
           throw new Error('staticPaths does not match with slug pattern');
         }
-        if (staticPath.length === 0 && numWildcards > 0) {
-          console.warn(
-            `Empty staticPaths entry is not supported for wildcard routes. ` +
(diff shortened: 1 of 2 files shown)


dai-shi:
So, it was just a false warning. Do you think we need a test to check the non-existence of false warnings???

tylersayshi:
good point. removed the test. i'll follow up on that issue though

tylersayshi:
closing this, we'll wait for https://github.com/wakujs/waku/issues/1978#issuecomment-4254006843

PR #2022, opened 2026-04-16
fix(router): incorrect warnings for wildcard API routes

fuma-nama:
see #1978

### DIFF
--- a/packages/waku/src/router/define-router.tsx
+++ b/packages/waku/src/router/define-router.tsx
@@ -179,7 +179,7 @@ const assertNonReservedSlotId = (slotId: SlotId) => {
 
 type RendererOption = { routePath: string; query: string | undefined };
 
-interface RouteConfig {
+type RouteConfig  = {
   type: 'route';
   path: PathSpec;
   isStatic: boolean;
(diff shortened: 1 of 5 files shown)


packages/waku/src/router/define-router.tsx
```diff
@@ -175,7 +179,7 @@ const assertNonReservedSlotId = (slotId: SlotId) => {
 
 type RendererOption = { routePath: string; query: string | undefined };
 
-type RouteConfig = {
+export interface RouteConfig {
```
dai-shi: Please do not export types arbitrary. We only do it when it's really necessary and reasonable.

dai-shi (changes requested):

packages/waku/src/router/define-router.tsx
fuma-nama: yep was added when writing unit tests, should not be needed now

tylersayshi (approved):
lgtm 👍

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/24649262014

dai-shi (approved):

PR #2023, opened 2026-04-17
feat(router): support building routes selectively

### LINKED ISSUE #1790: Is partial mode build safe?

dai-shi:
close #1790


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/24650223139

PR #2035, opened 2026-04-23
feat(router): layout slug props [breaking behavior]

### LINKED ISSUE #1665: layouts should use router props as pages do

dai-shi:
close #1665 

## Summary

This PR adds slug props support for layouts by introducing a breaking change in static layout behavior.

## Layout slug props

Before this PR, layouts only received `children`, even when the layout path itself contained dynamic segments.

With this PR, layouts receive slug props, but only for parent path.

Examples:

```tsx
createLayout({
  render: 'dynamic',
  path: '/[lang]',
  component: ({ lang, children }) => (
    <div>
      <h1>{lang}</h1>
      <p>{children}</p>
    </div>
  ),
});

createPage({
  render: 'dynamic',
  path: '/[lang]/blog/[slug]',
  component: ({ slug }) => <h2>{slug}</h2>,
});
```

For `/en/blog/hello`:
- the layout receives `{ lang: 'en' }`
- the layout does **not** receive `slug`
- the page receives `{ lang: 'en', slug: 'hello' }`

## Breaking behavioral change

This PR also changes how static layouts are cached.

Previously, a static layout under a dynamic segment was effectively shared. That is no longer the case. Static layout is rendered and cached for each path.

So if we have routes like,
- `/en/blog/hello`
- `/ja/blog/hello`

The layout example above will render twice for `/en` and `/ja`.

## Migration path for shared static UI: use a static slice

If your intent was to share one static wrapper across many slug variants, the recommended migration path is to move that shared part into a static slice.

### Before

```tsx
createLayout({
  render: 'static',
  path: '/[lang]',
  component: ({ children }) => (
    <div>
      <nav>Shared navigation</nav>
      <main>{children}</main>
    </div>
  ),
});
```

### After

```tsx
createPage({
  render: 'dynamic',
  path: '/[lang]/blog/[slug]',
  slices: ['shared-shell'],
  component: ({ slug }) => (
    <Slice id="shared-shell">
      <article>{slug}</article>
    </Slice>
  ),
});

createSlice({
  render: 'static',
  id: 'shared-shell',
  component: ({ children }) => (
    <div>
      <nav>Shared navigation</nav>
      <main>{children}</main>
    </div>
  ),
});
```

This keeps the shared wrapper static and reusable, while avoiding incorrect sharing of slug-dependent layout output. You can use the slice in a layout too rather than the page.

## Notes

- This change affects both dynamic and static layouts in `createPages`.
- The key behavioral change is only for **static** layouts under dynamic segments.
- This works for fs-router and managed mode too.

### DIFF
--- a/e2e/create-pages.spec.ts
+++ b/e2e/create-pages.spec.ts
@@ -502,7 +502,11 @@ test.describe(`create-pages`, () => {
   }) => {
     await page.goto(`http://localhost:${port}/layout-props/dynamic/en/post-1`);
     await expect(
-      page.getByRole('heading', { name: 'Dynamic Parent Layout' }),
+      page.getByRole('heading', {
+        name: 'Dynamic Parent Layout',
+        exact: true,
+        level: 2,
+      }),
     ).toBeVisible();
     await expect(page.getByTestId('dynamic-layout-props-keys')).toHaveText(
       'lang',
@@ -523,7 +527,11 @@ test.describe(`create-pages`, () => {
   }) => {
     await page.goto(`http://localhost:${port}/layout-props/static/en/post-1`);
     await expect(
-      page.getByRole('heading', { name: 'Static Parent Layout' }),
+      page.getByRole('heading', {
+        name: 'Static Parent Layout',
+        exact: true,
+        level: 2,
+      }),
     ).toBeVisible();
     await expect(page.getByTestId('static-layout-props-keys')).toHaveText(
       'lang',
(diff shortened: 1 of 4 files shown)


dai-shi:
@copilot resolve the merge conflicts in this pull request

dai-shi:
@copilot resolve the merge conflicts in this pull request

PR #2036, opened 2026-04-25
refactor(lib): build metadata

dai-shi:
just a little simplification.


dai-shi:
The change was actually breaking cloudflare vite plugin use case. Fix by #2085.

PR #2037, opened 2026-04-25
fix: cache static elements

dai-shi:
#1744 was incomplete. There was some missing cases and the bug.


packages/waku/src/router/define-router.tsx
```diff
@@ -744,14 +744,11 @@ export function unstable_defineRouter(fns: {
     const path2moduleIds: Record<string, string[]> = {};
     const htmlRenderTasks = new Set<() => Promise<void>>();
 
-    // static route
+    // for each route, cache static elements and generate files for full static page
```
dai-shi:
```suggestion
    // for each route, cache static elements and generate files for full static route
```

PR #2043, opened 2026-04-30
refactor(minimal): (de)serialize rsc utils [internal breaking]

dai-shi:
When I did #1984, I wasn't happy with `renderRscForParse` workaround. serialize/deserialize should be cleaner apis.


packages/waku/src/server.ts
```diff
@@ -21,3 +25,11 @@ export function getEnv(key: string): string | undefined {
 export function unstable_getHeaders(): Readonly<Record<string, string>> {
   return Object.fromEntries(getContext().req.headers.entries());
 }
+
+export async function serializeRsc(element: unknown): Promise<Uint8Array> {
+  return streamToBytes(renderToReadableStream(element, {}));
```
dai-shi, line 30: Actually, this second arg is a lie.
Maybe we should give up importing RSDW directly and follow #1995.

PR #2044, opened 2026-04-30
feat: prune build output

### LINKED ISSUE #1912: Remove fully static routes from production bundle

dai-shi:
close #1912


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/25217852138

dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/25348231625

PR #2066, opened 2026-05-06
support node 26 and drop node 20

dai-shi:


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/25437086440

PR #2067, opened 2026-05-07
Update react

dai-shi:
and vitejs/plugin-rsc.


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/25502562765

PR #2071, opened 2026-05-08
feat(cloudflare-example): configure assets caching

nihalgonsalves:
The hashed asset output should be cached instead of Cloudflare's [default](https://developers.cloudflare.com/workers/static-assets/headers/#default-headers), which is `Cache-Control: public, max-age=0, must-revalidate`.

### DIFF
--- a/examples/07_cloudflare/public/_headers
+++ b/examples/07_cloudflare/public/_headers
@@ -1,2 +1,4 @@
 /RSC/*
   X-Robots-Tag: noindex
+/assets/*
+  Cache-Control: public, max-age=31536000, immutable


examples/07_cloudflare/public/_headers
```diff
@@ -1,2 +1,4 @@
 /RSC/*
   X-Robots-Tag: noindex
+/assets/*
+  Cache-Control: public, max-age=31536000, immutable
```
dai-shi, line 4: Is `max-age` value coming from any de-facto configs?
https://developers.cloudflare.com/workers/static-assets/headers/#configure-custom-browser-cache-behavior uses `31556952`.

nihalgonsalves: just de-facto 1 year, e.g. from https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control#caching_static_assets_with_cache_busting – not sure why Cloudflare recommends the "accounting for leap years" figure. doesn't matter either way and happy to change, but the 000 value is more recognisable to me.

dai-shi: I'm fine either way.

PR #2074, opened 2026-05-09
fix: set charset=utf-8 on SSR HTML responses

withxat:
**Problem**: 
SSR HTML used Content-Type: text/html without charset.
On a full reload, some browsers defaulted to a Western encoding and corrupted UTF‑8 text (I hit this with Chinese). Client routing still looked fine.

**Fix**:
Use `text/html; charset=utf-8`, consistent with the existing fallback HTML path.

**Test**:
Added a unit check on the response header; optional manual hard refresh with non‑ASCII content.

### DIFF
--- a/packages/waku/src/lib/utils/render.ts
+++ b/packages/waku/src/lib/utils/render.ts
@@ -93,7 +93,7 @@ export function createRenderUtils(
       });
       return new Response(htmlResult.stream, {
         status: htmlResult.status || options.status || 200,
-        headers: { 'content-type': 'text/html' },
+        headers: { 'content-type': 'text/html; charset=utf-8' },
       });
     },
   };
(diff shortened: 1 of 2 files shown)


withxat:
PTAL: @dai-shi

dai-shi (approved):
Yeah, I was wondering what should be best. It's reasonable.

dai-shi:
@withxat Could you check the CI failure?

withxat:
@dai-shi Sure, I’ll take a look. :)

Not sure why there are so many failures since I only touched two files, but I’ll dig into it.

withxat:
Oh, just noticed that most of them are warnings.
Let me take a closer look at the actual error.

withxat:
Hi @dai-shi 

Fixed. 

It seems the CI needs an approval to re-run, but I’ve tested the same command locally and it passes now.

dai-shi (approved):
Thanks!

Issue #2075, opened 2026-05-09
CSS `?url` import breaks HMR


fuma-nama:
I was experiencing a problem with HMR that freezes the entire React VDOM unless manual refresh, took a long time to debug and narrowed it down to the CSS import in my root component.

```ts
import type { ReactNode } from 'react';
import { Provider } from '@/components/provider';
import appCss from '@/styles/globals.css?url';

export default async function RootElement({ children }: { children: ReactNode }) {
  return (
    <html lang="en" suppressHydrationWarning>
      <head>
        <link rel="stylesheet" href={appCss} />
      </head>
      <body data-version="1.0" className="flex flex-col min-h-screen">
        <Provider>{children}</Provider>
      </body>
    </html>
  );
}

export async function getConfig() {
  return {
    render: 'static',
  } as const;
}
```

For example, once I hit the save button, it should trigger HMR from my `import.meta.glob()`, then `index.mdx` -> `[...slugs].tsx` -> HMR. But instead, the React VDOM is somehow freezing and have no response even if I click/interact with the UI. It works again if I refresh manually.

https://github.com/user-attachments/assets/46a0b6be-31d9-417c-a25e-5503110d0766

If I use `?inline` or without any query params (rely on auto injection), the problem no longer exists. The same usage works on Tanstack Start & React Router, I currently suspect it's a problem between vite/rsc and waku. I feel it's a complicated problem to solve from my end, I'm not really familiar with vite/rsc so I am not able to narrow it down further.

dai-shi:
@hi-ogawa Do you have any instant hint?

hi-ogawa:
It would be good to know if it can be reproduced on https://github.com/vitejs/vite-plugin-react/tree/main/packages/plugin-rsc/examples/starter

`css?url` is in main e2e suite but don't remember if HMR is properly tested specifically for this component https://github.com/vitejs/vite-plugin-react/blob/65d378fc4d9bd8d383dc7598817261b3bdcb0861/packages/plugin-rsc/examples/basic/src/routes/style-server/server.tsx#L2

fuma-nama:
Hmm I found it only happens if I'm using it in my monorepo, the reproduction is at https://github.com/fuma-nama/fumadocs/tree/dev/examples/waku-reproduce, I removed all workspace deps.

Can you fork https://github.com/fuma-nama/fumadocs and try `pnpm i` & `pnpm dev --filter=example-waku-reproduce`?

```
Operating System:
  Platform: darwin
[...]

PR #2077, opened 2026-05-11
fix: hot reload styles

dai-shi:
for #2075


dai-shi:
https://github.com/wakujs/waku/issues/2075#issuecomment-4423422190
this approach doesn't seem valid. closing.

PR #2100, opened 2026-05-25
fix(router): warm prefetched route navigations

### LINKED ISSUE #2099: Router prefetch does not fully warm route navigations
## Summary

`waku@1.0.0-beta.0` can leave prefetched router navigations cold. In the repro, routes are static and links use `unstable_prefetchOnView`, but clicking a visible, already-prefetched route still triggers route-specific loading work.

There seem to be two related issues:

1. Production router prefetch code can preload the wrong client chunks.
2. RSC prefetch stores the response, but does not reuse decoded elements on navigation, so client module references can still be resolved at click time.

## Reproduction

Minimal repro: https://github.com/jxom/waku-prefetch-repro

```sh
pnpm install
pnpm build
pnpm start
```

Then:

1. Open the local URL in a browser.
2. Open DevTools Network.
3. Load `/`.
4. Wait for the visible links to prefetch.
5. Click `Route A`.

## Actual behavior

Clicking `Route A` still triggers route-specific loading work, such as a delayed `/assets/a-*.js` request, even though the route link was visible and had prefetched.

## Expected behavior

A prefetched route navigation should be warm by the time the user clicks it. Route-specific client chunks and RSC decoding work needed for the navigation should already have been started or reused from the prefetch.

## Environment

- `waku@1.0.0-beta.0`
- `react@19.2.4`
- `react-dom@19.2.4`
- Production build via `waku build` and `waku start`

### Router prefetch table uses route-local indexes

The generated router prefetch code builds a global `ids` array, but computes each route's indexes against that route's local module ID list. At runtime, those route-local indexes are used against the global `ids` array, which can point at earlier routes' chunks instead.

```diff
- Object.entries(path2moduleIds).forEach(([path, ids]) => {
-   path2idxs[path] = ids.map((id) => ids.indexOf(id))
+ Object.entries(path2moduleIds).forEach(([path, pathIds]) => {
+   path2idxs[path] = pathIds.map((id) => ids.indexOf(id))
  })
```

### RSC prefetch response is not decoded until navigation

`unstable_prefetchRsc` caches the fetched response, but the prefetch path returns before calling `createFromFetch`. On navigation, Waku consumes the prefetched response and then creates elements for the first time. That means client module references in the Flight payload may still be resolved during click-time navigation.

A complete fix likely needs to decode/create the elements during prefetch, store them on the prefetched entry, and reuse those elements when the matching navigation consumes the prefetched entry.

jxom:
Fixes https://github.com/wakujs/waku/issues/2099.

Router prefetch now maps route module ids against the shared prefetch id table, which keeps production preload callbacks aligned with the chunks for the requested route.

RSC prefetch now starts decoding the prefetched response and stores the resulting elements with the prefetched entry. Matching navigations reuse those elements instead of creating them for the first time on click.


packages/waku/src/lib/utils/prefetch.ts
```diff
@@ -8,6 +9,7 @@ const KEY_DEBUG_ID = 'd';
 export type INTERNAL_PrefetchedEntry = {
   [KEY_RSC_PATH]: string;
   [KEY_RESPONSE]: Promise<Response>;
+  [KEY_ELEMENTS]?: unknown;
```
dai-shi: Now, it feels like we should drop `KEY_RESPONSE` and make `KEY_ELEEMENTS` required. This will require separating initial server prefetch (bad naming) structure from client prefetch structure. I'll try that in a follow-up PR.

packages/waku/src/minimal/client.tsx
```diff
@@ -207,23 +208,31 @@ const fetchRscInternal: FetchRscInternal = (
...
+  const createElements = () =>
+    createFromFetch<Elements>(checkStatus(responsePromise), {
+      callServer: (funcId: string, args: unknown[]) =>
+        unstable_callServerRsc(funcId, args, () => fetchRscStore),
+      debugChannel: debug?.debugChannel,
+      temporaryReferences,
+    });
   if (prefetchOnly) {
+    const elements = createElements();
+    Promise.resolve(elements).catch(() => {});
```
dai-shi: Do we need this?

packages/waku/src/router/define-router.tsx
```diff
@@ -401,8 +401,8 @@ const getRouterPrefetchCode = (path2moduleIds: Record<string, string[]>) => {
   );
   const ids = Array.from(moduleIdSet);
   const path2idxs: Record<string, number[]> = {};
-  Object.entries(path2moduleIds).forEach(([path, ids]) => {
-    path2idxs[path] = ids.map((id) => ids.indexOf(id));
+  Object.entries(path2moduleIds).forEach(([path, pathIds]) => {
+    path2idxs[path] = pathIds.map((id) => ids.indexOf(id));
```
dai-shi, line 409: Oh my... This is such a fundamental bug... Thanks for the catch.

dai-shi (commented):
Thanks for working on it!! Do you think you can add some tests to cover the fix? I would like to refactor based on it.

packages/waku/src/minimal/client.tsx
dai-shi: > // Observe failed prefetches until a matching navigation consumes them.

Hmmm, I'm not sure how it helps.

dai-shi: Is it to hide unhandled promise error? Makes sense, but the comment can be different?

dai-shi: I think the error is handled if the prefetched entry is actually "used". So, this workaround is only for the case when the prefetched entry is never used. Is it correct?

jxom: yeah, I think if the prefetched entry is consumed by navigation, `unstable_fetchRsc` reuses the decoded elements promise and the rejection is handled through that path. This catch is only for abandoned prefetches that are never consumed, where nothing else observes the promise. can update the comment.

dai-shi:
@jxom https://github.com/wakujs/waku/issues/2099#issuecomment-4531781548

jxom:
Closing in favor of the focused replacement PR: https://github.com/wakujs/waku/pull/2102. It keeps only the router prefetch module id mapping fix.

dai-shi:
#2102 is merged, so reopened.

dai-shi (approved):
Now, I'm comfortable with this.
(I think `Unstable_FetchRscStore` should be revisited entirely in the near future.)

PR #2101, opened 2026-05-25
fix(plugins): allow overriding vite plugins

### LINKED ISSUE #2092: Feature Request: disable router typegen from config
Probably not a common use case, but in short, I am looking for an option to disable the router typegen (`pages.gen.ts`). Like:

```ts
export default defineConfig({
  routerTypegen: false
})
```

For context, I am overriding the default virtual module for server entry (via my own Vite plugin), but Waku generates `pages.gen.ts` when there's files under `src/pages`, even if fs router is not necessarily enabled in the overriden server entry. This currently breaks types for `<Link>` unless overriding it again from another `.d.ts` file.

dai-shi:
in addition to add custom vite plugins, this allows to override waku vite plugins.


fuma-nama:
Hmm interesting idea, looks very good to me.

PR #2102, opened 2026-05-25
fix(router): map prefetch module ids

### LINKED ISSUE #2099: Router prefetch does not fully warm route navigations

jxom:
Fixes https://github.com/wakujs/waku/issues/2099.

Router prefetch now maps route module ids through the shared prefetch id table when generating `__WAKU_ROUTER_PREFETCH__`. This keeps production preload callbacks aligned with the chunks for the requested route.

### DIFF
--- a/packages/waku/src/router/define-router.tsx
+++ b/packages/waku/src/router/define-router.tsx
@@ -401,8 +401,8 @@ const getRouterPrefetchCode = (path2moduleIds: Record<string, string[]>) => {
   );
   const ids = Array.from(moduleIdSet);
   const path2idxs: Record<string, number[]> = {};
-  Object.entries(path2moduleIds).forEach(([path, ids]) => {
-    path2idxs[path] = ids.map((id) => ids.indexOf(id));
+  Object.entries(path2moduleIds).forEach(([path, pathIds]) => {
+    path2idxs[path] = pathIds.map((id) => ids.indexOf(id));
   });
   return `
 globalThis.__WAKU_ROUTER_PREFETCH__ = (path, callback) => {
--- a/packages/waku/tests/define-router-static-build.test.ts
+++ b/packages/waku/tests/define-router-static-build.test.ts
@@ -1,4 +1,6 @@
+import { runInNewContext } from 'node:vm';
 import { describe, expect, it, vi } from 'vitest';
+import { ROUTE_ID } from '../src/router/common.js';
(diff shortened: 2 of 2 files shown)


dai-shi (approved):
Thanks for your contribution.

PR #1997, opened 2026-03-22
fix(canary-ci): a workaround for closing rsc stream

dai-shi:
I'm not sure if this is really a good solution...

- [x] timeout hack to see canary-ci passing
- [x] proper workaround

PR #2152, opened 2026-06-22
Fix hanging when a server component returns an object with the key status and the value pending

SaraVieira:
I had this very specific error that when using cloudflare the json response contained:

```
{
	"status": "pending"
}
```

And this would make my whole page non interactive and hang. This fix checks if when sometimes comes with this key and value it has a `.then` function attached to it.


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/27986105010

dai-shi (commented):
Thanks for the catch! It was really a hack.

Let me check if #1997 change is still necessary for react@canary later.

dai-shi (approved):

PR #2169, opened 2026-06-28
fix(router): loosen prefetch typing

dai-shi:
`router.prefetch()` only accepted known route paths, so a computed URL like `/posts/${slug}` was a type error (even though it worked at runtime). It now also accepts any string, while keeping autocomplete for known routes. Still href-only (no structured `{ to, params }` form).


dai-shi:
Ooops, this was a wrong move. Let me revert it.

PR #2185, opened 2026-07-09
fix: return 400 for form action requests without an action reference

SaraVieira:
A multipart/form-data POST to any non-RSC path is treated as a progressive-enhancement server action, but decodeAction returns null when the form data contains no $ACTION_ reference. The result was invoked unconditionally, so such requests (typically crawlers or vulnerability scanners probing arbitrary paths) crashed the render with "TypeError: (intermediate value) is not a function" and produced a 500 with a stack trace in server logs.

Guard the decoded action and throw a custom error instead, which the handler already maps to a clean 400 response. Also widen the decodeAction parameter type to include null, matching React's actual signature.

### DIFF
--- a/packages/waku/src/lib/utils/request.ts
+++ b/packages/waku/src/lib/utils/request.ts
@@ -79,8 +79,7 @@ export async function getInput(
           const formData = (await getActionBody(req)) as FormData;
           const decodedAction = await decodeAction(formData);
           if (typeof decodedAction !== 'function') {
-            // multipart/form-data POST without an action reference
-            // (e.g. crawlers or vulnerability scanners probing arbitrary paths)
+            // multipart/form-data POST without an action reference (e.g. crawlers)
             throw createCustomError('Bad Request', { status: 400 });
           }
           const result = await decodedAction();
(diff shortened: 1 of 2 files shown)


packages/waku/tests/request.test.ts
```diff
@@ -127,6 +127,31 @@ describe('getInput server action request validation', () => {
...
+    const input = await getInput(
+      new Request('https://app.test/', {
+        method: 'POST',
+        body: formData,
+        headers: { origin: 'https://app.test' },
+      }),
+      makeConfig(),
+      undefined,
+      vi.fn(),
+      vi.fn().mockResolvedValue(null),
```
dai-shi:
```suggestion
      vi.fn().mockReturnValue(null),
```

to match with the type.

dai-shi (commented):
Thanks for your fix!

packages/waku/src/lib/utils/request.ts
```diff
@@ -78,6 +78,11 @@ export async function getInput(
         fn: async () => {
           const formData = (await getActionBody(req)) as FormData;
           const decodedAction = await decodeAction(formData);
+          if (typeof decodedAction !== 'function') {
+            // multipart/form-data POST without an action reference
+            // (e.g. crawlers or vulnerability scanners probing arbitrary paths)
```
dai-shi:
```suggestion
            // multipart/form-data POST without an action reference (e.g. crawlers)
```

dai-shi (approved):
I'm planning to change something related with this, but let's merge this first anyway.

Issue #2000, opened 2026-03-23
Is there any way to place chunk files under a different domain?


Allenyou1126:
As is said in the title, I'm going to staticly build a site, assuming it will be deployed under `https://a.example.com/`. When I was using Next.js, it provided a configuration item `assetPrefix`，which allows me to place chunks(like JavaScript chunks and CSS chunks) under a different path. For example, setting it to `https://b.example.com/`, then every references to these chunks will be replaced to `https://b.example.com/_next/xxxxx`.

Now since I'm migrating my site to Waku, I wander is there any similar way to place chunks under a different domain, just working as this `assetPrefix` configuration item in Next.js?

dai-shi:
Thanks for opening an issue.

I wonder if this can be done only with Vite configuration, or Waku needs to do something.  @hi-ogawa

hi-ogawa:
There's a `base` option https://vite.dev/config/shared-options#base

dai-shi:
Yeah, I think then the problem is we override it...

https://github.com/wakujs/waku/blob/076c49994a7528a2903d4094e68105b29d7ac272/packages/waku/src/lib/vite-plugins/main.ts#L22

hi-ogawa:
Waku's `basePath` concept is technically different and having two same is merely for the sake of default convenience.

I think testing out something like this and see if it works is the starting point.

```js
export default defineConfig({
  vite: {
    plugins: [
      {
         name: 'override-vite',
         apply: "build",
         config: { order: 'post', handler: => ({ base: ""https://b.example.com/ }) }
      },
    ],
  },
})
```

dai-shi:
Yeah, I know Vite's `base` and Waku's `basePath` are technically different things.

@Allenyou1126 Could you try the workaround above and see how it goes?

Allenyou1126:
Well I think that may work, I'll try it later.

Thanks for your kindly help!

Allenyou1126:
That actually solves my problem, but I still hope it could be added as a Waku configuration item.

Thanks for your kindly help!

dai-shi:
I think Waku should not override it if it's already defined by a user. Should it be possible? @hi-ogawa

hi-ogawa:
Waku can do `config(userConfig) { return { base: userConfig.base ?? wakuConfig.basePath } }`

dai-shi:
I was trying to apply that, and noticed we have mergeConfig:

[...]

PR #2188, opened 2026-07-11
test(vite): cover user base override

Osraka:
## Summary

Adds coverage for the Vite config path that sets `basePath` as Waku's default Vite `base`, while still allowing a user-provided `vite.base` to override it.

## Why

#2000 asks whether client chunks can be served from a different domain. The current `mergeConfig` path already preserves `vite.base`; this test locks that behavior so it does not regress while the Vite environment plugin keeps changing.

## Validation

- `../../node_modules/.pnpm/node_modules/.bin/vitest run tests/vite-plugin-environments.test.ts`
- `./node_modules/.bin/prettier -c packages/waku/tests/vite-plugin-environments.test.ts`
- `git diff --check`

### DIFF
--- a/packages/waku/tests/vite-plugin-environments.test.ts
+++ b/packages/waku/tests/vite-plugin-environments.test.ts
@@ -1,0 +1,40 @@
+import type { UserConfig } from 'vite';
+import { expect, test } from 'vitest';
+import type { Config } from '../src/config.js';
+import { resolveConfig } from '../src/lib/utils/config.js';
+import { environmentsPlugin } from '../src/lib/vite-plugins/environments.js';
+
+const runConfigHook = async (config: Config): Promise<UserConfig> => {
+  const plugin = environmentsPlugin(resolveConfig(config));
+  const hook = plugin.config;
+  if (!(typeof hook === 'function')) {
+    throw new Error('Plugin config is not defined');
+  }
+  return (await hook.call(
+    undefined,
+    {},
+    {
+      command: 'build',
+      mode: 'production',
+    },
+  )) as UserConfig;
+};
+
+test('uses basePath as the default Vite base', async () => {
+  const config = await runConfigHook({
+    basePath: '/custom/base/',
+  });
+
+  expect(config.base).toBe('/custom/base/');
+});
+
+test('lets user-provided Vite base override basePath', async () => {
+  const config = await runConfigHook({
+    basePath: '/custom/base/',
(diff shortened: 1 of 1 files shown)


dai-shi (commented):
reviewed by chatgpt

----

## Review result: **Request changes**

### [P1] Pass a valid context to the Vite `config` hook

`runConfigHook` invokes:

```ts
hook.call(undefined, {}, {
  command: 'build',
  mode: 'production',
})
```

Vite 8 declares `Plugin.config` with `this: ConfigPluginContext`, so `undefined` is not assignable under the repository’s strict TypeScript settings. The root typecheck includes `packages/waku/tests`, and the pull-request lint workflow runs `pnpm test:lint`, which executes `tsc -b`; therefore this should fail lint CI even though Vitest itself can transpile and run the test. ([[GitHub](https://github.com/wakujs/waku/pull/2188/files)][1])

A minimal fix, consistent with other plugin-hook tests in the repository:

```ts
return (await hook.call(
  {} as never,
  {},
  {
    command: 'build',
    mode: 'production',
  },
)) as UserConfig;
```

Alternatively, provide a properly typed minimal `ConfigPluginContext` mock. Existing Waku tests already use `{} as never` when manually invoking Vite hooks. ([[GitHub](https://raw.githubusercontent.com/wakujs/waku/main/packages/waku/tests/vite-plugin-allow-server.test.ts)][2])

Apart from that, the two assertions correctly capture the current merge behavior: `basePath` supplies the default Vite `base`, while `config.vite.base` wins through `mergeConfig`. I found no other blocking correctness issues in the change. ([[GitHub](https://raw.githubusercontent.com/wakujs/waku/main/packages/waku/src/lib/vite-plugins/environments.ts)][3])

[1]: https://github.com/wakujs/waku/pull/2188/files "test(vite): cover user base override by Osraka · Pull Request #2188 · wakujs/waku · GitHub"
[2]: https://raw.githubusercontent.com/wakujs/waku/main/packages/waku/tests/vite-plugin-allow-server.test.ts "raw.githubusercontent.com"
[3]: https://raw.githubusercontent.com/wakujs/waku/main/packages/waku/src/lib/vite-plugins/environments.ts "raw.githubusercontent.com"

dai-shi (approved):
To be honest, I don't mean to preserve the tested behavior if we have another solution, but it would be nice to notice the change accidentally. Let's merge.

PR #2189, opened 2026-07-11
fix(minimal): etags with prefetch

dai-shi:
A repeat prefetch of a route previously re-downloaded everything the session store already held, because prefetch requests only sent the live page's etags.

`unstable_prefetchRsc` now takes an `unstable_base` option: elements the client already holds for the path. Their etags are sent with the request, so the server only renders and sends what changed. The router passes the session store's entry and merges the response over it, so every claimed copy is kept, and a repeat prefetch is nearly free on the wire.

This also closes a gap in instant navigation: a navigation that adopts a prefetched response serves stored copies for omitted keys, and those omissions are now proven against the stored copies' etags.

Follow-up to #2184.


dai-shi:
👉 https://github.com/wakujs/waku/actions/runs/29220350510

PR #2191, opened 2026-07-14
feat(core): action marker

dai-shi:
Builds on #2185. `type: 'action'` is currently guessed from the content-type
because the real signal (`$ACTION_*`) lives in the single-read body.

This PR puts the signal in the URL: a custom `encodeFormAction` (SSR and
browser) reproduces React's default encoding via `encodeReply` and sets the
form target to the current URL plus `?__waku_action=1`. `getInput` classifies
`action` only for multipart POSTs carrying the marker, so crawlers and plain
HTML forms flow through as `custom` with an untouched body. Marked POSTs
without an action reference hit #2185's 400 as before. Static builds emit a
relative marker so no-JS actions work on prerendered pages.

Known limitation (documented in `minimal-api.mdx`): if the same action is
used both unbound and client-side-bound on one page, the bound form's no-JS
fallback may drop its arguments (JS submissions unaffected; a warning is
logged during SSR). This is because custom `encodeFormAction` implementations
receive no stable identity for unbound references.

Currently, this draft depends on https://github.com/vitejs/vite-plugin-react/pull/1280 (pinned to
its pkg.pr.new build for now).


dai-shi:
I originally thought it's a neat idea to put the action marker in the URL, but it's visible to users after POST, which is probably unacceptable...

dai-shi:
Closing in favor of #2200.

PR #2201, opened 2026-07-18
feat(core): handler actions

dai-shi:
alternative to #2200


dai-shi:
Closing in favor of #2200.

PR #2221, opened 2026-07-23
feat(router): apply fetched elements atomically with the route commit

dai-shi:
Adds an unstable_onApply option to refetch in the minimal client. The router uses it to merge the fetched elements again in the same batch as the route commit, so the committed route always has its elements even when the first merge render is delayed. The eager merge is kept as is.


dai-shi:
superseded by https://github.com/wakujs/waku/pull/2227

PR #2232, opened 2026-07-26
chore: update comment in patch-rsdw again

dai-shi:
half reverting #2231


hi-ogawa:
Good catch. Sorry for misleading with my comment 😅 
I've also verified rsc payload shape doesn't matter and adjusted the example https://github.com/vitejs/vite-plugin-react/pull/1321

PR #2072, opened 2026-05-09
feat: version skew

dai-shi:
close #1037

PR #2239, opened 2026-07-30
fix: recover from the entry chunk itself failing to load

### LINKED ISSUE #2238: Version-skew recovery (vite:preloadError) can't fire when the entry chunk itself fails to load
## Summary

The version-skew recovery added in #2072 (from discussion #1037) registers its `vite:preloadError` listener inside `entry.browser.tsx` itself:

```ts
// src/lib/vite-entries/entry.browser.tsx (compiled to dist/lib/vite-entries/entry.browser.js)
if (import.meta.env.WAKU_BUILD_ID) {
  window.addEventListener('vite:preloadError', () => {
    window.location.reload();
  });
}
```

But the bootstrap script that loads this file is a bare, unwrapped `import()` written directly by `@vitejs/plugin-rsc`, not by anything Waku's own recovery code touches:

```js
// @vitejs/plugin-rsc/dist/plugin-*.js
bootstrapScriptContent = `import(${JSON.stringify(entryUrl)})`;
```

(`entryUrl` resolves to `VIRTUAL_ENTRIES.browser`, i.e. this same `entry.browser.js`.)

`import()` never throws synchronously — a fetch failure surfaces only as a rejected promise. Vite's `__vitePreload` helper (the thing that dispatches the recoverable, cancelable `vite:preloadError` event) only wraps *dynamic imports it rewrites within the already-loaded module graph* — lazy route/component splits reached after hydration. It does not wrap this raw bootstrap `import()`, since nothing has loaded yet for it to rewrite.

**The result is a self-referential hole**: the file containing the recovery listener is the exact file whose failure to load needs recovering. If `entry.browser.js` 404s — which is exactly what happens during the deploy window version-skew recovery exists to handle, since Cloudflare (and presumably other platforms) doesn't atomically flip a new Worker/asset-manifest across every edge colo — the listener that would reload the page never gets registered in the first place, because the code registering it never ran. The promise rejects with nothing catching it, `vite:preloadError` never fires, and the page is left blank with only an unhandled `TypeError: Failed to fetch dynamically imported module` in the console.

## Reproduction context

We hit this in production on Cloudflare Workers (Workers Static Assets): a `wrangler deploy` briefly left one edge colo serving fresh HTML (correct new content hash) while its asset manifest hadn't yet started serving that hash's JS file. When the 404 landed on the entry chunk specifically (rather than a chunk loaded later, which version-skew recovery does correctly catch), the page went permanently blank until a user manually reloaded.

We worked around this in our own app by adding an `unhandledrejection` listener alongside `vite:preloadError`, registered from an inline `<script>` in `<head>` — i.e. *before* the bootstrap `import()` runs at all, rather than from code reachable only through that same import. Matching cross-browser wording was also necessary, since each engine phrases a failed dynamic import differently (Chromium: "Failed to fetch dynamically imported module"; Firefox: "error loading dynamically imported module"; Safari: "Importing a module script failed.").

## Prior art

Astro hit the identical shape of bug — an uncaught `import()` bootstrapping island hydration, no recovery — and fixed it upstream as a framework bug: #16341 → #16412 (retry + a dispatched `astro:hydration-error` event).

## Suggested fix

Waku already owns the hook point: `getBootstrapPreamble` (`src/lib/vite-rsc/ssr.js`) is prepended ahead of the bootstrap script content on the server side. A recovery listener registered from *there* — before the bootstrap `import()` runs, the same way our app-level workaround does it from `<head>` — would cover the entry chunk the same way `entry.browser.tsx`'s listener already covers everything loaded after it. It would need to listen for `unhandledrejection` (not `vite:preloadError`, which nothing dispatches for a raw, unrewritten `import()`) and match cross-browser wording as above, since the entry import is what's actually failing.

## Environment

- `waku@1.0.0-beta.7`
- `@vitejs/plugin-rsc@0.5.27`
- Deployment target: Cloudflare Workers (Workers Static Assets)

christopherwxyz:
Fixes #2238.

## Problem

Version-skew recovery (#2072, from discussion #1037) registers its `vite:preloadError` listener inside `entry.browser.tsx`:

```ts
if (import.meta.env.WAKU_BUILD_ID) {
  window.addEventListener('vite:preloadError', () => {
    window.location.reload();
  });
}
```

But the bootstrap script that loads this very file is a bare `import(entryUrl)` written by `@vitejs/plugin-rsc`, never wrapped by Vite's `__vitePreload` helper (the thing that dispatches `vite:preloadError`) — nothing has loaded yet for that helper to rewrite. `import()` never throws synchronously, so a failed fetch of the entry chunk surfaces only as an `unhandledrejection`, which nothing was listening for.

The result is a self-referential hole: the file containing the recovery listener is the one file whose own failure to load needs recovering. During the same version-skew window this feature exists to handle (a deploy host that hasn't atomically flipped the served document and its asset manifest together), if the 404 lands on the entry chunk rather than a chunk loaded later, the page is left permanently blank instead of self-healing.

## Fix

Add an `unhandledrejection` listener to `getBootstrapPreamble` (`packages/waku/src/lib/utils/ssr.ts`), which runs immediately ahead of the bootstrap `import()` on both the hydrate and SSR-fallback paths in `renderHtmlStream` — before that import can fail, rather than inside the file it's trying to load. Each JS engine phrases a failed dynamic import differently, so the listener matches substrings that hold across Chromium, Firefox, and Safari rather than trusting one exact string:

- Chromium: "Failed to fetch dynamically imported module"
- Firefox: "error loading dynamically imported module"
- Safari: "Importing a module script failed."

Bounded to one reload per document via `sessionStorage` — the same reasoning the existing `preloadError` recovery leans on implicitly, made explicit here since this runs ahead of hydration and so has nothing to hand a budget back to on success.

Unconditional rather than gated on `WAKU_BUILD_ID` the way the existing listener is: that gate is truthy in dev too (`buildIdPlugin` defines it `'dev'`, not empty, outside a real build), so it already guards nothing in practice — and this listener is inert unless the exact failure it matches actually occurs, so there's nothing to gain by conditioning it.

## Verification

- Added 7 new unit tests to `tests/ssr-utils.test.ts` (all three engines' wording, an unrelated-rejection is ignored, a malformed reason is ignored, the one-reload budget, storage-blocked graceful degradation) plus one test confirming the listener is present on both the hydrate and SSR-fallback paths — 9/9 pass; confirmed 5 fail against the pre-fix code.
- `pnpm run compile`, `tsc -b`, `eslint`, and `prettier -c` all clean.
- Full unit suite: 648/648 passing.
- Built the `01_basic` template end-to-end, ran it with `waku start`, and confirmed the listener is present verbatim in the actual served HTML, immediately ahead of the bootstrap `import()`.

Prior art: Astro hit the identical shape of bug (an uncaught `import()` bootstrapping island hydration, no recovery) and fixed it upstream as a framework bug — withastro/astro#16341 → withastro/astro#16412.

### DIFF
--- a/packages/waku/src/lib/utils/ssr.ts
+++ b/packages/waku/src/lib/utils/ssr.ts
@@ -1,5 +1,49 @@
 import { createInitialRscEntryCode } from './initial-rsc.js';
 
+/**
+ * Recovers from the entry chunk itself failing to load — the one dynamic
+ * import in this document (`@vitejs/plugin-rsc`'s bootstrap
+ * `import(entryUrl)`, appended right after this preamble) that Vite's own
+ * `__vitePreload` helper never wraps, since nothing has loaded yet for it to
+ * rewrite. During the version-skew window this build id exists to detect,
+ * that import can 404 (a deploy host hasn't atomically flipped the served
+ * document and its asset manifest together), and a failed `import()` never
+ * throws synchronously — it surfaces only as `unhandledrejection`, never
+ * `vite:preloadError`, so `entry.browser.tsx`'s own listener (which recovers
+ * a *lazy* chunk failing once the entry itself is already running) cannot
+ * fire for this one. Each engine phrases a failed dynamic import
+ * differently, so this matches substrings that hold across Chromium,
+ * Firefox, and Safari rather than one exact string.
+ *
+ * Bounded to one reload per document: this runs ahead of hydration, so
+ * there's nothing yet to hand a budget back to on success the way a mounted
+ * app could. A version-skew window is measured in seconds, so one attempt is
+ * what it needs.
+ *
+ * Unconditional rather than gated on `WAKU_BUILD_ID` the way
+ * `entry.browser.tsx`'s own listener is: that gate is truthy in dev too (the
+ * plugin defines it `'dev'`, not empty, outside a real build), so it already
+ * guards nothing in practice, and this listener is inert unless the exact
+ * failure it matches actually occurs.
+ */
+const ENTRY_IMPORT_RECOVERY = `
+addEventListener("unhandledrejection",function(e){
+  var m=e&&e.reason&&e.reason.message;
+  if(typeof m!=="string")return;
+  m=m.toLowerCase();
+  if(m.indexOf("failed to fetch dynamically imported module")<0&&
+     m.indexOf("error loading dynamically imported module")<0&&
+     m.indexOf("importing a module script failed")<0)return;
+  try{
+    var k="waku:entry-import-reload";
+    if(sessionStorage.getItem(k))return;
+    sessionStorage.setItem(k,"1");
+  }catch(_){return}
+  e.preventDefault();
+  location.reload();
+});
+`.trim();
+
 export function getBootstrapPreamble(options: {
   hydrate: boolean;
   debugId?: string | undefined;
@@ -9,5 +53,6 @@ export function getBootstrapPreamble(options: {
     globalThis.__WAKU_INITIAL_RSC__ = ${createInitialRscEntryCode(
       options.debugId,
     )};
+    ${ENTRY_IMPORT_RECOVERY}
   `;
 }
--- a/packages/waku/tests/ssr-utils.test.ts
+++ b/packages/waku/tests/ssr-utils.test.ts
@@ -1,4 +1,5 @@
-import { describe, expect, it } from 'vitest';
+// @vitest-environment happy-dom
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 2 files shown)


christopherwxyz:
Closing in favor of #2240, which the maintainer opened directly for this issue — it unifies recovery onto a single `vite:preloadError` event (wrapping the bootstrap `import()` in `.catch()` rather than needing a separate cross-browser `unhandledrejection` string-match), is properly production-gated, uses a per-build-id retry budget, defers the reload until `DOMContentLoaded` for streaming SSR, covers `renderHtmlFallback` too, and ships a real end-to-end Playwright test simulating the deploy race. A strictly better fix than what I proposed here — thanks for picking this up quickly.

PR #2240, opened 2026-07-30
fix(version-skew): recover when the entry chunk fails to load

### LINKED ISSUE #2238: Version-skew recovery (vite:preloadError) can't fire when the entry chunk itself fails to load

dai-shi:
Fixes #2238.

The vite:preloadError listener lived inside entry.browser.tsx, so it could not cover a load failure of the entry chunk itself, and the plugin-rsc bootstrap import() dispatches no event on rejection.

- Register the listener from the bootstrap preamble, before the entry import runs
- Update @vitejs/plugin-rsc to 0.5.33 and build the bootstrap import from getClientEntryUrl() (vitejs/vite-plugin-react#1360), dispatching vite:preloadError from its catch like Vite's __vitePreload does
- Allow one recovery reload per build id, tracked in sessionStorage. Without  the cap a broken build reloads forever. If storage is blocked, recovery is  skipped instead of looping
- Wait for DOMContentLoaded before reloading, since the entry chunk fails  while the HTML is still streaming

Behavior notes: recovery is production only (dev no longer reloads on  vite:preloadError), and the one retry per build id can be spent by a transient chunk failure.


packages/waku/src/lib/utils/ssr.ts
```diff
@@ -1,13 +1,58 @@
...
+  return `
+    window.addEventListener('vite:preloadError', () => {
+      window.location.reload();
+    });
+  `;
+}
+
+// The plugin-rsc bootstrap is a bare `import("...")` whose failure dispatches no `vite:preloadError`.
+// TODO: remove once @vitejs/plugin-rsc handles this natively. https://github.com/wakujs/waku/issues/2238
+const BOOTSTRAP_IMPORT_RE = /^(import\("(?:[^"\\]|\\.)*"\));?$/;
```
hi-ogawa: What do you think would be the solution? I think we can rework current `loadBootstrapScriptContent` API to return the bare entry script source like `"/my-entry.js"` instead of the content `import("/some-entry.js")`.

dai-shi: Or, plugin-rsc should attach `.catch(...)` as `vite:preloadError` is a Vite thing.

dai-shi:
@codex review

PR #2243, opened 2026-08-01
ci: name React override install steps for readability

MGPOCKY:
## Problem

In `.github/workflows/e2e.yml` and `.github/workflows/test.yml`, the optional React-version override `run` steps are unnamed and longer than **300 characters**. In the GitHub Actions UI those steps collapse into generic labels, which hurts readability when `pnpm-workspace.yaml` override / install fails.

## Changes

Semantics are unchanged: same `if:` conditions, same env vars, and the same `yq` / `pnpm install` commands.

- Add `name: Override React versions and install` in `e2e.yml` (`build` and `e2e` jobs)
- Add `name: Override React versions and install` in `test.yml`

## Test plan
- [ ] Confirm the steps still run only when `inputs.react_version` is set
- [ ] Confirm React / react-dom / react-server-dom-webpack overrides are unchanged
- [ ] Confirm the steps show the new names in the Actions UI

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -125,7 +125,7 @@ jobs:
           cache: ${{ !inputs.react_version && 'pnpm' || '' }}
       - if: runner.os == 'Windows'
         run: choco install yq -y
-      - name: Override React versions and install
+      - name: Override React versions
         if: ${{ inputs.react_version }}
         env:
           REACT_VERSION: ${{ inputs.react_version }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -53,7 +53,8 @@ jobs:
         with:
           node-version: 'lts/*'
           cache: ${{ !inputs.react_version && 'pnpm' || '' }}
-      - if: ${{ inputs.react_version }}
+      - name: Override React versions and install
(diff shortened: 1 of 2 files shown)


dai-shi (approved):
Have you ever experienced any issue? I don't believe we have a real issue, but it's harmless. Next time, please open a new discussion first.

.github/workflows/e2e.yml
```diff
@@ -124,7 +125,8 @@ jobs:
           cache: ${{ !inputs.react_version && 'pnpm' || '' }}
       - if: runner.os == 'Windows'
         run: choco install yq -y
-      - if: ${{ inputs.react_version }}
+      - name: Override React versions and install
```
dai-shi:
```suggestion
      - name: Override React versions
```

.github/workflows/test.yml
```diff
@@ -40,7 +40,8 @@ jobs:
         with:
           node-version: ${{ matrix.version }}
           cache: ${{ !inputs.react_version && 'pnpm' || '' }}
-      - if: ${{ inputs.react_version }}
+      - name: Override React versions and install
```
dai-shi:
```suggestion
      - name: Override React versions
```

PR #2245, opened 2026-08-03
refactor(router): assign link refs without eslint-disable comments

dai-shi:
Move RefObject writes into assignRef so useSharedRef keeps a ref-based
link node without react-hooks/immutability exceptions or an extra render.

### DIFF
--- a/packages/waku/src/router/client.tsx
+++ b/packages/waku/src/router/client.tsx
@@ -535,7 +535,7 @@ export function useSetSearch_UNSTABLE<Path extends RoutePath>({
   );
 }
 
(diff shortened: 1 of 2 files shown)


packages/waku/src/router/client.tsx
```diff
@@ -535,34 +535,37 @@ export function useSetSearch_UNSTABLE<Path extends RoutePath>({
   );
 }
 
+// Keeps ref.current mutation out of render scope for react-hooks/immutability.
```
dai-shi:
```suggestion
// HACK: Keeps ref.current mutation out of render scope for react-hooks/immutability.
```

PR #2250, opened 2026-08-05
fix(router): aim at a hash target on a first load too

dai-shi:
The browser stops looking for a hash target once the document has finished loading, so a target that streams in after that was never reached. The watch that client navigations already use now runs on mount as well.

The four hash scrolling functions move to client-utils/scroll.ts with the two scroll policy predicates, which is what gives them direct tests: decoding, anchor names, #top, and giving up when the reader scrolls. A keydown now only gives up when the key actually scrolls, so tabbing or typing no longer drops a pending target.


dai-shi:
As it turned out, I don't love the complexity this introduces as well as #2248. We are minimalist. Let's revisit it.

PR #2270, opened 2026-08-19
bugfix: Improve request validation for server action

lqhuang:
## Reasons and how to reproduce

While using SSR features, there is a `validateServerActionRequest` <https://github.com/wakujs/waku/blob/76cd032df3ab90ba6338b04fe622fc7fb06d3170/packages/waku/src/lib/utils/request.ts#L124-L141> to verify server action. When Waku itself acts as a server and serves endpoints directly, the `origin` header received in the request is taken directly from Hono. and whether it is http or https comes directly from the requesting client and hono to waku without **modification**. the check `originUrl.origin !== requestOrigin` works well in this case.

However, when Waku sits **behind a reverse proxy** (e.g. Caddy / Nginx), a very common pattern to host server, the `origin` read from the headers and the origin scheme read from new `URL(req.url).origin` DO NOT match. The header forwarded by the reverse proxy will be `https`, while the URL remains `http`. **This causes validateServerActionRequest to fail validation**, show 403/500 error on front pages.

That's

1. When accessing the server hosted by Waku directly:
   - plain server case (http)
     - `originUrl = new URL(req.headers.get('origin'))` -> `http://example.org`
     - `const requestUrl = new URL(req.url).origin` -> `http://example.org`
     - `originUrl.origin !== requestUrl.origin` -> true ✅
   - Secured by SSL
     - `originUrl = new URL(req.headers.get('origin'))` -> `https://example.org`
     - `const requestUrl = new URL(req.url).origin` -> `https://example.org`
     - `originUrl.origin === requestUrl.origin` -> true ✅
2. When putting caddy before waku
   - SSL provided by Caddy, waku still keep plain http/1.1
     - `originUrl = new URL(req.headers.get('origin'))` -> `https://example.org`
     - `const requestUrl = new URL(req.url).origin` -> `http://example.org`
     - `originUrl.origin === requestUrl.origin` -> false ❌
     - server action calls are rejected

## What I changed in this PR

Modify the validation logic so that 

1. when the scheme read from the URL is `http`, the origin is allowed to be either `http` or `https`
2. when the URL scheme is `https`, the origin must not be downgraded to `http`, preserves exact-origin validation. 

Add the necessary regression tests to verify the correctness of the logic.

## That's all

Looking for your opinions

Best,
Lanqing

### DIFF
--- a/packages/waku/src/lib/utils/request.ts
+++ b/packages/waku/src/lib/utils/request.ts
@@ -126,14 +126,24 @@ function validateServerActionRequest(req: Request) {
     if (origin === 'null') {
       throw createCustomError('Forbidden', { status: 403 });
     }
-    const requestOrigin = new URL(req.url).origin;
+
+    const requestUrl = new URL(req.url);
     let originUrl: URL;
     try {
       originUrl = new URL(origin);
     } catch {
       throw createCustomError('Forbidden', { status: 403 });
     }
-    if (originUrl.origin !== requestOrigin) {
+
+    if (
+      requestUrl.protocol === 'https:' &&
+      originUrl.origin !== requestUrl.origin
+    ) {
+      throw createCustomError('Forbidden', { status: 403 });
+    } else if (
+      requestUrl.protocol === 'http:' &&
+      originUrl.host !== requestUrl.host
+    ) {
       throw createCustomError('Forbidden', { status: 403 });
     }
   } else if (req.headers.get('sec-fetch-site') === 'cross-site') {
--- a/packages/waku/src/lib/utils/request.ts
+++ b/packages/waku/src/lib/utils/request.ts
@@ -126,7 +126,6 @@ function validateServerActionRequest(req: Request) {
     if (origin === 'null') {
       throw createCustomError('Forbidden', { status: 403 });
     }
-
     const requestUrl = new URL(req.url);
     let originUrl: URL;
     try {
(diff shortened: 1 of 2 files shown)


packages/waku/src/lib/utils/request.ts
```diff
@@ -126,14 +126,24 @@ function validateServerActionRequest(req: Request) {
...
-    if (originUrl.origin !== requestOrigin) {
+
+    if (
+      requestUrl.protocol === 'https:' &&
+      originUrl.origin !== requestUrl.origin
+    ) {
+      throw createCustomError('Forbidden', { status: 403 });
+    } else if (
+      requestUrl.protocol === 'http:' &&
+      originUrl.host !== requestUrl.host
```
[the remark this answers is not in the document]
lqhuang: Good catch, add more tests and improved

lqhuang:
@dai-shi Hi, Daishi,

I think the failed one job is unrelated to my changes. What would you suggest I do to rerun the CI?

And what do you think about this PR?

Thanks a lot :)

Best,
Lanqing

packages/waku/src/lib/utils/request.ts
```diff
@@ -126,14 +126,26 @@ function validateServerActionRequest(req: Request) {
     if (origin === 'null') {
       throw createCustomError('Forbidden', { status: 403 });
     }
-    const requestOrigin = new URL(req.url).origin;
+
+    const requestUrl = new URL(req.url);
```
dai-shi:
```suggestion
    const requestUrl = new URL(req.url);
```

packages/waku/src/lib/utils/request.ts
```diff
@@ -126,14 +126,26 @@ function validateServerActionRequest(req: Request) {
...
       throw createCustomError('Forbidden', { status: 403 });
     }
-    if (originUrl.origin !== requestOrigin) {
+
+    // Reject if the request's origin is different from the request URL's origin.
+    if (
+      originUrl.origin !== requestUrl.origin &&
+      // Only allow same-host requests upgrading from http to https (e.g. behind a reverse proxy),
+      // but not the other way around
+      !(
```
dai-shi:
```suggestion
    if (
      originUrl.origin !== requestUrl.origin &&
      !(
```

dai-shi (approved):
This should be a reasonable change for no configs, which is my preference.

PR #2287, opened 2026-09-02
docs: add Authentication and Migrating from Next.js guides

dai-shi:
Two guides grounded in four migrations of real Next.js apps.

Authentication shows where each part of an auth flow goes in Waku: reading a session from request headers, protecting routes in the layout that wraps them, setting a cookie from a server action through a request-scoped jar that middleware writes onto the response, signing out, and keeping session-dependent fragments in a lazy slice so the page around them stays static. It also names the two approaches that fail silently: a path-matching Hono middleware, which never sees the /RSC/R/<path>.txt request a client navigation makes and so serves the protected payload; and c.header() from server code, which Waku's handler discards when it replaces c.res. The jar's read-through is explained by the case that needs it — redirect() renders the destination in the same request, where the just-queued cookie is otherwise invisible.

Migrating from Next.js maps every App Router API to its Waku counterpart: project shape, getConfig, params and search params, metadata (and why layout and page tags do not merge), routing hooks and typed routes, the file conventions that become plain React, caching removal, parallel and intercepting routes via layout persistence, and the pieces with no equivalent.


MrDaves24:
Coming from next.js and being tired of it, I just stumbled in this very issue with authentication. Thanks for the guide and explanations about the migration!
In your guide about auth, you show how to protect every page that needs to be. I was wondering if there was a way to protect a "group" of pages instead of protecting pages one by one. Is there a way to group pages using "(group)" or even routing under "/user" (in a folder `user`) and write one check at one place that protects every pages under such folder?

dai-shi:
> Coming from next.js and being tired of it, I just stumbled in this very issue with authentication. Thanks for the guide and explanations about the migration! In your guide about auth, you show how to protect every page that needs to be. I was wondering if there was a way to protect a "group" of pages instead of protecting pages one by one. Is there a way to group pages using "(group)" or even routing under "/user" (in a folder `user`) and write one check at one place that protects every pages under such folder?

@MrDaves24 That's good input. I'm working on improving docs and features. RSC level isn't a good place for authentication. You can't simply protect a page or a group of pages. You should either protect data or routes.

PR #2294, opened 2026-09-07
fix(build): fail static generation on render errors

fuma-nama:
Currently, when a component throws inside a Suspense while pre-rendering, it only logs the errors but don't fail the build.

This PR throws an error to fail the build explicitly. I only noticed this problem after I found some pages were blank after build.

### DIFF
--- a/docs/create-pages.mdx
+++ b/docs/create-pages.mdx
@@ -241,7 +241,7 @@ export default adapter(pages);
 
 #### Disabling SSR
 
-`unstable_disableSSR` disables document SSR for a page. Runtime document requests return Waku's fallback HTML shell, and build output uses fallback HTML for that route. The RSC payload is still available to the client router.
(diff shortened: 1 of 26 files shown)


dai-shi (changes requested):
Thanks for working on it. I may want to refactor a bit, but before that could you take a look at CI result?

fuma-nama:
I think it's a pre-existing problem surfaced by the PR, the noSSR page was still running SSR and throwing error, this PR surfaced the error which caused the test to fail.

dai-shi:
Interesting. Thanks for the fix.

dai-shi:
Hm, on second thought, the proposed change is not acceptable as it false alarms (fails) with expected error boundary (or actually suspense boundary) use case. Let me try something else.

dai-shi:
On another thought, what I thought valid use cases are not technically meaningful (mostly because error boundary is not server rendered.)
Moving forward with this direction. I just added a fix in serializeRsc, which is mostly unrelated, but revealed with this exploration.

dai-shi:
> I just added a fix in serializeRsc

Reverted. Will open a dedicated PR.

dai-shi (approved):

